PR #36, opened 2024-09-30
Refactor `parser` and add comprehensive edge case tests for it

a-hilaly:
This patch refactors the `parser.Parse*` functions, by split the validation
logic from the extraction logic. Now we use helper functions for schema validation
and type checking. Allowing reading to distinguish recursive walk logic, from
other mechanics used for parsing.

- Refactor `parseResource` function to use type specific parsing functions
- Introduce `TestParserEdgeCases` to cover various type mimatches and
  error conditions..
- Add `TestParseWithExpectedSchema` to validate schema extraction from
  objects

### DIFF
--- a/internal/typesystem/parser/parser.go
+++ b/internal/typesystem/parser/parser.go
@@ -57,129 +57,150 @@ func ParseResource(resource map[string]interface{}, resourceSchema *spec.Schema)
 // from a resource. It uses a depthh first search to traverse the resource and
 // extract expressions from string fields
 func parseResource(resource interface{}, schema *spec.Schema, path string) ([]ExpressionField, error) {
-	var expressionsFields []ExpressionField
-	if schema == nil {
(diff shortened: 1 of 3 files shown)


internal/typesystem/parser/parser.go
```diff
@@ -57,129 +57,150 @@ func ParseResource(resource map[string]interface{}, resourceSchema *spec.Schema)
 // from a resource. It uses a depthh first search to traverse the resource and
 // extract expressions from string fields
 func parseResource(resource interface{}, schema *spec.Schema, path string) ([]ExpressionField, error) {
```
a-hilaly, line 59: i'm still mixed on whether we need to sort the expression fields by path. The order of expressions isn't guaranteed here.

a-hilaly: The graph runtime doesn't really care about the order of expressions. Probably something to deal with later if it becomes problematic.

PR #54, opened 2024-10-21
Refactor resourcegroup package: introduce graph and runtime interfaces

a-hilaly:
This commit significantly refactors the resourcegroup package, improving
architecture and separation of concerns.

The new interfaces provide a clear contract for implement Symphony's
core functionalities - they will mainly allow for multiple implementations
and easier testing componenets generation.

Key changes include:

1. New `graph` package for resource group graph construction and analysis.
2. New `runtime` package with clear interfaces for resource group operations.
3. Improved type safety and error handling throughou the codebase
4. Introduction of a `variable` package to centralize variable related structures.
5. Refactored resource group builder using new graph and runtime interfaces
6. update controller logic to utilize the new runtime interface.


internal/resourcegroup/graph/graph_builder.go
```diff
@@ -11,17 +11,18 @@
 // express or implied. See the License for the specific language governing
```
michaelhtm: Why are we changing the name of this file?

config/crd/bases/x.symphony.k8s.aws_resourcegroups.yaml
```diff
@@ -113,7 +113,6 @@ spec:
                     readyOn:
                       items:
                         type: string
-                        x-kubernetes-preserve-unknown-fields: true
```
a-hilaly, line 116: @michaelhtm i don't why my code-gen removed this, i'm curious to know what changed between your PR and mine

internal/resourcegroup/runtime/runtime.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Interface defines the main runtime interface for managing and synchronizing
+// resources.
+//
+// Note: The use of interfaces here is to allow for easy testing and mocking of
+// the runtime the instance controller uses. e.g we can create a fake runtime
+// that returns a specific set of resources for testing purposes.
+//
+// The other reason for the interface is to allow for different implementations
+type Interface interface {
```
a-hilaly, line 30: While i'm not a super fan of using `Interface` as a name, it kinda make sense here. `resourcegroup/runtime.Inteface`...

a-hilaly: NOTE(a-hilaly): I very rarely find myself defining interfaces and contracts.. when writing controllers. In this case, we're leveraging interfaces to achieve three objectives:
1. Abstracting the state management from the instance reconciliation logic: The controller now only cares about knowing if a certain resource is ready to be resolved (created/deployed) and is completely isolated from the details of CEL expressions and readiness conditions.
2. Decoupling the Graph build stage from the runtime stage: It was challenging to keep both processes aligned, as they used similar concepts but had different goals. For instance, a Graph builder extracts the CEL expressions, while the runtime evaluates them. Splitting these concerns into two packages necessitated the creation of a `variable` package to manage separate data structures for post processing and post evaluation variables.
3. Enabling user generated mocks: this isn't my n1 approach when it comes to testing, but the symphony systems are getting really complex and are only going to grow more so. We need to be able to isolate components and test them thoroughly, before even getting into the e2e tests, as often e2e tests will require running near-real-life production systems and will probably take a long time to execute.

config/crd/bases/x.symphony.k8s.aws_resourcegroups.yaml
michaelhtm: I added that line manually in the beginning..when i generated it the line remained 🤔.

internal/resourcegroup/graph/graph_builder.go
a-hilaly: Good call i changed it back to builder :) - probably that was me overthinking it

config/crd/bases/x.symphony.k8s.aws_resourcegroups.yaml
a-hilaly: What controller-gen version are you using? i'm using v0.16.2

PR #58, opened 2024-10-23
celinspector: fix variable scope conflicts and resource detection edge cases

a-hilaly:
Fix several edge cases in the CEL inspector around variable scoping and resource detection:

- Fix variable scope conflicts between top-level declarations and loop variables
  (e.g when 'i' is both a declared resource and used as a loop variable in filter)
- Add proper tracking of loop variables to avoid false positives in unknown resource detection
- Fix chain method calls on unknown resources to correctly track the dependency chain
- Handle ambiguous identifiers more robustly in nested scopes
- Add comprehensive test coverage for edge cases and variable conflicts

The changes make the inspector more reliable when dealing with complex expressions
that mix resource references with loop variables and method chains.

### DIFF
--- a/internal/typesystem/celinspector/celinspector.go
+++ b/internal/typesystem/celinspector/celinspector.go
@@ -21,38 +21,87 @@ import (
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 )
 
+// ResourceDependency represents a resource and its accessed path within a CEL expression.
+// For example, in the expression "deployment.spec.replicas > 0",
+// Name would be "deployment" and Path would be "deployment.spec.replicas"
 type ResourceDependency struct {
+	// Name is the root resource identifier (e.g "deployment", "service", "pod")
 	Name string
+	// Path is the full access path including nested fields
+	// For example: "deployment.spec.replicas" or "service.metadata.name"
 	Path string
 }
 
+// FunctionCall represents an invocation of a declared function within a CEL expression.
+// This tracks both the function name and its arguments as they appear in the expression
+//
+// The arguments are string representations of the AST nodes. We mainly ignore them for
+// now, but they could be used to further analyze the expression.
 type FunctionCall struct {
-	Name      string
+	// Name is the function identifier
+	// For example: "hash" "toLower"
+	Name string
+
+	// Arguments contains the string representation of each argument passed to the function
+	// For example: ["deployment.name", "'frontend'"] for a call like concat(deployment.name, "frontend")
 	Arguments []string
 }
 
+// UnknownResource represents a resource reference in the expression that wasn't
+// declared in the known resources list. This helps identify potentially missing
+// or misspelled resource names.
 type UnknownResource struct {
+	// Name is the undeclared resource identifier that was referenced
 	Name string
+	// Path is the full access path that was attempted with this unknown resource
+	// For example: "unknown_resource.field.subfield"
 	Path string
 }
 
+// UnknownFunction represents a function call in the expression that wasn't
+// declared in the known functions list and isn't a CEL built in function.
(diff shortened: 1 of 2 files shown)


internal/typesystem/celinspector/celinspector.go
```diff
@@ -80,9 +129,12 @@ func NewInspector(resources []string, functions []string) (*Inspector, error) {
 		env:       env,
 		resources: resourceMap,
 		functions: functionMap,
+		loopVars:  make(map[string]struct{}),
 	}, nil
 }
 
+// NewInspectorWithEnv creates a new Inspector instance with the given resources and functions
+// using the provided CEL environment.
```
michaelhtm, line 137: Should we add loopVars here too?

internal/typesystem/celinspector/celinspector.go
```diff
@@ -202,88 +278,219 @@ func (a *Inspector) inspectIdent(ident *exprpb.Expr_Ident, currentPath string) E
...
+	case *exprpb.Constant_NullValue:
+		return "null"
+	default:
+		return "<unknown constant>"
+	}
+}
+
+// callToString converts a function call expression to its string representation.
+// This includes both regular function calls and operator calls.
+func (a *Inspector) callToString(call *exprpb.Expr_Call) string {
```
michaelhtm: nit: could this be called `callExpressionToString`?

internal/typesystem/celinspector/celinspector.go
```diff
@@ -133,45 +193,63 @@ func (a *Inspector) inspectAst(expr *exprpb.Expr, currentPath string) Expression
...
+		// This is an unknown function, but not an internal one
+		inspection.UnknownFunctions = append(inspection.UnknownFunctions, UnknownFunction{Name: call.Function})
 	}
 
 	return inspection
 }
 
+// inspectIdent analyzes identifier expressions in CEL and determines if they are known resources
+// or unknown references. It handles the base identifiers in field access chains and distinguishes
+// between declared resources and unknown/internal identifiers.
```
michaelhtm, line 247: i think we can find a better name for this function as well...
`inspectVariable`, `inspectSymbol`..

internal/typesystem/celinspector/celinspector.go
```diff
@@ -133,45 +193,63 @@ func (a *Inspector) inspectAst(expr *exprpb.Expr, currentPath string) Expression
 	}
 }
 
+// inspectCall analyzes function calls and method invocations within a CEL expression.
+// It tracks three types of calls:
+// 1. Custom functions (declared in Inspector initialization)
+// 2. Method calls on resources ( list.filter(...))
+// 3. Unknown functions (neither custom nor internal)
```
michaelhtm, line 201: `inspectCallExpression`

internal/typesystem/celinspector/celinspector.go
a-hilaly: Great catch!

internal/typesystem/celinspector/celinspector.go
a-hilaly: I agree there are better names to use here, however the naming choice here was to make as close to the CEL AST library (the `Ident` name comes from the CEL library 🤷 )

internal/typesystem/celinspector/celinspector.go
a-hilaly: ditto

internal/typesystem/celinspector/celinspector.go
a-hilaly: I think CEL libraries are going to evolve and change their naming standards with time (before stabilizing and releasing v1.0.0), we'll need to keep an eye on that and update our library accordingly.

Issue #15, opened 2024-09-21
ResourceGroup conditional resources


a-hilaly:
Add support for conditional resource creation in `ResourceGroups` allowing resources to be created only when specified conditions are met.

- Implement condition specification using CEL expressions
- Allow referencing of spec fields and other contextual information in conditions
- Add condition evaluation during the reconciliation process

Conditional example (still not decided on the field name and type):
```yaml
apiVersion: x.symphony.k8s.aws/v1alpha1
kind: ResourceGroup
metadata:
  name: conditionalExample.x.symphony.k8s.aws
spec:
  apiVersion: v1alpha1
  kind: ConditionalExample
  definition:
    spec:
      enabled: 
  resources:
  - name: service
    conditional:
    - ${spec.enabled}
    - ${123 > 100 && 123 < 200}
    definition:
      apiVersion: v1
      kind: Service
      metadata:
        name: myservice
      spec: {}
```

michaelhtm:
We can extend this feature and support conditional fields, allowing for certain fields to be null when specified conditions are not met. 

An example 
```yaml
metadata:
  name: ${spec.name}
  namespace: ${if(spec.namespace != "", spec.namespace, null)}
```

a-hilaly:
Grazie @michaelhtm !

PR #64, opened 2024-10-28
(feat) Add conditional expressions for resources

michaelhtm:
Issue #15

Description of changes:
Conditional expressions allow the user to dictate whether 
they want to create a resource or not via a series of 
expressions that evaluate to a boolean. The variables in the 
expressions need to be declared in the instance spec for 
now. If all the expressions evaluate to true, we will create 
the resource, if at least one returns false, we will ignore 
the resource and all its dependents.


internal/controller/instance/controller_reconcile.go
```diff
@@ -160,8 +160,14 @@ func (igr *instanceGraphReconciler) reconcileResource(ctx context.Context, resou
...
 
+	if want, err := igr.runtime.WantToCreateResource(resourceID); err != nil || !want {
+		log.V(1).Info("Skipping resource creation", "reason", err)
+		resourceState.State = "SKIPPED"
+		return nil
+	}
+
 	rUnstructured, state := igr.runtime.GetResource(resourceID)
-	if state != runtime.ResourceStateResolved {
+	if state != runtime.ResourceStateResolved && state != runtime.ResourceStateIgnoredByConditions {
```
a-hilaly: What's the difference between `WantToCreateResource` returning false and `state != runtime.ResourceStateIgnoredByConditions` - directly and indirectly ignored?

internal/resourcegroup/graph/builder.go
```diff
@@ -313,18 +314,24 @@ func (b *Builder) buildRGResource(rgResource *v1alpha1.Resource, namespacedResou
 		return nil, fmt.Errorf("failed to parse readyOn expressions: %v", err)
 	}
 
+	conditional, err := parser.ParseConditionExpressions(rgResource.Conditional)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse consitional expressions: %v", err)
+	}
```
a-hilaly, line 320: I like that we're using the same functionality with ReadyOn and Conditional 👍

internal/resourcegroup/graph/builder.go
```diff
@@ -749,6 +757,36 @@ func validateResourceCELExpressions(resources map[string]*Resource, instance *Re
...
+					},
+				}
+
+				output, err := dryRunExpression(instanceEnv, conditionalExpression, context)
+				if err != nil {
+					return fmt.Errorf("failed to dry-run expression %s: %w", conditionalExpression, err)
+				}
+				if !celutil.IsBoolType(output) {
+					return fmt.Errorf("output of conditional expression %s can only be of type bool", conditionalExpression)
+				}
```
a-hilaly, line 788: Should refactor this logic in a seperate function that we can use for both conditionals and readyOn expressions?

internal/resourcegroup/graph/resource.go
```diff
@@ -138,6 +141,10 @@ func (r *Resource) GetReadyOnExpressions() []string {
 	return r.readyOnExpressions
 }
 
+func (r *Resource) GetConditionalExpressions() []string {
+	return r.conditionalExpressions
+}
```
a-hilaly, line 147: nit: godoc

internal/resourcegroup/runtime/resourcegroup.go
```diff
@@ -160,6 +160,9 @@ type ResourceGroupRuntime struct {
 	// dependencies, preventing circular dependencies and ensuring efficient
 	// synchronization.
 	topologicalOrder []string
+
+	// ignoredByConditionalResources holds the resources whos defined conditionals returned false
+	ignoredByConditionalResources map[string]bool
```
a-hilaly, line 167: I can't figure out where is the map getting populated

internal/resourcegroup/runtime/resourcegroup.go
```diff
@@ -189,6 +192,9 @@ func (rt *ResourceGroupRuntime) GetResource(name string) (*unstructured.Unstruct
 		return rt.resources[name].Unstructured(), ResourceStateResolved
 	}
 
+	if _, ignored := rt.ignoredByConditionalResources[name]; ignored {
+		return nil, ResourceStateIgnoredByConditions
+	}
```
a-hilaly: Do we want to make this statement before L190? what if a resource is resolved but has a conditional=false?

internal/resourcegroup/runtime/resourcegroup.go
```diff
@@ -538,6 +544,77 @@ func (rt *ResourceGroupRuntime) IsResourceReady(resourceID string) (bool, string
 	return true, "", nil
 }
 
+// IsIgnored will returns true if we're ignoring a resource due to a conditional
+// or for being a dependent of a resource that is ignored
+func (rt *ResourceGroupRuntime) areDependenciesIgnored(resourceID string) bool {
```
a-hilaly:
```suggestion
// areDependenciesIgnored will returns true if we're ignoring a resource due to a conditional
// or for being a dependent of a resource that is ignored
func (rt *ResourceGroupRuntime) areDependenciesIgnored(resourceID string) bool {
```

internal/resourcegroup/runtime/runtime.go
```diff
@@ -96,6 +100,10 @@ type ResourceDescriptor interface {
 	// evaluated before the resource is considered ready.
 	GetReadyOnExpressions() []string
 
+	// GetConditionalExpressions returns the list of expressions that need to
+	// be evaluated before t
+	GetConditionalExpressions() []string
```
a-hilaly: before what :)

internal/resourcegroup/runtime/state.go
```diff
@@ -40,6 +40,9 @@ const (
 	// according to its defined readiness criteria. e.g waiting for a Pod to
 	// be running and ready, a PVC to be bound ...
 	ResourceStateWaitingOnReadiness ResourceState = "WaitingOnReadiness"
+
+	// ResourceStateIgnoredByConditions
```
a-hilaly: can we expand on what does this condition indicate

internal/typesystem/variable/variable.go
```diff
@@ -114,6 +114,16 @@ const (
...
+	// ResourceVariableKindConditional represents a conditional variable.
+	// Conditional variables are resolved at the beginning of the execution and
+	// their value is constant. They decide whether we are going to create
+	// a resource or not
+	//
+	// For example:
+	//   name: deployment
+	//   conditional:
+	//   - ${spec.replicas > 1}
+	ResourceVariableKindConditional ResourceVariableKind = "conditional"
```
a-hilaly, line 126: 👍

internal/typesystem/variable/variable.go
```diff
@@ -130,3 +140,7 @@ func (r ResourceVariableKind) IsStatic() bool {
 func (r ResourceVariableKind) IsDynamic() bool {
 	return r == ResourceVariableKindDynamic
 }
+
+func (r ResourceVariableKind) IsConditional() bool {
+	return r == ResourceVariableKindConditional
+}
```
a-hilaly, line 147: nit: godoc

internal/controller/instance/controller_reconcile.go
michaelhtm: Good call, had to use this to test a bug. It never does get here if ignored

internal/resourcegroup/graph/builder.go
michaelhtm: I thought of that, even for dynamic/static variable resolution

internal/resourcegroup/runtime/resourcegroup.go
michaelhtm: Removing it entirely since there is no case where the method is called is the resource is ignored

internal/resourcegroup/runtime/resourcegroup.go
michaelhtm: Forgot that :|.. Adding population

test/integration/suites/core/conditional_test.go
```diff
@@ -0,0 +1,249 @@
...
+		Expect(env.Client.Create(ctx, instance)).To(Succeed())
+
+		// Check if instance is created
+		Eventually(func(g Gomega) {
+			err := env.Client.Get(ctx, types.NamespacedName{
+				Name:      name,
+				Namespace: namespace,
+			}, instance)
+			g.Expect(err).ToNot(HaveOccurred())
+		}, 20*time.Second, time.Second).Should(Succeed())
```
a-hilaly: I think we need to wait for the instance to be `ACTIVE` before checking on the deployment and serviceaccount. The reason for this is that the controller creates a resource every 3 seconds, meaning that for 3 seconds deployment won't be created, and for 6 seconds it won't be created. We want to check that the instance was full reconciled before checking anything below.

internal/resourcegroup/graph/builder.go
michaelhtm: Can't really refactor this section by a lot, since readyOn and conditionals handle contexts differently.
Not sure how to handle that

PR #73, opened 2024-10-30
Safe faulty ResourceGroup cleanup

michaelhtm:
Description of changes:
When the ResourceGroup graph is faulty, the cleanup tends to panic.
With this change, during cleanup, we avoid creating the entire 
resourceGroup, but we just retrieve the values we need to delete such 
as the GVR and the CRD name.

### DIFF
--- a/test/integration/suites/core/validation_test.go
+++ b/test/integration/suites/core/validation_test.go
@@ -324,6 +324,42 @@ var _ = Describe("Validation", func() {
 			}
 		})
 	})
+
+	Context("Proper Cleanup", func() {
+		It("should not panic when deleting an inactive ResourceGroup", func() {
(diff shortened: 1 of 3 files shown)


test/integration/suites/core/validation_test.go
```diff
@@ -324,6 +324,42 @@ var _ = Describe("Validation", func() {
...
+					map[string]interface{}{
+						"name": "string",
+					},
+					nil,
+				),
+				generator.WithResource("testResource", map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ServiceAccount",
+					"metadata": map[string]interface{}{
+						"name": "${Bad expression}",
```
a-hilaly, line 343: Clever one!

PR #96, opened 2024-11-08
Make readyWhen variables start with their own name

michaelhtm:
*Description of changes:*
This change ensures that all variables referenced in readyWhen 
expressions are referencing their own resource by name

This change will also allow us in the future to reference other 
resource fields if needed

changing from:
```yaml
name: cluster
readyWhen:
- ${status.state == 'available'}
```

to
```yaml
name: cluster
readyWhen:
- ${cluster.status.state == 'available'}
```

### DIFF
--- a/internal/graph/builder.go
+++ b/internal/graph/builder.go
@@ -731,27 +731,26 @@ func validateResourceCELExpressions(resources map[string]*Resource, instance *Re
 			// I would also suggest separating the dryRuns of readyWhenExpressions
 			// and the resourceExpressions.
 			for _, readyWhenExpression := range resource.readyWhenExpressions {
-				fieldNames := schema.GetResourceTopLevelFieldNames(resource.schema)
-				fieldEnv, err := scel.DefaultEnvironment(scel.WithResourceNames(fieldNames))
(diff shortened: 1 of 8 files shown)


internal/graph/builder.go
```diff
@@ -731,26 +731,31 @@ func validateResourceCELExpressions(resources map[string]*Resource, instance *Re
...
 				// add resource fields to the context
-				context := map[string]*Resource{}
-				for _, n := range fieldNames {
-					context[n] = &Resource{
-						emulatedObject: &unstructured.Unstructured{
-							Object: resource.emulatedObject.Object[n].(map[string]interface{}),
-						},
+				resourceEmulatedCopy := resource.emulatedObject.DeepCopy()
+				if resourceEmulatedCopy != nil && resourceEmulatedCopy.Object != nil {
+					if _, ok := resourceEmulatedCopy.Object["apiVersion"]; ok {
```
a-hilaly: nit: I think there is a guarantee that these objeects will have an apiVersion and a kind https://github.com/awslabs/kro/blob/main/internal/graph/emulator/emulator.go#L69-L70

PR #121, opened 2024-11-11
Change resource entrypoint from `name` to `id`

michaelhtm:
**Description of changes:**
This gives the perception that the entrypoint should be unique

### DIFF
--- a/internal/graph/validation.go
+++ b/internal/graph/validation.go
@@ -65,9 +65,9 @@ var (
 	}
 )
 
-// isValidResourceName checks if the given name is a valid KRO resource name (loawercase)
-func isValidResourceName(name string) bool {
-	return lowerCamelCaseRegex.MatchString(name)
+// isValidResourceID checks if the given id is a valid KRO resource id (loawercase)
+func isValidResourceID(id string) bool {
+	return lowerCamelCaseRegex.MatchString(id)
 }
 
 // isValidKindName checks if the given name is a valid KRO kind name (uppercase)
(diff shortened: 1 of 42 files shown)


internal/graph/validation.go
```diff
@@ -109,18 +109,18 @@ func validateResourceGroupNamingConventions(rg *v1alpha1.ResourceGroup) error {
 func validateResourceNames(rg *v1alpha1.ResourceGroup) error {
 	seen := make(map[string]struct{})
 	for _, res := range rg.Spec.Resources {
-		if isKROReservedWord(res.Name) {
-			return fmt.Errorf("name %s is a reserved keyword in KRO", res.Name)
+		if isKROReservedWord(res.ID) {
+			return fmt.Errorf("name %s is a reserved keyword in KRO", res.ID)
```
a-hilaly:
```suggestion
			return fmt.Errorf("id %s is a reserved keyword in KRO", res.ID)
```

internal/graph/validation.go
```diff
@@ -109,18 +109,18 @@ func validateResourceGroupNamingConventions(rg *v1alpha1.ResourceGroup) error {
...
 		}
 
-		if !isValidResourceName(res.Name) {
-			return fmt.Errorf("name %s is not a valid KRO resource name: must be lower camelCase", res.Name)
+		if !isValidResourceName(res.ID) {
+			return fmt.Errorf("name %s is not a valid KRO resource name: must be lower camelCase", res.ID)
```
a-hilaly:
```suggestion
		if !isValidResourceName(res.ID) {
			return fmt.Errorf("id %s is not a valid KRO resource id: must be lower camelCase", res.ID)
```

internal/graph/validation.go
```diff
@@ -109,18 +109,18 @@ func validateResourceGroupNamingConventions(rg *v1alpha1.ResourceGroup) error {
...
 		}
 
-		if _, ok := seen[res.Name]; ok {
-			return fmt.Errorf("found duplicate resource name %s", res.Name)
+		if _, ok := seen[res.ID]; ok {
+			return fmt.Errorf("found duplicate resource name %s", res.ID)
```
a-hilaly:
```suggestion
			return fmt.Errorf("found duplicate resource ids %s", res.ID)
```

Issue #139, opened 2024-11-14
Resources not accessible in multi-line string fields


giedri:
### Description

**Observed Behavior**:
Synchronization fails if resource refers another resource's field in the string value, for example:
```
        apiVersion: iam.services.k8s.aws/v1alpha1
        kind: Role
...
        spec:
...
          inlinePolicies:
            s3-policy: |
              {
...
                    "Resource": [ 
                        "arn:aws:s3:::${inputBucket.status.location}/${schema.spec.input.prefix}/*"
                    ]
...
                  }
```
fails, while 
```
        apiVersion: iam.services.k8s.aws/v1alpha1
        kind: Role
...
        spec:
...
          inlinePolicies:
            s3-policy: |
              {
...
                    "Resource": [ 
                        "arn:aws:s3:::${schema.spec.input.bucket}/${schema.spec.input.prefix}/*"
                    ]
...
                  }
```
works as expected
**Expected Behavior**:
I would expect fields of the other resources in the RG to be accessible throughout the template
 
**Reproduction Steps** (Please include `ResourceGroup` and `Instances` files):

**Versions**:
- kro version: 0.1.0
- Kubernetes Version (`kubectl version`):
Client Version: v1.29.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.31.0-eks-a737599
**Involved Controllers**:
- Controller URLs and Versions (if applicable):
iam.services.k8s.aws/v1alpha1
**Error Logs** (if applicable)**:
```
Message:               failed to synchronize reconciling resource inputBucket: failed to evaluate dynamic variables: failed compiling expression schema.spec.input.prefix: ERROR: <input>:1:1: undeclared reference to 'schema' (in container '')
 | schema.spec.input.prefix
```

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

a-hilaly:
This is no longer an issue in 0.4.1.

/close

PR #145, opened 2024-11-18
Fix: allow referencing dynamic and static variables in one field

michaelhtm:
**Issue #139**

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -346,6 +346,7 @@ func (rt *ResourceGroupRuntime) evaluateDynamicVariables() error {
 	// and are resolved after all the dependencies are resolved.
 
 	resolvedResources := maps.Keys(rt.resolvedResources)
+	resolvedResources = append(resolvedResources, "schema")
 	env, err := krocel.DefaultEnvironment(krocel.WithResourceIDs(resolvedResources))
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


internal/runtime/runtime.go
```diff
@@ -346,6 +346,7 @@ func (rt *ResourceGroupRuntime) evaluateDynamicVariables() error {
 	// and are resolved after all the dependencies are resolved.
 
 	resolvedResources := maps.Keys(rt.resolvedResources)
+	resolvedResources = append(resolvedResources, "schema")
```
a-hilaly, line 349: looks we allowed this during the RG parsing layer, but we didn't during the runtime execution layer. Great catch!

a-hilaly:
@michaelhtm looks like we need to inject real unstrcutured instance objects for the unit tests - i'm also wondering if we can have integration tests for this already, if it's not already the case.

internal/runtime/runtime_test.go
```diff
@@ -1690,6 +1690,9 @@ func Test_evaluateDynamicVariables(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			rt := &ResourceGroupRuntime{
+				instance: newTestResource(
+					withObject(map[string]interface{}{}),
+				),
```
a-hilaly, line 1695: I wonder how is this working when object is empty?

michaelhtm: It just looks for object to not be nil..it seems like this test does not need object

PR #148, opened 2024-11-19
parser: Add support for `x-kubernetes-preserve-unknown-fields` schema extension

a-hilaly:
Add support for the `x-kubernetes-preserve-unknown-fields` Kubernetes schema extension.
When this extension is present and set to true in a schema, skip parsing the object
and return an empty list of expressions.

This extension is commonly used in Kubernetes CRDs to allow for arbitrary nested
fields that shouldn't be validated against the schema.

### DIFF
--- a/internal/graph/parser/parser.go
+++ b/internal/graph/parser/parser.go
@@ -22,6 +22,10 @@ import (
 	"github.com/awslabs/kro/internal/graph/variable"
 )
 
+const (
+	xKubernetesPreserveUnknownFields = "x-kubernetes-preserve-unknown-fields"
+)
+
 // ParseResource extracts CEL expressions from a resource based on
 // the schema. The resource is expected to be a map[string]interface{}.
 //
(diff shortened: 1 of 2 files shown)


internal/graph/parser/parser.go
```diff
@@ -88,6 +92,15 @@ func parseObject(field map[string]interface{}, schema *spec.Schema, path, expect
 		return nil, fmt.Errorf("expected object type or AdditionalProperties allowed for path %s, got %v", path, field)
 	}
 
+	// Look for vendor schema extensions first
+	if len(schema.VendorExtensible.Extensions) > 0 {
+		// If the schema has the x-kubernetes-preserve-unknown-fields extension, we should not
+		// parse the object and return an empty list of expressions.
+		if enabled, ok := schema.VendorExtensible.Extensions[xKubernetesPreserveUnknownFields]; ok && enabled.(bool) {
+			return nil, nil
```
a-hilaly, line 100: Instead of ignoring all CEL expressions inside RawExtension fields, we could just turtle it down using the schemaless parser https://github.com/awslabs/kro/blob/main/internal/graph/parser/schemaless.go#L23-L25

internal/graph/parser/parser_test.go
```diff
@@ -618,6 +618,21 @@ func TestParserEdgeCases(t *testing.T) {
 			},
 			expectedError: "",
 		},
+		{
+			name: "shema with x-kubernetes-preserve-unknown-fields",
```
michaelhtm: nit: 
```suggestion
			name: "schema with x-kubernetes-preserve-unknown-fields",
```

michaelhtm (commented):
These are nice changes 🔥
I like the idea handling it with schemaless parser!
left one nit

PR #173, opened 2024-12-23
feat(core): track and apply resource changes through deep diffing

a-hilaly:
Up until now, we weren't really doing proper diffing when deciding whether to update
managed resources - we just marked everything as "SYNCED". This patch adds a new delta
package that does a proper deep comparison between desired and observed resource
states, being careful about all the k8s metadata fields that we should ignore during
comparisons (e.g `metadata.generation`, `metadta.revision` etc...)

A key design aspect of this implementation is that `kro` only diffs fields that are
 explicitly defined in `ResourceGroup` - any field that is defaulted by other
controllers or mutating webhooks are ignored. This ensures `kro` coexists seamlessly
with other reconciliation systems without fighting over field ownership.

The delta `Compare` function takes the desired and observed `unstructured.Unstructured`
objects and returns a list of structured "differences", where each difference contains
the full path to the changed field and its desired/observed values. It recursively
walks both object trees in parallel, building pathstring like `spec.containers[0].image`
to precisely identify where values diverge. The comparison handles type mismatches,
nil vs empty maps/slices, value differences etc...

Also patch also adds integration tests in suites/core for generic resource
updates and in suites/ackekscluster for EKS-clutser-specific version updates. note that
in both tests we mainly rely on the `metadata.revision` to validate that indeed the
controller did make an spec update call.

### DIFF
--- a/pkg/controller/instance/controller_reconcile.go
+++ b/pkg/controller/instance/controller_reconcile.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 
+	"github.com/awslabs/kro/pkg/controller/instance/delta"
 	"github.com/awslabs/kro/pkg/metadata"
 	"github.com/awslabs/kro/pkg/requeue"
 	"github.com/awslabs/kro/pkg/runtime"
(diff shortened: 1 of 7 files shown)


internal/controller/resourcegroup/controller.go
```diff
@@ -80,6 +81,7 @@ func NewResourceGroupReconciler(
 func (r *ResourceGroupReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.ResourceGroup{}).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
a-hilaly: i've observed some weird infinite-loops during integration test logs... While i know that predicate event filters can fix it, i'm still slightly hesitant, it should be needed..

Need to do some careful sleuthing here

internal/controller/instance/delta/delta.go
```diff
@@ -0,0 +1,181 @@
...
+		if len(annotations) == 0 {
+			delete(metadata, "annotations")
+		}
+	}
+
+	if annotations, exists := metadata["labels"].(map[string]interface{}); exists {
+		if len(annotations) == 0 {
+			delete(metadata, "labels")
+		}
+	}
```
michaelhtm: nit:
```suggestion
	if labels, exists := metadata["labels"].(map[string]interface{}); exists {
		if len(labels) == 0 {
			delete(metadata, "labels")
		}
	}
```

internal/controller/instance/delta/delta.go
```diff
@@ -0,0 +1,181 @@
...
+	cleanMetadata(desiredCopy)
+	cleanMetadata(observedCopy)
+
+	var differences []Difference
+	walkCompare(desiredCopy.Object, observedCopy.Object, "", &differences)
+	return differences, nil
+}
+
+// ignoredFields are Kubernetes metadata fields that should not trigger updates.
+var ignoredFields = []string{
```
michaelhtm: Will we add more fields to ignore here? or just the ones in metadata?
Because instead i'm thinking we can call it ignoredMetadataFields...

internal/controller/instance/delta/delta.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	if annotations, exists := metadata["labels"].(map[string]interface{}); exists {
+		if len(annotations) == 0 {
+			delete(metadata, "labels")
+		}
+	}
+
+	for _, field := range ignoredFields {
+		delete(metadata, field)
```
michaelhtm: 👍🏼

internal/controller/instance/delta/delta.go
```diff
@@ -0,0 +1,181 @@
...
+
+// Compare takes desired and observed unstructured objects and returns a list of
+// their differences. It performs a deep comparison while being aware of Kubernetes
+// metadata specifics. The comparison:
+//
+// - Cleans metadata from both objects to avoid spurious differences
+// - Walks object trees in parallel to find actual value differences
+// - Builds path strings to precisely identify where differences occurs
+// - Handles type mismatches, nil values, and empty vs nil collections
+func Compare(desired, observed *unstructured.Unstructured) ([]Difference, error) {
```
michaelhtm, line 42: Does this allow all possible updates? Including removing a resource, adding new ones, etc.?

michaelhtm (commented):
Great Job @a-hilaly 🔥. I left a few comments.

michaelhtm:
🔥
/lgtm

PR #174, opened 2024-12-23
refactor: move core packages from internal to pkg directory

a-hilaly:
Move controller, graph, metadata, and runtime packages from `internal/`
to `pkg/` directory to make them available for external consumption.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #175, opened 2024-12-23
feat: parse expressions in `x-kubernetes-preserve-unknown` fields

a-hilaly:
Fixes https://github.com/awslabs/kro/issues/162
Merge after #174 

Previously, fields marked with `x-kubernetes-preserve-unknown-fields` were skipped
during parsing. Now these fields are parsed to extract expressions, allowing for
variable substitution in schemaless fields while still preserving unknown fields
as specified by the k8s-open-api extension.

### DIFF
--- a/pkg/graph/parser/parser.go
+++ b/pkg/graph/parser/parser.go
@@ -96,10 +96,17 @@ func parseObject(field map[string]interface{}, schema *spec.Schema, path, expect
 
 	// Look for vendor schema extensions first
 	if len(schema.VendorExtensible.Extensions) > 0 {
-		// If the schema has the x-kubernetes-preserve-unknown-fields extension, we should not
-		// parse the object and return an empty list of expressions.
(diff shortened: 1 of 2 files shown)


pkg/graph/parser/parser.go
```diff
@@ -99,7 +99,11 @@ func parseObject(field map[string]interface{}, schema *spec.Schema, path, expect
 		// If the schema has the x-kubernetes-preserve-unknown-fields extension, we should not
 		// parse the object and return an empty list of expressions.
```
michaelhtm: nit:
let's change comments to reflect the code change?

michaelhtm (commented):
Nice feature @a-hilaly 🔥
left one nit

pkg/graph/parser/parser.go
a-hilaly: Good catch! Fixed~

Issue #164, opened 2024-12-09
Make Group name configurable within ResourceGroup


amirahav:
### Feature Description

**Problem Statement**:

The `Group` name of the CRD created by a `ResourceGroup` is currently hardcoded. It would be beneficial to be able to define the group name to assist in differentiating kro specific CRDs and user/customer defined CRDs.
 
**Proposed Solution**:

Add a `group` attribute within `spec.schema` which should default to `kro.run`, similar to this:

```
apiVersion: kro.run/v1alpha1
kind: ResourceGroup
metadata:
  name: sqs-queue
spec:
  schema:
    apiVersion: v1alpha1
    kind: Queue
    group: somegroup.xyz.com
    spec:
      ...
```

**Alternatives Considered**:

**Additional Context**:

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

RinkiyaKeDad:
I'm happy to pick this up!

RinkiyaKeDad:
Here's how I was thinking this could be implemented:

1. Adding `Group string` to the schema in [`kro/api/v1alpha1/resource_group.go`](https://github.com/awslabs/kro/blob/main/api/v1alpha1/resource_group.go#L50)
2. Passing this information to the `buildInstanceResource` function in [`kro/internal/graph
/builder.go`](https://github.com/awslabs/kro/blob/main/internal/graph/builder.go#L192)
3. Further passing this info to `SynthesizeCRD` in [`kro/internal/graph
/builder.go`](https://github.com/awslabs/kro/blob/main/internal/graph/builder.go#L454) to get the new CRD which would have the group info.
4. Updating the [`extractCRDName`](https://github.com/awslabs/kro/blob/main/internal/controller/resourcegroup/controller_cleanup.go) function to accomodate the newly added Group field during deletion. 
5. Updating the `GetResourceGroupInstanceGVK` and `GetResourceGroupInstanceGVR` functions in [`kro/internal/metadata/groupversion.go`](https://github.com/awslabs/kro/blob/main/internal/metadata/groupversion.go#L71-L87)


Things I'm not sure about:
1. Do we need to do something about [`internal/metadata/labels.go`](https://github.com/awslabs/kro/blob/main/internal/metadata/labels.go)?


@a-hilaly can you please verify if I'm in the right direction before I start working on this? 

[...]

PR #176, opened 2024-12-26
made group name configurable within resourcegroup

RinkiyaKeDad:
*Issue #, if available:* #164

*Description of changes:* 

- Adding `Group string` to the schema in [kro/api/v1alpha1/resource_group.go](https://github.com/awslabs/kro/blob/main/api/v1alpha1/resource_group.go#L50)
- Passing this information to the `buildInstanceResource` function in [kro/internal/graph /builder.go](https://github.com/awslabs/kro/blob/main/internal/graph/builder.go#L192)
- Further passing this info to `SynthesizeCRD` in [kro/internal/graph /builder.go](https://github.com/awslabs/kro/blob/main/internal/graph/builder.go#L454) to get the new CRD which would have the group info.
- Updating the [extractCRDName](https://github.com/awslabs/kro/blob/main/internal/controller/resourcegroup/controller_cleanup.go) function to accomodate the newly added Group field during deletion.
- Updating the `GetResourceGroupInstanceGVK` and `GetResourceGroupInstanceGVR` functions in [kro/internal/metadata/groupversion.go](https://github.com/awslabs/kro/blob/main/internal/metadata/groupversion.go#L71-L87)

### DIFF
--- a/api/v1alpha1/resource_group.go
+++ b/api/v1alpha1/resource_group.go
@@ -60,6 +60,12 @@ type Schema struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="apiVersion is immutable"
 	APIVersion string `json:"apiVersion,omitempty"`
+	// The group of the resourcegroup. This is used to set the API group
+	// of the generated CRD. If omitted, it defaults to "kro.run".
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="kro.run"
+	Group string `json:"group,omitempty"`
 	// The spec of the resourcegroup. Typically, this is the spec of
 	// the CRD that the resourcegroup is managing. This is adhering
 	// to the SimpleSchema spec
--- a/config/crd/bases/kro.run_resourcegroups.yaml
+++ b/config/crd/bases/kro.run_resourcegroups.yaml
@@ -103,6 +103,12 @@ spec:
                     x-kubernetes-validations:
                     - message: apiVersion is immutable
                       rule: self == oldSelf
+                  group:
+                    default: kro.run
+                    description: |-
+                      The group of the resourcegroup. This is used to set the API group
+                      of the generated CRD. If omitted, it defaults to "kro.run".
+                    type: string
                   kind:
                     description: |-
                       The kind of the resourcegroup. This is used to generate
--- a/helm/crds/kro.run_resourcegroups.yaml
+++ b/helm/crds/kro.run_resourcegroups.yaml
@@ -103,6 +103,12 @@ spec:
                     x-kubernetes-validations:
                     - message: apiVersion is immutable
                       rule: self == oldSelf
+                  group:
+                    default: kro.run
+                    description: |-
+                      The group of the resourcegroup. This is used to set the API group
+                      of the generated CRD. If omitted, it defaults to "kro.run".
+                    type: string
                   kind:
                     description: |-
                       The kind of the resourcegroup. This is used to generate
--- a/pkg/controller/resourcegroup/controller_cleanup.go
+++ b/pkg/controller/resourcegroup/controller_cleanup.go
@@ -35,13 +35,13 @@ func (r *ResourceGroupReconciler) cleanupResourceGroup(ctx context.Context, rg *
 	log.V(1).Info("cleaning up resource group", "name", rg.Name)
 
 	// shutdown microcontroller
-	gvr := metadata.GetResourceGroupInstanceGVR(rg.Spec.Schema.APIVersion, rg.Spec.Schema.Kind)
+	gvr := metadata.GetResourceGroupInstanceGVR(rg.Spec.Schema.Group, rg.Spec.Schema.APIVersion, rg.Spec.Schema.Kind)
 	if err := r.shutdownResourceGroupMicroController(ctx, &gvr); err != nil {
 		return fmt.Errorf("failed to shutdown microcontroller: %w", err)
 	}
 
 	// cleanup CRD
-	crdName := extractCRDName(rg.Spec.Schema.Kind)
+	crdName := extractCRDName(rg.Spec.Schema.Group, rg.Spec.Schema.Kind)
 	if err := r.cleanupResourceGroupCRD(ctx, crdName); err != nil {
 		return fmt.Errorf("failed to cleanup CRD %s: %w", crdName, err)
 	}
@@ -75,8 +75,8 @@ func (r *ResourceGroupReconciler) cleanupResourceGroupCRD(ctx context.Context, c
 
 // extractCRDName generates the CRD name from a given kind by converting it to plural form
 // and appending the Kro domain name.
-func extractCRDName(kind string) string {
+func extractCRDName(group, kind string) string {
 	return fmt.Sprintf("%s.%s",
 		flect.Pluralize(strings.ToLower(kind)),
-		v1alpha1.KroDomainName)
+		group)
 }
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -163,7 +163,7 @@ func (b *Builder) NewResourceGroup(originalCR *v1alpha1.ResourceGroup) (*Graph, 
 	// That's because the instance status schema is inferred from the CEL expressions
 	// in the status field of the instance resource. Those CEL expressions refer to
 	// the resources defined in the resource group. Hence, we need to build the resources
-	// first, to be able ot generate a proper schema for the instance status.
+	// first, to be able to generate a proper schema for the instance status.
 
 	//
 
@@ -190,6 +190,7 @@ func (b *Builder) NewResourceGroup(originalCR *v1alpha1.ResourceGroup) (*Graph, 
 	// 4. Infer the status schema based on the CEL expressions.
 
 	instance, err := b.buildInstanceResource(
+		rg.Spec.Schema.Group,
 		rg.Spec.Schema.APIVersion,
 		rg.Spec.Schema.Kind,
 		rg.Spec.Schema,
@@ -415,7 +416,7 @@ func (b *Builder) buildDependencyGraph(
 // Since instances are defined using the "SimpleSchema" format, we use a different
 // approach to build the instance resource. We need to:
 func (b *Builder) buildInstanceResource(
-	apiVersion, kind string,
+	group, apiVersion, kind string,
 	rgDefinition *v1alpha1.Schema,
 	resources map[string]*Resource,
 ) (*Resource, error) {
@@ -428,7 +429,7 @@ func (b *Builder) buildInstanceResource(
 	// CRD declarations.
 
 	// The instance resource is a Kubernetes resource, so it has a GroupVersionKind.
-	gvk := metadata.GetResourceGroupInstanceGVK(apiVersion, kind)
+	gvk := metadata.GetResourceGroupInstanceGVK(group, apiVersion, kind)
 
 	// We need to unmarshal the instance schema to a map[string]interface{} to
 	// make it easier to work with.
@@ -451,7 +452,7 @@ func (b *Builder) buildInstanceResource(
 
 	// Synthesize the CRD for the instance resource.
 	overrideStatusFields := true
-	instanceCRD := crd.SynthesizeCRD(apiVersion, kind, *instanceSpecSchema, *instanceStatusSchema, overrideStatusFields)
+	instanceCRD := crd.SynthesizeCRD(group, apiVersion, kind, *instanceSpecSchema, *instanceStatusSchema, overrideStatusFields)
 
 	// Emulate the CRD
 	instanceSchemaExt := instanceCRD.Spec.Versions[0].Schema.OpenAPIV3Schema
--- a/pkg/graph/crd/crd.go
+++ b/pkg/graph/crd/crd.go
@@ -17,7 +17,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/awslabs/kro/api/v1alpha1"
 	"github.com/gobuffalo/flect"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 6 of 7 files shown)


RinkiyaKeDad:
@a-hilaly I need some help figuring out how do I run `kro` locally to test my changes. I created an issue to document this process as well: https://github.com/awslabs/kro/issues/177

This is what I tried:

- Installed `kro` using the helm chart in a local [k3d](https://k3d.io/stable/) cluster. 
- Ran `make run` after making my changes so that I could test them

I assumed this would have the controller running locally on my machine and the cluster would use that instead of the helm installed one but that was not the case because I kept getting this error:

```
❯ k apply -f resourcegroup.yaml
Error from server (BadRequest): error when creating "resourcegroup.yaml": ResourceGroup in version "v1alpha1" cannot be handled as a ResourceGroup: strict decoding error: unknown field "spec.schema.group"
```

Can you please help out with how do I run these changes I made to `kro` locally? I can then create a guide and add to the CONTRIBUTING.md as well for future contributors.

michaelhtm:
Hey @RinkiyaKeDad, thanks for contributing to **kro**
Good catch, I will take a look at the Contributing doc.

After making changes, you can run the controller locally with the following command:
```bash
go run -tags=dev cmd/controller/main.go 
```

Make sure to install the updated CRD when making changes to it, if you haven't already.

RinkiyaKeDad:
Thanks for clearing that @michaelhtm! 

Let's say I'm developing a feature, the flow should be:
- I apply the updated CRDs using `kubectl apply -f config/crd/bases/` (this is the command right?)
- Run `go run -tags=dev cmd/controller/main.go`

When I want to go back to my default `kro` installation I terminate the controller process but how do I go back to the original CRDs? Will I have to do a helm install again? Or check out into the `main` branch and apply the CRDs from there? Is there a better way?

newtondev:
Hi there, and thanks so much for adding this. I am currently testing it out on my own fork and everything looks good. This feature is great since the idea here is that we have a few separate teams that will be creating abstractions and it would be good to know which team is responsible for the building block. 

I managed to get it all running with:

```shell
make build
kubectl apply -k config/crd
make run
```

I am in the process of writing some tests just to be sure it works completely as expected.

The only difference in my code base is that I do not have an `internal` folder.

newtondev:
> Thanks for clearing that @michaelhtm!
> 
> Let's say I'm developing a feature, the flow should be:
> 
> * I apply the updated CRDs using `kubectl apply -f config/crd/bases/` (this is the command right?)
> * Run `go run -tags=dev cmd/controller/main.go`
> 
> When I want to go back to my default `kro` installation I terminate the controller process but how do I go back to the original CRDs? Will I have to do a helm install again? Or check out into the `main` branch and apply the CRDs from there? Is there a better way?

I tend to checkout main and then just build and install the CRDs from there. You could also just keep a backup of the CRDs in a local folder too.

RinkiyaKeDad:
Hey @newtondev! Thanks for testing it. I was just doing that as well. 

I applied the following `ResourceGroup`:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGroup
metadata:
  name: my-application
spec:
  # kro uses this simple schema to create your CRD schema and apply it
  # The schema defines what users can provide when they instantiate the RG (create an instance).
  schema:
    apiVersion: v1alpha1
    kind: AppWithGroup
    group: arshsharma.com
    spec:
      name: string
      ingress:
        enabled: boolean | default=false
    status:
      # Fields the controller will inject into instances status.
      deploymentConditions: ${deployment.status.conditions}
      availableReplicas: ${deployment.status.availableReplicas}

  # Define the resources this API will manage.
(same as the getting started sample)
```

And these CRDs got created:

```
❯ k get crds
NAME                                    CREATED AT
addons.k3s.cattle.io                    2024-12-19T07:36:40Z
appwithgroups.arshsharma.com            2025-01-02T14:00:57Z
appwithgroups.kro.run                   2025-01-02T14:00:57Z
```
There should only be one CRD right? Can you help me figure out where the lifecycle of the created resource starts so I can know where I'm missing to add some code. I thought it started in [builder.go here](https://github.com/awslabs/kro/blob/main/pkg/graph/builder.go#L192) but I've already added code there and that doesn't seem to do the trick. 

Also the reason you don't see an `internal` folder is probably because I think it got renamed to the `pkg` directory: https://github.com/awslabs/kro/tree/main/pkg

newtondev:
Hi @RinkiyaKeDad, yes indeed, there should only be the 1 crd, you might have to remove the old one manually:

```shell
k get crds
NAME                        CREATED AT
applications.newtondev.nl   2025-01-02T14:22:12Z
resourcegroups.kro.run      2025-01-02T12:28:43Z
```

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGroup
metadata:
  name: my-application
spec:
  # kro uses this simple schema to create your CRD schema and apply it
  # The schema defines what users can provide when they instantiate the RG (create an instance).
  schema:
    group: newtondev.nl
    apiVersion: v1alpha1
    kind: Application
....
```

```yaml
apiVersion: newtondev.nl/v1alpha1
kind: Application
metadata:
  name: my-application-instance
spec:
  name: my-awesome-app
  ingress:
    enabled: false
```

```shell
k get application
NAME                      STATE    SYNCED   AGE
my-application-instance   ACTIVE   True     2m12s

k get deployments,services,ingresses
NAME                             READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/my-awesome-app   3/3     3            3           2m34s

NAME                             TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
service/kubernetes               ClusterIP   10.96.0.1      <none>        443/TCP   132m
service/my-awesome-app-service   ClusterIP   10.96.99.126   <none>        80/TCP    2m31s
```

```shell
k delete application my-application-instance
application.newtondev.nl "my-application-instance" deleted

k get deployments,services,ingresses
NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
service/kubernetes   ClusterIP   10.96.0.1    <none>        443/TCP   135m
```

RinkiyaKeDad:
I resolved the conflicts it had with the main branch. 

> Hi @RinkiyaKeDad, yes indeed, there should only be the 1 crd, you might have to remove the old one manually:

What I meant was that 2 CRDs are getting created currently when I create a new `ResourceGroup` with the newly added `Group` field. Which means something more needs to be added in the code :)

newtondev:
Hi @RinkiyaKeDad , I am not seeing an extra one being created. I also added a printcolumn to the group:

```
git diff api/v1alpha1/resource_group.go
diff --git a/api/v1alpha1/resource_group.go b/api/v1alpha1/resource_group.go
index b26b1ab..2b51f7e 100644
--- a/api/v1alpha1/resource_group.go
+++ b/api/v1alpha1/resource_group.go
@@ -60,6 +60,12 @@ type Schema struct {
        // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="apiVersion is immutable"
        APIVersion string `json:"apiVersion,omitempty"`
+       // The group of the resourcegroup. This is used to set the API group
+       // of the generated CRD. If omitted, it defaults to "kro.run".
+       //
+       // +kubebuilder:validation:Optional
+       // +kubebuilder:default="kro.run"
+       Group string `json:"group,omitempty"`
        // The spec of the resourcegroup. Typically, this is the spec of
        // the CRD that the resourcegroup is managing. This is adhering
        // to the SimpleSchema spec
@@ -120,6 +126,7 @@ type Dependency struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="GROUP",type=string,priority=0,JSONPath=`.spec.schema.group`
 // +kubebuilder:printcolumn:name="APIVERSION",type=string,priority=0,JSONPath=`.spec.schema.apiVersion`
 // +kubebuilder:printcolumn:name="KIND",type=string,priority=0,JSONPath=`.spec.schema.kind`
 // +kubebuilder:printcolumn:name="STATE",type=string,priority=0,JSONPath=`.status.state`
```

```shell
k get rg
NAME             GROUP          APIVERSION   KIND          STATE    AGE
my-application   newtondev.nl   v1alpha1     Application   Active   121m
```

RinkiyaKeDad:
Thanks @newtondev, that's useful! Could you commit that to this PR? I don't want to take the credit for your contribution :) 

> I am not seeing an extra one being created.

That's really weird. Here's what I'm experiencing:

https://github.com/user-attachments/assets/b7346ed4-5f9b-4106-94bf-763f48338481

Let's wait for someone else to test I as well I guess. Can you share the resource group you tried creating?

newtondev:
Hi @RinkiyaKeDad , I believe it will not automatically delete the CRD when you delete the RG. There is a flag `allowCRDDeletion` that when not specified defaults to `false`. I believe you might have an old CRD installed alongside your custom one and will require manual deletion.

Here is my branch with the changes that you can grab from: https://github.com/newtondev/kro/tree/feat-configurable-group-name - don't worry about credit :) 

```shell
vscode ➜ /workspaces/kro (feat-configurable-group-name) $ k get crd
NAME                     CREATED AT
resourcegroups.kro.run   2025-01-02T12:28:43Z

vscode ➜ /workspaces/kro (feat-configurable-group-name) $ cd tmp
vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k apply -f resourcegroup.yaml 
resourcegroup.kro.run/my-application created

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get crd
NAME                        CREATED AT
applications.newtondev.nl   2025-01-03T15:22:13Z
resourcegroups.kro.run      2025-01-02T12:28:43Z

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get rg
NAME             GROUP          APIVERSION   KIND          STATE    AGE
my-application   newtondev.nl   v1alpha1     Application   Active   18s

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k apply -f instance.yaml 
application.newtondev.nl/my-application-instance created

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get application
NAME                      STATE         SYNCED   AGE
my-application-instance   IN_PROGRESS   False    4s

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get application
NAME                      STATE    SYNCED   AGE
my-application-instance   ACTIVE   True     16s

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k delete -f instance.yaml 
application.newtondev.nl "my-application-instance" deleted

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get application
No resources found in default namespace.

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get rg
NAME             GROUP          APIVERSION   KIND          STATE    AGE
my-application   newtondev.nl   v1alpha1     Application   Active   60s

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k delete -f resourcegroup.yaml 
resourcegroup.kro.run "my-application" deleted

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get rg
No resources found in default namespace.

vscode ➜ /workspaces/kro/tmp (feat-configurable-group-name) $ k get crd
NAME                        CREATED AT
applications.newtondev.nl   2025-01-03T15:22:13Z
resourcegroups.kro.run      2025-01-02T12:28:43Z
```

a-hilaly:
Catching up with this thread :)

>What I meant was that 2 CRDs are getting created currently when I create a new ResourceGroup with the newly added Group field. Which means something more needs to be added in the code :) 

@RinkiyaKeDad Yeah.. the controller shouldn't be creating two CRDs. Could you check if there are two controller instances running simultaneously - perhaps an old version installed through helm alongside the newly patched one (locally)? This might explain why they're both trying to reconcile the RG and create CRDs.

RinkiyaKeDad:
> Could you check if there are two controller instances running simultaneously

How do I check that? What you're saying seems very likely. But I'm not sure how to stop my cluster from not using the Helm installed one and only use the one I'm running locally. Should I just uninstall it using Helm?

a-hilaly:
You can just check if you already have a kro pod running in cluster, either via `helm` or `kubectl` CLI. 

> Should I just uninstall it using Helm?

Yes, ideally you only want one controller running.

a-hilaly:
Now thinking about it - we should enable leader election and lease claiming by default for kro controllers to help prevent these conflicts. we already have this supported https://github.com/awslabs/kro/blob/main/cmd/controller/main.go#L80

api/v1alpha1/resource_group.go
```diff
@@ -60,6 +60,12 @@ type Schema struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="apiVersion is immutable"
 	APIVersion string `json:"apiVersion,omitempty"`
+	// The group of the resourcegroup. This is used to set the API group
+	// of the generated CRD. If omitted, it defaults to "kro.run".
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="kro.run"
+	Group string `json:"group,omitempty"`
```
a-hilaly, line 68: Wondering if it makes sense to leverage the existing `apiVersion` field to optionally specify the group  (like how k8s can have "apps/v1" or just "v1")? This would align with k8s conventions while keep our schema simpler. Just an idea.. what do you think about this approach? 

cc @rushmash91 @michaelhtm @newtondev

a-hilaly (commented):
@RinkiyaKeDad Thank you so much for working on this! Overall the change looks good to me. I have one question inline about relying on the existing `spec.schema.apiVersion` field.

api/v1alpha1/resource_group.go
RinkiyaKeDad: I don't have a preference on this, I'm fine with implementing whatever approach you folks see fit :)

newtondev: I believe that makes sense if you were trying to group the objects into some logical group like `apps`, but it does not solve the use case there you want a custom group though. Some companies like to use their own name in the group name sometimes :) 

`networking.mycompany.com/v1beta1`

Whereas if group is not configurable then it would look something like:

`kro.run/networking/v1beta1`

Where apiVersion="networking/v1beta1"

Also happy to not make the group name configurable.

rushmash91: I am happy with either but leaning towards using apiVersion for the simple schema. I do see the flexibility with group though.

michaelhtm: not sure if this is what you mean @a-hilaly, what i understand is, as default we use an empty group, and if someone wants to specify their own group, they should be free to do so. Is that correct? If so I do like that idea!

a-hilaly: @michaelhtm what I meant is we could've used `"spec.schema.apiVersion"` to contain both the `apiVersion` AND optionally a domain, meaning that you could've have:
- `apiVersion: v1alpha1` (which defaults the group to "kro.run")
- `apiVersion: my.custom.group/v1alpha1` (which overrides the "kro.run" with "my.custom.group")

However after some thought on this, I think it makes a lot of sense to use a separate field for the group, and the reason is that one day we might want kro to automatically manage `apiVersions`. In that case it would make a lot of sense to ask users to keep `apiVersion` empty and only ask them to provide a `spec.schema.group`.

a-hilaly (approved):
Neatt!! Thank you sir!
/lgtm

PR #182, opened 2025-01-09
adding support for `x-kubernetes-int-or-string` in dummy CR

### LINKED ISSUE #168: Support `x-kubernetes-int-or-string: true` from schema
### Feature Description

**Problem Statement**:

I would like to use flagger (and other) controllers that have fields configured with `x-kubernetes-int-or-string: true` meaning that the field can be an integer or a string. KRO doesn't seem to work with this right now.

The dummy CR generation doesn't know how to handle this configuration yet and fails.

**Proposed Solution**:

The dummy CR generator should use an integer for CR generation.

**Alternatives Considered**:

I've patched the flagger schema to force int or string and it works, but since their schema is valid and widely used, kro should probably handle their case.

**Additional Context**:

RinkiyaKeDad:
Fixes #168 

*Description of changes:* Adding support for `x-kubernetes-int-or-string: true` in the dummy CR generator and a test case for that.  Based on discussion of changes [here](https://github.com/awslabs/kro/issues/168#issuecomment-2578021187)

### DIFF
--- a/pkg/graph/emulator/emulator.go
+++ b/pkg/graph/emulator/emulator.go
@@ -79,6 +79,11 @@ func (e *Emulator) generateValue(schema *spec.Schema) (interface{}, error) {
 		return nil, fmt.Errorf("schema is nil")
 	}
 
+	if enabled, ok := schema.VendorExtensible.Extensions["x-kubernetes-int-or-string"]; ok && enabled.(bool) {
+		// Default to integer for dummy CRs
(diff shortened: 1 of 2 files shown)


RinkiyaKeDad:
Pinging @a-hilaly for review :)

a-hilaly (approved):
Nice! Thank you Arsh!
/lgtm

PR #197, opened 2025-01-10
add  `x-kubernetes-int-or-string` support for parser

rushmash91:
Fixes https://github.com/awslabs/kro/issues/168

*Description of changes:*
	•	Added the constant xKubernetesIntOrString to detect if the CRD schema has the extension.
	•	Updated parseString and parseScalarTypes to handle fields annotated with x-kubernetes-int-or-string, allowing them to be treated as either a string or an integer.
	•	Added Test Cases function: TestXKubernetesIntOrString and TestNestedXKubernetesIntOrString

### DIFF
--- a/pkg/graph/parser/parser.go
+++ b/pkg/graph/parser/parser.go
@@ -24,6 +24,8 @@ import (
 
 const (
 	xKubernetesPreserveUnknownFields = "x-kubernetes-preserve-unknown-fields"
+	xKubernetesIntOrString           = "x-kubernetes-int-or-string"
+	schemaTypeAny                    = "any"
 )
 
 // ParseResource extracts CEL expressions from a resource based on
@@ -45,7 +47,10 @@ func parseResource(resource interface{}, schema *spec.Schema, path string) ([]va
 		return nil, err
 	}
 
-	expectedType := getExpectedType(schema)
+	expectedType, err := getExpectedType(schema, resource)
+	if err != nil {
+		return nil, err
+	}
 
 	switch field := resource.(type) {
 	case map[string]interface{}:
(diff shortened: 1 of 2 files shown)


michaelhtm:
Great job @rushmash91!
/lgtm
cc: @a-hilaly

pkg/graph/parser/parser.go
```diff
@@ -24,6 +24,8 @@ import (
 
 const (
 	xKubernetesPreserveUnknownFields = "x-kubernetes-preserve-unknown-fields"
+	xKubernetesIntOrString           = "x-kubernetes-int-or-string"
+	schemaTypeAny                    = "any"
```
a-hilaly, line 28: 👍

pkg/graph/parser/parser.go
```diff
@@ -75,18 +80,38 @@ func validateSchema(schema *spec.Schema, path string) error {
...
+	if enabled {
+		switch resource.(type) {
+		case string:
+			return "string", nil
+		case int:
+			return "integer", nil
+		default:
+			return "", nil
+		}
+	}
```
a-hilaly: Since this is related to to the logic above, should we call it right after L90?

pkg/graph/parser/parser.go
```diff
@@ -75,18 +80,38 @@ func validateSchema(schema *spec.Schema, path string) error {
...
 	if schema.AdditionalProperties != nil && schema.AdditionalProperties.Allows {
 		// NOTE(a-hilaly): I don't like the type "any", we might want to change this to "object"
 		// in the future; just haven't really thought about it yet.
 		// Basically "any" means that the field can be of any type, and we have to check
 		// the ExpectedSchema field.
-		return "any"
+		return schemaTypeAny, nil
 	}
-	return ""
+	return "", nil
```
a-hilaly: Should return an error here? Thinking
```suggestion
	return "", fmt.Errorf("unknown schema type")
```

pkg/graph/parser/parser.go
```diff
@@ -75,18 +80,38 @@ func validateSchema(schema *spec.Schema, path string) error {
...
+	}
+
+	if enabled {
+		switch resource.(type) {
+		case string:
+			return "string", nil
+		case int:
+			return "integer", nil
+		default:
+			return "", nil
```
a-hilaly: Ditto i think we can error here as well
```suggestion
			return "", fmt.Errorf("found `xKubernetesIntOrString` extension but field value is neither a string nor an integer: %v, resourceType)
```

pkg/graph/parser/parser.go
```diff
@@ -75,18 +80,38 @@ func validateSchema(schema *spec.Schema, path string) error {
...
+	var enabled bool
+	if ext, ok := schema.VendorExtensible.Extensions[xKubernetesIntOrString]; ok {
+		enabled, ok = ext.(bool)
+		if !ok {
+			return "", fmt.Errorf("xKubernetesIntOrString extension is not a boolean")
+		}
+	}
+
+	if enabled {
+		switch resource.(type) {
```
a-hilaly:
```suggestion
		switch resourceType := resource.(type) {
```

pkg/graph/parser/parser.go
```diff
@@ -153,6 +178,7 @@ func parseString(field string, schema *spec.Schema, path, expectedType string) (
 	if err != nil {
 		return nil, err
 	}
+
```
a-hilaly: nit:
```suggestion
```

a-hilaly (commented):
Good start, thank you @rushmash91 !! left a few comments in line

a-hilaly (approved):
Good stuff Arush! Thank you!
/lgtm

PR #200, opened 2025-01-13
Add support for multiple expected types by converting `expectedtype` from `string` to `[]string`

### LINKED ISSUE #181: Support multiple valid types
#### Enhancing Type Validation in CEL Expression Parser

Currently, our parser only stores a single `expectedType` (a string) when validating or extracting CEL expressions. However, certain Kubernetes schemas (including CRDs) can specify multiple valid types using `anyOf` etc, which is not fully supported by the existing design.

#### Single-type assumption
The parser uses `expectedType` string internally, so it assumes only one type is valid (e.g., "string", "object", etc.).

#### Multiple-type scenarios
In many CRDs or Kubernetes resources, a field might be valid as an "object" or a "string" (e.g., `anyOf: [ {type: "string"}, {type: "object"} ]`). The current code cannot represent this because `expectedType` is not a list, leading to erroneous failures or loss of details when trying to parse fields.

## Proposed Solution

1. **handle multiple types**
   This allows us to store multiple valid types (e.g., `["string", "object"]`) instead of a single string.

2. Update parsing and validation logic

rushmash91:
fixes #181 

*Description of changes:*
- In`FieldDescriptor` convert `ExpectedType` to `ExpectedTypes` of type `[]string` 
- `getExpectedType` returns `[]string` of all valid types
- `isExpectedType` utility function verifies is the type is valid

### DIFF
--- a/pkg/graph/parser/parser.go
+++ b/pkg/graph/parser/parser.go
@@ -15,6 +15,7 @@ package parser
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
 	"k8s.io/kube-openapi/pkg/validation/spec"
@@ -47,22 +48,22 @@ func parseResource(resource interface{}, schema *spec.Schema, path string) ([]va
 		return nil, err
 	}
 
-	expectedType, err := getExpectedType(schema, resource)
+	expectedTypes, err := getExpectedTypes(schema)
 	if err != nil {
 		return nil, err
 	}
 
 	switch field := resource.(type) {
 	case map[string]interface{}:
-		return parseObject(field, schema, path, expectedType)
+		return parseObject(field, schema, path, expectedTypes)
 	case []interface{}:
-		return parseArray(field, schema, path, expectedType)
+		return parseArray(field, schema, path, expectedTypes)
 	case string:
-		return parseString(field, schema, path, expectedType)
+		return parseString(field, schema, path, expectedTypes)
 	case nil:
 		return nil, nil
 	default:
-		return parseScalarTypes(field, schema, path, expectedType)
+		return parseScalarTypes(field, schema, path, expectedTypes)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/graph/parser/parser.go
```diff
@@ -80,40 +80,45 @@ func validateSchema(schema *spec.Schema, path string) error {
...
 }
 
-func getExpectedType(schema *spec.Schema, resource interface{}) (string, error) {
+func getExpectedTypes(schema *spec.Schema, resource interface{}) ([]string, error) {
 	// handle "x-kubernetes-int-or-string"
 	if ext, ok := schema.VendorExtensible.Extensions[xKubernetesIntOrString]; ok {
 		enabled, ok := ext.(bool)
 		if !ok {
-			return "", fmt.Errorf("xKubernetesIntOrString extension is not a boolean")
+			return []string{""}, fmt.Errorf("xKubernetesIntOrString extension is not a boolean")
```
a-hilaly: nit: we can return a nil slice instead of an empty string?

pkg/graph/parser/parser.go
```diff
@@ -80,40 +80,45 @@ func validateSchema(schema *spec.Schema, path string) error {
...
-				return "string", nil
-			case int:
-				return "integer", nil
-			default:
-				return "", fmt.Errorf("found `xKubernetesIntOrString` extension but field value is neither a string nor an integer: %v", resourceType)
-			}
+			return []string{"string", "integer"}, nil
+		}
+		if !enabled {
+			return []string{""}, fmt.Errorf("found `xKubernetesIntOrString` extension but field value is neither a string nor an integer: %T", resource)
```
a-hilaly: Do we want to error if a CRD has `x-kubernetes-int-or-string: false`? seems like a case where we shouldn't error

pkg/graph/parser/parser.go
```diff
@@ -80,40 +80,45 @@ func validateSchema(schema *spec.Schema, path string) error {
...
-			switch resourceType := resource.(type) {
-			case string:
-				return "string", nil
-			case int:
-				return "integer", nil
-			default:
-				return "", fmt.Errorf("found `xKubernetesIntOrString` extension but field value is neither a string nor an integer: %v", resourceType)
-			}
+			return []string{"string", "integer"}, nil
+		}
```
a-hilaly: What if the `resource.(type)` is a boolean?

pkg/graph/parser/parser.go
```diff
@@ -80,40 +80,45 @@ func validateSchema(schema *spec.Schema, path string) error {
...
-func parseObject(field map[string]interface{}, schema *spec.Schema, path, expectedType string) ([]variable.FieldDescriptor, error) {
-	if expectedType != "object" && (schema.AdditionalProperties == nil || !schema.AdditionalProperties.Allows) {
+func isExpectedType(expectedTypes []string, expectedType string) bool {
+	for _, exceptedType := range expectedTypes {
+		if exceptedType == expectedType {
+			return true
+		}
+	}
+	return false
+}
```
a-hilaly: I think this function could be named `sliceInclude`  - we can also just use https://pkg.go.dev/slices#Contains

pkg/graph/parser/schemaless_test.go
```diff
@@ -39,6 +39,27 @@ func areEqualExpressionFields(a, b []variable.FieldDescriptor) bool {
 	return true
 }
 
+func compareUnorderedSlices(slice1, slice2 []string) bool {
```
a-hilaly:
```suggestion
func areEqualSlices(slice1, slice2 []string) bool {
```

pkg/graph/parser/schemaless_test.go
```diff
@@ -39,6 +39,27 @@ func areEqualExpressionFields(a, b []variable.FieldDescriptor) bool {
...
+
+	// Check if all elements from slice2 are in the set
+	for _, s := range slice2 {
+		if !elementSet[s] {
+			return false
+		}
+		// Remove the element to ensure uniqueness
+		delete(elementSet, s)
+	}
+	return len(elementSet) == 0
```
a-hilaly, line 70: nice!

pkg/graph/parser/schemaless_test.go
a-hilaly: can you also add a godoc comment to this function?

a-hilaly (commented):
Great stuff Arush! i have a few comments in line

a-hilaly (approved):
Nicee!
/lgtm

PR #202, opened 2025-01-15
added `OneOf` and `AnyOf` support for parser

rushmash91:
*Description of changes:*
- Added support for `OneOf` and `AnyOf` to `getExpectedTypes` ->  returns a slice of the options
- Updated `validateSchema` to remove  `OneOf` and `AnyOf` logic -> moved to `getExpectedTypes`
- added simple and nested testcases func `TestOneOfAndAnyOf`

### DIFF
--- a/pkg/graph/parser/parser_test.go
+++ b/pkg/graph/parser/parser_test.go
@@ -663,6 +663,21 @@ func TestParserEdgeCases(t *testing.T) {
 			}},
 			expectedError: "",
 		},
+		{
+			name: "invalid schema",
+			schema: &spec.Schema{
+				SchemaProps: spec.SchemaProps{
+					Type: []string{"object"},
+					Properties: map[string]spec.Schema{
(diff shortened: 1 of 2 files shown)


pkg/graph/parser/parser_test.go
```diff
@@ -879,3 +879,224 @@ func TestNestedXKubernetesIntOrString(t *testing.T) {
 		}
 	})
 }
+
+func TestOneOfAndAnyOf(t *testing.T) {
```
a-hilaly, line 898: tiny nit: can we have a test case triggering L122? or even a seperate unit test function for `validateSchema`

a-hilaly (dismissed):
Neat!! Good job on this @rushmash91 !
i left one tiny suggestion below

pkg/graph/parser/parser_test.go
rushmash91: added an edge case for invalid schema

a-hilaly (approved):
👍

PR #204, opened 2025-01-15
update ACK controllers to combine 7 controllers into a single resource

rushmash91:
*Issue #, if available:*

*Description of changes:*
- Update the ACK Resource group example to have a combined RG that can use `includeWhen` to select the once to deploy
- added a script to setup IRSA for IAM controller (which is used by other controllers to set up roles)


a-hilaly (approved):
Thank you @rushmash91 !!

PR #205, opened 2025-01-17
adding validation for immutable fields in schema

### LINKED ISSUE #165: Disallow creation of ResourceGroups if value of immutable field is invalid
### Description

**Observed Behavior**:
kro currently allows a ResourceGroup with an invalid value of an immutable field to be created.
This then requires the deletion and recreation of the ResourceGroup.

**Expected Behavior**:
The kro controller returns an error and doesn't create the ResourceGroup.

**Reproduction Steps** (Please include `ResourceGroup` and `Instances` files):

```
apiVersion: kro.run/v1alpha1
kind: ResourceGroup
metadata:
  name: sqs-queue
spec:
  schema:
    apiVersion: v1alpha1
    kind: sqs-queue
    spec:
      ...
```

**Versions**:
- kro version:
- Kubernetes Version (`kubectl version`):

**Involved Controllers**:
- kro v0.1.0

```
Paste any relevant logs here
```

_No response_

RinkiyaKeDad:
Fixes #165 

*Description of changes:* Added validtion for `Kind` and `ApiVersion` so that kro shows a message during time of creation instead of having to delete and re-create those resources. 

Also fixed the `make` command to be run during development in the contributing doc. 


a-hilaly (approved):
Neat! Thank you @RinkiyaKeDad ~

RinkiyaKeDad:
I think the test is failing because the `Kind` of the `ResourceGroup` in the test starts with a lower case letter "test-k8s-valid" - should I change the test case `Kind` to start with an uppercase letter?

a-hilaly:
@RinkiyaKeDad i think you're right, some tests need to be updated, now that we validate things at admission, we need to assert an error on create.

RinkiyaKeDad:
So should I update the tests to start with an upper case letter in the name for the `Kind` field?

api/v1alpha1/types.go
```diff
@@ -52,12 +52,14 @@ type Schema struct {
 	// and create the CRD for the resourcegraphdefinition.
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^[A-Z][a-zA-Z0-9]*$`
```
cheftako: I believe CRD kind are supposed to match https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L244 with an addition length limit of 63 characters.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L762

a-hilaly: TIL, thanks Walter!

a-hilaly (approved):
Thanks @RinkiyaKeDad !
/lgtm

PR #208, opened 2025-01-23
resourcegroup: use default KroDomainName when group is empty

a-hilaly:
Add fallback to KroDomainName when group is not specified in
resource group controller and CRD synthesis. This ensures
consistent behavior when creating CRDs without an explicit group name.

The change affects two components:
- ResourceGroup reconciler: Set default group during cleanup
- CRD synthesis: Use default group when generating new CRDs

### DIFF
--- a/pkg/controller/resourcegroup/controller_cleanup.go
+++ b/pkg/controller/resourcegroup/controller_cleanup.go
@@ -40,8 +40,12 @@ func (r *ResourceGroupReconciler) cleanupResourceGroup(ctx context.Context, rg *
 		return fmt.Errorf("failed to shutdown microcontroller: %w", err)
 	}
 
+	group := rg.Spec.Schema.Group
+	if group == "" {
+		group = v1alpha1.KroDomainName
+	}
 	// cleanup CRD
-	crdName := extractCRDName(rg.Spec.Schema.Group, rg.Spec.Schema.Kind)
+	crdName := extractCRDName(group, rg.Spec.Schema.Kind)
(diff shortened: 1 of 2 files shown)


rushmash91:
Nicee!
/lgtm

pkg/controller/resourcegroup/controller_cleanup.go
```diff
@@ -40,8 +40,12 @@ func (r *ResourceGroupReconciler) cleanupResourceGroup(ctx context.Context, rg *
 		return fmt.Errorf("failed to shutdown microcontroller: %w", err)
 	}
 
+	group := rg.Spec.Schema.Group
```
RinkiyaKeDad, line 43: I understand what's being done here but I'm just curious why the `group` isn't populated with `kro.run` by default because in the schema we've specified this: https://github.com/kro-run/kro/blob/main/api/v1alpha1/types.go#L67

```
// +kubebuilder:default="kro.run"
```

cc @a-hilaly

a-hilaly: That's definitely a better approach :), wanna send a PR reverting my change and using kubebuilder markers instead?

RinkiyaKeDad: But I already added the markers in my PR 🤔 that's why I'm confused why this was needed, were the markers not working this means? 

You can see they're already in the code base: https://github.com/kro-run/kro/blob/main/api/v1alpha1/types.go#L67

PR #209, opened 2025-01-23
Rename module to `github.com/kro-run/kro`

a-hilaly:

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -31,11 +31,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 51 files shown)


cmd/controller/main.go
```diff
@@ -31,11 +31,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
```
michaelhtm, line 31: Should we also make these changes for docs/website?

michaelhtm (commented):
Nice one @a-hilaly 
Left one comment

cmd/controller/main.go
a-hilaly: yesir https://github.com/kro-run/kro/pull/210

nicslatts (approved):

PR #223, opened 2025-01-29
update readme, drop aws-isms

jlbutler:
We don't need the AWS questions in the README any longer. Longer term, we could use a more broad update of the README.


a-hilaly (approved):
Sweet, thanks @jlbutler !

PR #233, opened 2025-01-29
Fix resource namespaced check

matthchr:
This check did not function properly for resources with multiple versions. Rather than depending on the GVK, we use just the GK, as a GK is either namespaced or not, it can't change between versions.


a-hilaly (approved):
Nicee! Thanks @matthchr !

PR #234, opened 2025-01-29
Tweak topological sort to preserve client-specified order

justinsb:
If there are no explicit dependencies between two nodes,
try to keep them in the same order.

By doing this, we can minimize the need for synthetic dependencies.

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -142,12 +142,17 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
 
 	// we'll also store the resources in a map for easy access later.
 	resources := make(map[string]*Resource)
-	for _, rgResource := range rgd.Spec.Resources {
-		r, err := b.buildRGResource(rgResource, namespacedResources)
+	for i, rgResource := range rgd.Spec.Resources {
+		id := rgResource.ID
+		order := i
+		r, err := b.buildRGResource(rgResource, namespacedResources, order)
 		if err != nil {
-			return nil, fmt.Errorf("failed to build resource '%v': %v", rgResource.ID, err)
+			return nil, fmt.Errorf("failed to build resource %q: %w", id, err)
 		}
-		resources[rgResource.ID] = r
+		if resources[id] != nil {
+			return nil, fmt.Errorf("found resources with duplicate id %q", id)
+		}
+		resources[id] = r
 	}
 
 	// At this stage we have a superficial understanding of the resources that are
@@ -246,7 +251,7 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
 // It provides a high-level understanding of the resource, by extracting the
 // OpenAPI schema, emualting the resource and extracting the cel expressions
 // from the schema.
-func (b *Builder) buildRGResource(rgResource *v1alpha1.Resource, namespacedResources map[k8sschema.GroupKind]bool) (*Resource, error) {
+func (b *Builder) buildRGResource(rgResource *v1alpha1.Resource, namespacedResources map[k8sschema.GroupKind]bool, order int) (*Resource, error) {
 	// 1. We need to unmashal the resource into a map[string]interface{} to
 	//    make it easier to work with.
 	resourceObject := map[string]interface{}{}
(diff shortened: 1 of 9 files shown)


pkg/graph/builder.go
```diff
@@ -397,10 +403,8 @@ func (b *Builder) buildDependencyGraph(
 				resource.addDependencies(resourceDependencies...)
 				resourceVariable.AddDependencies(resourceDependencies...)
 				// We need to add the dependencies to the graph.
-				for _, dependency := range resourceDependencies {
-					if err := directedAcyclicGraph.AddEdge(resourceName, dependency); err != nil {
-						return nil, err
-					}
+				if err := directedAcyclicGraph.AddDependencies(resource.id, resourceDependencies); err != nil {
```
justinsb, line 406: This should be (very) marginally faster than adding one at a time, but really it was because I was confused on the direction of the "dependency arrow"

pkg/graph/dag/dag.go
```diff
@@ -42,134 +56,128 @@ func NewDirectedAcyclicGraph() *DirectedAcyclicGraph {
...
+		fromNode.DependsOn[dependency] = struct{}{}
+	}
 
 	// Check if the graph is still a DAG
-	hasCycle, cycle := d.HasCycle()
+	hasCycle, cycle := d.hasCycle()
 	if hasCycle {
 		// Ehmmm, we have a cycle, let's remove the edge we just added
-		delete(fromNode.Edges, to)
+		for _, dependency := range dependencies {
```
justinsb, line 111: We might want to get rid of this cleanup, and just declare the graph "broken" in this case

pkg/graph/dag/dag.go
```diff
@@ -42,134 +56,128 @@ func NewDirectedAcyclicGraph() *DirectedAcyclicGraph {
...
+		vertices = append(vertices, vertex)
+	}
+	sort.Slice(vertices, func(i, j int) bool {
+		return vertices[i].Order < vertices[j].Order
+	})
 
-	var dfs func(string)
-	dfs = func(node string) {
-		visited[node] = true
+	for len(visited) < len(vertices) {
```
justinsb, line 137: This algorithm is less efficient, but more straightforward and I think can detect cycles also (so we might be able to remove cycle detection from adding the edges).

test/integration/suites/ackekscluster/generator.go
```diff
@@ -43,8 +43,8 @@ func eksCluster(
 				"clusterARN": "${cluster.status.ackResourceMetadata.arn}",
 			},
 		),
+		generator.WithResource("clusterRole", clusterRoleDef(namespace), nil, nil),
```
justinsb, line 46: I figured out that it is much easier to reorder the objects in the "manifest" than it is to reorder the mocking behaviour :-)

a-hilaly: clever, looove it :)

pkg/graph/dag/dag.go
```diff
@@ -23,9 +24,22 @@ import (
...
+}
+
+func (v Vertex) String() string {
+	var dependsOn strings.Builder
+	for id := range v.DependsOn {
+		if dependsOn.Len() != 0 {
+			dependsOn.WriteString(",")
+		}
+		dependsOn.WriteString(id)
+	}
```
a-hilaly: nit: maybe `strings.Join`? i guess strings.Builder is theoretically faster?

pkg/graph/dag/dag.go
```diff
@@ -23,9 +24,22 @@ import (
...
 	ID string
-	// Edges stores the IDs of the nodes that this node has an outgoing edge to.
-	// In kro, this would be the children of a resource.
-	Edges map[string]struct{}
+	// Order records the original order, and is used to preserve the original user-provided ordering as far as posible.
+	Order int
+	// DependsOn stores the IDs of the nodes that this node depends on.
+	// If we depend on another vertex, we must appear after that vertex in the topological sort.
+	DependsOn map[string]struct{}
+}
```
a-hilaly, line 34: This makes the code base easier to understand 👍

a-hilaly (dismissed):
Wonderful! Thank you sir!

pkg/graph/dag/dag.go
justinsb: Good point, I have no idea why I did it this way :joy:

a-hilaly (approved):

PR #237, opened 2025-01-30
remove UUID-based finalizers in favor of static string

a-hilaly:
The previous implementation used resource UUIDs as part of finalizer names,
which was about as stable as building a house on quicksand (I really thought
using UUIDs that change on resource recreation was a brilliant idea at the
time


pkg/metadata/finalizers.go
```diff
@@ -63,15 +63,14 @@ func HasInstanceFinalizer(obj metav1.Object, uid types.UID) bool {
 }
 
 // SetInstanceFinalizerUnstructured adds an instance-specific finalizer to an unstructured object.
-func SetInstanceFinalizerUnstructured(obj *unstructured.Unstructured, uid types.UID) error {
-	finalizerName := getInstanceFinalizerName(uid)
```
matthchr, line 67: Planning to remove the `getInstanceFinalizerName` too?

matthchr (dismissed):

pkg/metadata/finalizers.go
a-hilaly: removed along with 3 other non used functions

matthchr (approved):

PR #238, opened 2025-01-30
make `ResourceGraphDefinition` cluster scoped 

a-hilaly:
Make `ResourceGraphDefinition` a cluster-scoped resource instead of namespace-scoped:
- Add `scope=Cluster` to RGD kubebuilder resource marker
- Regenerate CRD
- Remove namespace field and related code from tests and generators
- Update test cases to use non namespacde resource names


barney-s (approved):
Lgtm

PR #239, opened 2025-01-30
chore: bump version to 0.2.0

a-hilaly:


pkg/client/set.go
```diff
@@ -62,7 +62,7 @@ func NewSet(cfg Config) (*Set, error) {
 	if config.Burst == 0 {
 		config.Burst = cfg.Burst
 	}
-	config.UserAgent = "kro/0.1.0"
+	config.UserAgent = "kro/0.2.0"
```
a-hilaly, line 65: to address after this PR: we probably want to us ldflags to inject this version

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -72,7 +72,7 @@ func NewResourceGraphDefinitionReconciler(
 		allowCRDDeletion:  allowCRDDeletion,
 		crdManager:        crdWrapper,
 		dynamicController: dynamicController,
-		metadataLabeler:   metadata.NewKroMetaLabeler("0.1.0", "kro-pod"),
+		metadataLabeler:   metadata.NewKroMetaLabeler("0.2.0", "kro-pod"),
```
a-hilaly, line 75: same for the pod id, but using k8s downward api

barney-s (approved):

PR #259, opened 2025-02-01
Copy finalizers from observed to desired object on resource update

amirahav:
Testing the changes in https://github.com/kro-run/kro/pull/173 - even though resource updates do seem to work, the finalizers are removed from the updated resource.
This will add finalizers, if any are defined.


a-hilaly (approved):
Great catch, and thank you for your contribution @amirahav :)

a-hilaly:
i'm thinking that we definitely need a v0.2.1 with this patch

a-hilaly:
https://github.com/kro-run/kro/pkgs/container/kro%2Fcontroller/347572012?tag=dev-45adaf5 is live with this patch

PR #260, opened 2025-02-02
fix: schema parser improvements for complex CRD patterns

a-hilaly:
This patch fixes schema validation and expression extraction for CRDs that
define schemas like `map[string]struct{properties any}` using a combination
of `additionalProperties` and `x-kubernetes-preserve-unknown-fields`
vendor extension.

this commit improves structured resource parsing to:
- safely handle empty/nil schema types with proper length checks
- Support `x-kubernetes-preserve-unknown-fields` at any schema level by validating
  extensions before type checking
- Support patterns like `map[string]struct{Properties any}`

Example of schemas we parse correctly now:
```yaml
  resources:
    additionalProperties:  # map[string]->struct
      properties:
        properties:       # any
```

links:
- https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/operator/api/pulumi/v1/program_types.go#L64-L82
- https://github.com/pulumi/pulumi-kubernetes-operator/blob/master/operator/config/crd/bases/pulumi.com_programs.yaml


a-hilaly:
cc @rushmash91 can you please take a look here?

rushmash91:
LGTM

PR #261, opened 2025-02-02
fix: preserve owner references during resource updates

a-hilaly:
During resource updates, we now maintain owner references on resources managed by
the controller, similar to how we already preserve finalizers. This ensures that
existing ownership relationships are not disrupted during reconciliation. Added
integration tests to verify both finalizers and owner references are preserved
when resources are updated.

### DIFF
--- a/pkg/controller/instance/controller_reconcile.go
+++ b/pkg/controller/instance/controller_reconcile.go
@@ -273,6 +273,8 @@ func (igr *instanceGraphReconciler) updateResource(
 	// TODO: Handle annotations
 	desired.SetResourceVersion(observed.GetResourceVersion())
 	desired.SetFinalizers(observed.GetFinalizers())
+	desired.SetOwnerReferences(observed.GetOwnerReferences())
+
 	_, err = rc.Update(ctx, desired, metav1.UpdateOptions{})
 	if err != nil {
 		resourceState.State = "ERROR"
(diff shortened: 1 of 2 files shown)


amirahav:
Do `OwnerReferences` get added during resource creation?
I was going to open a PR adding them during resource creation but decided not to do so based on https://github.com/kro-run/kro/issues/48

barney-s:
Cluster scoped resources cannot use namespace scoped ownerrefs. We definitely need an alternative mechanism. Or treat Cluster scoped resources as special cases.

barney-s:
Having ownerrefs is important as a fallback for cleanup. K8s ownerrefs are namespace scoped by design.

pkg/controller/instance/controller_reconcile.go
```diff
@@ -273,6 +273,8 @@ func (igr *instanceGraphReconciler) updateResource(
 	// TODO: Handle annotations
 	desired.SetResourceVersion(observed.GetResourceVersion())
 	desired.SetFinalizers(observed.GetFinalizers())
+	desired.SetOwnerReferences(observed.GetOwnerReferences())
```
barney-s, line 276: Is this for adoption use cases ? 
What about setting ownerref to the instance ?

n3wscott, line 276: Are you sure you can flat copy the observed into the desired? Has a merge already happened?

a-hilaly:
closing in favor of applysets

PR #262, opened 2025-02-02
add 'schema' to kro reserved keywords list

a-hilaly:
'schema' is reserved as it's used to reference instance fields (e.g `schema.spec.name` || `schema.metadata.name`)


barney-s:
LGTM

barney-s (approved):

PR #264, opened 2025-02-03
test(core): RGDs all the way down

a-hilaly:
Adds integration tests that verify:
- Nested `ResourceGraphDefinition` lifecycle including creation, status updates, and cleanup
- Dynamic creation of RGDs with different schema field types (integer, string, boolean)

### DIFF
--- a/test/integration/suites/core/nested_test.go
+++ b/test/integration/suites/core/nested_test.go
@@ -1,0 +1,251 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package core_test
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	krov1alpha1 "github.com/kro-run/kro/api/v1alpha1"
+	"github.com/kro-run/kro/pkg/testutil/generator"
+)
+
+var _ = Describe("Nested ResourceGraphDefinition", func() {
+	var (
+		ns *corev1.Namespace
+	)
+
+	BeforeEach(func(ctx SpecContext) {
+		ns = &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: fmt.Sprintf("test-%s", rand.String(5)),
+			},
+		}
+		Expect(env.Client.Create(ctx, ns)).To(Succeed())
+	})
+
+	AfterEach(func(ctx SpecContext) {
+		Expect(env.Client.Delete(ctx, ns)).To(Succeed())
+	})
+
+	It("should handle nested ResourceGraphDefinition lifecycle", func(ctx SpecContext) {
+		// Create parent ResourceGraphDefinition
+		rg, genInstance := nestedResourceGraphDefinition("testnestedrg")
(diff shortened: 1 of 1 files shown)


jakobmoellerdev (commented):
@a-hilaly maybe take a look here again and rebase close as needed

a-hilaly:
PR is rebased and tests are passing (at least locally) @jakobmoellerdev

test/integration/suites/core/nested_test.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+var _ = Describe("Nested ResourceGraphDefinition", func() {
+	var (
+		ctx       context.Context
+		namespace string
+	)
+
+	BeforeEach(func() {
+		ctx = context.Background()
```
jakobmoellerdev: I think it is best to not use context.Background for context init in Ginkgo. Instead use SpecContext and use the ginkgo nodes properly https://onsi.github.io/ginkgo/#spec-timeouts-and-interruptible-nodes

test/integration/suites/core/nested_test.go
```diff
@@ -0,0 +1,279 @@
...
+		namespace = fmt.Sprintf("test-%s", rand.String(5))
```
jakobmoellerdev: this can fail with improper characters, no?

test/integration/suites/core/nested_test.go
```diff
@@ -0,0 +1,279 @@
...
+		Expect(env.Client.Create(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: namespace,
+			},
+		})).To(Succeed())
+	})
+
+	It("should handle nested ResourceGraphDefinition lifecycle", func() {
+		ctx := context.Background()
+		namespace := fmt.Sprintf("test-%s", rand.String(5))
```
jakobmoellerdev: looks like oyu are just recreating another namespace even though you have one in before each

test/integration/suites/core/nested_test.go
```diff
@@ -0,0 +1,279 @@
...
+			},
+		}
+		Expect(env.Client.Create(ctx, ns)).To(Succeed())
+
+		// Create parent ResourceGraphDefinition
+		rg, genInstance := nestedResourceGraphDefinition("testnestedrg")
+		Expect(env.Client.Create(ctx, rg)).To(Succeed())
+
+		// Wait for parent ResourceGraphDefinition to be ready
+		Eventually(func(g Gomega) {
```
jakobmoellerdev, line 58: all eventually blocks should use WithContext() to be safely cancellable

test/integration/suites/core/nested_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+		// Delete parent ResourceGraphDefinition
+		Expect(env.Client.Delete(ctx, rg)).To(Succeed())
+
+		// Cleanup namespace
+		Expect(env.Client.Delete(ctx, ns)).To(Succeed())
+	})
+
+	It("should dynamically create RGDs with different schema field types", func() {
+		ctx := context.Background()
```
jakobmoellerdev: override on ctx and namespace again

test/integration/suites/core/nested_test.go
a-hilaly: It uses alphanum without vowels => `bcdfghjklmnpqrstvwxz2456789`. It's used all over the place and i'venever seen an issue with it.

test/integration/suites/core/nested_test.go
a-hilaly: TIL, i think we should address this seperately, many other tests are using ctx.Background()

jakobmoellerdev: Hmmm, TBH I would like to not merge more tech debt than we have already in. If you wanna readdress other tests in a separate PR (I can take care of this) thats totally fine, but adding more tests with this for me is not really helpful. You can swap this describe block over without any other tests / runs being affected I believe

test/integration/suites/core/nested_test.go
a-hilaly: Similar i think we need a dedicated PR addressing this problem in all the existing tests

test/integration/suites/core/nested_test.go
a-hilaly: good catch!

test/integration/suites/core/nested_test.go
a-hilaly: Fair enough, let's raise the bar :)

a-hilaly:
thanks for the review @jakobmoellerdev - I addressed all your comments

jakobmoellerdev (approved):
/lgtm

jakobmoellerdev:
I think this PR is flaky as the parent doesnt seem to get into active state after 30s everytime: https://github.com/kubernetes-sigs/kro/actions/runs/18093052611/job/51477908854 I am gonna try to remediate this by increasing timeouts but that should likely be revisited. 
EDIT: Even after 60s we dont run green here so I believe something is broken.

PR #266, opened 2025-02-03
Adding instructions on how to start kro in kind with ko locally

n3wscott:
I did not see a documented way to use `KinD` locally, so figured it out and I made a page to help anyone to do this too.


website/docs/examples/kind/01-Installation.md
```diff
@@ -0,0 +1,92 @@
...
+   
+   > _Note_, if not using the default kind cluster name, set KIND_CLUSTER_NAME
+
+   ```sh
+   export KIND_CLUSTER_NAME=my-other-cluster
+   ```
+4. Apply the Kro CRDs.
+
+   ```sh
+   make manifests
```
barney-s: Do you think setting DOCKER_REPO to kind.local allows us to build and run controller as a pod in kind ? 

https://github.com/kro-run/kro/blob/main/CONTRIBUTING.md#setting-up-a-local-development-environment

n3wscott: yes, Some people are a fan of running the controller outside the cluster like the link, but personally my dev loop is inside a cluster where I apply the controller into the cluster, even if it is kind. I prefer a real external cluster to do dev tasks though and in that case the only difference is that `KO_DOCKER_REPO` is an external registry but all the other flows are the same.

website/docs/examples/kind/01-Installation.md
a-hilaly, line 1: I'm thinking if we should move this to a dedicated section (in /docs) for contributions/local dev setup instead of putting it in /examples?

helm/templates/deployment.yaml
```diff
@@ -33,7 +33,11 @@ spec:
         - name: {{ .Chart.Name }}
           securityContext:
             {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+          {{- if .Values.image.ko }}
+          image: "ko://github.com/kro-run/kro/cmd/controller"
```
a-hilaly, line 37: interesting, TIL.

a-hilaly (commented):
Thank you @n3wscott !! i have one tiny comment below

website/docs/examples/kind/01-Installation.md
n3wscott: I was thinking it really should go in a developers tab, but you tell me. Happy to do whatever.

helm/templates/deployment.yaml
n3wscott: oh if you are new to ko, buckle up. It is going to change your life.

website/docs/examples/kind/01-Installation.md
barney-s: Given similar comments below, would you mind adding a part of the blurb in CONTRIBUTION.md as well ?

website/docs/examples/kind/01-Installation.md
barney-s: IMHO: adding a separate tab in website, it may be a bit early. That website is currently focussing on users and not developers. Developer docs can reside in the repo itself.

website/docs/examples/kind/01-Installation.md
n3wscott: how about I move it all to `CONTRIBUTION.md`? I don't care really where it ends up. we can always move it to the website when the dev effort gets bigger.

barney-s: SGTM. would be a good start.

barney-s (dismissed):

n3wscott:
Thanks for the feedback, moved the steps to the `CONTRIBUTING.md` file, I had not noticed that one. PTAL

website/docs/examples/kind/01-Installation.md
n3wscott: Moved to CONTRIBUTING.md

a-hilaly (approved):
Thank you Scott :)

PR #273, opened 2025-02-05
Add support for external references

### LINKED ISSUE #72: Mega Feature: Add support for external references
Add ability to reference fields from external to the ResourceGroups resources that exist in the cluster.

For example, reference `data` from an existing ConfigMap: 

Sample ConfigMap:
```bash
❯ k get configmap this -o yaml
apiVersion: v1
kind: ConfigMap
data:
  awsAccountID: "111122223333"
  clusterName: test-cluster
  eksOIDC: oidc.eks.us-east-1.amazonaws.com/id/EXAMPLE0123456789
  region: us-east-1
  subnetIDs: subnet-1111aaaa2222bbbbb,subnet-1111aaaa2222bbbbb,subnet-1111aaaa2222bbbbb
  vpcID: <vpc-1111aaaa2222bbbbb>
```
Option 1: Import an external resource explicitly without taking over control of it and reference values from other resources. Import by one of three options 1/ `name`, 2/ `labels`, 3/ `annotations`
```yaml
spec:
  schema:
    apiVersion: v1alpha1
    kind: s3podidentity
    spec:
...
  resources: 
  - name: externalConfigmap
    externalRef: 
      apiVersion: v1
      kind: ConfigMap
      metadata:
          namespace: default
          name: my-config-map-name
          labels:
             prod: my-prod-label
          annotations:
             prod: my-prod-annotation
....
  - name: anyResource
    var: ${externalConfigmap.data.region}
```    

Option 2: just reference it inline.
```yaml
...
spec:
  schema:
    apiVersion: v1alpha1
    kind: s3podidentity
    spec:
      name: string
      namespace: string | default="default"
      region: string | default=${externalref.configmap.default.this.data.region}
      clusterName: string | default=${externalref.configmap(default, name).data.clusterName}
```
Option 1 contains more lines of yaml but is readable and easier to implement hence it is the preferred option.

barney-s:
- Add ReadOnly flag to resource. True indicates an external resource we want to read (no create/update/delete)
- Reuse template  to describe the external resource (gvkn). Reason: we could templatize the name part of the external resource
- Use as much of existing resource reconcile flow as possible. Reason: we can optionally support ReadyWhen and IncludeWhen rules for the external resource

Fixes #72


pkg/controller/instance/controller_reconcile.go
```diff
@@ -177,6 +177,12 @@ func (igr *instanceGraphReconciler) handleResourceReconciliation(
 	observed, err := rc.Get(ctx, resource.GetName(), metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			// For read-only resources, we don't create
+			if igr.runtime.ResourceDescriptor(resourceID).IsReadOnly() {
```
matthchr: Can we make sure that the user is only setting name/namespace (+ the resource types).

pkg/controller/instance/controller_reconcile.go
```diff
@@ -142,9 +142,9 @@ func (igr *instanceGraphReconciler) reconcileResource(ctx context.Context, resou
 	resourceState := &ResourceState{State: "IN_PROGRESS"}
 	igr.state.ResourceStates[resourceID] = resourceState
 
-	// Check if resource should be created
-	if want, err := igr.runtime.WantToCreateResource(resourceID); err != nil || !want {
-		log.V(1).Info("Skipping resource creation", "reason", err)
+	// Check if resource should be processed (create or get)
+	if want, err := igr.runtime.ReadyToProcessResource(resourceID); err != nil || !want {
```
matthchr, line 161: I do sorta wonder if we want something _other_ than template (but which we also template?) `externalRef` that only has `name`/`namespace`/`kind`/`apiversion`.

I am also fine keeping it `template` as well, I don't feel super strongly about the above.

api/v1alpha1/types.go
```diff
@@ -90,6 +90,9 @@ type Resource struct {
 	ID string `json:"id,omitempty"`
 	// +kubebuilder:validation:Required
 	Template runtime.RawExtension `json:"template,omitempty"`
+	// ReadOnly indicates an external resource we want to read and use in the Graph
+	// +kubebuilder:validation:Optional
+	ReadOnly bool `json:"readOnly,omitempty"`
```
a-hilaly: There are two other approaches that we can consider for this feature:
1/ use the `externalRef: true` (readonly insinuates that there will be updateOnly, createOnly, and why not a deleteOnly
2/ having something similar to `Template`, but `ExternalRef`. Where in external ref we only allow metadata.name/namespacec kind and apiVersion? and maybe eventually support a `labelSelector`

pkg/controller/instance/controller_reconcile.go
```diff
@@ -142,9 +142,9 @@ func (igr *instanceGraphReconciler) reconcileResource(ctx context.Context, resou
 	resourceState := &ResourceState{State: "IN_PROGRESS"}
 	igr.state.ResourceStates[resourceID] = resourceState
```
a-hilaly, line 159: How about making sure that all externalReferences ar available before kicking off the dag execution?

a-hilaly (commented):
Thanks @barney-s ! i have two suggestions/question bellow

pkg/controller/instance/controller_reconcile.go
barney-s: I did not want to do that because we may have an externalRef that is being materialized as a side-effect of the dagExecution. Indirectly created resource.

pkg/controller/instance/controller_reconcile.go
barney-s: I can add a validation that allows only name/namespace/kind/apiVersion in the template field. That way we keep the definition tight and set the correct expectation as well. WDYT ?

pkg/controller/instance/controller_reconcile.go
barney-s: can do.

api/v1alpha1/types.go
barney-s: i like `externalRef` instead of `readOnly`. FYI  i believe we should support `updateOnly`, `createOnly` for some use cases. We can deal with this as the need arises.

barney-s: done.

pkg/controller/instance/controller_reconcile.go
barney-s: added

pkg/controller/instance/controller_reconcile.go
barney-s: will make it a typed struct parallel to template

barney-s:
@a-hilaly  @matthchr - Would you please take a look at this again. I have addressed the review comments.

api/v1alpha1/types.go
```diff
@@ -86,12 +86,31 @@ type Validation struct {
...
+	Kind string `json:"kind,omitempty"`
+	// +kubebuilder:validation:Required
+	Version string `json:"version,omitempty"`
+	// +kubebuilder:validation:Required
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Optional
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// +kubebuilder:validation:XValidation:rule="(has(self.template) && !has(self.externalRef)) || (!has(self.template) && has(self.externalRef))",message="exactly one of template or externalRef must be provided"
```
a-hilaly, line 103: Neat! can we have an integration test validating this?

pkg/graph/builder.go
```diff
@@ -250,21 +250,44 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
...
-	err := yaml.UnmarshalStrict(rgResource.Template.Raw, &resourceObject)
-	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal resource %s: %w", rgResource.ID, err)
+	if len(rgResource.Template.Raw) > 0 {
+		err := yaml.UnmarshalStrict(rgResource.Template.Raw, &resourceObject)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal resource %s: %w", rgResource.ID, err)
+		}
+	} else if rgResource.ExternalRef != nil {
+		b.buildExternalRefResource(rgResource.ExternalRef, &resourceObject)
```
a-hilaly: i do think we're still missing emulation here. We do need integration tests validating that CEL expressiosn that reference an "externalRef" are working.

api/v1alpha1/types.go
```diff
@@ -86,12 +86,31 @@ type Validation struct {
...
+// ExternalRef is a reference to an external resource.
+// It allows the user to specify the Kind, Version, Name and Namespace of the resource
+// to be read and used in the Graph.
+type ExternalRef struct {
+	// +kubebuilder:validation:Optional
+	Group string `json:"group,omitempty"`
+	// +kubebuilder:validation:Required
+	Kind string `json:"kind,omitempty"`
+	// +kubebuilder:validation:Required
+	Version string `json:"version,omitempty"`
```
a-hilaly: i'm curious here, why do we seperate Group and version? `template` does use `apiVersion`, maybe we can keep it consistent?

api/v1alpha1/types.go
```diff
@@ -86,12 +86,31 @@ type Validation struct {
...
+	// +kubebuilder:validation:Required
+	Name string `json:"name,omitempty"`
```
a-hilaly: nit: you do not need omitempty when a field is required (and i might have missed that as well in other fields/structs)

api/v1alpha1/types.go
barney-s: Would you point me to where this test can be inserted ?

api/v1alpha1/types.go
barney-s: makes sense. done.

api/v1alpha1/types.go
barney-s: done

barney-s:
> Thanks @barney-s ! I left a few comments/questions bellow

PTAL @a-hilaly

joshuabaird:
@barney-s Could we potentially include an [example](https://github.com/kro-run/kro/tree/main/examples) of using an external reference in this PR?

barney-s:
> @barney-s Could we potentially include an [example](https://github.com/kro-run/kro/tree/main/examples) of using an external reference in this PR?

Done. #537

test/integration/suites/core/externalref_test.go
a-hilaly, line 1: Really cool, thank you for adding this!

pkg/testutil/generator/resourcegraphdefinition.go
```diff
@@ -68,6 +68,24 @@ func WithSchema(kind, version string, spec, status map[string]interface{}) Resou
...
+) ResourceGraphDefinitionOption {
+	return func(rgd *krov1alpha1.ResourceGraphDefinition) {
+		rgd.Spec.Resources = append(rgd.Spec.Resources, &krov1alpha1.Resource{
+			ID:          id,
+			ReadyWhen:   readyWhen,
+			IncludeWhen: includeWhen,
+			ExternalRef: externalRef,
+		})
+	}
+}
```
a-hilaly, line 87: 👍

pkg/controller/instance/controller_reconcile.go
```diff
@@ -369,6 +381,18 @@ func (igr *instanceGraphReconciler) deleteResourcesInOrder(ctx context.Context)
...
+		if igr.runtime.ResourceDescriptor(resourceID).IsExternalRef() {
+			igr.state.ResourceStates[resourceID].State = "SKIPPED"
+			continue
+		}
+
+		// Skip deletion for read-only resources
+		if igr.runtime.ResourceDescriptor(resourceID).IsExternalRef() {
+			igr.state.ResourceStates[resourceID].State = "SKIPPED"
+			continue
+		}
```
a-hilaly: repeated block?

pkg/controller/instance/controller_reconcile.go
```diff
@@ -369,6 +381,18 @@ func (igr *instanceGraphReconciler) deleteResourcesInOrder(ctx context.Context)
 			continue
 		}
 
+		// Skip deletion for read-only resources
+		if igr.runtime.ResourceDescriptor(resourceID).IsExternalRef() {
+			igr.state.ResourceStates[resourceID].State = "SKIPPED"
```
a-hilaly: shall we use enums since this patch is merged https://github.com/kro-run/kro/blob/main/pkg/controller/instance/controller_reconcile.go#L35-L47

pkg/graph/builder.go
```diff
@@ -250,21 +250,40 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
 	return resourceGraphDefinition, nil
 }
 
+func (b *Builder) buildExternalRefResource(externalRef *v1alpha1.ExternalRef, resourceObject *map[string]interface{}) {
```
a-hilaly: can you please add a comment stating what are the responsibilities of this function?

a-hilaly: also i think this function is just easier to read through if it was returning a `map[string]interface{}`

pkg/graph/builder.go
```diff
@@ -250,21 +250,40 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
...
+	(*resourceObject)["metadata"] = metadata
+}
+
 // buildRGResource builds a resource from the given resource definition.
 // It provides a high-level understanding of the resource, by extracting the
 // OpenAPI schema, emulating the resource and extracting the cel expressions
 // from the schema.
-func (b *Builder) buildRGResource(rgResource *v1alpha1.Resource, namespacedResources map[k8sschema.GroupKind]bool, order int) (*Resource, error) {
+func (b *Builder) buildRGResource(rgResource *v1alpha1.Resource,
+	namespacedResources map[k8sschema.GroupKind]bool, order int) (*Resource, error) {
```
a-hilaly: nit (just keep the style consistent with the rest of the code base):
```suggestion
func (b *Builder) buildRGResource(
    rgResource *v1alpha1.Resource,
	namespacedResources map[k8sschema.GroupKind]bool,
	order int,
) (*Resource, error) {
```

a-hilaly (commented):
Thanks Barni! i left a few suggestions bellow

pkg/controller/instance/controller_reconcile.go
barney-s: looks like a bad merge resolution. fixed.

pkg/controller/instance/controller_reconcile.go
barney-s: missed it during merge.

pkg/graph/builder.go
barney-s: done.

barney-s:
@a-hilaly PTAL. Its good to be merged now.

a-hilaly (approved):
Thanks a lot @barney-s !
cc @matthchr wanna take a final look?

PR #275, opened 2025-02-05
Add helm support for existing controller flags

Gonfidel:
The current helm chart values file is missing multiple controller argument flags. This merge requests adds the missing flags as options in the chart.

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -57,6 +57,18 @@ spec:
               value: {{ .Values.config.dynamicControllerConcurrentReconciles | quote }}
             - name: KRO_LOG_LEVEL
               value: {{ .Values.config.logLevel | quote }}
+            - name: KRO_DYNAMIC_CONTROLLER_DEFAULT_RESYNC_PERIOD
+              value: {{ .Values.config.dynamicControllerDefaultResyncPeriod | quote }}
+            - name: KRO_DYNAMIC_CONTROLLER_DEFAULT_QUEUE_MAX_RETRIES
+              value: {{ .Values.config.dynamicControllerDefaultQueueMaxRetries | quote }}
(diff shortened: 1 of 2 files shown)


helm/templates/deployment.yaml
```diff
@@ -57,6 +57,18 @@ spec:
               value: {{ .Values.config.dynamicControllerConcurrentReconciles | quote }}
             - name: KRO_LOG_LEVEL
               value: {{ .Values.config.logLevel | quote }}
+            - name: KRO_DYNAMIC_CONTROLLER_DEFAULT_RESYNC_PERIOD
```
n3wscott, line 60: why set both the args and env vars?

nic-george: Transparently, I was following the pre-existing pattern but I'm guessing it's for cases where `helm template` is used in tandem another rendering tool. It would allow the flexibility to override the ENV without replacing the args block.

Open to changing it if we prefer a one-to-one relationship.

a-hilaly: That, and it's a common pattern I've seen in quite a few projects to let users tweak args by overriding env vars with `kubectl set env` 😄

a-hilaly (approved):
Thank you very much @Gonfidel !

helm/templates/deployment.yaml
n3wscott: interesting. But with this setup, the `kubectl set env` var is ignored if the binary is using a tool like viper:

> Viper uses the following precedence order. Each item takes precedence over the item below it:
> - explicit call to Set
> - flag
> - env
> - config
> - key/value store
> - default

https://github.com/spf13/viper

a-hilaly: Ah yeah good point about the Viper precedence!

a-hilaly: For now though I think this setup works - maybe we can do a follow up PR to align the env var handling? Don't want to block this one.

PR #278, opened 2025-02-06
Update(01-Installation.md): Helm login ghcr

tzahimizrahi:


website/docs/docs/getting-started/01-Installation.md
```diff
@@ -13,6 +13,7 @@ Before you begin, ensure you have the following:
 
 1. `Helm` 3.x installed
 2. `kubectl` installed and configured to interact with your Kubernetes cluster
+3. Login to ghcr.io
```
n3wscott: it is not an anonymous pull?

n3wscott:
What issue did you bump into that needed to login to `ghcr.io`?

website/docs/docs/getting-started/01-Installation.md
a-hilaly: Right - i think users who previously used helm to auth to private registries can run into this issue. We can add a note in the docs explaining that running `helm registry logout ghcr.io` usually fixes it.

PR #279, opened 2025-02-07
Update dev docs and Makefile 

barney-s:
- Create a docs/ folder and add dev getting started
- move contents from contribution.md to getting started doc
- add make targets to make it easy to build and deploy local code to local kind cluster


barney-s:
@n3wscott PTAL

Makefile
```diff
@@ -199,4 +211,30 @@ publish-helm: ## Helm publish
...
+
+.PHONY: deploy-kind
+deploy-kind:
+	$(KIND) delete clusters ${KIND_CLUSTER_NAME} || true
+	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
+	$(KUBECTL) create namespace kro-system
+	make install
+	make build-image # TODO: how do you kind.local here ?
+	kind load docker-image ${KO_DOCKER_REPO}:${RELEASE_VERSION} --name ${KIND_CLUSTER_NAME}
+	helm template kro ./helm --namespace kro-system --set image.pullPolicy=Never --set image.repository=${KO_DOCKER_REPO} --set image.tag=${RELEASE_VERSION} | $(KUBECTL) apply -f -
```
n3wscott: kind sets and overwrites the context, but to be consistent, I would add `--context kind-${KIND_CLUSTER_NAME}` to this line, or remove it from the next line.

Makefile
```diff
@@ -199,4 +211,30 @@ publish-helm: ## Helm publish
...
+	sleep 5
+	kubectl --context kind-${KIND_CLUSTER_NAME} get pods -A
```
n3wscott: nit: needs a newline at the end of the file.

docs/developer-getting-started.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Running the controller inside a [`KinD`][kind] cluster with [`ko`][ko]
+
+[ko]: https://ko.build
+[kind]: https://kind.sigs.k8s.io/
+
+Use the make target to (re)create a kind cluster and build and deploy local code to it.
+
+   ```sh
+   KIND_CLUSTER_NAME=kro make deploy-kind
```
n3wscott: personally, I would show the real commands and then also say there is some magic in the make file. I personally dislike makefiles because it hides what is happening and the syntax is just ever so different from bash. I would take a bash script all day over a makefile.

docs/developer-getting-started.md
```diff
@@ -0,0 +1,89 @@
...
+    go run ./cmd/controller --log-level 2
+    ```
+    This will connect to the default Kubernetes context in your local kubeconfig (`~/.kube/config`). Ensure the context is pointing to your local cluster.
+
+### Running the controller inside a [`KinD`][kind] cluster with [`ko`][ko]
+
+[ko]: https://ko.build
+[kind]: https://kind.sigs.k8s.io/
+
+Use the make target to (re)create a kind cluster and build and deploy local code to it.
```
n3wscott: In most of my dev loops, I absolutely do not want to delete the cluster and start over. I want to create a bunch of resources and then have the next install of the controller do whatever feature I am working on. If I have to also get back to where I was that slows my process down. I think it would be better to call out and teach the Kro developers how to loop effectively vs teaching the "blow it away and start again" way every time.

Makefile
```diff
@@ -1,15 +1,13 @@
 
-AWS_ACCOUNT_ID ?= $(shell aws sts get-caller-identity --query Account --output text)
-AWS_REGION ?= us-west-2
 RELEASE_VERSION ?= dev-$(shell git rev-parse --short HEAD)
 OCI_REPO ?= ghcr.io/kro-run/kro
```
n3wscott, line 3: This is likely a bad default, the only thing that can push to that target is the automation.

Makefile
```diff
@@ -151,14 +149,26 @@ $(LOCALBIN):
 
 ## Tool Binaries
 KUBECTL ?= kubectl
+KIND ?= kind
 KUSTOMIZE ?= $(LOCALBIN)/kustomize
+KO ?= $(LOCALBIN)/ko
```
n3wscott, line 154: will this work with my ko install via brew?

Makefile
```diff
@@ -151,14 +149,26 @@ $(LOCALBIN):
...
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
+KO_VERSION ?= v0.17.1
 KUSTOMIZE_VERSION ?= v5.2.1
```
n3wscott, line 160: nit: Kustomize comes with `kubectl`. No need to install it independently.

Makefile
```diff
@@ -199,4 +211,30 @@ publish-helm: ## Helm publish
...
+uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config
+	$(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f ./helm/crds
+
+.PHONY: deploy-kind
+deploy-kind:
+	$(KIND) delete clusters ${KIND_CLUSTER_NAME} || true
+	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
+	$(KUBECTL) create namespace kro-system
+	make install
+	make build-image # TODO: how do you kind.local here ?
```
n3wscott: re:TODO, don't build the image like this here, do it on the line 238: `helm template | kubectl apply` line, change `kubectl apply` to `ko apply`

The whole point of using ko in this way with kind is to use the node image loading feature of ko and not need a secondary registry. Sure you can get it to work other ways too and really ko does not care, just set the correct `KO_DOCKER_REPO` and you are gtg. 

Another reason why I don't like this kind of stuff inside the make file, it is too complex to expose all the ways you might want to do this.

Makefile
n3wscott: I added the helm template ko support in the last PR to allow for 


   ```sh
    helm template kro ./helm \
      --namespace kro-system \
      --set image.pullPolicy=Never \
      --set image.ko=true | ko apply -f -
    ```

docs/developer-getting-started.md
barney-s: Sounds good. Will add back the commands.

Makefile
barney-s: It will not use the brew installed ko. It will use local bin/ko

Makefile
barney-s: I wasn’t sure how that works. Will try that and update it here.

Makefile
barney-s: Will try that.

barney-s: Does ko add the image using kind load image

docs/developer-getting-started.md
barney-s: added text and note to use this make target for getting started or in case you want to start over.

docs/developer-getting-started.md
barney-s: done.

Makefile
barney-s: for lack of a better alternative leaving it as is. I tried pushing to this and could not due to authn. So i guess its ok to leave it in.

barney-s: resolving this convo. Happy to take other suggestions.

Makefile
barney-s: apparently it does.

barney-s: done.

Makefile
barney-s: this is a sideeffect of kubebuilder tooling. Also there was a brief period where in the kustomize files generated by a kubebuilder version were dependent on specific kustomize version.  Not sure if is a remnant from that time or if it is still valid.

Makefile
```diff
@@ -1,15 +1,13 @@
...
 RELEASE_VERSION ?= dev-$(shell git rev-parse --short HEAD)
 OCI_REPO ?= ghcr.io/kro-run/kro
 
-CONTROLLER_IMAGE ?= ${OCI_REPO}/controller:${RELEASE_VERSION}
 HELM_IMAGE ?= ${OCI_REPO}
 KO_DOCKER_REPO ?= ${OCI_REPO}/kro
 
 KOCACHE ?= ~/.ko
 KO_PUSH ?= true
+export KIND_CLUSTER_NAME ?= kro
```
a-hilaly, line 10: is the `export` needed here?

Makefile
a-hilaly: just wondering - what's the benefit of using a local ko binary vs the brew installed one?

Makefile
```diff
@@ -151,14 +149,26 @@ $(LOCALBIN):
...
 
 ## Tool Versions
+KO_VERSION ?= v0.17.1
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.16.2
 
+.PHONY: ko
+ko: $(KO) ## Download ko locally if necessary. If wrong version is installed, it will be removed before downloading.
+$(KO): $(LOCALBIN)
+	@if test -x $(LOCALBIN)/ko && ! $(LOCALBIN)/kustomize version | grep -q $(KO_VERSION); then \
```
a-hilaly:
```suggestion
	@if test -x $(LOCALBIN)/ko && ! $(LOCALBIN)/ko version | grep -q $(KO_VERSION); then \
```

Makefile
```diff
@@ -175,15 +185,17 @@ $(CONTROLLER_GEN): $(LOCALBIN)
...
 
 .PHONY: image
-build-image: ## Build the kro controller images using ko build
-	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) \
-		ko build --bare github.com/kro-run/kro/cmd/controller \
+build-image: ko ## Build the kro controller images using ko build
+	echo $(RELEASE_VERSION)
+	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) KO_DOCKER_REPO=$(KO_DOCKER_REPO) \
+		$(KO) build --bare github.com/kro-run/kro/cmd/controller \
+		--local\
```
a-hilaly, line 192: QQ about the `--local` flag in build-image rule for publishing official kro images - can this affect our build/release pipeline? https://github.com/kro-run/kro/blob/main/.github/workflows/build-push-image.yaml#L62-L66 (we don't have any docker daemon running in those jobs AFAIK)

Makefile
```diff
@@ -175,15 +185,17 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
 
 .PHONY: image
-build-image: ## Build the kro controller images using ko build
-	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) \
-		ko build --bare github.com/kro-run/kro/cmd/controller \
+build-image: ko ## Build the kro controller images using ko build
+	echo $(RELEASE_VERSION)
+	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) KO_DOCKER_REPO=$(KO_DOCKER_REPO) \
```
a-hilaly, line 190: `KO_DOCKER_REPO` is already defined in L6
```suggestion
	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) \
```

Makefile
```diff
@@ -175,15 +185,17 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
 
 .PHONY: image
-build-image: ## Build the kro controller images using ko build
-	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) \
-		ko build --bare github.com/kro-run/kro/cmd/controller \
+build-image: ko ## Build the kro controller images using ko build
+	echo $(RELEASE_VERSION)
```
a-hilaly: nit:
```suggestion
	echo "Building kro image $(RELEASE_VERSION).."
```

a-hilaly (commented):
Thanks @barney-s ! Had a couple questions/comments below~

Makefile
barney-s: workflow - e2e testing. Please look at https://github.com/kro-run/kro/actions/runs/13272926960/job/37056355649?pr=290

Makefile
barney-s: Yes needed for KO, https://ko.build/configuration/

Makefile
barney-s: This sets the env variable from make variable. Alternatively we need to export it.

Makefile
barney-s: done

Makefile
barney-s: It is optional. The local is required to work with kind. 

Doesnt affect the build-push image workflow: https://github.com/kro-run/kro/actions/runs/13280535130/job/37077790345?pr=279

barney-s:
> Thanks @barney-s ! Had a couple questions/comments below~

Thanks. Addressed the comments. PTAL @a-hilaly

Makefile
```diff
@@ -199,4 +211,31 @@ publish-helm: ## Helm publish
...
+deploy-kind: export KO_DOCKER_REPO=kind.local
+deploy-kind:
+	$(KIND) delete clusters ${KIND_CLUSTER_NAME} || true
+	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
+	$(KUBECTL) --context kind-$(KIND_CLUSTER_NAME) create namespace kro-system
+	make install
+	# This generates deployment with ko://... used in image. 
+	# ko then intercepts it builds image, pushes to kind node, replaces the image in deployment and applies it
+	helm template kro ./helm --namespace kro-system --set image.pullPolicy=Never --set image.ko=true | $(KO) apply -f -
+	sleep 5
```
matthchr: Is this sleep waiting for the pod to come up?

You can actually do that wait in kubectl instead? Something like this:
`kubectl wait --for=condition=ready --timeout=3m pod -n "$OPERATOR_NAMESPACE" -l control-plane=controller-manager`

You'll need to modify this based on the labels on the KRO pod though.

barney-s: done

a-hilaly (approved):
Nice, thanks @barney-s !

Issue #185, opened 2025-01-09
label CRDs with owning ResourceGroup information


a-hilaly:
### Feature Description

Currently when kro creates CRDs from ResourceGroups, these CRDs lack metadata about their origin. ideally we to  automatically label CRDs with their owning `ResourceGroup` information. This allows kro to track which ResourceGroup owns which CRD, enabling better conflict detection and resource management. 

We can leverage the metadata library that we already use for sub resources https://github.com/awslabs/kro/blob/main/pkg/metadata/labels.go

gerardocorea:
Instead of labels should owner references be set on the child resources so should the parent resource be deleted the child ones will also be cleaned up? i wouldnt mind looking into this.

nic-george:
I have created MR #282 in response to this issue. Please let me know if this meets the issues requirements!

Jayanta111:
@a-hilaly 
Can you assign me for this issue

Sanjeevliv:
May I know if the issue is completely solved or something needs to be done as its still open

a-hilaly:
This was added in https://github.com/kro-run/kro/pull/282

PR #282, opened 2025-02-08
Implement KRO Labels For RGD CRDs

Gonfidel:
In relation to issue #185, we should add KRO management labels to CRD's created via ResourceGraphDefinitions.

Example label output from `example/webapp/rg.yaml`
```
metadata:
  creationTimestamp: "2025-02-08T04:53:40Z"
  generation: 1
  labels:
    kro.run/controller-pod-id: kro-pod
    kro.run/kro-version: 0.2.1
    kro.run/owned: "true"
    kro.run/resource-graph-definition-id: f25ffd07-df0c-4c7e-b90e-507c80d317fc
    kro.run/resource-graph-definition-name: webapp.kro.run
  name: webapps.kro.run
  resourceVersion: "33637"
  uid: 0e3fdd78-bafe-4bca-b44e-9cb6e23dce58
```

### DIFF
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -117,7 +117,6 @@ func NewResourceGraphDefinitionLabeler(rgMeta metav1.Object) GenericLabeler {
 	return map[string]string{
 		ResourceGraphDefinitionIDLabel:        string(rgMeta.GetUID()),
 		ResourceGraphDefinitionNameLabel:      rgMeta.GetName(),
-		ResourceGraphDefinitionNamespaceLabel: rgMeta.GetNamespace(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/metadata/labels.go
```diff
@@ -117,7 +117,6 @@ func NewResourceGraphDefinitionLabeler(rgMeta metav1.Object) GenericLabeler {
 	return map[string]string{
 		ResourceGraphDefinitionIDLabel:        string(rgMeta.GetUID()),
 		ResourceGraphDefinitionNameLabel:      rgMeta.GetName(),
-		ResourceGraphDefinitionNamespaceLabel: rgMeta.GetNamespace(),
```
nic-george, line 120: Is there a reason we need the namespace label for ResourceGraphDefinitions since they're cluster scoped?

a-hilaly (commented):
Neat! Thanks a lot @Gonfidel !
/lgtm

a-hilaly (approved):

pkg/metadata/labels.go
a-hilaly: it's just a leftover from when RGDs were namespace-scoped `ResourceGroups` :)

Issue #283, opened 2025-02-08
RGD - Failing on field type parsing


khalilj-appsflyer:
### Description

**Observed Behavior**: Defining a resource with below resource definition fails on: "expected number type for path renotifyInterval, got int64"
`optionsList:
- monitorOptions:
  - renotifyInterval: 0
`

**Expected Behavior**: The resource definition should be parsed successfully

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files): The following GRD when applied should the mentioned error on its description:
`
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: synthtestwithslo.kro.run
spec:
  schema:
    apiVersion: v1alpha1
    kind: SynthTestWithSlo
    spec:
      name: string
      namespace: string | default=default
  resources:
  - id: test
    template:
      apiVersion: synthetics.datadog.upbound.io/v1alpha1
      kind: Test
      metadata:
        name: ${schema.spec.name}-test
        labels:
          app.kubernetes.io/name: ${schema.spec.name}
      spec:
        forProvider:
          assertion:
          - operator: lessThan
            target: "1000"
            type: responseTime
          locations:
          - pl:platform-private-location-7997d828d18ffe96cbd5c7ba76f198da
          message: "test for xyz"
          name: my-test
          requestHeaders: {}
          requestDefinition:
          - method: GET
            url: https://www.google.com
          tags:
          - service:MyService
          status: live
          subtype: http
          type: api
          optionsList:
          - monitorOptions:
            - renotifyInterval: 0
            retry:
            - count: 3
              interval: 60
            tickEvery: 60
            monitorPriority: 2
`

**Versions**:
- kro version: 0.2.1
- Kubernetes Version (`kubectl version`):
Client Version: v1.30.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.31.4-eks-2d5f260

**Involved Controllers**:
- Controller URLs and Versions (if applicable): crossplane DataDog provider. https://github.com/upbound/provider-datadog

**Error Logs** (if applicable)**:
```
Paste any relevant logs here
[...]

PR #284, opened 2025-02-08
fix: improve number type validation to support int values

khalilj:
This PR solves issue #283 
Tested using local Kro development environment.

### DIFF
--- a/pkg/graph/parser/parser.go
+++ b/pkg/graph/parser/parser.go
@@ -53,16 +53,24 @@ func parseResource(resource interface{}, schema *spec.Schema, path string) ([]va
 		return nil, err
 	}
 
+	fmt.Printf("resource: %v\n", resource)
+	fmt.Printf("expectedTypes: %v\n", expectedTypes)
+
 	switch field := resource.(type) {
 	case map[string]interface{}:
+		fmt.Printf("map field: %v\n", field)
 		return parseObject(field, schema, path, expectedTypes)
 	case []interface{}:
+		fmt.Printf("array field: %v\n", field)
 		return parseArray(field, schema, path, expectedTypes)
 	case string:
+		fmt.Printf("string field: %v\n", field)
 		return parseString(field, schema, path, expectedTypes)
 	case nil:
+		fmt.Printf("nill field: %v\n", field)
 		return nil, nil
 	default:
+		fmt.Printf("default field: %v\n", field)
 		return parseScalarTypes(field, schema, path, expectedTypes)
 	}
 }
@@ -214,6 +222,9 @@ func parseString(field string, schema *spec.Schema, path string, expectedTypes [
 		}}, nil
 	}
 
+	fmt.Printf("expectedTypes: %v\n", expectedTypes)
+	fmt.Printf("schemaTypeAny: %v\n", schemaTypeAny)
+
 	if !slices.Contains(expectedTypes, "string") && !slices.Contains(expectedTypes, schemaTypeAny) {
 		return nil, fmt.Errorf("expected string type or AdditionalProperties for path %s, got %v", path, field)
 	}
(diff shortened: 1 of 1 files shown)


pkg/graph/parser/parser.go
```diff
@@ -286,9 +286,22 @@ func getArrayItemSchema(schema *spec.Schema, path string) (*spec.Schema, error)
 	return nil, fmt.Errorf("invalid array schema for path %s: neither Items.Schema nor Properties are defined", path)
 }
 
+func isNumber(v interface{}) bool {
+	return isInteger(v) || ifFloat(v)
+}
+
+func ifFloat(v interface{}) bool {
```
a-hilaly:
```suggestion
func isFloat(v interface{}) bool {
```

a-hilaly:
Thank you @khalilj !

pkg/graph/parser/parser.go
```diff
@@ -236,7 +236,7 @@ func parseScalarTypes(field interface{}, _ *spec.Schema, path string, expectedTy
 	// perform type checks for scalar types
 	switch {
 	case slices.Contains(expectedTypes, "number"):
-		if _, ok := field.(float64); !ok {
+		if !isNumber(field) {
```
a-hilaly, line 239: So if i understand correctly, in openapispecv3, `number` values can be either `int` or `float` in go?

pkg/graph/parser/parser.go
khalilj: Fixed.

pkg/graph/parser/parser.go
khalilj: Actually, Open api v3 number refers to float and double. https://spec.openapis.org/oas/v3.0.0.html#openapi-specification:~:text=The%20formats%20defined%20by%20the%20OAS%20are%3A
Which might make the issue not only related to this method only, because the error i got refers that it expected to get the type 'number' but received an int64: Defining a resource with below resource definition fails on: "expected number type for path renotifyInterval, got int64".
parseScalarTypes func gets 'number' at expectedTypes slice. But the 'field' variable held an int64 type.

khalilj: The open api resource field type is really a number: kubectl get crd tests.synthetics.datadog.upbound.io -o yaml

> renotifyInterval:
>   description: (Number) Specify a renotification frequency
>     in minutes. Values available by default are 0, 10,
>     20, 30, 40, 50, 60, 90, 120, 180, 240, 300, 360,
>     720, 1440. Defaults to 0. Specify a renotification
>     frequency in minutes. Values available by default
>     are `0`, `10`, `20`, `30`, `40`, `50`, `60`, `90`,
>     `120`, `180`, `240`, `300`, `360`, `720`, `1440`.
>     Defaults to `0`.
>   type: number

Which leaves the question, why `resource interface{}` type is int64.

khalilj: @a-hilaly did you get the chance to check this?

a-hilaly: Hey @khalilj ! Thanks for the making the changes. Yep LGTM, i'm just running some tests with float/int to understand more why `resource interface{}` type is int64 - otherwise this is ready to go :)

khalilj: Thanks @a-hilaly

khalilj: Hi @a-hilaly , did you get the chance to test it?

a-hilaly: After a quick deep dive, looks like this is related to the JSON unmarshallers client-go is using https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/serializer/json/json.go - i think this is ready to go. Thanks a lot @khalilj !

khalilj: Great news @a-hilaly , Thanks!

Issue #286, opened 2025-02-10
Callout ACK controllers dependency for the AWS Networking stack example


premdass:
### Documentation Issue

For the AWS Networking stack example, please callout that AWS ACK EC2 controllers need to be running in the clustser for the  VPC, Subnets and security groups to be provisioned.

**Location**:
https://kro.run/examples/aws/ack-networking-stack

**Current State**:
It doesnt talk about the AWS ACK controller dependency

**Suggested Changes**:
Please callout the requirement in https://kro.run/examples/aws/aws-examples

PR #287, opened 2025-02-10
Add callout for ACK controller #286

premdass:
Add callout for ACK controller #286 for the Networking stack example


a-hilaly (approved):
Thank you @premdass !

PR #290, opened 2025-02-11
Add E2E test framework and 2 simple tests

barney-s:


barney-s:
Please look at https://github.com/kubernetes-sigs/e2e-framework and see if we can use it.

barney-s:
@justinsb - Can you please take a look at this PR. Please comment on the approach being used for e2e testing.

barney-s:
@matthchr - Please review this.

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,157 @@
...
+	// DefaultWaitTimeout is the default timeout for waiting operations
+	DefaultWaitTimeout = 2 * time.Minute
+	// DefaultWaitInterval is the default interval for waiting operations
+	DefaultWaitInterval = 5 * time.Second
+)
+
+// Framework provides utilities for e2e testing
+type Framework struct {
+	// KindCluster is the kind cluster provider (optional)
+	KindCluster *KindClusterProvider
```
justinsb, line 39: It looks like the direction here is to have a ClusterProvider interface?

test/e2e/framework/manifest.go
```diff
@@ -0,0 +1,102 @@
...
+		// Decode yaml
+		decoder := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
+		obj := &unstructured.Unstructured{}
+		_, _, err = decoder.Decode(data, nil, obj)
+		if err != nil {
+			return fmt.Errorf("failed to decode yaml %s: %w", path, err)
+		}
+
+		// Create or update resource
+		err = c.Create(ctx, obj)
```
justinsb, line 68: In my experience, switching from Create/Update to SSA is a real pain.  Shall we just use SSA from the get-go?

test/e2e/framework/yaml.go
```diff
@@ -0,0 +1,121 @@
...
+	filename        string
+}
+
+// String returns a string representation of the resource
+func (r *resourceInfo) String() string {
+	return fmt.Sprintf("%s/%s[%s]", r.obj.GetKind(), r.obj.GetName(), r.filename)
+}
+
+// LoadObjectsFromDirectory reads and applies Kubernetes manifests from a directory with proper ordering
+func LoadObjectsFromDirectory(path string) ([]resourceInfo, error) {
```
justinsb, line 57: This feels like a duplicate of the code above that applies the objects?

test/e2e/suites/advanced/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"testing"
+
+	"github.com/kro-run/kro/test/e2e/framework"
+)
+
+var (
+	f *framework.Framework
+)
+
+func TestMain(m *testing.M) {
```
justinsb, line 29: My 2c: it's much easier to understand tests that use a Harness object in the test.  The downside is that it can be slower if we're creating an envtest or kind cluster every time.

test/e2e/suites/advanced/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Option 2: Create a kind cluster. we dont support local image build yet.
+	/*
+		framework.WithKindCluster(
+		framework.WithName("kro-basic-test"),
+		framework.WithWaitTimeout(5*time.Minute),
+	*/
+
+	// Option 3: Do nothing. kubeconfig is loaded from standard path
+	f, err = framework.New()
```
justinsb, line 51: I think this should probably be determined by env-vars or flags.  I prefer env-vars because they flags + tests are confusing, but either way lets us have multiple behaviours and switch between them with a script.

justinsb (commented):
I think maybe this would be much clearer if we started with one simple test, in one simple mode of operation (envtest or pre-existing cluster KUBECONFIG), and then built-up to the full complexity (e.g. I think we can introduce multi-step etc later)

test/e2e/suites/basic/simple_deployment_test.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+func TestSimpleDeploymentRGD(t *testing.T) {
+
+	tc, err := framework.NewTestCase(t, f, "simple-deployment-rgd")
+	require.NoError(t, err)
+
+	// Run the test
+	// Option 1: Run all steps automatically using the manifests to verify
+	tc.RunTest(t, func(ctx context.Context) error {
```
matthchr, line 32: Discussed this with @barney-s and @a-hilaly in a call - my big question(s) on this PR are:
1. What do we want our tests to look like?
2. How do we want to perform assertions?

**Things I like about using YAML for defining the inputs:**
1. It's what most users will be doing
2. You can kubectl apply it easily
3. It serves as an additional sample for users

**Things I don't like about using YAML for inputs:**
1. It splits the test up, with the Go test being "empty" and then you have to go elsewhere to see the actual test.

**Things I like about using YAML for checking the outputs:**
1. Very explicit

**Things I don't like about using YAML for outputs:**
1. Very verbose - "common things" like checking conditions which could be shared/hidden in a Go test helper clutter up the various output YAMLs in a way that could be avoided in Go.
2. Further splits the test up
3. Difficult to assert things like "absence of a certain field", "absence of a certain object", "the field matches this regex", etc. Need to fall back to code for this.

The "steps in the files/folders" (versus steps in the tests) approach has similar questions around it. I also don't have a good feel for how many tests are going to be single-step and how many tests are going to be multi-step.

I see two directions here:
1. We lean hard into the "everything in YAML" approach, which seems to be what this PR is showcasing, and most tests are written with their inputs/outputs/steps defined in YAML on the filesystem rather than in the Go test.
2. We push more on the Go code (with maybe continuing to define the inputs in YAML in files alongside the test or in testdata, or inline in the test for small YAMLs), and do most/all of the assertions and step-building in the Go tests rather than in the filesystem.

My personal inclination is more towards 2 I think, but I want to hear what other people think.

matthchr: @justinsb since you were just looking at this PR as well.

test/e2e/framework/framework.go
barney-s: We are parking this PR and switching to Chainsaw as suggested by a community user. We may need cluster creation scripts as part of tests at some point when we move beyond kind only tests.

test/e2e/framework/manifest.go
barney-s: We are parking this PR and switching to Chainsaw as suggested by a community user.

test/e2e/suites/basic/simple_deployment_test.go
barney-s: We are parking this PR and switching to Chainsaw as suggested by a community user.

barney-s:
Closing this in favor of https://github.com/kro-run/kro/pull/590

Issue #291, opened 2025-02-11
Kro is not supporting CEL raw string 


dirien:
### Description

**Observed Behavior**:

The Pulumi YAML support uses ${...} to evaluate expressions inside the `yaml`. When using this with we get following issue in our `rgd`: 

```bash
failed to validate resource CEL expressions: failed to dry-run expression project.id: failed to compile expression: ERROR: <input>:1:1: undeclared reference to 'project' (in container '')
```

Have a look into the `projectId` property. 

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: my-pko-app
spec:
  schema:
    apiVersion: v1alpha1
    kind: MyApplication
    spec:
      name: string
      stack: string
      gitRepo: string
      region: string | default="nyc1"
      instanceCount: integer | default=1
      instanceSizeSlug: string | default="apps-s-1vcpu-1gb"
  resources:
  - id: program
    template:
      apiVersion: pulumi.com/v1
      kind: Program
      metadata:
        name: ${schema.spec.name}
      program:
        resources:
          project:
            type: digitalocean:Project
            properties:
              name: ${schema.spec.name}-project
              description: A project to represent development resources.
              purpose: Web Application
              environment: Development
          app:
            type: digitalocean:App
            properties:
              projectId: ${project.id}
              spec:
                name: app-${schema.spec.name}-${schema.spec.region}
                region: ${schema.spec.region}
                services:
                - name: app-service-${schema.spec.name}-${schema.spec.region}
                  instanceCount: ${schema.spec.instanceCount}
                  instanceSizeSlug: ${schema.spec.instanceSizeSlug}
                  git:
                    repoCloneUrl: ${schema.spec.gitRepo}
                    branch: "main"
  - id: stack
    template:
      apiVersion: pulumi.com/v1
      kind: Stack
      metadata:
        name: ${schema.spec.name}
      spec:
        stack: ${schema.spec.stack}
        envRefs:
          DIGITALOCEAN_TOKEN:
            type: Env
            env:
              name: DIGITALOCEAN_TOKEN
        programRef:
          name: ${schema.spec.name}
        destroyOnFinalize: true
```

More details about the Pulumi Property Access: https://www.pulumi.com/docs/iac/languages-sdks/yaml/yaml-language-reference/#property-access

[...]

PR #297, opened 2025-02-12
feat: allow escaped nested expressions

dirien:
This PR is Related to #291

This commit enhances the expression parser to properly handle nested expressions when they appear inside string literals, adding string literal tracking and escape sequence handling while still rejecting unquoted nested expressions.

### DIFF
--- a/pkg/graph/parser/cel.go
+++ b/pkg/graph/parser/cel.go
@@ -42,26 +42,48 @@ func extractExpressions(str string) ([]string, error) {
 		// Adjust the start index to the actual position in the string
 		startIdx += start
 
-		// We need to find the matching end bracket. we have to be careful about
-		// nested expressions and dictionary building expressions. For example:
-		// a user can have an expression like "${{"key": 123}}". For this reason,
-		// we need to keep track of the bracket count and make sure that we only
-		// consider the outermost expression
+		// We need to find the matching end bracket, being careful about
+		// nested expressions, dictionary building expressions, and string literals
 		bracketCount := 1
 		endIdx := startIdx + len(exprStart)
+		inStringLiteral := false
+		escapeNext := false
+
 		for endIdx < len(str) {
-			if str[endIdx] == '{' {
-				bracketCount++
-			} else if str[endIdx] == '}' {
-				bracketCount--
-				// If we have reached the end of the expression, break
-				if bracketCount == 0 {
-					break
+			c := str[endIdx]
+
+			// Handle escape sequences inside string literals
+			if escapeNext {
+				escapeNext = false
+				endIdx++
+				continue
+			}
+
+			// Check for escape character inside string literals
+			if inStringLiteral && c == '\\' {
+				escapeNext = true
+				endIdx++
+				continue
+			}
+
+			// Handle string literal boundaries
+			if c == '"' {
+				inStringLiteral = !inStringLiteral
+			} else if !inStringLiteral {
+				// Only count braces when not inside a string literal
+				if c == '{' {
+					bracketCount++
+				} else if c == '}' {
+					bracketCount--
+					if bracketCount == 0 {
+						break
+					}
+				} else if endIdx+1 < len(str) && str[endIdx:endIdx+2] == "${" {
+					// Allow nested expressions, but only if they are escaped with quotes
+					if str[endIdx-1] != '"' {
+						return nil, ErrNestedExpression
+					}
 				}
-			} else if endIdx+1 < len(str) && str[endIdx:endIdx+2] == "${" {
-				// We do not allow nested expressions. I'm not sure if this is a
-				// good idea, but its sounds like a reasonable restriction.
-				return nil, ErrNestedExpression
 			}
 			endIdx++
 		}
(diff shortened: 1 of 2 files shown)


dirien:
Resolves #297

barney-s:
What about supporting R"" raw string instead of this ?

a-hilaly:
>What about supporting R"" raw string instead of this ?

Are you suggesting using `R""` as in `${"hello" + R"${something}"}`? I like it! and it looks like it's CEL compatible.

In general I think our parser should indeed allow strings that contain`${}` characters, and we should also consider cases where strings appear between single quotes `'`.

a-hilaly:
I think for now @dirien's solution handles the current use case well while maintaining backward compatibility with existing implementations. so i'm with merging this and later on add `""`, `''`, R"", and `R''` support

dirien:
@a-hilaly, for now I wait with the Pulumi example contribution for this to land. Don't want to use the workaround with string concat.

dirien:
Hi @a-hilaly, do you need any input from me for this PR?

dirien:
smol ping @a-hilaly for this PR.

dirien:
@a-hilaly is this PR still relevant?

pkg/graph/parser/cel.go
```diff
@@ -49,19 +49,23 @@ func extractExpressions(str string) ([]string, error) {
...
-				// We do not allow nested expressions. I'm not sure if this is a
-				// good idea, but its sounds like a reasonable restriction.
-				return nil, ErrNestedExpression
+				// Allow nested expressions, but only if they are escaped with quotes, CEL will
+				// treat them as a string literal
+				if str[endIdx-1] != '"' {
+					// We do not allow nested expressions. I'm not sure if this is a
+					// good idea, but its sounds like a reasonable restriction.
+					return nil, ErrNestedExpression
+				}
```
a-hilaly: Thanks @dirien! I like the approach of allowing quoted nested expressions. one thing to consider though - the current implementation might have trouble with braces inside quoted strings. since we're counting all `{` and `}` characters regardless of whether they're in quotes. e.g an expression like `${"text with } inside"}` could end prematurely, and we'll only extract "`text with }`" as a CEL expression.

I'm thinking that  we could track whether we're inside a string literal as we parse? That way we'd only count braces when they're actually structural elements of the CEL expression

a-hilaly:
>@a-hilaly is this PR still relevant?

Hey @dirien , yes this is relevant and I would love to have it for 0.2.2 or 0.3.0, let me know if you have time to address to comment above, otherwise i can also give it a shot, thanks :)

pkg/graph/parser/cel.go
dirien: @a-hilaly made the changes. Have a look if you like it.

dirien:
@a-hilaly Made some changes to your comment. Please review this and let me know what you think.

pkg/graph/parser/cel.go
```diff
@@ -78,6 +101,7 @@ func extractExpressions(str string) ([]string, error) {
 		expressions = append(expressions, expr)
 		start = endIdx + 1
 	}
+	fmt.Println("Expressions: ", expressions)
```
a-hilaly: @dirien  Looks like this debug statement was left in :)

a-hilaly (commented):
This looks correct to me, thank you @dirien ! left one minor comment above

pkg/graph/parser/cel.go
dirien: @a-hilaly oops!

dirien:
@a-hilaly Thanks for the feedback. I removed the line! 😅

pkg/graph/parser/cel.go
```diff
@@ -42,26 +42,48 @@ func extractExpressions(str string) ([]string, error) {
...
+				} else if c == '}' {
+					bracketCount--
+					if bracketCount == 0 {
+						break
+					}
+				} else if endIdx+1 < len(str) && str[endIdx:endIdx+2] == "${" {
+					// Allow nested expressions, but only if they are escaped with quotes
+					if str[endIdx-1] != '"' {
+						return nil, ErrNestedExpression
+					}
```
rushmash91, line 85: can we update the nested error as well?

```var ErrNestedExpression = errors.New("nested expressions are not allowed unless inside string literals")```

File: pkg/graph/parser/cel.go [hunk 4]
```diff
@@ -42,26 +42,48 @@ func extractExpressions(str string) ([]string, error) {
...
+				// Only count braces when not inside a string literal
+				if c == '{' {
+					bracketCount++
+				} else if c == '}' {
+					bracketCount--
+					if bracketCount == 0 {
+						break
+					}
+				} else if endIdx+1 < len(str) && str[endIdx:endIdx+2] == "${" {
+					// Allow nested expressions, but only if they are escaped with quotes
```
rushmash91, line 82: nit: might be cleaner to have a small example
```suggestion
					// Allow nested expressions, but only if they are escaped with quotes i.e. ${outer("${inner}")} is allowed, but ${outer(${inner})} is not
```

rushmash91 (commented):
Thank you @dirien 🚀! This is great! left two minor comments.

a-hilaly (approved):
Thanks a lot @dirien !

a-hilaly:
@rushmash91 @dirien we can address the comments in a separate PR :), merging for now.

PR #299, opened 2025-02-12
Handle Secrets that are transformed by server and hence show up as diff

### LINKED ISSUE #272: I have Kubernetes Secrets correctly created, but seing with delta by kro
### Description

**Observed Behavior**:

I created a Ressource with kro which is a kubernetes secret.
Kro is correctly creating the secret, but then it continuously find deltas in it and won't create the other resources

**Expected Behavior**:

I expect kro to see that there is no difference because the secret has been correctly created, and then it continue and create other resources.

Note: I didn't saw this behaviour in the 0.1 of kro.

**Versions**:
- kro version: 0.2
- Kubernetes Version (`kubectl version`): 1.30

**Involved Controllers**:

**Error Logs** (if applicable)**: This is the log line but in better readibility
```


Pod: kro-6d5b8f8b8f-7zxt9
Application: kro
Timestamp: 2025-02-04T10:31:08.327Z
Log Level: DEBUG
Component: controller.eksclusters
Message: Found deltas for resource

Resource Details:
{
    "namespace": "workload-cluster3",
    "name": "workload-cluster3",
    "resourceID": "argocdSecret",
    "delta": [
        {
            "path": "stringData",
            "desired": {
                "config": {
                    "awsAuthConfig": {
                        "clusterName": "workload-cluster3",
                        "roleARN": "arn:aws:iam::798082067117:role/workload-cluster3-argocd-role"
                    },
                    "tlsClientConfig": {
                        "caData": "LS0tLS1CRUdJTiBDRVJUSUZ....[truncated]....FLS0tLS0K",
                        "insecure": false
                    }
                },
                "name": "workload-cluster3",
                "server": "https://2D0E5C77B3C.truncated.4553EA49070B.gr7.us-west-2.eks.amazonaws.com"
            },
            "observed": null
        }
    ]
}
```

_No response_

barney-s:
- Add code infra to handle secrets that are transformed by the server
- Add logic to allow future special case handling as generic object transformers
- Add unit tests

Fixes #272 

Using tests defined in https://github.com/kro-run/kro/pull/290

## Tests Run
```
❯ go test -v ./test/e2e/suites/advanced/...
=== RUN   TestSecretsInResources
    using_secrets_test.go:27: Using namespace: e2e-secrets-in-resources-1739398489, from: testdata/../../../../test/e2e/testdata/secrets-in-resources
    testcase.go:193: Running step0: Install RGD and verify that the RGD is active
    testcase.go:359:    Applying ResourceGraphDefinition/deployment-with-secret-rgd[rgd.yaml]
    verify.go:57:       Verifying ResourceGraphDefinition/deployment-with-secret-rgd[0rgd.yaml]
    verify.go:57:       Verifying CustomResourceDefinition/deploymentwithsecrets.kro.run[1instancecrd.yaml]
    testcase.go:193: Running step1: Create instance DeploymentWithSecret with replicas=2 and tokenID=sometokenid and tokenSecret=kq4gihvszzgn1p0r
    testcase.go:359:    Applying DeploymentWithSecret/test-instance[instance.yaml]
    verify.go:57:       Verifying Deployment/test-instance[deployment.yaml]
    verify.go:57:       Verifying DeploymentWithSecret/test-instance[instance.yaml]
    verify.go:57:       Verifying Secret/test-instance-secret[secret.yaml]
    testcase.go:193: Running step2: Update instance DeploymentWithSecret with tokenID=newtokenid and tokenSecret=newtokensecret
    testcase.go:359:    Applying DeploymentWithSecret/test-instance[instance.yaml]
    verify.go:57:       Verifying Deployment/test-instance[deployment.yaml]
    verify.go:57:       Verifying DeploymentWithSecret/test-instance[instance.yaml]
    verify.go:57:       Verifying Secret/test-instance-secret[secret.yaml]
    testcase.go:134: Deleting namespace: e2e-secrets-in-resources-1739398489
    testcase.go:148: Deleting cluster-scoped object: ResourceGraphDefinition/deployment-with-secret-rgd
--- PASS: TestSecretsInResources (40.19s)
PASS
ok      github.com/kro-run/kro/test/e2e/suites/advanced 40.258s
```

### DIFF
--- a/pkg/controller/instance/delta/delta.go
+++ b/pkg/controller/instance/delta/delta.go
@@ -36,6 +36,7 @@ type Difference struct {
 // their differences. It performs a deep comparison while being aware of Kubernetes
 // metadata specifics. The comparison:
 //
+// - Applies type-specific transformations to match server-side representation
 // - Cleans metadata from both objects to avoid spurious differences
 // - Walks object trees in parallel to find actual value differences
 // - Builds path strings to precisely identify where differences occurs
@@ -44,6 +45,10 @@ func Compare(desired, observed *unstructured.Unstructured) ([]Difference, error)
 	desiredCopy := desired.DeepCopy()
 	observedCopy := observed.DeepCopy()
 
+	// Transform objects to match server-side representation
+	if err := TransformObjectToServerSideRepresentation(desiredCopy); err != nil {
+		return nil, fmt.Errorf("failed to transform desired object: %w", err)
+	}
 	cleanMetadata(desiredCopy)
 	cleanMetadata(observedCopy)
 
--- a/pkg/controller/instance/delta/transform.go
+++ b/pkg/controller/instance/delta/transform.go
@@ -1,0 +1,104 @@
+// Copyright 2025 The Kube Resource Orchestrator Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package delta
+
+import (
+	"encoding/base64"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// ObjectTransformer is an interface for transforming objects before comparison
+type ObjectTransformer interface {
+	// Transform modifies the object to match server-side representation
+	Transform(obj *unstructured.Unstructured) error
+	// CanTransform returns true if this transformer can handle the given object
+	CanTransform(obj *unstructured.Unstructured) bool
+}
+
+// transformerRegistry holds all registered transformers
+var transformerRegistry []ObjectTransformer
+
+// RegisterTransformer adds a new transformer to the registry
+func RegisterTransformer(t ObjectTransformer) {
+	transformerRegistry = append(transformerRegistry, t)
+}
+
+// TransformObjectToServerSideRepresentation applies all applicable transformers to the object
+func TransformObjectToServerSideRepresentation(obj *unstructured.Unstructured) error {
+	for _, transformer := range transformerRegistry {
+		if transformer.CanTransform(obj) {
+			if err := transformer.Transform(obj); err != nil {
+				return fmt.Errorf("failed to transform object: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+// SecretTransformer handles transformation of Secret objects
+type SecretTransformer struct{}
+
+// CanTransform returns true if the object is a Secret
+func (t *SecretTransformer) CanTransform(obj *unstructured.Unstructured) bool {
+	return obj.GetAPIVersion() == "v1" && obj.GetKind() == "Secret"
+}
+
+// Transform modifies the Secret object to match server-side representation:
+// - Moves stringData to data with base64 encoding
+// - Merges any existing data with encoded stringData (stringData takes precedence)
+func (t *SecretTransformer) Transform(obj *unstructured.Unstructured) error {
+	// Get stringData if it exists
+	stringData, exists, err := unstructured.NestedMap(obj.Object, "stringData")
+	if err != nil {
+		return fmt.Errorf("failed to get stringData: %w", err)
+	}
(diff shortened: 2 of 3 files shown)


allamand:
@barney-s thanks for the PR, I tested it, and it do not loop anymore. For updates, It did do the update but only If I manually delete kro pod, but this is more general and I guess it should be tracked in another issue ?

So seems you're PR worked fine with my use case, thanks

allamand:
What's needed to get this merged ?

allamand:
@a-hilaly with this fix, the problem with secrets is gone, is it possible to get it merged ?

barney-s:
> @barney-s thanks for the PR, I tested it, and it do not loop anymore. For updates, It did do the update but only If I manually delete kro pod, but this is more general and I guess it should be tracked in another issue ?
> 
> So seems you're PR worked fine with my use case, thanks

Thanks for verifying. We have testcases for this use case in #290 . Would it be possible to describe a test case for the failing update usecase.

mrwulf:
Solves #272 for me as well. Please review and merge!

pkg/controller/instance/delta/transform.go
```diff
@@ -0,0 +1,104 @@
...
+			if err := transformer.Transform(obj); err != nil {
+				return fmt.Errorf("failed to transform object: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+// SecretTransformer handles transformation of Secret objects
+type SecretTransformer struct{}
```
matthchr, line 52: ConfigMaps will do something similar with their two options? Do we need to handle them too?

pkg/controller/instance/delta/transform.go
```diff
@@ -0,0 +1,104 @@
...
+// ObjectTransformer is an interface for transforming objects before comparison
+type ObjectTransformer interface {
+	// Transform modifies the object to match server-side representation
+	Transform(obj *unstructured.Unstructured) error
+	// CanTransform returns true if this transformer can handle the given object
+	CanTransform(obj *unstructured.Unstructured) bool
+}
+
+// transformerRegistry holds all registered transformers
+var transformerRegistry []ObjectTransformer
```
matthchr, line 32: style, curious @a-hilaly on this:

I'm in general not a huge fan of package-scope mutable state. I think I'd rather see this just be a fixed list of transformers, not something that you can dynamically register more of.

i.e. like:
```
var transformerRegistry = []ObjectTransformer{
    &SecretTransformer{}
}
```

and just no dynamic registration?

pkg/controller/instance/delta/delta.go
```diff
@@ -43,6 +44,10 @@ func Compare(desired, observed *unstructured.Unstructured) ([]Difference, error)
 	desiredCopy := desired.DeepCopy()
 	observedCopy := observed.DeepCopy()
 
+	// Transform objects to match server-side representation
```
matthchr, line 48: I wonder if we actually need something more than this.
The problem you're solving here is: "The shape of the resource changes when you submit it k8s, and we detect those changes as differences"

The solution you have here solves that problem for Secrets (and maybe ConfigMaps, if we add a mapper for that), but the problem is more general than that, because it can impact any resource with a mutating webhook.

Possible solutions that are more generic:
1. Take the result of the `Create` or `Update` call and store them fully in an annotation or status (either of the resource itself or the RGD). Then use this as the base for our diffs.
2. Same as 1, but instead of storing the whole object, strip the metadata (like we do for diffing) and then compute a hash and store that someplace (an annotation?). Then recompute a hash from the observed state via GET and if the hashes differ we Update.
3. Similar to 1/2, but rather than persisting the information into an annotation/status we just store the last seen create/update result in-memory and then use that as the expected state and diff on that. This has the advantage of being easy/clean and the disadvantage that it'll be lost at pod restart, although maybe re-applying resources on pod restart isn't the worst thing, I am not sure.

Another thing that occurs to me here is that we probably should be excluding at least CRD status (but realistically probably all status fields), as I don't think we care about status when diffing?

matthchr: The more I look at the way we do things now the more I think we need a more comprehensive solution to this problem. I don't know about KCC/ACK, but ASO definitely has resources with mutating webhooks. _and_ status fields that can change (conditions, etc) even if the spec is fixed, that if not accounted for will trigger continuous reconciliation, so I don't think the problem is theoretical.

allamand:
@barney-s in the mentime waiting for a more global solution, Can you rebase your PR so it have latest kro improvments also ?

barney-s:
@allamand  - Could you test with #561 for this use case ?

barney-s:
Closing this in favor of #561

allamand:
@barney-s awesome! I’ll do

PR #302, opened 2025-02-13
chore: update golangci-lint version

ldez:
Update golangci-lint to v1.64 to supoort go1.24


a-hilaly (approved):
Nice catch, and thanks a lot @ldez !

PR #303, opened 2025-02-13
Bump Go version to 1.24

a-hilaly:


JulienBreux:
Maybe need to add also https://github.com/kro-run/kro/blob/main/go.mod#L3 ?

a-hilaly:
@JulienBreux Ohhh, is that why golangcilint jobs are failing in everyother PR? could be..

JulienBreux (approved):
LGTM, enjoy Go 1.24 🚀

Issue #305, opened 2025-02-15
Improper GVR formatting causing failed informer deregistration


Gonfidel:
### Description

**Observed Behavior**:
When testing RGD creation and deletion using the following sample: `example/kubernetes/webapp/rg.yaml`, informers are successfully registered but fail to be cleaned up due to `pkg\metadata\groupversion.go#GetResourceGraphDefinitionInstanceGVR` returning an improperly formatted GVR. This finding also blocks #198 due to relying on the same gvr reference to deregister queues or gvrSets. 

Proposed GVR:
```diff
schema.GroupVersionResource{
-		Group:    fmt.Sprintf("%s.%s", pluralKind, group),
+		Group:    group,
		Version:  apiVersion,
		Resource: pluralKind,
	}
```

**Expected Behavior**:
Successfully deregister informers on `DynamicController#StopServiceGVK`

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):
Apply `examples\kubernetes\webapp\rg.yaml` and delete it with a debugger in `DynamicController#StopServiceGVK`.  We perform an early return after not finding the informer due to the name mismatch

**Versions**:
- kro version: 0.2.1
- Kubernetes Version (`kubectl version`):  v1.30.2

**Involved Controllers**:
- Controller URLs and Versions (if applicable): KRO RGD Controller passing GVR to Dynamic Controller

**Error Logs** (if applicable)**:
No logs, just early termination and no errors or warnings

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Which option describes the most your issue?

ResourceGraphDefinition (Create, Update, Deletion)

PR #307, opened 2025-02-15
Implement weighted queues

### LINKED ISSUE #198: Add Support for weight based queue management in dynamiccontroller
### Feature Description

Currently, the DynamicController uses a single queue for all GVRs, which doesn't allow for prioritization of resource processing based on user requirements. This can lead to resource starvation and unpredictable processing times for important resources.

One of the optimisations we can work on is a weight based queue system where resources are processed according to weights specified in ResourceGroup specs. Instead of maintaining a queue per GVR, we'll maintain queues per unique weight value.

We can add a `spec.weight` field to the ResourceGroup CRD to allow users to specify the processing priority of their resources. 

Thoughts: the weight will be an integer value between 1 and 1000, with a default value of 100 if not specified. Resources with higher weights will receive proportionally more processing time and workers from the controller.

Sketching:

```go
type WeightedQueue struct {
    weight    int
    queue     workqueue.RateLimitingInterface 
    gvrSet    map[schema.GroupVersionResource]struct{}
}

type DynamicController struct {
    weightedQueues map[int]*WeightedQueue
    gvrWeights     map[schema.GroupVersionResource]int
    mu             sync.RWMutex
}
```

Gonfidel:
Closes #198


pkg/metadata/groupversion.go
```diff
@@ -71,7 +71,7 @@ func GetResourceGraphDefinitionInstanceGVK(group, apiVersion, kind string) schem
 func GetResourceGraphDefinitionInstanceGVR(group, apiVersion, kind string) schema.GroupVersionResource {
 	pluralKind := flect.Pluralize(strings.ToLower(kind))
 	return schema.GroupVersionResource{
-		Group:    fmt.Sprintf("%s.%s", pluralKind, group),
+		Group:    group,
```
nic-george, line 74: This change remediates a bug related #305, causing informers to never deregister.

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -146,18 +168,88 @@ func NewDynamicController(
...
+		return false
+	}
+
+	if _, exists := dc.weightedQueues[weight]; !exists {
+		dc.mu.Lock()
+		defer dc.mu.Unlock()
+		// TODO(a-hilaly): Make the queue size configurable.
+		dc.weightedQueues[weight] = &WeightedQueue{
+			weight: weight,
+			queue: workqueue.NewNamedRateLimitingQueue(
```
nic-george, line 245: SAME COMMENT AS ABOVE: The old queue rate limiting types were failing linter tests due to being deprecated. These new limiters support type specific limiting. I believe we want these configured as `any` but feel free to provide feedback

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -98,6 +98,18 @@ type Config struct {
...
 }
 
+// DynamicController (DC) manages multiple GVRs and needs a way to prioritize the importance
+// of each GVR. Weighted queues provide a flexible way for the DynamicController (DC) to prioritize
+// GVRs based on their weights.
+type WeightedQueue struct {
+	// queue items will be prioritized based on GVR weight, between 1 and 1000
+	weight int
+	// queues are the workqueue used to process items
+	queue workqueue.TypedRateLimitingInterface[any]
```
nic-george, line 108: SAME COMMENT AS BELOW: The old queue rate limiting types were failing linter tests due to being deprecated. These new limiters support type specific limiting. I believe we want these configured as `any` but feel free to provide feedback

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -230,26 +337,73 @@ func (dc *DynamicController) worker(ctx context.Context) {
...
+	duration := time.Since(startTime)
+	dc.log.V(1).Info("Time to select queue", "duration", duration.String(), "weight", selectedQueue.weight)
+
+	return selectedQueue
+}
+
 // processNextWorkItem processes a single item from the queue.
 func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
-	obj, shutdown := dc.queue.Get()
+	weightedQueue := dc.selectQueueByWeight()
```
nic-george, line 387: This function performs an `RLock` and might be contentious due it's call frequency. Any thoughts or feedback on how best to improve this?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -230,26 +337,73 @@ func (dc *DynamicController) worker(ctx context.Context) {
...
+
+	obj, shutdown := weightedQueue.queue.Get()
 	if shutdown {
-		return false
+		return true
```
nic-george, line 391: I believe this returned false previously to exist the processing loop when our queue was shutting down because we only had one. Now that we have multiple, I think returning `true` is what we want, allowing us to short-circuit the processing event but not break the processing loop. Does that make the most sense here?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -146,18 +168,94 @@ func NewDynamicController(
...
+// setGVRWeight sets the weight for a given GroupVersionResource (GVR) and updates the
+// corresponding weighted queue.
+func (dc *DynamicController) setGVRWeight(gvr schema.GroupVersionResource, weight int) {
+	dc.ensureWeightedQueue(weight)
+
+	dc.mu.Lock()
+	defer dc.mu.Unlock()
+
+	dc.gvrWeights[gvr] = weight
+	dc.weightedQueues[weight].gvrSet[gvr] = struct{}{}
```
n3wscott, line 206: I don't understand why the gvr set is being set to an anonymous struct. If the value does not matter, why not a bool?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -146,18 +168,94 @@ func NewDynamicController(
...
+	delete(wq.gvrSet, gvr)
+	if len(wq.gvrSet) < 1 {
+		wq.queue.ShutDown()
+		delete(dc.weightedQueues, weight)
+	}
+}
+
+// Ensure weighted queue exists with the specified weight
+func (dc *DynamicController) ensureWeightedQueue(weight int) bool {
+	if weight < 1 || weight > 1000 {
```
n3wscott, line 234: this kind of protection should happen at RGD creation and should be bounds set inside of the DC or the env, not magic values inside this method. What happens if we want to add tightening the bounds of the range? Or expanding the range? Does it matter to the implementation inside of DC that the weight is a 1-1000 range at all? It seems it would just work if it treated it as a number genetically.

n3wscott (commented):
I don't think this is the direction the weighted queue should go. We should continue to have a discussion in the linked issue.

pkg/dynamiccontroller/dynamic_controller.go
nic-george: We probably don't need it. It's already defaulted and set on the CRD when we reconcile the RGD. It was just being extra cautious. We can definitely remove it

pkg/dynamiccontroller/dynamic_controller.go
nic-george: This is a WIP and it was set this way before adding the safety check for queue deletion. If a queue isn't managing any gvrs, we should delete it. The controller fails without at least one queue. This was a work around that can be removed now that we protect again deleting the default queue

nic-george:
@n3wscott I'm totally cool with taking it a different direction. This was a first pass and gave us some insight into potential hiccups. Happy to continue discussing until we can come up with a pattern that works for us 🙏

PR #311, opened 2025-02-18
Remove controller-pod-id label

a-hilaly:
This label is not really useful since pod IDs change in HA setups, making the label meaningless.

### DIFF
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -29,9 +29,8 @@ const (
 const (
 	NodeIDLabel = LabelKroPrefix + "node-id"
 
-	OwnedLabel           = LabelKroPrefix + "owned"
-	KroVersionLabel      = LabelKroPrefix + "kro-version"
-	ControllerPodIDLabel = LabelKroPrefix + "controller-pod-id"
+	OwnedLabel      = LabelKroPrefix + "owned"
+	KroVersionLabel = LabelKroPrefix + "kro-version"
 
 	InstanceIDLabel        = LabelKroPrefix + "instance-id"
 	InstanceLabel          = LabelKroPrefix + "instance-name"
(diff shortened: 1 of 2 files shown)


pkg/metadata/labels.go
```diff
@@ -135,12 +134,10 @@ func NewInstanceLabeler(instanceMeta metav1.Object) GenericLabeler {
 // KroVersion, and ControllerPodID labels on a resource.
 func NewKroMetaLabeler(
 	kroVersion string,
-	controllerPodID string,
 ) GenericLabeler {
 	return map[string]string{
-		OwnedLabel:           "true",
-		KroVersionLabel:      kroVersion,
-		ControllerPodIDLabel: controllerPodID,
```
barney-s, line 143: Just to confirm. We are removing "controller-pod-id: kro-pod" from the label ?

pkg/metadata/labels.go
```diff
@@ -135,12 +134,10 @@ func NewInstanceLabeler(instanceMeta metav1.Object) GenericLabeler {
 // KroVersion, and ControllerPodID labels on a resource.
```
barney-s: nit: please remove ControllerPodID from the comment.

pkg/metadata/labels.go
a-hilaly: Yes, it was hard coded, and I don't think it's bringing any value :)

pkg/metadata/labels.go
```diff
@@ -29,9 +29,8 @@ const (
 const (
 	NodeIDLabel = LabelKroPrefix + "node-id"
 
-	OwnedLabel           = LabelKroPrefix + "owned"
-	KroVersionLabel      = LabelKroPrefix + "kro-version"
-	ControllerPodIDLabel = LabelKroPrefix + "controller-pod-id"
+	OwnedLabel      = LabelKroPrefix + "owned"
+	KroVersionLabel = LabelKroPrefix + "kro-version"
```
a-hilaly, line 33: @matthchr i saw your comment on Go names intialisms i will address it in a seperate PR for all the kro repository. Looks like my s/Symphony/Kro/ wasn't a good idea :)

matthchr (approved):

PR #313, opened 2025-02-18
feat: add go ldflags for version info injection

a-hilaly:
Add support for injecting version information at build time via ldflags. This includes:
- Version, git commit hash, and build date injection in `.ko.yaml`
- Updated metadata labeler to use dynamic version
- Makefile updates to support version info injection

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -5,3 +5,6 @@ defaultPlatforms:
 defaultLdflags:
   - -s
   - -w
+  - -X github.com/kro-run/kro/pkg.Version={{.Env.RELEASE_VERSION}}
+  - -X github.com/kro-run/kro/pkg.GitCommit={{.Env.GIT_COMMIT}}
+  - -X github.com/kro-run/kro/pkg.BuildDate={{.Env.BUILD_DATE}}
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 
-RELEASE_VERSION ?= dev-$(shell git rev-parse --short HEAD)
+GIT_COMMIT ?= $(shell git rev-parse --short HEAD)
+BUILD_DATE ?= $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+RELEASE_VERSION ?= dev-$(GIT_COMMIT)
 OCI_REPO ?= ghcr.io/kro-run/kro
 
 HELM_IMAGE ?= ${OCI_REPO}
@@ -8,6 +10,10 @@ KO_DOCKER_REPO ?= ${OCI_REPO}/kro
 KOCACHE ?= ~/.ko
 KO_PUSH ?= true
 export KIND_CLUSTER_NAME ?= kro
+
+LDFLAGS ?= -X github.com/kro-run/kro/pkg.Version=$(RELEASE_VERSION) \
+           -X github.com/kro-run/kro/pkg.GitCommit=$(GIT_COMMIT) \
+           -X github.com/kro-run/kro/pkg.BuildDate=$(BUILD_DATE)
 
 WITH_GOFLAGS = GOFLAGS="$(GOFLAGS)"
 
@@ -106,7 +112,7 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 
 .PHONY: build
 build: manifests generate fmt vet ## Build controller binary.
-	go build -o bin/controller ./cmd/controller/main.go
+	go build -ldflags=${LDFLAGS} -o bin/controller ./cmd/controller/main.go
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -188,13 +194,15 @@ $(CONTROLLER_GEN): $(LOCALBIN)
...
 	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) KO_DOCKER_REPO=$(KO_DOCKER_REPO) \
+		GIT_COMMIT=$(GIT_COMMIT) RELEASE_VERSION=$(RELEASE_VERSION) BUILD_DATE=$(BUILD_DATE) \
 		$(KO) build --bare github.com/kro-run/kro/cmd/controller \
 		--local\
 		--push=false --tags ${RELEASE_VERSION} --sbom=none
 
 .PHONY: publish
 publish-image: ko ## Publish the kro controller images to ghcr.io
-	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) \
+	$(WITH_GOFLAGS) KOCACHE=$(KOCACHE) KO_DOCKER_REPO=$(KO_DOCKER_REPO) \
```
matthchr, line 204: Is the fact the `KO_DOCKER_REPO` wasn't here before a bug?

pkg/metadata/labels.go
```diff
@@ -29,9 +30,8 @@ const (
 const (
 	NodeIDLabel = LabelKroPrefix + "node-id"
 
-	OwnedLabel           = LabelKroPrefix + "owned"
-	KroVersionLabel      = LabelKroPrefix + "kro-version"
-	ControllerPodIDLabel = LabelKroPrefix + "controller-pod-id"
+	OwnedLabel      = LabelKroPrefix + "owned"
+	KroVersionLabel = LabelKroPrefix + "kro-version"
```
matthchr: I'm being a super pedant here but:
https://google.github.io/styleguide/go/decisions#initialisms

`KRO` vs `Kro`.

matthchr: This applies elsewhere in this file too.

pkg/metadata/labels.go
```diff
@@ -133,14 +133,10 @@ func NewInstanceLabeler(instanceMeta metav1.Object) GenericLabeler {
 
 // NewKroMetaLabeler returns a new labeler that sets the OwnedLabel,
 // KroVersion, and ControllerPodID labels on a resource.
-func NewKroMetaLabeler(
-	kroVersion string,
-	controllerPodID string,
-) GenericLabeler {
+func NewKroMetaLabeler() GenericLabeler {
```
matthchr: What's the experience when running tests locally (in a KIND cluster, for example).

Are the labels set correctly? I looked at the makefile and I _think_ they will be?
What about in envtest (less sure they will be there)?

Does it make sense to have this still be settable on the labeller and then have the envtest context specifically pass "something"? Or maybe we need envtest contexts to just set pkg.Version/etc correctly?

matthchr (approved):

pkg/metadata/labels.go
```diff
@@ -133,14 +133,10 @@ func NewInstanceLabeler(instanceMeta metav1.Object) GenericLabeler {
 
 // NewKroMetaLabeler returns a new labeler that sets the OwnedLabel,
 // KroVersion, and ControllerPodID labels on a resource.
```
barney-s: minor: remove ControllerPodID from the comment. I guess this also applies to https://github.com/kro-run/kro/pull/311

.ko.yaml
```diff
@@ -7,3 +7,6 @@ defaultPlatforms:
 defaultLdflags:
   - -s
   - -w
+  - -X github.com/kro-run/kro/pkg.Version={{.Env.RELEASE_VERSION}}
+  - -X github.com/kro-run/kro/pkg.GitCommit={{.Env.GIT_COMMIT}}
```
barney-s, line 9: I assume this automagic is handled by KO ?

a-hilaly: Correct https://ko.build/configuration/#overriding-go-build-settings

Makefile
a-hilaly: Correct, i was testing it with a custom OCI registry, looks like this is needed

pkg/metadata/labels.go
a-hilaly: I like it!! please keep them coming :)

pkg/metadata/labels.go
a-hilaly: addressed in https://github.com/kro-run/kro/pull/311

pkg/metadata/labels.go
a-hilaly: I did some tests locally for this - I like the idea of injecting some in envtest, i'll send another commit add more tests here.

matthchr (approved):

PR #315, opened 2025-02-18
ci: add Go path filters to Go related workflows

a-hilaly:
Add path filters to only trigger Go related workflows when relevant files change:
   - **.go files
   - go.mod
   - go.sum

(except golangci-lint which is only needed for .go file changes)

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -3,8 +3,16 @@ name: Integration Tests
 on:
   push:
     branches: [ main ]
+    paths:
+      - "**.go"
+      - go.mod
+      - go.sum
   pull_request:
     branches: [ main ]
+    paths:
+      - "**.go"
+      - go.mod
+      - go.sum
 
 jobs:
   integration:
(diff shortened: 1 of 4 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -2,31 +2,39 @@ name: Integration Tests
...
 
-    - name: Set up Go
-      uses: actions/setup-go@v5
-      with:
-        go-version-file: 'go.mod'
-        cache: true
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: "go.mod"
```
matthchr: minor: why `"` this and (basically) nothing else?

matthchr (approved):

.github/workflows/integration-tests.yaml
a-hilaly: That's my yaml formatter, and i don't like it too..

a-hilaly: Removed the formatting for now. We want to address that seperately!

matthchr (approved):

PR #316, opened 2025-02-18
Refactor dynamic controller queue to use typed rate limiting interface

DanielKlt:
RateLimitingInterface is deprecated and should be removed with TypedRateLimitingInterface


a-hilaly (approved):
Neat! Thanks a lot @DanielKlt !

PR #321, opened 2025-02-20
clusterName field addition to avoid podidentity error

7navyasa:
clusterName` is a required field to achieve pod identity association. If the user's cluster name is not 'kro' then pod identity association will fail and we will get output as "unhealthy"


candonov (approved):
lgtm!

a-hilaly (approved):
Thanks @7navyasa !

PR #325, opened 2025-02-20
fix: handle x-kubernetes-preserve-unknown-fields

jknutson:
Handle missing type in schema (generateObject) when x-kubernetes-preserve-unknown-fields is present.

CI checks are passing on my copy/fork of the repo: https://github.com/jknutson/kro/actions

I added a test, but did not include a type check; I did not feel that the goal here was to ensure a specific type.
If it would be valuable, I can add a test to ensure that we have an "object" returned.

This resolves https://github.com/kro-run/kro/issues/319; the failed RGD reconciles as soon as I start up the controller with this change.

### DIFF
--- a/pkg/graph/emulator/emulator.go
+++ b/pkg/graph/emulator/emulator.go
@@ -81,6 +81,11 @@ func (e *Emulator) generateValue(schema *spec.Schema) (interface{}, error) {
 		return nil, fmt.Errorf("schema is nil")
 	}
 
+	if enabled, ok := schema.VendorExtensible.Extensions["x-kubernetes-preserve-unknown-fields"]; ok && enabled.(bool) {
+		// Handle x-kubernetes-preserve-unknown-fields
+		return e.generateObject(schema)
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/graph/emulator/emulator.go
```diff
@@ -101,6 +106,7 @@ func (e *Emulator) generateValue(schema *spec.Schema) (interface{}, error) {
 			return e.generateValue(&schema.AnyOf[e.rand.Intn(len(schema.AnyOf))])
 		}
 
+		// return nil, fmt.Errorf("schema type is empty and has no properties: %+v", schema)
```
a-hilaly: nit: we can completly remove this comment

a-hilaly (commented):
Thank you @jknutson ! i have one tiny suggestion below :)

pkg/graph/emulator/emulator.go
jknutson: Oops! It will be gone momentarily 😄

a-hilaly (approved):
Thank you @jknutson ! :)

PR #331, opened 2025-02-27
Add support for explicit dependencies with `dependsOn` field

rushmash91:
Closes https://github.com/kro-run/kro/issues/110
Implements support for explicit dependencies between resources in a ResourceGraphDefinition using the dependsOn field. 

- Added support for the dependsOn field in the Resource struct
- Modified the dependency graph building logic to process explicit dependencies first
- Added validation to ensure referenced resources exist
- Ensured cycle detection works for both explicit and implicit dependencies
- Added unit tests

### DIFF
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -94,6 +94,9 @@ type Resource struct {
 	ReadyWhen []string `json:"readyWhen,omitempty"`
 	// +kubebuilder:validation:Optional
 	IncludeWhen []string `json:"includeWhen,omitempty"`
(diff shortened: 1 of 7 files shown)


ikchifo:
Overall, I think this looks really good. The tests you added are exhaustive and cover most use cases.
Adding an example of how to use this feature in the `examples` folder would also be great.

a-hilaly:
Thank you for your efforts on this @rushmash91 ! as discussed offline we won't proceed with this unless if we have a good reason to - where implicit dependencies aren't enough. Happy to continue the conversation in the github issue. Closing for now
/close

PR #349, opened 2025-03-04
Handle structural constraints in AnyOf schemas consistently with OneOf

rushmash91:
## Reason for Change
Previously, the parser only handled structural constraints (Required and Not) for OneOf schemas, but not for AnyOf schemas. This inconsistency could lead to incorrect type inference when parsing resources with AnyOf schemas that contain structural constraints.
## Description of Change
- Modified the schema parser to handle structural constraints in AnyOf schemas the same way as OneOf schemas
- Extracted common logic for handling subschemas into `collectTypesFromSubSchemas` function
- Added comprehensive unit tests to verify the behavior of structural constraint handling in both OneOf and AnyOf schemas


a-hilaly (approved):
Neat, thank you @rushmash91 !

PR #351, opened 2025-03-05
Update README.md to add star history graph

rushmash91:
- add `https://api.star-history.com/svg?repos=kro-run/kro&type=Date` graph


matthchr:
I'm not personally a huge fan of this. Pressuring people to star, or even giving the appearance of doing so, can be a bad look. I expect people to star if they like the project or are interested in it. I'm not sure they need a reminder to do so.

Showing the star history, which looks cool (who doesn't love graphs going up and to the right!), is really highlighting the wrong thing about the project, and I don't think it deserves such prominence in the README. I'd rather see the README laser focused on the user value and technical details. 

If we do that, I think stars and user engagement will go up naturally.

matthchr:
Closing this based on the above and some discussion with others (@a-hilaly, @nicslatts).

We appreciate the interest in the project though! If you're looking for things to contribute there's a bunch of possibilities in the issues list.

PR #352, opened 2025-03-05
feat(crd): Prevent breaking changes in RGD schemas

a-hilaly:
`ResourceGraphDefinitions` generate CRDs on the fly, which means one wrong
move in the schema can break existing resources. Deleting a critical field
or switching types can cause immediate deployment failures.

This patch introduces a `DiffSchema` function that systematically compares
old and new schemas, identifying breaking changes like property removals,
type modifications, and constraint updates. It categorizes changes as
breaking or non breaking, allowing safer schema evolution while preventing
accidental resource invalidation.

### DIFF
--- a/pkg/client/crd.go
+++ b/pkg/client/crd.go
@@ -25,6 +25,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	crdcompat "github.com/kro-run/kro/pkg/graph/crd/compat"
 )
 
 const (
@@ -92,27 +94,46 @@ func newCRDWrapper(cfg CRDWrapperConfig) *CRDWrapper {
 // Ensure ensures a CRD exists, up-to-date, and is ready. This can be
 // a dangerous operation as it will update the CRD if it already exists.
 //
-// The caller is responsible for ensuring the CRD, isn't introducing
-// breaking changes.
-func (w *CRDWrapper) Ensure(ctx context.Context, crd v1.CustomResourceDefinition) error {
-	_, err := w.Get(ctx, crd.Name)
+// If a CRD does exist, it will compare the existing CRD with the desired CRD
+// and update it if necessary. If the existing CRD has breaking changes, it
+// will return an error.
+func (w *CRDWrapper) Ensure(ctx context.Context, desired v1.CustomResourceDefinition) error {
+	current, err := w.Get(ctx, desired.Name)
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return fmt.Errorf("failed to check for existing CRD: %w", err)
 		}
 
-		w.log.Info("Creating CRD", "name", crd.Name)
-		if err := w.create(ctx, crd); err != nil {
+		w.log.Info("Creating CRD", "name", desired.Name)
+		if err := w.create(ctx, desired); err != nil {
 			return fmt.Errorf("failed to create CRD: %w", err)
 		}
 	} else {
-		w.log.Info("Updating existing CRD", "name", crd.Name)
-		if err := w.patch(ctx, crd); err != nil {
+		// CRD exists, check compatibility
+		diffResult, err := crdcompat.DiffSchema(current.Spec.Versions, desired.Spec.Versions)
+		if err != nil {
+			return fmt.Errorf("failed to check schema compatibility: %w", err)
+		}
+
+		// If there are no changes at all, we can skip the update
+		if !diffResult.HasChanges() {
+			w.log.Info("CRD is up-to-date", "name", desired.Name)
+			return nil
+		}
+
+		// Check for breaking changes
+		if diffResult.HasBreakingChanges() {
+			w.log.Info("Breaking changes detected in CRD update", "name", desired.Name, "breakingChanges", len(diffResult.BreakingChanges), "summary", diffResult)
+			return fmt.Errorf("cannot update CRD %s: breaking changes detected: %s", desired.Name, diffResult)
+		}
+
+		w.log.Info("Updating existing CRD", "name", desired.Name)
+		if err := w.patch(ctx, desired); err != nil {
 			return fmt.Errorf("failed to patch CRD: %w", err)
 		}
 	}
 
-	return w.waitForReady(ctx, crd.Name)
+	return w.waitForReady(ctx, desired.Name)
 }
 
 // Get retrieves a CRD by name
--- a/pkg/graph/crd/compat/changes.go
+++ b/pkg/graph/crd/compat/changes.go
@@ -1,0 +1,155 @@
+// Copyright 2025 The Kube Resource Orchestrator Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package compat
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ChangeType represents the type of schema change
+type ChangeType string
+
+const (
+	// Breaking change types
+	PropertyRemoved ChangeType = "PROPERTY_REMOVED"
+	TypeChanged     ChangeType = "TYPE_CHANGED"
+	RequiredAdded   ChangeType = "REQUIRED_ADDED"
+	EnumRestricted  ChangeType = "ENUM_RESTRICTED"
+	PatternChanged  ChangeType = "PATTERN_CHANGED"
+
+	// Non-breaking change types
+	PropertyAdded      ChangeType = "PROPERTY_ADDED"
+	DescriptionChanged ChangeType = "DESCRIPTION_CHANGED"
+	DefaultChanged     ChangeType = "DEFAULT_CHANGED"
+	RequiredRemoved    ChangeType = "REQUIRED_REMOVED"
+	EnumExpanded       ChangeType = "ENUM_EXPANDED"
+)
+
+// Change represents a single schema change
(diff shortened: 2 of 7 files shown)


pkg/client/crd.go
```diff
@@ -92,27 +94,46 @@ func newCRDWrapper(cfg CRDWrapperConfig) *CRDWrapper {
...
+		if !diffResult.HasChanges() {
+			w.log.Info("CRD is up-to-date", "name", desired.Name)
+			return nil
+		}
+
+		// Check for breaking changes
+		if diffResult.HasBreakingChanges() {
+			w.log.Info("Breaking changes detected in CRD update", "name", desired.Name, "breakingChanges", len(diffResult.BreakingChanges), "summary", diffResult)
+			return fmt.Errorf("cannot update CRD %s: breaking changes detected: %s", desired.Name, diffResult)
+		}
```
michaelhtm, line 175: could there be a way to override this? if someone doesn't care about the breaking changes

pkg/graph/crd/compat/changes.go
```diff
@@ -0,0 +1,155 @@
...
+	RequiredAdded   ChangeType = "REQUIRED_ADDED"
+	EnumRestricted  ChangeType = "ENUM_RESTRICTED"
+	PatternChanged  ChangeType = "PATTERN_CHANGED"
+
+	// Non-breaking change types
+	PropertyAdded      ChangeType = "PROPERTY_ADDED"
+	DescriptionChanged ChangeType = "DESCRIPTION_CHANGED"
+	DefaultChanged     ChangeType = "DEFAULT_CHANGED"
+	RequiredRemoved    ChangeType = "REQUIRED_REMOVED"
+	EnumExpanded       ChangeType = "ENUM_EXPANDED"
```
michaelhtm, line 38: Maybe we can add number changes as well..I saw a PR approved for adding maximum and minimum to the schema. We should also track those changes..

pkg/graph/crd/compat/changes.go
```diff
@@ -0,0 +1,155 @@
...
+
+	changeDescs := make([]string, 0, maxBreakingChangesSummary)
+
+	for i, change := range r.BreakingChanges {
+		// Cut off the summary if there are too many breaking changes
+		if i >= maxBreakingChangesSummary {
+			remaining := len(r.BreakingChanges) - i
+			if remaining > 0 {
+				changeDescs = append(changeDescs, fmt.Sprintf("and %d more changes", remaining))
+			}
```
michaelhtm, line 93: 💯

pkg/graph/crd/compat/changes.go
```diff
@@ -0,0 +1,155 @@
...
+// License is located at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package compat
```
michaelhtm, line 14: nit:?
```suggestion
package compare
```

michaelhtm (commented):
Great job on this @a-hilaly!! 🚀🚀🚀
I left a few inline comments

pkg/client/crd.go
```diff
@@ -92,27 +94,46 @@ func newCRDWrapper(cfg CRDWrapperConfig) *CRDWrapper {
...
-		if err := w.patch(ctx, crd); err != nil {
+		// CRD exists, check compatibility
+		diffResult, err := crdcompat.DiffSchema(current.Spec.Versions, desired.Spec.Versions)
+		if err != nil {
+			return fmt.Errorf("failed to check schema compatibility: %w", err)
+		}
+
+		// If there are no changes at all, we can skip the update
+		if !diffResult.HasChanges() {
+			w.log.Info("CRD is up-to-date", "name", desired.Name)
```
n3wscott: Might want to make this a trace level comment to reduce the log chatter

pkg/graph/crd/compat/compat.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"fmt"
+
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// DiffSchema compares schema versions and returns compatibility details.
+// This function expects exactly one version in each slice. If more versions are present,
+// it will return an error as multi-version support is not implemented.
+func DiffSchema(oldVersions []v1.CustomResourceDefinitionVersion, newVersions []v1.CustomResourceDefinitionVersion) (*DiffResult, error) {
```
n3wscott: Looking at https://github.com/kubernetes/kubernetes/blob/dd43c3d29d5203378ce456de9466597234a83e66/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/types.go#L178

- removal of sub-resources might matter?
- flopping Deprecated could be bad
- setting served or storage to false would break KRO

pkg/graph/crd/compat/changes.go
n3wscott: I think the current name is ok, it is not comparing, it is attempting to determine compatibility

pkg/graph/crd/compat/schema.go
```diff
@@ -0,0 +1,220 @@
...
+		)
+	}
+
+	// type changes are breaking
+	if oldSchema.Type != newSchema.Type {
+		result.AddBreakingChange(
+			path+".type",
+			TypeChanged,
+			string(oldSchema.Type),
+			string(newSchema.Type),
```
n3wscott: nit: .Type is a string already

pkg/graph/crd/compat/schema.go
```diff
@@ -0,0 +1,220 @@
...
+			}
+		}
+	}
+}
+
+// compareRequiredFields checks for changes to required fields, it only considers
+// existing properties, since new properties are handled in compareProperties.
+func compareRequiredFields(path string, oldSchema, newSchema *v1.JSONSchemaProps, result *DiffResult) {
+	// Use length checks instead of nil checks
+	if len(oldSchema.Required) == 0 && len(newSchema.Required) == 0 {
```
n3wscott, line 154: you might want to test that old and new schema are not nil as well just to protect yourself. It is likely someone attempts to delete the schema to work around something.

pkg/graph/crd/compat/schema.go
```diff
@@ -0,0 +1,220 @@
...
+	existingProps := make(map[string]bool)
+	for propName := range oldSchema.Properties {
+		existingProps[propName] = true
+	}
+
+	// Check for newly required fields ONLY for existing properties (breaking)
+	for req := range newRequiredSet {
+		// Only consider requirements for properties that already existed
+		if existingProps[req] && !oldRequiredSet[req] {
+			result.AddBreakingChange(path+".required", RequiredAdded, "", req)
```
n3wscott, line 172: it is too bad this is not easier to use but Knative made `FieldError` to help with pathed errrors: https://github.com/knative/pkg/blob/b7bbf4be5dbd9a8b8b60e06094bd8930cf241357/apis/field_error.go#L63

n3wscott (approved):
LGTM

pkg/graph/crd/compat/changes.go
a-hilaly: Are those considered breaking changes? should we really inspect the defaults/max/min and basically every attribute in https://kro.run/docs/concepts/simple-schema#supported-markers ? cc @jakobmoellerdev

crenshaw-dev:
Apologies for a very drive-by suggestion, but crdify might be useful to externalize some of the breaking change detection: https://github.com/kubernetes-sigs/crdify

jakobmoellerdev:
@crenshaw-dev I believe crdify is a CLI tool first. does it plan to support a library mode long term? If so lets take a look there!

crenshaw-dev:
@jakobmoellerdev good question... I used it as a go lib for a small project, and it seemed to work fine: https://github.com/crenshaw-dev/miaka/blob/8d568300f998f1028942ceee66d65cace12b8f09/pkg/build/validation/breaking_changes.go#L45

I'm not sure whether they plan to maintain stable APIs for use as a go library, but I asked the question here: https://github.com/kubernetes-sigs/crdify/issues/52

a-hilaly:
Thanks for the suggestion @crenshaw-dev! I actaully looked at crdify before writing this PR. At that time, it didn't strike me as something designed for library usage and looked primarily a CLI tool [global registries, `log.Fatalf` in erorr paths and heavy transitive dependencies]. 

That said, thank you for opening the github issue - happy to reconsider this in favor of using crdify. And since I was here, I just rebased this PR!

pkg/client/crd.go
a-hilaly: in future we could support this via annotations

pkg/graph/crd/compat/compat.go
a-hilaly: ~~kro does hard code there values so i'll skip them for now.~~ Added

crenshaw-dev:
Looks like [crdify is usable as a library](https://github.com/kubernetes-sigs/crdify/issues/52#issuecomment-3632156109)!

pkg/client/crd.go
```diff
@@ -110,52 +112,72 @@ func newCRDWrapper(cfg CRDWrapperConfig) *CRDWrapper {
...
+		}
+
+		// If there are no changes at all, we can skip the update
+		if !report.HasChanges() {
+			log.V(1).Info("CRD is up-to-date", "name", desired.Name)
+			return nil
+		}
+
+		// Check for breaking changes
+		if !report.IsCompatible() {
```
jakobmoellerdev, line 169: from community call: we should have a way to opt out of enforcing that no breaking changes are allowed

pkg/client/crd.go
```diff
@@ -42,7 +44,7 @@ var _ CRDClient = &CRDWrapper{}
 // CRDClient represents operations for managing CustomResourceDefinitions
 type CRDClient interface {
 	// EnsureCreated ensures a CRD exists and is ready
-	Ensure(ctx context.Context, crd v1.CustomResourceDefinition) error
+	Ensure(ctx context.Context, crd v1.CustomResourceDefinition, allowBreakingChanges bool) error
```
a-hilaly, line 47: Not sure happy with boolean params, planning on removing/simplifying this package later on.

PR #354, opened 2025-03-05
feat: add header configuration to allow vanity go imports

### LINKED ISSUE #353: Implement Go vanity Import path for kro (kro.run/)
### Feature Description

Currently, our Go module is being imported via the GitHub URL, which requires developers to use `github.com/kro-run/kro/pkg/...` in their import statements. We want to transition to a custom import path using our own domain, specifically `kro.run/pkg/...`.

To make this happen, we can leverage Go's vanity import path mechanism by adding meta tags to our Docusaurus site. This will allow developers to import our packages using our domain while still hosting the code on GitHub. 

Important links:
- https://pkg.go.dev/cmd/go#hdr-Remote_import_paths
- https://docusaurus.io/docs/api/docusaurus-config#headTags

slashexx:
Fixes #353 

This PR adds support for a Go vanity import path for kro.run/pkg/..., allowing developers to import Kro modules using our own domain instead of the GitHub URL.

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -32,7 +32,23 @@ const config: Config = {
     defaultLocale: "en",
     locales: ["en"],
   },
-
+  headTags: [
+    {
+      tagName: "meta",
+      attributes: {
+        name: "go-import",
+        content: "kro.run/pkg git https://github.com/kro-run/kro",
+      },
+    },
+    {
+      tagName: "meta",
+      attributes: {
+        name: "go-source",
+        content: "kro.run/pkg git https://github.com/kro-run/kro https://github.com/kro-run/kro/tree/main{/dir} https://github.com/kro-run/kro/blob/main{/dir}/{file}#L{line}",
+      },
+    },
+  ],
+  
   presets: [
     [
       "classic",


helm/crds/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -167,7 +167,7 @@ spec:
                       format: date-time
                       type: string
                     message:
-                      description: A human readable message indicating details about
+                      description: A human-readable message indicating details about
```
matthchr: Feels like this change should probably be split from the docusarus changes?

slashexx: Hi matthew !

I actually didn't make that change, it happened when I built the project locally :')

I'll remove those files from tracking and commit once more.

a-hilaly: Yep looks like we forgot to run `make generate` at some point :) 

>I'll remove those files from tracking and commit once more.

That would be great, thanks @slashexx !

slashexx: @a-hilaly Done !

a-hilaly (approved):
Neat, thanks @slashexx !

Issue #347, opened 2025-03-04
Implement SimpleSchema CEL validation support


fsommar:
### Feature Description

**Problem Statement**:
I want to use the [new-ish functionality in CRDs to validate fields with CEL expressions](https://opensource.googleblog.com/2023/11/kubernetes-crd-validation-using-cel.html).

**Proposed Solution**:
Add support in SimpleSchema for CEL validation

**Alternatives Considered**:
Don't support it, or allow merging an existing CRD with the SimpleSchema output. Or allow connecting Kro to an existing CRD and skip SimpleSchema altogether.

**Additional Context**:
BIkeshedding notwithstanding, here's what it could look like to support immutability on a field:
```yaml
immutableFoo: string | validation="self == oldSelf"
```

The `x-kubernetes-validations` field accepts an array, so it should be possibly to include more than one validation in SimpleSchema too. There are also some parsing considerations to keep in mind (e.g. escaping `"`). See [JSONSchemaProps reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#jsonschemaprops-v1-apiextensions-k8s-io) for more details on the field. 

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

MahmoudShakour:
@fsommar i would like to work on this issue, could you assign it to me ?

MahmoudShakour:
@a-hilaly @fsommar 
i have a question regarding having multiple validation rules:
should we separate rules by commas like that:
```
immutableFoo: string | validation="self == oldSelf , self.minReplicas <= self.replicas"
```

fsommar:
> i have a question regarding having multiple validation rules: [...]

I have a weak preference for one validation marker for each validation, as I think it is easier to parse for both humans and machines:

```yaml
immutableFoo: |
  string |
  validation="self == oldSelf"
  validation="self.minReplicas <= self.replicas"
```

note the use of yaml multi-line string. I haven't verified it works with Kro atm, but I believe it should; it depends on how much importance whitespace has

a-hilaly:
This was supported in https://github.com/kubernetes-sigs/kro/pull/484

PR #357, opened 2025-03-05
feat: add validation support in marker type

MahmoudShakour:
fixes: #347

### DIFF
--- a/pkg/simpleschema/markers_test.go
+++ b/pkg/simpleschema/markers_test.go
@@ -124,6 +124,15 @@ func TestParseMarkers(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name:  "validation marker with one value",
+			input: "validation=\"oldself=self\" required=true",
+			want: []*Marker{
+				{MarkerType: MarkerTypeValidation, Key: "validation", Value: "oldself=self"},
+				{MarkerType: MarkerTypeRequired, Key: "required", Value: "true"},
+			},
+			wantErr: false,
+		},
 	}
 
 	for _, tt := range tests {
--- a/pkg/simpleschema/markers.go
+++ b/pkg/simpleschema/markers.go
@@ -50,12 +50,14 @@ const (
 	MarkerTypeMinimum MarkerType = "minimum"
 	// MarkerTypeMaximum represents the `maximum` marker.
 	MarkerTypeMaximum MarkerType = "maximum"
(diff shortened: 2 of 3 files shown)


pkg/simpleschema/markers_test.go
```diff
@@ -124,6 +124,15 @@ func TestParseMarkers(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name:  "validation marker with one value",
```
fsommar, line 128: Can you add a test for multiple validation markers, and a test for using quotes (e.g. `self == "foobar"`)?

fsommar:
I didn't realize it when writing my feature request, but the fact that each [validation is a struct](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#validationrule-v1-apiextensions-k8s-io) with at least rule and typically also message means that this syntax needs to be updated to support that too

MahmoudShakour:
@fsommar 
got it. i had that question in my mind but thought you want the default message to be displayed.
what do you think about this format ? 
```
validation="rule:"self == oldSelf",message:"can't be changed"" 
```

fsommar:
@MahmoudShakour nested quotes is going to be hard to parse for both humans and machines, I think. What about updating the right-hand side to support either strings, like today, or JSON objects? The JSON object suggestion would look something like this:

```yaml
immutableFoo: |
  string |
  validation={
    "rule": "self == oldSelf",
    "message": "can't be changed"
  }
```

Another suggestion, where the index is arbitrary as long as it's unique (could be 0 and 1 also):

```yaml
immutableBar: |
  integer
  validation[immutability].rule="self == oldSelf"
  validation[immutability].message="can't be changed"
  validation[positive].rule="self > 0 || self == -1"
  validation[positive].message="has to be positive or -1"
```

a-hilaly:
Hi @MahmoudShakour - are you still working on this one?

DhairyaMajmudar:
@a-hilaly If @MahmoudShakour is okay with it, I’d be happy to take this on since I have priorly worked with Markers logic in PR https://github.com/kro-run/kro/pull/424

a-hilaly:
please go for it @DhairyaMajmudar - can you please make sure to add @MahmoudShakour as co-author or fork his branch so that we can preserve some of his work.

DhairyaMajmudar:
Thank you, yeah sure will add him as commit co-author

PR #360, opened 2025-03-06
Update website README file

yashisrani:
- I updated website/README.md file to link to docusaurus documentation in case users are using a different package manager than npm.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -5,26 +5,14 @@ This website is built using [Docusaurus](https://docusaurus.io/), a modern stati
 ### Installation
 If you're using other tools than `npm`, you can refer to the `docusaurs documentation` for build and installation https://docusaurus.io/docs/installation#build
 
-if you are using `npm`, then use below command to install all required dependencies.
-
-```
-$ npm install 
-```
-
 If you are using `yarn`, use:
 
 ```
 $ yarn
 ```
-
 ### Local Development
 
 To start a local development server, use:
-
-```
-$ npm run start
-```
-or use:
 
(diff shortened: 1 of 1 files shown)


website/README.md
```diff
@@ -3,13 +3,29 @@
 This website is built using [Docusaurus](https://docusaurus.io/), a modern static website generator.
 
 ### Installation
+Use `npm` or `yarn` to setup and test website on local environment.
```
a-hilaly: There are many ways to run and build this package (i personally build it with `bun`). Instead of linking all the possible tools, shall we just add a link, stating "If you're using other tools than npm, you can refer to the docusaurs documentation for build and installation https://docusaurus.io/docs/installation#build"

website/README.md
```diff
@@ -3,13 +3,29 @@
...
 ### Installation
+If you're using other tools than `npm`, you can refer to the `docusaurs documentation` for build and installation https://docusaurus.io/docs/installation#build
+
+if you are using `npm`, then use below command to install all required dependencies.
+
+```
+$ npm install 
+```
+
+If you are using `yarn`, use:
```
a-hilaly: Now that we have the instruction in L6, let's retain only the yarn installation instruction

yashisrani: The reason behind this PR was that the majority of users (around 80%) use npm to install dependencies. Since most users will likely follow npm commands, removing them from the README might create extra effort for them to look up the commands on the official website. This change ensures a smoother experience and eliminates that struggle.

PR #365, opened 2025-03-08
Remove website dockerfile

yashisrani:
- i improved the Dockerfile of website.


website/Dockerfile
a-hilaly, line 1: Actually, we do not use this docker file for anything. We can remove it.

a-hilaly: can we change this PR to remove the Dockerfile instead?

yashisrani: ok , no problem ..

a-hilaly (approved):
Thank you @yashisrani !

Issue #70, opened 2024-10-30
Support argocd tracking via annotations or labels


a-hilaly:
Add support for ArgoCD resource tracking using annotations or labels as described in the ArgoCD documentation:
https://argo-cd.readthedocs.io/en/latest/user-guide/resource_tracking/#resource-tracking

BerndFarkaDyna:
@a-hilaly : I'd like stand up for trying to implement this feature (as discussed on slack)

denzhel:
This will be an awesome feature !

BerndFarkaDyna:
PR has been submitted #366  now I'm happily waiting for feedback

Executioner1939:
Just hit this issue today while trying to use Kro with ArgoCD, won't be able to use Kro until this is merged. @BerndFarkaDyna thank you for submitting one! Glad the community is actively working on this.

a-hilaly:
Closing in favor of https://github.com/kubernetes-sigs/kro/pull/651
/close

PR #366, opened 2025-03-09
#70 Support ArgoCD Resource tracking

BerndFarkaDyna:
This PR enables Tracking of child resources of a Kro created Resource (#70)

* Implemented propagation of app.kubernetes.io/appinstance label
* implemented propagation of OnwerReferences 

This PR supports the Resources Tracking of ArgoCD via Labels and Annoations

### DIFF
--- a/config/crd/bases/kro.run_resourcegraphdefinitions.yaml
+++ b/config/crd/bases/kro.run_resourcegraphdefinitions.yaml
@@ -167,7 +167,7 @@ spec:
                       format: date-time
                       type: string
                     message:
-                      description: A human readable message indicating details about
+                      description: A human-readable message indicating details about
                         the transition.
                       type: string
                     observedGeneration:
--- a/helm/crds/kro.run_resourcegraphdefinitions.yaml
+++ b/helm/crds/kro.run_resourcegraphdefinitions.yaml
@@ -167,7 +167,7 @@ spec:
                       format: date-time
                       type: string
                     message:
-                      description: A human readable message indicating details about
+                      description: A human-readable message indicating details about
                         the transition.
                       type: string
                     observedGeneration:
--- a/pkg/controller/instance/controller.go
+++ b/pkg/controller/instance/controller.go
@@ -159,6 +159,12 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) error {
 		reconcileConfig:             c.reconcileConfig,
 		// Fresh instance state at each reconciliation loop.
 		state: newInstanceState(),
+		OwnerReference: metav1.OwnerReference{
+			APIVersion: instance.GetAPIVersion(),
+			Kind:       instance.GetKind(),
+			Name:       instance.GetName(),
+			UID:        instance.GetUID(),
+		},
 	}
 	return instanceGraphReconciler.reconcile(ctx)
 }
--- a/pkg/controller/instance/controller_reconcile.go
+++ b/pkg/controller/instance/controller_reconcile.go
@@ -55,6 +55,8 @@ type instanceGraphReconciler struct {
 	reconcileConfig ReconcileConfig
 	// state holds the current state of the instance and its sub-resources.
 	state *InstanceState
+	// owner reference to the owning ResourceGraph Instance
+	OwnerReference metav1.OwnerReference
 }
 
 // reconcile performs the reconciliation of the instance and its sub-resources.
@@ -211,6 +213,11 @@ func (igr *instanceGraphReconciler) getResourceClient(resourceID string) dynamic
 	return igr.client.Resource(gvr)
 }
 
+func applyValues(resource *unstructured.Unstructured, labeler metadata.Labeler, owner metav1.OwnerReference) {
+	labeler.ApplyLabels(resource)
+	resource.SetOwnerReferences([]metav1.OwnerReference{owner})
+}
+
 // handleResourceCreation manages the creation of a new resource
 func (igr *instanceGraphReconciler) handleResourceCreation(
 	ctx context.Context,
@@ -222,7 +229,8 @@ func (igr *instanceGraphReconciler) handleResourceCreation(
 	igr.log.V(1).Info("Creating new resource", "resourceID", resourceID)
 
 	// Apply labels and create resource
-	igr.instanceSubResourcesLabeler.ApplyLabels(resource)
+	applyValues(resource, igr.instanceSubResourcesLabeler, igr.OwnerReference)
+
 	if _, err := rc.Create(ctx, resource, metav1.CreateOptions{}); err != nil {
 		resourceState.State = "ERROR"
 		resourceState.Err = fmt.Errorf("failed to create resource: %w", err)
@@ -267,7 +275,7 @@ func (igr *instanceGraphReconciler) updateResource(
 		"resourceID", resourceID,
 		"delta", differences,
 	)
-	igr.instanceSubResourcesLabeler.ApplyLabels(desired)
+	applyValues(desired, igr.instanceSubResourcesLabeler, igr.OwnerReference)
 
 	// Apply changes to the resource
 	// TODO: Handle annotations
@@ -418,7 +426,7 @@ func (igr *instanceGraphReconciler) setManaged(ctx context.Context, obj *unstruc
 		return nil, fmt.Errorf("failed to set finalizer: %w", err)
 	}
 
-	igr.instanceLabeler.ApplyLabels(copy)
+	applyValues(copy, igr.instanceSubResourcesLabeler, igr.OwnerReference)
 
 	updated, err := igr.client.Resource(igr.gvr).
 		Namespace(obj.GetNamespace()).
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -24,6 +24,7 @@ import (
 const (
 	// LabelKROPrefix is the label key prefix used to identify KRO owned resources.
 	LabelKROPrefix = v1alpha1.KRODomainName + "/"
+	K8sAppPrefix   = "app.kubernetes.io/"
 )
 
 const (
@@ -40,6 +41,8 @@ const (
 	ResourceGraphDefinitionNameLabel      = LabelKROPrefix + "resource-graph-definition-name"
 	ResourceGraphDefinitionNamespaceLabel = LabelKROPrefix + "resource-graph-definition-namespace"
 	ResourceGraphDefinitionVersionLabel   = LabelKROPrefix + "resource-graph-definition-version"
+
+	K8sAppInstanceLabel = K8sAppPrefix + "instance"
 )
 
 // IsKROOwned returns true if the resource is owned by KRO.
@@ -119,16 +122,25 @@ func NewResourceGraphDefinitionLabeler(rgMeta metav1.Object) GenericLabeler {
 	}
 }
 
-// NewInstanceLabeler returns a new labeler that sets the InstanceLabel and
+// NewInstanceLabeler returns a new labeler that sets the K8sComponentId, InstanceLabel and
 // InstanceIDLabel labels on a resource. The InstanceLabel is the namespace
 // and name of the instance that was reconciled to create the resource.
+// The K8s Component Id gets propagated from the given given and only set when
+// available on it.
 func NewInstanceLabeler(instanceMeta metav1.Object) GenericLabeler {
-	return map[string]string{
+	retVal := map[string]string{
 		InstanceIDLabel:        string(instanceMeta.GetUID()),
 		InstanceLabel:          instanceMeta.GetName(),
 		InstanceNamespaceLabel: instanceMeta.GetNamespace(),
 	}
+
+	appInstance, exists := instanceMeta.GetLabels()[K8sAppInstanceLabel]
+	if exists {
+		retVal[K8sAppInstanceLabel] = appInstance
+	}
+	return retVal
 }
+
 
 // NewKROMetaLabeler returns a new labeler that sets the OwnedLabel and
 // KROVersion labels on a resource.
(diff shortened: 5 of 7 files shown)


a-hilaly (commented):
Thank you @BerndFarkaDyna for this PR!I appreciate the thorough implementation with tests.

I have some concerns about using owner references for resource tracking:
1/ Garbage collection and cluster scope resources: When referencing cluster scoped resources, owner references become problematic. Since owner references don't contain namespace information for cluster scoped resources, this can trigger immediate garbage collection of these resources since the reference points to non existing resourcse
2/ multi namespace orchestration: kro supports cross namespace orchestration, which doesn't work well with owner references for similar reasons. Owner references struggle with resources that span multiple namespaces

I think kro needs to perform static analysis on the resource graph before deciding whether to use owner references. We should only apply them when:
- all resources in an RGD are namespace scoped
- all resources are in the same namespace as the instance
- no cross namespace references exist

perhaps we could use owner references conditionally based on graph analysis, with labels/annotations tracking as the fallback for more complex cases. Would you be open to refining the implementation with this approach?

or maybe we should aim for tracking via annotations/labels instead for all RGDs/resources? https://argo-cd.readthedocs.io/en/latest/user-guide/resource_tracking/

BerndFarkaDyna:
@a-hilaly: I think there is a misunderstanding in how ArgoCD is doing the Resource tracking:

The labels / annotations feature only applies to the "root resources" in the ArgoCD application. If someone deploys any ResourceGraphDefinition Instance, those Objects will get the Labels/Annotations by ArgoCD set.

If this meta-data gets propagated to the corresponding child objects of an RGD Instance ArgoCD will show them as orphans and won't be able to show the parent <-> relations between the RGD Instance and created child objects. 

An Example can be seen here where the RGD Instance is creating a service:

With OwnerReference:
image
Without OwnerReference
image
details:
image



There are other projects which have similar problems as described here: https://github.com/argoproj/argo-cd/issues/5082

I'm open to adapting my PR in any way, but adding OnwerReferences is the only way to go if proper ArgoCD support is desired (at the moment), at least in terms of how I understand it.

BerndFarkaDyna:
@a-hilaly : can you please share any inputs on this topic from your point of view?

RafPe:
There is another PR open in the argocd community that could potentially help https://github.com/argoproj/argo-cd/pull/20908

thomaspetit:
@RafPe When creating a CR, I would generally expect the underlying resources to be linked via ownerReferences, as per standard Kubernetes behavior — no ArgoCD-specific handling should be needed here.

This is commonly used by operators like the Prometheus Operator. For example, the generated Secret includes an ownerReferences entry pointing back to the Prometheus CR:

https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

ArgoCD should respect these relationships during reconciliation and display them correctly without disrupting the GitOps flow. Let me know if you're seeing different behavior. Happy to help debug further if needed.

image

```
apiVersion: v1
data:
  prometheus.yaml.gz: ++++++++
kind: Secret
metadata:
  creationTimestamp: '2025-05-02T13:48:15Z'
  labels:
    managed-by: prometheus-operator
  name: prometheus-prometheus
  namespace: prometheus
  ownerReferences:
    - apiVersion: monitoring.coreos.com/v1
      blockOwnerDeletion: true
      controller: true
      kind: Prometheus
      name: prometheus
      uid: 1ab4a260-4a67-48af-8e6e-a35134027680
  resourceVersion: '39055'
  uid: 66a98323-0fbc-45d0-8412-47b9d1db17d1
type: Opaque
```

From the looks of it, the right approach is [implemented](https://github.com/kro-run/kro/pull/366/files#diff-8d2eccde733f8cd3afc6c96de351578b5cb840d3748b0f51301a18ef0ce61556).. or maybe I'm missing something of course

RafPe:
> @RafPe When creating a CR, I would generally expect the underlying resources to be linked via ownerReferences, as per standard Kubernetes behavior — no ArgoCD-specific handling should be needed here.
> 
> This is commonly used by operators like the Prometheus Operator. For example, the generated Secret includes an ownerReferences entry pointing back to the Prometheus CR:
> 
> https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/
> 
> ArgoCD should respect these relationships during reconciliation and display them correctly without disrupting the GitOps flow. Let me know if you're seeing different behavior. Happy to help debug further if needed.
> 
> image
> 
> ```
> apiVersion: v1
> data:
>   prometheus.yaml.gz: ++++++++
> kind: Secret
> metadata:
>   creationTimestamp: '2025-05-02T13:48:15Z'
>   labels:
>     managed-by: prometheus-operator
>   name: prometheus-prometheus
>   namespace: prometheus
>   ownerReferences:
>     - apiVersion: monitoring.coreos.com/v1
>       blockOwnerDeletion: true
>       controller: true
>       kind: Prometheus
>       name: prometheus
>       uid: 1ab4a260-4a67-48af-8e6e-a35134027680
>   resourceVersion: '39055'
>   uid: 66a98323-0fbc-45d0-8412-47b9d1db17d1
> type: Opaque
> ```
> 
> From the looks of it, the right approach is [implemented](https://github.com/kro-run/kro/pull/366/files#diff-8d2eccde733f8cd3afc6c96de351578b5cb840d3748b0f51301a18ef0ce61556).. or maybe I'm missing something of course

On my end I am using tracking of resources via annotations like 

```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    argocd.argoproj.io/tracking-id: nfs:apps/Deployment:storage/nfs-provisioner
    deployment.kubernetes.io/revision: "1"
  creationTimestamp: "2025-05-20T08:28:23Z"
  generation: 2
  name: nfs-provisioner
  namespace: storage
  resourceVersion: "2058"
  uid: 7713ebba-70b4-47a8-8d75-40f8c8e39397
spec:
``` 

This does not break any principal of ownerreferences and makes ArgoCD specific tracking agnostic from ownerRef ( concern also raised by @a-hilaly  ) , simply by leveraging annotations.

thomaspetit:
@RafPe Thanks for sharing this one! I learned something new you could do 😄

RafPe:
> @RafPe Thanks for sharing this one! I learned something new you could do 😄

Happy i could share some insights!  

I really hope the information we gather here can help KRO community to move this issue as not having properly tracked resources in Argo makes it really difficult to use.

thomaspetit:
@RafPe I agree, this should be straightforward. In its current form, it’s quite difficult to use in ArgoCD. It’s important that developers using these graphs can clearly see what’s happening so the platform team isn’t inundated with questions

@a-hilaly, are there any remaining open points on this implementation? I haven’t identified any in the code created by @BerndFarkaDyna

RafPe:
@a-hilaly this one is open for quite a while - can you let us know if there are any changes outstanding on this PR that we could help to address to get the functionality implemented ? Would love to get to use the powers of Argo in combination with Kro

n3wscott:
The Kubernetes recommended label is `app.kubernetes.io/instance` [ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

Executioner1939:
Hi, what can we do to get this merged? I am blocked from using KRO at my company until this is fixed.

HEP85:
Please merge this. We evaluated kro in the company and not having this feature is a non-starter for us, as it makes everything a "blackbox" for the developers. Right now putting an eye on Crossplane v2 preview, but still hoping kro makes some progress here...

n3wscott:
> Please merge this. We evaluated kro in the company and not having this feature is a non-starter for us, as it makes everything a "blackbox" for the developers. Right now putting an eye on Crossplane v2 preview, but still hoping kro makes some progress here...

I don't think it is currently mergable because this code path does not take into account that an RGD could compose a cluster level resource from a namespace resource, but that owner link is not valid in Kubernetes. Also, in KRO I can make cross Namespace resources, which is also not a valid ownership link.

RafPe:
> The Kubernetes recommended label is `app.kubernetes.io/instance` [ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

Yes that may be - but as we can see this does not really play well in the light of using the framework with ArgoCD. Maybe we are overcomplicating this setup and could just think of feature where we enable downsttream propagation of annotations - hence we are then enabling teams to use ArgoCD tracking via annotations if desired and does not close the door the implementing owner reference as mentioned ? 

I'm open for other suggestions , paths or form of help that I could offer so we could implement this feature

HEP85:
> > The Kubernetes recommended label is `app.kubernetes.io/instance` [ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
> 
> Yes that may be - but as we can see this does not really play well in the light of using the framework with ArgoCD. Maybe we are overcomplicating this setup and could just think of feature where we enable downsttream propagation of annotations - hence we are then enabling teams to use ArgoCD tracking via annotations if desired and does not close the door the implementing owner reference as mentioned ?
> 
> I'm open for other suggestions , paths or form of help that I could offer so we could implement this feature

I like the suggestion. We could still use ownerReferences for the namespace-level resources and for the others just have an option to propagate what ArgoCD injects anyways?

n3wscott:
> hence we are then enabling teams to use ArgoCD tracking via annotations if desired and does not close the door the implementing owner reference as mentioned ?
> 
> I'm open for other suggestions , paths or form of help that I could offer so we could implement this feature

I think it is totally reasonable to own resources from the kro rgd instance downward, it just needs to be filtered based on if the rgd is cluster or namespace scoped, and then if namespace scoped, only same namespace resources. An integrator can make a stance that they can't compose resources that would create children that have no owners ref as part of their adoption story. 

Re: the recommended or not label. I think kro should only pick labels that are the current k8s recommendations but that should not stop an RGD author from using an additional label or annotation in the template...

jakobmoellerdev:
Following up after a [slack ping](https://kubernetes.slack.com/archives/C081TMY9D6Y/p1756378957945779) and some talks with @fabianburth. I support @n3wscott and @a-hilaly:

1. Tracking resources via labels like `app.kubernetes.io/instance` is already possible: arbitrary metadata can be passed into the graph and cross-referenced.
2. Owner references can be set manually in the RGD today, so the absence of automatic handling is not a blocker.
3. Automatic owner references are not just metadata—they change GC behavior. If you prepare a PR for this feature, it
    - must be opt-in
    - thoroughly tested
    - not break cross-namespace orchestration.
    - All this is not present here (That’s why the PR cannot be merged in its current form.)
4. To progress: keep rebasing your PR, join community calls for live discussion, and consider submitting a [design proposal](https://github.com/kro-run/kro/tree/main/docs/design/proposals) so we can work on something together 🙌 .

RafPe:
> Following up after a [slack ping](https://kubernetes.slack.com/archives/C081TMY9D6Y/p1756378957945779) and some talks with @fabianburth. I support @n3wscott and @a-hilaly:
> 
> 1. Tracking resources via labels like `app.kubernetes.io/instance` is already possible: arbitrary metadata can be passed into the graph and cross-referenced.
> 2. Owner references can be set manually in the RGD today, so the absence of automatic handling is not a blocker.
> 3. Automatic owner references are not just metadata—they change GC behavior. If you prepare a PR for this feature, it
>    
>    * must be opt-in
>    * thoroughly tested
>    * not break cross-namespace orchestration.
>    * All this is not present here (That’s why the PR cannot be merged in its current form.)
> 4. To progress: keep rebasing your PR, join community calls for live discussion, and consider submitting a [design proposal](https://github.com/kro-run/kro/tree/main/docs/design/proposals) so we can work on something together 🙌 .

I just wanted to follow up here an clarify a little bit the context for my drive of this PR. Since I use ArgoCD in majority of cases to manage the kubernetes resources/apps ability to track it is extremely important. 

Since in ArgoCD the default tracking method is done via annotations ( https://argo-cd.readthedocs.io/en/stable/user-guide/resource_tracking/ ) my ultimate goal is to have the annotation `argocd.argoproj.io/tracking-id` replicated downstream to instances of my resource. 

As a solution that works in the meantime for any1 using ArgoCD and planning to use KRO - you can use the following syntax to replicate the annotation

```
annotations:
  argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"]}
```

Big kudos for helping on this one to @fabianburth

jakobmoellerdev:
Glad to see that there are some alternatives here. Considering this, if you want we can always add a FAQ part for argo users. Considering this ability, I would propose to close this PR out though. Happy to hear other thoughts

RafPe:
> Glad to see that there are some alternatives here. Considering this, if you want we can always add a FAQ part for argo users. Considering this ability, I would propose to close this PR out though. Happy to hear other thoughts

I can add a PR for that - no problem!

jakobmoellerdev:
/close 
Closing this PR as we have documentation in place for how to integrate with Argo

crenshaw-dev:
> Since in ArgoCD the default tracking method is done via annotations

Argo CD maintainer here: I just want to clarify that Argo CD does **its own** tracking via annotations. These annotations are explicitly _not_ meant to be constructed or maintained by users. The proper way to have Argo CD track a resource created by another resources is to set an `ownerRef`, as [discussed](https://github.com/kubernetes-sigs/kro/pull/366#issuecomment-3233811161) above. You can manually set the tracking annotation, and it may cause Argo CD to behave in a way that works for your use case. But it's a hack and is at your own risk.

RafPe:
> > Since in ArgoCD the default tracking method is done via annotations
> 
> Argo CD maintainer here: I just want to clarify that Argo CD does **its own** tracking via annotations. These annotations are explicitly _not_ meant to be constructed or maintained by users. The proper way to have Argo CD track a resource created by another resources is to set an `ownerRef`, as [discussed](https://github.com/kubernetes-sigs/kro/pull/366#issuecomment-3233811161) above. You can manually set the tracking annotation, and it may cause Argo CD to behave in a way that works for your use case. But it's a hack and is at your own risk.

@crenshaw-dev so the only way for us to properly have this as a supported setup is to use ownerRef on the resources that are being created from the RGD instance ? I know there are cross-namespaces constraints so on this side as when we would be creating resources in multiple namespaces ( CC @a-hilaly / @jakobmoellerdev )

Issue #138, opened 2024-11-14
Mega Feature: Add kro cli


candonov:
### Feature Description

**Problem Statement**:
Add a cli for kro and integrate it with `kubectl`.

**Proposed Solution**:

The cli should support a diff functionality using `--dry-run`:
```
kubectl kro apply -f <updatdRG>.yaml --dry-run
kubectl kro apply -f <updatedInstance>.yaml --dry-run
```
It should also show the chain dependent resources of an instance and their statuses: 
```
NAME                                  STATE    SYNCED   AGE
webappInstance                        ACTIVE   True     42h
       ├── deployment                 ACTIVE   True     42h
       ├── service                    ACTIVE   True     42h
       └── ingress                    ACTIVE   True     42h
```

**Alternatives Considered**:

**Additional Context**:

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

amirahav:
We believe that the biggest benefits of having the CLI would be around schema validation, linting of the expression language, having an ability of doing a dry-run on the ResourceGroup as well as an ability to produce the complete CRD that would be generated by the ResourceGroup.

In addition, it would be beneficial if there would be a way to view the effects of changes to the `ResourceGroup` on the instances prior to the `ResourceGroup` changes being made. eg, there is a proposed change to the `ResourceGroup` it would be very helpful to know how that change would affect all the instances of that `ResourceGroup`.

ikchifo:
I'm interested in working on this. @a-hilaly @candonov

darrenlavery91:
I would like to work on this also, @candonov I would also like to work on ansible plugins, once the cli bit works great...

PR #371, opened 2025-03-10
feat: Fine-grained RBAC in the Helm chart

### LINKED ISSUE #230: Fine-grained RBAC controls in the Helm chart
### Feature Description

**Problem Statement**:
The Helm chart currently creates a cluster role with full permissions on everything, and a cluster role binding to the KRO service account for that cluster role; while practical, it is also going to be way overspecified in most production environments. The Helm chart should support more fine-grained control of permissions assigned to KRO - or, at least, support turning off the super-admin permissions and allow the user to manage RBAC separately.

**Proposed Solution**:
A parameter to the Helm chart that turns off cluster-level RBAC for KRO, with documentation for how to set up permissions manually.

**Alternatives Considered**:
The Helm chart could also be given parameters for exactly what permissions to include in the role, but that is probably no easier to specify than just using the RBAC resources directly, so not sure if it's worth the complexity in the chart.

tomasaschan:
This PR introduces a new scheme for configuring RBAC through the Helm chart.

The basic approach mimics [what Kyverno does](https://kyverno.io/docs/installation/customization/#role-based-access-controls), so there is some industry precedent for this approach, and the aggregation feature it relies on is [long-time supported in Kubernetes](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles).

To enable this in your deployment, set

```yaml
rbac:
  legacy: false
```

in your `values.yaml` file. You'll then need to create `ClusterRole`s with the label `rbac.kro.run/aggregate-to-controller: "true"` to grant any access you want to give Kro outside the bare-minimum set of permissions the Helm chart will now include.

Fixes #230.

### DIFF
--- a/helm/templates/cluster-role.yaml
+++ b/helm/templates/cluster-role.yaml
@@ -4,12 +4,12 @@ metadata:
   annotations:
     kubernetes.io/description: |
 {{- if .Values.rbac.legacy }}
-      This ClusterRole grants cluster-wide admin access for the Kro controller.
+      This ClusterRole grants cluster-wide admin access for the kro controller.
 {{- else }}
-      This ClusterRole grants access for the Kro controller.
+      This ClusterRole grants access for the kro controller.
 
-      The Kro installation comes with enough for the controller to run, and start watching the relevant resources
-      (e.g. ResourceGraphDefinitions), but in order to allow Kro to create and manage other types of resoruces you
+      The kro installation comes with enough for the controller to run, and start watching the relevant resources
+      (e.g. ResourceGraphDefinitions), but in order to allow kro to create and manage other types of resoruces you
       must manually grant access to those separately.
 
       The easiest way to accomplish this is to create a new ClusterRole with the label `rbac.kro.run/aggregate-to-controller: "true"`
@@ -37,7 +37,7 @@ kind: ClusterRole
 metadata:
   annotations:
     kubernetes.io/description: |
-      This ClusterRole grants access for the Kro controller to resources it always needs access to.
+      This ClusterRole grants access for the kro controller to resources it always needs access to.
   labels:
     {{- include "kro.labels" . | nindent 4 }}
     rbac.kro.run/aggregate-to-controller: "true"
--- a/helm/templates/cluster-role.yaml
+++ b/helm/templates/cluster-role.yaml
@@ -15,10 +15,10 @@ metadata:
       The easiest way to accomplish this is to create a new ClusterRole with the label `rbac.kro.run/aggregate-to-controller: "true"`
       which will be automatically aggregated into this one.
 {{- end }}
-  name: {{ include "kro.fullname" . }}{{- if .Values.rbac.legacy -}}-cluster-role{{- else -}}:controller{{- end }}
+  name: {{ include "kro.fullname" . }}{{- if eq .Values.rbac.mode "unrestricted" -}}-cluster-role{{- else -}}:controller{{- end }}
   labels:
     {{- include "kro.labels" . | nindent 4 }}
-{{ if .Values.rbac.legacy -}}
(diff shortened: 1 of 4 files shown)


helm/values.yaml
```diff
@@ -31,6 +31,9 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name: ""
 
+rbac:
+  legacy: true
```
a-hilaly: how about we use a different name than `.Values.rbac.legacy`? perhaps something like `rbac.fullAccess`/`rbac.restrictive/rbac.permissive`

since kro is still in alpha, the term "legacy" implies the current approach is outdated. we could use a slightly more neutral terminology to frame these as two valid configuration options during the alpha phase rather than suggesting one is deprecated.

Otherwise, the implementation looks clean :)

helm/templates/cluster-role.yaml
```diff
@@ -1,13 +1,72 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ include "kro.fullname" . }}-cluster-role
+  annotations:
+    kubernetes.io/description: |
+{{- if .Values.rbac.legacy }}
+      This ClusterRole grants cluster-wide admin access for the Kro controller.
+{{- else }}
+      This ClusterRole grants access for the Kro controller.
```
a-hilaly: tiny nit: we "try" to standardize on using `kro` (lowercase) throughout our documentation and code.
```suggestion
      This ClusterRole grants access for the kro controller.
```

(I just did a quick search and it looks like we still using `Kro` in some other areas of the code base.. we'll have to fix that)

a-hilaly (commented):
Awesome, thank you @tomasaschan !

I left two comment below, PTAL

helm/values.yaml
tomasaschan: Yeah, I felt already when creating this that the naming here isn't great. I'll come up with something less value-laden!

tomasaschan: Revamped this to be `rbac.mode: unrestricted` (current behavior) or `rbac.mode: aggregation` (new, recommended, behavior) instead; that's more descriptive and less value-laden.

tomasaschan:
I realized that the static role might not quite contain all it needs yet - for example, permission to manage CRDs is missing. I'm gonna run some tests to see what breaks when I try to actually run this with a few RGDs in place.

helm/values.yaml
```diff
@@ -31,6 +31,18 @@ serviceAccount:
...
+  # Specifies how to manage access for kro in the cluster
+  # Two modes are currently supported:
+  # - unrestricted: Grants cluster-wide full access to all resources
+  # - aggregation: Grants a minimal set of permissions for kro to manage its own
+  #                resources, and additionally aggregates ClusterRoles with the
+  #                label `rbac.kro.run/aggregate-to-controller: "true"` to allow
+  #                kro users to extend kro's permissions.
+  # Recommended setting is aggregation, but unrestricted remains the default for
+  # backwards compatibility.
+  mode: unrestricted
```
a-hilaly, line 44: Love this, i'm thinking in the future we can add another option `impersonation-only` that only kro permissions to impersonate other service accounts and delegates that to admins.

tomasaschan:
@a-hilaly Unless you have further comments or suggestions, I think this is ready to go :)

a-hilaly:
Thank you @tomasaschan!

Slightly related, i think this will play nicely with https://github.com/a-hilaly/kro/blob/kro-cli/cmd/kubectl-kro/generate/rbac/rbac.go.

tomasaschan:
@a-hilaly That looks very nice!

Any chance you'd wanna do whatever patches you want to those CLI features and send them as a PR to here, so they don't just live in your fork? I think that'd make it easier to ensure there's not too much double-work happening in the context of #138.

PR #373, opened 2025-03-10
Fix for issue 359

burdena:
https://github.com/kro-run/kro/issues/359

### DIFF
--- a/pkg/graph/emulator/emulator.go
+++ b/pkg/graph/emulator/emulator.go
@@ -82,6 +82,10 @@ func (e *Emulator) generateValue(schema *spec.Schema) (interface{}, error) {
 	}
 
 	if enabled, ok := schema.VendorExtensible.Extensions["x-kubernetes-preserve-unknown-fields"]; ok && enabled.(bool) {
+		// Handle x-kubernetes-preserve-unknown-fields with no properties or type
+		if len(schema.Properties) == 0 && schema.Type == nil {
+			return e.generateString(schema), nil
+		}
 		// Handle x-kubernetes-preserve-unknown-fields
 		return e.generateObject(schema)
 	}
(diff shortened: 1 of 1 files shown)


a-hilaly:
Hi @burdena, thank you for your contribution! can you please confirm if this is still an issue with one of the nightly builds? https://github.com/kro-run/kro/pkgs/container/kro%2Fkro. If it's still an issue can we add some unit tests to validate that this is working?

a-hilaly:
also, quick suggestion: to help other reviewers understand your changes better, could you please update your PR with a more descriptive title and body?

pkg/graph/emulator/emulator.go
```diff
@@ -137,7 +141,7 @@ func (e *Emulator) generateObject(schema *spec.Schema) (map[string]interface{},
 	if schema == nil {
 		return nil, fmt.Errorf("schema is nil")
 	}
-
+
```
a-hilaly, line 144: Can you please run a gofmt on this file?

pkg/graph/emulator/emulator.go
```diff
@@ -82,6 +82,10 @@ func (e *Emulator) generateValue(schema *spec.Schema) (interface{}, error) {
 	}
 
 	if enabled, ok := schema.VendorExtensible.Extensions["x-kubernetes-preserve-unknown-fields"]; ok && enabled.(bool) {
+		// Handle x-kubernetes-preserve-unknown-fields with no properties or type
+		if len(schema.Properties) == 0 && schema.Type == nil {
+			return e.generateString(schema), nil
+		}
```
a-hilaly, line 88:
```suggestion
		// Handle x-kubernetes-preserve-unknown-fields with no properties or type
		if len(schema.Properties) == 0 && len(schema.Type) == 0 {
			return e.generateString(schema), nil
		}
```

a-hilaly (commented):
Thanks @burdena - looks like indeed this is still an issue. I'm curious to known which CRD are you trying to use with kro? haven't seen this case before :)

Can you please provide a description and title for this PR? these are used to generate our changelog

burdena:
@a-hilaly https://github.com/dapr/dapr/blob/master/charts/dapr/crds/components.yaml. 

This seems to be working in 0.2.2, https://github.com/kro-run/kro/pull/325/files fixes it ?

a-hilaly:
@burdena indeed looks like @jknutson fixed it in #325 :). Also, i'm glad to see dapr and kro usage!

Issue #361, opened 2025-03-06
Add support for custom functions


fsommar:
### Feature Description

**Problem Statement**:
I have repeat transformations that I want to do in various places, such as transforming a Google Cloud Project (GCP) region into a GCP zone (effectively a lookup table), or concatenating strings to create identifiers.

**Proposed Solution**:
Introduce a new spec-level field called `functions` (I have no preferences on naming convention) that looks something like the following:

```yaml
functions:
- id: greeting
  inputs:
  - string
  - string
  template: ${_0 + " " + _1 + "!"}
```

It can then be used as such

```yaml
resources:
- id: configmap
  template:
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: ${schema.metadata.name}
    data:
      greeting: ${greeting("Hello", "world")}
```

Which results in

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: foobar
data:
  greeting: Hello world!
```

**Alternatives Considered**:
Support hardcoded lookup tables as a map from string to string. This only solves a small subset of use cases.

**Additional Context**:
I haven't looked too much into the possibilities of CEL, but it supports maps and objects, so it could theoretically be possible to have a function return an object and have that reified as a yaml object. Is it feasible in the context of Kro? Supporting complex objects could definitely be a v2, starting with supporting primitive arguments. It may also be dependent on the approach chosen in https://github.com/kro-run/kro/issues/144.

With object support, it could be used to solve the use cases highlighted in https://github.com/kro-run/kro/issues/17 to expand input into a list in the output (such as the containers field in Deployment kinds) as opposed to expanding input into a list of resources.

Here's a PoC to prove the feasibility of dynamically registering functions
```go
package main

import (
	"fmt"
	"log"

	"github.com/google/cel-go/cel"
	"github.com/google/cel-go/common/types"
	"github.com/google/cel-go/common/types/ref"
)

type Function struct {
	id       string
	types    []string
	template string
}

func (f *Function) GetCelFunction() (cel.EnvOption, error) {
	envOpts := make([]cel.EnvOption, len(f.types))
	celTypes := make([]*cel.Type, len(f.types))
	for i, inputType := range f.types {
		celType, err := getCelType(inputType)
		if err != nil {
			return nil, err
		}

		envOpts[i] = cel.Variable(fmt.Sprintf("_%d", i), celType)
		celTypes[i] = celType
	}

	env, err := cel.NewEnv(envOpts...)
	if err != nil {
[...]

PR #375, opened 2025-03-11
Support custom functions

HeeManSu:
#361 

- [x] Implement support for custom functions
- [ ] Write tests 
- [ ] Update docs

### DIFF
--- a/pkg/cel/function/function.go
+++ b/pkg/cel/function/function.go
@@ -1,0 +1,108 @@
+package function
+
+import (
+	"fmt"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	xv1alpha1 "github.com/kro-run/kro/api/v1alpha1"
+)
+
+// FunctionRegistry manages custom CEL functions
+type FunctionRegistry struct {
+	functions map[string]*xv1alpha1.Function
+}
+
+// NewFunctionRegistry creates a new function registry
+func NewFunctionRegistry() *FunctionRegistry {
+	return &FunctionRegistry{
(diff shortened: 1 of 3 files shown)


HeeManSu:
@fsommar Can you please check the implementation?

pkg/cel/function/function.go
```diff
@@ -0,0 +1,108 @@
...
+	case "double", "float":
+		return cel.DoubleType, nil
+	case "bytes":
+		return cel.BytesType, nil
+	case "any":
+		return cel.AnyType, nil
+	case "list":
+		return cel.ListType(cel.AnyType), nil
+	case "map":
+		return cel.MapType(cel.StringType, cel.AnyType), nil
```
fsommar, line 104: If we're ok with making backwards incompatible changes to this (@a-hilaly your input please), we can start including list and map. If not, I suggest we start with just the primitives and add support for lists and maps later, probably when the design for complex types https://github.com/kro-run/kro/issues/144 is finished

a-hilaly (commented):
thanks for your contribution @HeeManSu. While I see this feature will be useful, it's important that we align and have consensus on how to implement it. we also need to clearly define the scope as it might impact many other features in kro. Please share your thoughts first in a GH issue, and once it's accepted and validated by maintainers/community, we need it to be assigned according to our contribution process before implementation.

pkg/cel/function/function.go
a-hilaly: @fsommar given that kro is experimental, we're very open to aggressively introducing breaking changes :) I really like the lambda idea.. really cool.  I'm wondering about the right place for these definitions.. does it make sense to put it in an RGD or should we consider a separate abstraction? I'm also curious if we can come up with better semantics. Let's pick this up in the main thread/next. community meeting

Issue #372, opened 2025-03-10
Simplify Developer Tooling Setup in the repo


tomasaschan:
### Feature Description

**Problem Statement**:
There is currently _a lot_ going on in the repo, tooling wise, and there's quite a lot of duplicated source of truths for e.g. installation manifests (should I look at `config` or `helm`?).

Furthermore, running the various tools, building the project, etc, requires not only figuring out which make targets are relevant ones vs implementation details, but also sometimes which ones have implicit dependencies on each other (e.g. the image-publishing ones don't have a dependency declared on the image-building ones...).

There are also some tooling versions managed in the Makefile itself, which is bound to cause issues when folks forget to update those tools accordingly (the binary management in the Makefile simply detects existence, but not versions...).

**Proposed Solution**:

If we take these actions, we should be able to get rid of the Makefile entirely:
* [ ] Port references to the third-party tools to be installable as Go 1.24 `go tool`s to simplify version management
* [ ] Port as many as possible of the make targets to instead be invoked by e.g. `go generate ./...` or `go test ./...`
* [ ] Port remaining make targets, e.g. to build and deploym Kro, into plain shellscripts with clear and documented usage instructions

We should also ensure to
* [ ] Implement a way to generate the Helm chart based on output from `controller-tools`, so that things like RBAC and metrics/webhooks configuration is automatically kept up-to-date in the chart versus what the controller code expects

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue

PR #376, opened 2025-03-11
feat: Create CLI skeleton

tomasaschan:
This is a start at tackling #138, but by no means even close to implementing the full feature set.

However, as both I and others find this an interesting thing to work on, and the mega-feature is big enough that it probably should not be tackled in a single PR anyway, this is a start that opens up for collaborating by splitting the work by subcommands etc.

My intention is that we could implement, at least, the following commands (API of course subject to future discussion!):

```sh
# validate a RGD to ensure CEL expressions etc are valid
kubectl kro validate resource-graph-definition --resource-graph-definition rgd.yaml

# validate an instance against an RGD
kubectl kro validate instance --resource-graph-definition rgd.yaml --instance instance.yaml

# take a RGD, validate its definition, and output the CRD it would create to stdout
kubectl kro render crd --resource-graph-definition rgd.yaml

# take a RGD and an instance, and output the resulting resources to stdout
kubectl kro render instance --resource-graph-definition rgd.yaml --instance instance.yaml

# inspect a live instance in a cluster, and output the rendered resource graph
kubectl kro inspect --namespace my-namespace my-resource-type my-resource
```


tomasaschan:
CC @ikchifo @darrenlavery91 @candonov

cmd/kubectl-cro/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"os"
+
+	"github.com/kro-run/kro/cmd/kubectl-cro/render"
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "kubectl-cro",
```
a-hilaly: How about just `kro` for a CLI name? also curious why cro and not kro?

cmd/kubectl-cro/main.go
a-hilaly, line 1: You can use https://github.com/kro-run/kro/blob/main/hack/boilerplate.sh to inject License headers for the Go files. I'm thinking we could include this in the Makefile

a-hilaly (commented):
Thank you @tomasaschan ! left a few comments inline.

Also worth taking a look at this POC i've buitl sometime ago https://github.com/a-hilaly/kro/tree/kro-cli/cmd/kubectl-kro

a-hilaly:
@tomasaschan also can we start by introduce the root command first, and port the conversation about CLI verbs/nouns to the github issue?

cmd/kubectl-cro/main.go
tomasaschan: `kubectl-kro` was the intention - the `c` is a typo :)

I basically just took that from the issue, which used the CLI as a `kubectl` extension. But I have no strong opinions (in fact I might also prefer to just call it `kro`, and have it live side by side with `kubectl` rather than "within" it).

a-hilaly: Sweet! I'm not set on any specific way either. I've been using `kubectl kro` (k kro) for a while now and it feels a bit clunky. I was thinking what if we just call it "kro" and let people decide for themselves if they want to use it as a kubectl plugin or standalone?

cmd/kubectl-cro/main.go
tomasaschan: I would love for it if we could work toward getting rid of the Makefile 😅 See #372 for some discussion in that direction. Maybe something like [`nwa`](https://github.com/B1NARY-GR0UP/nwa) could be utilized as a `go tool` instead?

This is getting OT for this PR though; I'll make sure to add the license header to all files I add, and we can discuss mechanics in #372.

cmd/kubectl-cro/main.go
tomasaschan: Making it runnable as a plugin is just a matter of renaming the binary - but it might also have implications on how we desing the CLI itself; most plugins tend to accept the same types of flags as `kubectl` itself (e.g. `--namespace`, `--context`, perhaps `--watch`, etc) so if we want it to _feel_ like a plugin we should probably decide that upfront.

I'm not convinced we want that, though, so starting out as `kro` is fine by me!

tomasaschan:
> also can we start by introduce the root command first

@a-hilaly Are you saying you want me to remove the `render crd` command from here, so we can merge just the root? Sure. I was planning on working on the `render crd` feature in our hackweek, but I can split it out :)

a-hilaly:
>@a-hilaly Are you saying you want me to remove the render crd command from here, so we can merge just the root? Sure. I was planning on working on the render crd feature in our hackweek, but I can split it out :)

@tomasaschan Ideally in seperate PRs if possible so that we can iterate and move fast with multiple commands in parallel! On the `render` command, i personally prefer `generate` but no strong opinions here. I was thinking long term, we can have
```
kro generate crd
kro generete instance
kro generate rbac
kro generate docs
kro generate diagram
```
@justinsb suggested to add a `kro generate diagram` which is super interesting we could generate images/svgs like: https://github.com/kro-run/kro/blob/main/examples/gcp/eventarc/gcsbucket-with-finalizer-trigger.png !

a-hilaly (dismissed):
Thanks @tomasaschan ! 👍

a-hilaly:
@tomasaschan  for future PRs, feel free to port/use some of the stuff in https://github.com/a-hilaly/kro/tree/kro-cli/cmd/kubectl-kro - i never had the time to fully refine it and submit a PR for it.

a-hilaly (approved):
Thank you @tomasaschan !

Issue #377, opened 2025-03-11
Support referencing metadata fields in RGD


tomasaschan:
### Feature Description

**Problem Statement**:
When defining a new resource type, the CEL expressions today can only reference fields specified in the `spec.schema` section of the RGD. This is severely limiting in several ways:

* It requires the user to explicitly specify a name for their instance, even though they already have one in `.metadata.name`, if any additional logic is to be applied to naming (e.g. deriving three new resource names based on the instance), since there is no way for the CEL expressions that generate new resources to reference the metadata properties of the instance itself

* It makes it impossible to reference fields not explicitly specified in the schema or subresource definition, but which we know will be there - e.g. one cannot set `status.created: ${myresource.metadata.creationTimestamp}` to have the creation timestamp of the underlying resource be displayed in the instance status, because `creationTimestamp` is not a known field on the subresource (and we can't add it to the resource definition, as creating resources with it set is not supported).

**Proposed Solution**:
The instance `metadata` node (with underlying type `ObjectMeta` and the schema that follows from there) is exposed to the CEL expressions in the resource graph under some appropriate top-level name.

Furthermore, the resource graph entries also expose _their_ `metadata` under that name, so that `myresource.metadata` works, even if `metadata` only has e.g. `name` specified in the resource definition.

**Additional Context**:
Here's a sample RGD that uses these features, that I would like to have working but it does not:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: using-metadata
spec:
  schema:
    apiVersion: v1alpha1
    kind: UsingMetadata
    spec:
      data:
        foo: string
        bar: integer
    status:
      created: ${configmap.metadata.creationTimestamp} # doesn't work - but should!
  resources:
    - id: configmap
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${instance.metadata.name + "-cm"} # doesn't work - but should!
        data:
          foo: ${schema.spec.data.foo}
          bar: ${string(schema.spec.data.bar)}
```
* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

SanjayUG:
👍🏻

Yes, i am glad and interested in working on this feature.

a-hilaly:
[...]

PR #378, opened 2025-03-11
fix: Support referencing metadata fields in RGD. #377

SanjayUG:
Hi developer, 

This contribution fixes the #377 issue,

Support metadata field access in RGD CEL expressions
Adds support for referencing metadata fields in Resource Graph Definition (RGD) CEL expressions. Users can now access:
- Instance metadata via ${instance.metadata.*}
- Resource metadata via ${resourceId.metadata.*}
- Standard K8s metadata fields (name, namespace, creationTimestamp, etc.)

### DIFF
--- a/examples/test-metadata.yaml
+++ b/examples/test-metadata.yaml
@@ -1,0 +1,24 @@
+apiVersion: kro.run/v1alpha1
+kind: ResourceGraphDefinition
+metadata:
+  name: using-metadata
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: UsingMetadata
+    spec:
+      data:
+        foo: string
+        bar: integer
+    status:
+      created: ${configmap.metadata.creationTimestamp}
+  resources:
+    - id: configmap
+      template:
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: ${instance.metadata.name + "-cm"}
+        data:
+          foo: ${schema.spec.data.foo}
+          bar: ${string(schema.spec.data.bar)} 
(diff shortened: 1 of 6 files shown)


examples/test-metadata.yaml
a-hilaly, line 1: Are you willing to keep this file in your PR? if yes, can we move it somewhere under `examples/kuberetes/*`?

pkg/graph/builder.go
```diff
@@ -292,14 +292,24 @@ func (b *Builder) buildRGResource(rgResource *v1alpha1.Resource, namespacedResou
...
+		// Add metadata fields to emulated resource
+		if emulatedResource.Object["metadata"] == nil {
+			emulatedResource.Object["metadata"] = map[string]interface{}{
+				"name": "dummy-name",
+				"creationTimestamp": "2024-01-01T00:00:00Z",
+				"namespace": "default",
+				"labels": map[string]interface{}{},
+				"annotations": map[string]interface{}{},
+			}
+		}
```
a-hilaly, line 311: I believe that we already set fake metadata on emulated resources here https://github.com/kro-run/kro/blob/main/pkg/graph/emulator/emulator.go#L70-L74

a-hilaly (commented):
Hey @SanjayUG, thank you for your PR! Suggestion for this PR: can you intially address the bug for instance statuses first? as mentioned here https://github.com/kro-run/kro/issues/377#issuecomment-2717054342

We need to agree whether renaming schema to instance is the right call before starting the implementation.

examples/test-metadata.yaml
SanjayUG: Yes, i will update it.

pkg/graph/builder.go
SanjayUG: Thank you for deep diving,

I believe it helps in future development in the project. 

If it is really not required then i welcome to notify me.

a-hilaly:
closing this PR due to innactivity please feel free to re-open it if you're still interrested in upstreaming this.

/close

PR #382, opened 2025-03-12
:sparkles: Improve developer tooling

tomasaschan:
The purpose of this PR is to make a bunch of improvements to the developer tooling, over what was scaffolded by Kubebuilder and hacked together on top of that.

## Guiding principles

These are a couple of principles I'm trying to use for guidance on how to set things up:

### Expose tools through `go tool`

By adding tools as dependencies in `go.mod`, through `go get -tool ...`, we make it so that as little as possible, other than a modern Go toolchain, is required to start contributing to the project. In addition,
this makes it easier for automation to manage version bumps etc.

### Make as much as possible go through the `go` toolchain

We should be able to run tests with `go test ./...`, generate code and other artifacts using `go generate ./...`, and so on. The dependency on a Makefile to document how to run third-party tools is a smell.

Sometimes it cannot be avoided; e.g. there is currently no way to hook into `go vet` that would allow us to configure e.g. `golangci-lint` and `govulncheck` to run through that, but where possible we should try to avoid using other CLI entrypoints than `go` itself.

### Progress and ideas

- [x] Automate generating license headers with `go generate`
- [x] Automate running `attribution-gen` with `go generate`
- [ ] Automate running `controller-gen` with `go generate`
- [ ] Make `controller-gen` emit only the stuff we care about (CRDs, perhaps manager config, but not RBAC) and emit it into the Helm chart directly
- [ ] Register any other developer tools in `go.mod` and get rid of `hack/toolchain.sh`
- [ ] Get rid of the dependency on `setup-envtest` to run tests, by means of hooking into https://github.com/kubernetes-sigs/controller-runtime/pull/3135
- [ ] Get rid of Make targets for building images and instead document how to do it with `go tool ko`; also set up Github Actions to do it in CI
- [ ] Get rid of Make targets for deploying to local test clusters, and instead just document how to set up a test environment
- [ ] Get rid of Make targets for packaging and publishing the Helm chart, and instead do that in Github Actions

We'll see how much of this I do in this PR and how much will be punted for later. The diff is already quite large, but basically anything that's not in the `tool` section of `go.mod`, or in the new files `codegen.go` and `.nwa-config.yaml`, is generated.


tomasaschan:
@a-hilaly The tests are now failing because the bump of `sigs.k8s.io/controller-runtime` from https://github.com/kro-run/kro/pull/382/commits/ae0b66db815de5f255115d2e49eb38c56e1ae07a also brings in `github.com/google/cel-go@v0.23.2` which seems to have a breaking change compared to what the tests assume. Not sure I understand enough of what's going on in those tests to fix them 😞 Do you have any input?

tomasaschan:
@a-hilaly I figured out the test failures; #388 should fix it. I'll incorporate that change here to unblock this work.

tomasaschan:
I think this is ready for a proper review now!

The change set is huge, because the automated changes touch almost every file. Most of it is whitespace changes in the license headers (🤦🏻) but I've tried to separate out the generated changes into separate commits to make them easier to ignore when reviewing.

There were also a few tools being mentioned and installed that weren't actually used anywhere - e.g. `cosign` and `govulncheck`; I'm assuming here that those were installed aspirationally, and I'm punting adding those into the project for a future time when we actually want to integrate them in the other workflows.

Most of the local development workflow should now be covered by `go <cmd>` - e.g. `go vet ./...`, `go test ./...`, `go generate ./...` and `go run ./...`; most of what is not possible to with native commands, but which you might want to do locally anyway, is encapsulated in `go tool`, e.g. `go tool golangci-lint run` or `go tool ko build ...`.

Some things, e.g. releasing new versions, is no longer "easy" to do locally, because I feel they _shouldn't_ be done locally - they should be done from GH actions, in response to merges or tags being pushed. That's why basically all `helm` stuff is just removed.

(I realize as the build turns red that I might have to iterate a couple of times on the GH actions here... please bear with me!)

tomasaschan:
(There; rebased to get rid of conflicts in `go.{mod,sum}` after merging #388)

config/crd/kustomization.yaml
n3wscott, line 1: are these deleted because they are no longer generated?

go.mod
```diff
@@ -4,83 +4,352 @@ go 1.24.0
...
+	github.com/stretchr/testify v1.10.0
+	go.uber.org/zap v1.27.0
+	golang.org/x/exp v0.0.0-20240909161429-701f63a606c0
+	golang.org/x/time v0.11.0
+	google.golang.org/genproto/googleapis/api v0.0.0-20241209162323-e6fa225c2576
+	k8s.io/api v0.33.0-beta.0
+	k8s.io/apiextensions-apiserver v0.33.0-beta.0
+	k8s.io/apimachinery v0.33.0-beta.0
+	k8s.io/apiserver v0.33.0-beta.0
+	k8s.io/client-go v0.33.0-beta.0
```
n3wscott, line 22: might not want to upgrade to a pre release of client go yet.

go.mod
```diff
@@ -4,83 +4,352 @@ go 1.24.0
...
+	k8s.io/kube-openapi v0.0.0-20250304201544-e5f78fe3ede9
+	sigs.k8s.io/controller-runtime v0.20.1-0.20250313041745-d82dcd80ca96
```
n3wscott, line 24: Likely want to roll this back to a stable release

test/integration/environment/setup.go
```diff
@@ -66,15 +68,16 @@ func New(controllerConfig ControllerConfig) (*Environment, error) {
 	env.TestEnv = &envtest.Environment{
 		CRDDirectoryPaths: []string{
 			// resourcegraphdefinition CRD
-			filepath.Join("../../../..", "config", "crd", "bases"),
+			filepath.Join("..", "..", "..", "..", "helm", "templates", "crds"),
```
n3wscott, line 71: this is a good catch, the old way would make windows sad

go.mod
tomasaschan: Both client-go and controller-runtime are bumped to pre-release to get access to https://github.com/kubernetes-sigs/controller-runtime/pull/3135 (and the follow-up fix 3137); they should be released in v0.20.4 shortly, and then we should use that.

config/crd/kustomization.yaml
tomasaschan: Correct; we now put all generated manifests (and some hand-crafted) in the helm chart directly, so there is no longer any kustomization going on.

tomasaschan: (Note that there is a corresponding CRD file generated in `helm/templates/crds`; the rest of the stuff in here is basically unused in the default scaffolding from kubebuilder.)

a-hilaly:
Hi @tomasaschan - are you still planning on breaking down this PR into a fewer smaller ones? if not we can have someone pick it up and make sure commits are co-authored with your email.

tomasaschan:
@a-hilaly Yes, I've been meaning to follow up on this! But I've held off for a couple of reasons - partly because I've been swamped at work, but also because [this question in Slack](https://kubernetes.slack.com/archives/C081TMY9D6Y/p1742481615800009?thread_ts=1742403039.171739&cid=C081TMY9D6Y) never really got answered.

In short: this PR uses https://github.com/B1NARY-GR0UP/nwa to manage license headers, but in the Community Meeting a couple of weeks ago we discussed using https://github.com/google/addlicense/ instead; I'm thinking https://github.com/google/addlicense/issues/173 is somewhat of a blocker for that, and want input from the maintainers on how to proceed.

jakobmoellerdev:
Closing this for now due to inactivity, feel free to reopen if you still wanna push for this

tomasaschan:
@jakobmoellerdev This PR became a behemoth too big to review anyway, and the intention was no longer to ever actually get it merged in this form - but it has served as a guide for what steps to take to accomplish the tooling changes proposed. A couple of those steps have been taken in separate PRs, but the more impactful (and therefore more potentially contentious) things are still to do.

As many of the changes I'm making are either very opinionated, or touch a lot of files (or both!) I have tried to drive discussions in Slack to achieve some level of maintainer buy-in before opening a PR for each of these steps - it will be practically impossible to keep the PRs in a mergeable state for any long period while discussion happens, so maintainers should already be in favor of the approach, and preferably ready to review and merge quickly. However, I'm on parental leave and so have not been able to champion these changes as actively as I did at the start, and most of these ideas have been left hanging in the meantime.

Assuming my employer is still interested in using Kro, and I can therefore justify spending work hours on contributing to the project, I expect to return to these proposals when I'm back sometime mid-Q1, 2026.

PR #396, opened 2025-03-15
feat: add GitOps with Flux CD example

samsonkolge:
# GitOps with Flux CD Example

This PR adds a comprehensive example demonstrating how to implement GitOps practices using KRO with Flux CD.

## Features
- Complete GitOps workflow implementation
- Multi-environment setup (dev/prod)
- ResourceGraphDefinition for GitOps applications
- Integration with Flux CD
- Environment-specific configurations


gecube:
Hi! Loos nice. But I still have some questions:

- is it complete example for bootstrap?
- as I see I need to bootstrap cluster first (into which repo) and then somehow add this repo with `GitRepo` & `Kustomization`?
- Where is the dependency between flux - kro - kro custom resources and the target application? I'd prefer to split everything into several kustomizations and make a proper dependencies between them avoiding direct kubectl against cluster and making everything working in one go. The order is very important as is `kro` did not inject a new API into k8s then you can't create your custom application and fluxcd synchronisation of corresponding `Kustomization` will fail.

examples/gitops-flux/base/flux-system/resource-graphs/gitops-app.yaml
```diff
@@ -0,0 +1,82 @@
...
+                - name: app
+                  image: "{{ .Parameters.image }}"
+                  ports:
+                    - containerPort: 8080
+                  resources:
+                    limits:
+                      cpu: "1"
+                      memory: "512Mi"
+                    requests:
+                      cpu: "200m"
```
a-hilaly: was this RGD generated by an LLM? It doesn't look like a correct kro RGD

samsonkolge: Shoot! I didnt add the appropriate file.. let me fix that

PR #412, opened 2025-03-18
fix: correct patches array format in `config/crd/kustomization.yaml`

### LINKED ISSUE #411: Expected `array` incorrect type in `config/crd/kustomization.yaml`
Image

patches : 
expects an array but it has nothing so its showing incorrect type , so adding an empty array can solve the issue

Rudra-Sankha-Sinhamahapatra:
Fixes #411 

- Changes patches field from '[]' to proper YAML array format
- Fixes YAML schema validation error
- Maintains kubebuilder scaffold structure

## Before 

Screenshot 2025-03-18 at 12 33 10 PM

## After

Screenshot 2025-03-18 at 12 40 36 PM

cc: @a-hilaly

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -5,8 +5,7 @@ resources:
 - bases/kro.run_resourcegraphdefinitions.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
-patches:
-[]
+patches: []
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
 #- path: patches/webhook_in_resourcegraphdefinitions.yaml
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -6,6 +6,7 @@ resources:
 #+kubebuilder:scaffold:crdkustomizeresource
 
 patches:
+[]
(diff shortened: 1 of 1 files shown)


config/crd/kustomization.yaml
```diff
@@ -6,6 +6,7 @@ resources:
 #+kubebuilder:scaffold:crdkustomizeresource
 
 patches:
+[]
```
a-hilaly: Thanks @Rudra-Sankha-Sinhamahapatra - two comments:
1 - This file was generated by controller-gen
2 - i think `patches:\n[]` isn't a valid yaml syntax

Rudra-Sankha-Sinhamahapatra: So will it be patches: [ ] ? To make it valid

@a-hilaly

a-hilaly: I'm not sure really, most of the examples i've seen in the wild keep this field empty

a-hilaly: Did some digging and looks like this was generated by kubebuilder see https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/crd/kustomization.go#L104-L119 - i don't believe we need to make this change in our code base, at least for now :)

Rudra-Sankha-Sinhamahapatra: Thanks for clearing my doubt

PR #413, opened 2025-03-18
Added crd watcher

### LINKED ISSUE #386: Trigger RGD reconciliation on external CRD Modifications
### Feature Description

Currently, kro does not actively watch for modifications to CRDs it manages. This means that if an RGD is modified outside of kro (via kubectl edit or another tool), the controller is not immediately notified, and reconciliation is only triggered based on the default resync period (10hours / configurable)

We can improve this by reacting to changes immediately instead of waiting for the next reconciliation. 

hint: since kro labels all CRDs it manages, we can leverage this by setting up an informer w/ label selectors to watch on CRDs managed by kro.

yashisrani:
fix #386

### DIFF
--- a/pkg/client/crd_watcher.go
+++ b/pkg/client/crd_watcher.go
@@ -1,0 +1,226 @@
+// Copyright 2025 The Kube Resource Orchestrator Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


Rudra-Sankha-Sinhamahapatra (changes requested):
Your code fails the golang lint test , Please fix them

a-hilaly (commented):
thanks for your contribution @yashisrani. I'm closing this PR. before implementing this feature, please create a GH issue first to discuss the approach with the community. Once the issue is triaged and assigned to you according to our contribution guidelines, we can revisit the implementation. I also do believe there are simpler ways to handle this in controller-runtime.

PR #419, opened 2025-03-18
chore: migrate dag to cmp.Ordered

jakobmoellerdev:
allows usage of the DAG with potentially other data types if used as a library.
This is useful if dependency graphs are used for vertexes other than strings (e.g. ints which are used as identities or hashes).

I know this doesn't really benefit KROs internal use directly but it should not harm it either and opens up potential for reuse.

### DIFF
--- a/pkg/graph/dag/dag.go
+++ b/pkg/graph/dag/dag.go
@@ -17,10 +17,10 @@ import (
 	"cmp"
 	"errors"
 	"fmt"
+	"maps"
+	"slices"
 	"sort"
 	"strings"
-
-	"golang.org/x/exp/maps"
 )
 
 // Vertex represents a node/vertex in a directed acyclic graph.
@@ -37,7 +37,7 @@ type Vertex[T cmp.Ordered] struct {
 func (v Vertex[T]) String() string {
 	var builder strings.Builder
 	builder.Grow(len(v.DependsOn))
-	for i, s := range maps.Keys(v.DependsOn) {
+	for i, s := range slices.Collect(maps.Keys(v.DependsOn)) {
 		builder.WriteString(fmt.Sprintf("%v", s))
 		if i < len(v.DependsOn)-1 {
 			builder.WriteString(",")
--- a/pkg/graph/dag/dag.go
+++ b/pkg/graph/dag/dag.go
@@ -14,6 +14,7 @@
 package dag
 
 import (
+	"cmp"
 	"errors"
 	"fmt"
 	"sort"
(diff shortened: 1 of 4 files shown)


pkg/graph/dag/dag.go
```diff
@@ -23,62 +24,75 @@ import (
...
-	dependsOn := strings.Join(maps.Keys(v.DependsOn), ",")
-	return fmt.Sprintf("Vertex[ID: %s, Order: %d, DependsOn: %s]", v.ID, v.Order, dependsOn)
+func (v Vertex[T]) String() string {
+	var builder strings.Builder
+	for i, s := range maps.Keys(v.DependsOn) {
+		builder.WriteString(fmt.Sprintf("%v", s))
+		if i < len(v.DependsOn)-1 {
+			builder.WriteString(",")
+		}
+	}
```
a-hilaly: Not sure how different this is from just using `strings.Join(maps.Keys(v.DependsOn), ",")`

pkg/graph/dag/dag.go
```diff
@@ -23,62 +24,75 @@ import (
...
-	return strings.Join(cycle, " -> ")
+func formatCycle[T cmp.Ordered](cycle []T) string {
+	var builder strings.Builder
+	for i, s := range cycle {
+		builder.WriteString(fmt.Sprintf("%v", s))
+		if i < len(cycle)-1 {
+			builder.WriteString(" -> ")
+		}
+	}
+	return builder.String()
```
a-hilaly: Same comment as above

pkg/graph/dag/dag.go
a-hilaly: The only thing i could think of is performance, but looks like `strings.Join` uses `strings.Builder` under the hood https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/strings/strings.go;l=481-513

a-hilaly (commented):
Thanks a lot @jakobmoellerdev ! left a few comments inline.

cc @justinsb thoughts on this change?

pkg/graph/dag/dag.go
jakobmoellerdev: I was concerned about that too. I didnt want to compare the entire slice to strings so Thats why I ended up with the builder. I adopted the Growing of the Builder capacity from strings.Join now to avoid additional allocs like in the stdlib function

n3wscott (approved):
LGTM, with fun generics!

a-hilaly:
Thanks @jakobmoellerdev !

PR #424, opened 2025-03-19
pkg simpleschema: adds SimpleSchema enum support

### LINKED ISSUE #304: Implement SimpleSchema enum support
### Feature Description

Currently the `enum` marker type is not implemented in SimpleSchema, causing validation errors when users try to use it:
```
failed to parse field schema for mode: invalid marker key 'enum': unknown marker type: enum
```

Example usage that fails:
```yaml
mode: string | enum="debug,info,warn,error" default="info"
```

The implementation can start here, where the enum marker is defined but not implemented yet: https://github.com/kro-run/kro/blob/fe885014289a7cbad1eaff76d7e0109414f1412b/pkg/simpleschema/markers.go#L40-L49

DhairyaMajmudar:
## Description

1. Adds Enum Support to pkg simpleschema
2. Adds relevant tests cases for the same

fixes #304

### DIFF
--- a/pkg/simpleschema/atomic.go
+++ b/pkg/simpleschema/atomic.go
@@ -34,11 +34,13 @@ const (
 	AtomicTypeFloat AtomicType = "float"
 	// AtomicTypeString represents a string value.
 	AtomicTypeString AtomicType = "string"
+	// AtomicTypeNumber represents a number value.
+	AtomicTypeNumber AtomicType = "number"
 )
 
 func isAtomicType(s string) bool {
 	switch AtomicType(s) {
-	case AtomicTypeBool, AtomicTypeInteger, AtomicTypeFloat, AtomicTypeString:
+	case AtomicTypeBool, AtomicTypeInteger, AtomicTypeFloat, AtomicTypeString, AtomicTypeNumber:
 		return true
 	default:
 		return false
--- a/pkg/simpleschema/atomic.go
+++ b/pkg/simpleschema/atomic.go
@@ -34,13 +34,11 @@ const (
 	AtomicTypeFloat AtomicType = "float"
 	// AtomicTypeString represents a string value.
 	AtomicTypeString AtomicType = "string"
-	// AtomicTypeNumber represents a number value.
-	AtomicTypeNumber AtomicType = "number"
 )
 
 func isAtomicType(s string) bool {
 	switch AtomicType(s) {
-	case AtomicTypeBool, AtomicTypeInteger, AtomicTypeFloat, AtomicTypeString, AtomicTypeNumber:
+	case AtomicTypeBool, AtomicTypeInteger, AtomicTypeFloat, AtomicTypeString:
 		return true
 	default:
 		return false
--- a/pkg/simpleschema/markers.go
+++ b/pkg/simpleschema/markers.go
@@ -50,7 +50,6 @@ const (
 	MarkerTypeMinimum MarkerType = "minimum"
 	// MarkerTypeMaximum represents the `maximum` marker.
 	MarkerTypeMaximum MarkerType = "maximum"
-
 	// MarkerTypeEnum represents the `enum` marker.
 	MarkerTypeEnum MarkerType = "enum"
 )
--- a/pkg/simpleschema/markers.go
+++ b/pkg/simpleschema/markers.go
@@ -50,12 +50,15 @@ const (
 	MarkerTypeMinimum MarkerType = "minimum"
 	// MarkerTypeMaximum represents the `maximum` marker.
 	MarkerTypeMaximum MarkerType = "maximum"
+
+	// MarkerTypeEnum represents the `enum` marker.
+	MarkerTypeEnum MarkerType = "enum"
 )
 
 func markerTypeFromString(s string) (MarkerType, error) {
 	switch MarkerType(s) {
 	case MarkerTypeRequired, MarkerTypeDefault, MarkerTypeDescription,
-		MarkerTypeMinimum, MarkerTypeMaximum:
+		MarkerTypeMinimum, MarkerTypeMaximum, MarkerTypeEnum:
 		return MarkerType(s), nil
 	default:
 		return "", fmt.Errorf("unknown marker type: %s", s)
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -231,7 +231,7 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
 
 				switch schema.Type {
 				case "string":
-					rawValue = []byte(fmt.Sprintf("\"%s\"", val))
+					rawValue = []byte(fmt.Sprintf("%q", val))
 				case "integer":
 					if _, err := strconv.ParseInt(val, 10, 64); err != nil {
 						return fmt.Errorf("failed to parse integer enum value: %w", err)
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -216,19 +216,16 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
 				schema.Maximum = &val
 			}
 		case MarkerTypeEnum:
-			enumValues := strings.Split(marker.Value, ",")
-
 			var enumJSONValues []extv1.JSON
 
+			enumValues := strings.Split(marker.Value, ",")
 			for _, val := range enumValues {
 				val = strings.TrimSpace(val)
-
 				if val == "" {
 					return fmt.Errorf("empty enum values are not allowed")
 				}
-
+				
 				var rawValue []byte
-
 				switch schema.Type {
 				case "string":
 					rawValue = []byte(fmt.Sprintf("%q", val))
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -117,7 +117,9 @@ func (tf *transformer) parseFieldSchema(key, fieldValue string, parentSchema *ex
 		fieldJSONSchemaProps = &preDefinedType
 	}
 
-	tf.applyMarkers(fieldJSONSchemaProps, markers, key, parentSchema)
+	if err := tf.applyMarkers(fieldJSONSchemaProps, markers, key, parentSchema); err != nil {
+		return nil, fmt.Errorf("failed to apply markers: %w", err)
+	}
 
 	return fieldJSONSchemaProps, nil
 }
@@ -185,7 +187,7 @@ func (tf *transformer) handleSliceType(key, fieldType string) (*extv1.JSONSchema
 	return fieldJSONSchemaProps, nil
 }
 
-func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Marker, key string, parentSchema *extv1.JSONSchemaProps) {
+func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Marker, key string, parentSchema *extv1.JSONSchemaProps) error {
 	for _, marker := range markers {
 		switch marker.MarkerType {
 		case MarkerTypeRequired:
@@ -206,11 +208,15 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
 		case MarkerTypeDescription:
 			schema.Description = marker.Value
 		case MarkerTypeMinimum:
-			if val, err := strconv.ParseFloat(marker.Value, 64); err == nil {
+			if val, err := strconv.ParseFloat(marker.Value, 64); err != nil {
+				return fmt.Errorf("failed to parse minimum enum value: %w", err)
+			} else {
 				schema.Minimum = &val
 			}
 		case MarkerTypeMaximum:
-			if val, err := strconv.ParseFloat(marker.Value, 64); err == nil {
+			if val, err := strconv.ParseFloat(marker.Value, 64); err != nil {
+				return fmt.Errorf("failed to parse maximum enum value: %w", err)
+			} else {
 				schema.Maximum = &val
 			}
 		case MarkerTypeEnum:
@@ -226,20 +232,20 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
 				case "string":
 					rawValue = []byte(fmt.Sprintf("\"%s\"", val))
 				case "integer":
-					if _, err := strconv.ParseInt(val, 10, 64); err == nil {
+					if _, err := strconv.ParseInt(val, 10, 64); err != nil {
+						return fmt.Errorf("failed to parse integer enum value: %w", err)
+					} else {
 						rawValue = []byte(val)
-						enumJSONValues = append(enumJSONValues, extv1.JSON{Raw: rawValue})
-						continue
 					}
+
 				case "number":
-					if _, err := strconv.ParseInt(val, 10, 64); err == nil {
+					if _, err := strconv.ParseInt(val, 10, 64); err != nil {
+						return fmt.Errorf("failed to parse number enum value: %w", err)
+					} else {
 						rawValue = []byte(val)
-						enumJSONValues = append(enumJSONValues, extv1.JSON{Raw: rawValue})
-						continue
 					}
 				default:
 					rawValue = []byte(val)
-					enumJSONValues = append(enumJSONValues, extv1.JSON{Raw: rawValue})
 				}
 				enumJSONValues = append(enumJSONValues, extv1.JSON{Raw: rawValue})
 			}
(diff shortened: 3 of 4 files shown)


pkg/simpleschema/markers.go
```diff
@@ -50,12 +50,15 @@ const (
 	MarkerTypeMinimum MarkerType = "minimum"
 	// MarkerTypeMaximum represents the `maximum` marker.
 	MarkerTypeMaximum MarkerType = "maximum"
+
```
a-hilaly:
```suggestion
```

pkg/simpleschema/transform.go
```diff
@@ -212,6 +213,45 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+				case "boolean":
+					if val == "true" || val == "false" {
+						rawValue = []byte(val)
+						enumJsonValues = append(enumJsonValues, extv1.JSON{Raw: rawValue})
+						continue
+					}
+				default:
+					rawValue = []byte(val)
+					enumJsonValues = append(enumJsonValues, extv1.JSON{Raw: rawValue})
+				}
```
a-hilaly: Not sure if we want to support JSON and boolean enums

pkg/simpleschema/transform.go
```diff
@@ -212,6 +213,45 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+				switch schema.Type {
+				case "string":
+					rawValue = []byte(fmt.Sprintf("\"%s\"", val))
+				case "integer":
+					if _, err := strconv.ParseInt(val, 10, 64); err == nil {
+						rawValue = []byte(val)
+						enumJsonValues = append(enumJsonValues, extv1.JSON{Raw: rawValue})
+						continue
+					}
+				case "number":
```
matthchr: This case is missing a test

pkg/simpleschema/transform.go
```diff
@@ -212,6 +213,45 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
 			if val, err := strconv.ParseFloat(marker.Value, 64); err == nil {
 				schema.Maximum = &val
 			}
+		case MarkerTypeEnum:
+			enumValues := strings.Split(marker.Value, ",")
+
+			var enumJsonValues []extv1.JSON
```
matthchr: minor: 
```suggestion
			var enumJSONValues []extv1.JSON
```

pkg/simpleschema/transform.go
```diff
@@ -212,6 +213,45 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+
+			for _, val := range enumValues {
+				val = strings.TrimSpace(val)
+				var rawValue []byte
+
+				switch schema.Type {
+				case "string":
+					rawValue = []byte(fmt.Sprintf("\"%s\"", val))
+				case "integer":
+					if _, err := strconv.ParseInt(val, 10, 64); err == nil {
```
matthchr: We should do something if err is not nil - probably amend this function to return an error and then handle/return that error in `parseFieldSchema`?

pkg/simpleschema/transform.go
```diff
@@ -212,6 +213,45 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+						enumJsonValues = append(enumJsonValues, extv1.JSON{Raw: rawValue})
+						continue
+					}
+				case "number":
+					if _, err := strconv.ParseInt(val, 10, 64); err == nil {
+						rawValue = []byte(val)
+						enumJsonValues = append(enumJsonValues, extv1.JSON{Raw: rawValue})
+						continue
+					}
+				case "boolean":
```
matthchr: IMO boolean enums are nonsensical - remove?

a-hilaly (commented):
Thanks @DhairyaMajmudar ! i have two nits below:

pkg/simpleschema/transform_test.go
```diff
@@ -279,6 +279,64 @@ func TestBuildOpenAPISchema(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name: "Schema with multiple enum types",
+			obj: map[string]interface{}{
+				"logLevel": "string | enum=\"debug,info,warn,error\" default=\"info\"",
+				"features": map[string]interface{}{
+					"debugMode": "boolean | enum=\"true,false\" default=false",
+					"logFormat": "string | enum=\"json,text,csv\" default=\"json\"",
```
matthchr, line 287: I think you're missing some negative test cases where users pass invalid values to the enum - i.e. it's an int but they pass strings, or its a string and they pass a totally empty value (`,,`)

pkg/simpleschema/transform.go
matthchr: Agree, IMO no boolean/json enums.

DhairyaMajmudar: Removed in the commit ee57ea0783a0c90936fb49bd69adde1bd98df7f1

pkg/simpleschema/transform.go
DhairyaMajmudar: Modified the `parseFieldSchema` and `applyMarkers`  with proper error handling in the commit  5db9e88ba900b9682fe647184dafc4fb7ea012ae

pkg/simpleschema/transform.go
DhairyaMajmudar: Added in the commit 74fd90c0c87f18cdc0272db0bf758b71b517d3e3

pkg/simpleschema/transform_test.go
DhairyaMajmudar: Done in the commit 763c707d013ee6b4b35bef01041131c8baf1980e

DhairyaMajmudar:
Thank you @a-hilaly @matthchr for the review, updated my PR following the same.

pkg/simpleschema/transform.go
```diff
@@ -205,15 +208,52 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+				case "string":
+					rawValue = []byte(fmt.Sprintf("\"%s\"", val))
+				case "integer":
+					if _, err := strconv.ParseInt(val, 10, 64); err != nil {
+						return fmt.Errorf("failed to parse integer enum value: %w", err)
+					} else {
+						rawValue = []byte(val)
+					}
+				case "number":
+					if _, err := strconv.ParseFloat(val, 64); err != nil {
```
matthchr: FWIW there is a general guidance to avoid floats in k8s APIs.

I am not sure if we want to support float enums - support floats (which we already support AFAIK) is one thing, but a float enum is a pretty weird/niche case and I'm wondering if we just want to drop this for now, at least until some user has a concrete ask/reason for why they might want something like this. I'm not sure I've ever seen a float enum in the wild.

@a-hilaly wdyt?

pkg/simpleschema/transform_test.go
```diff
@@ -279,6 +279,73 @@ func TestBuildOpenAPISchema(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "invalid enum type",
```
matthchr, line 332: one more test for empty enum value?

`"threshold": "int | enum=\"1,,3\"",` - I think we expect error here too, no?

DhairyaMajmudar: Yeah that one is required, Thanks for suggestion. Added one 👍🏻

matthchr: I think this needs to be in a separate test because otherwise the `threshold` will alawys fail. For failure testscases you always need 1 failure-type per test.

pkg/simpleschema/transform.go
DhairyaMajmudar: Attaching some references for why not to use floats

1. https://github.com/zecke/Kubernetes/blob/master/docs/devel/api-conventions.md#primitive-types
2. https://github.com/kubernetes-sigs/controller-tools/issues/245

POV: We should remove the case for number as these references tells not support floating values.

pkg/simpleschema/transform_test.go
DhairyaMajmudar: Got you point, added both of them into one as they are same outputs `nil` 😁

NVM separated both of them now.

pkg/simpleschema/transform.go
a-hilaly: Agreed, we can drop support for enum floats. We can remove the `number` case

pkg/simpleschema/transform.go
```diff
@@ -205,15 +208,52 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+		case MarkerTypeEnum:
+			enumValues := strings.Split(marker.Value, ",")
+
+			var enumJSONValues []extv1.JSON
+
+			for _, val := range enumValues {
+				val = strings.TrimSpace(val)
+				var rawValue []byte
+
+				switch schema.Type {
```
a-hilaly, line 229: Shall we just keep `string` and `integer` here? for the default we can just error saying that enums are only supported for string and integer types

pkg/simpleschema/atomic.go
```diff
@@ -34,11 +34,13 @@ const (
...
 	// AtomicTypeString represents a string value.
 	AtomicTypeString AtomicType = "string"
+	// AtomicTypeNumber represents a number value.
+	AtomicTypeNumber AtomicType = "number"
 )
 
 func isAtomicType(s string) bool {
 	switch AtomicType(s) {
-	case AtomicTypeBool, AtomicTypeInteger, AtomicTypeFloat, AtomicTypeString:
+	case AtomicTypeBool, AtomicTypeInteger, AtomicTypeFloat, AtomicTypeString, AtomicTypeNumber:
```
a-hilaly: Can introduce this change in a separate PR instead? feels like it's addressing a different problem and will very likely be a breaking change.

a-hilaly: Now that i'm thinking about it we might have some miss leading documentation here https://kro.run/docs/concepts/simple-schema - it should be `float` not `number`, so maybe not that breaking of a change :)

DhairyaMajmudar: This won't be required now, as it was associated with the number case in `transform.go` file, but as we are dropping it I've undoed these changes.

pkg/simpleschema/transform.go
DhairyaMajmudar: Done, removed the number cases.

pkg/simpleschema/transform.go
```diff
@@ -205,15 +208,46 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+				case "string":
+					rawValue = []byte(fmt.Sprintf("\"%s\"", val))
+				case "integer":
+					if _, err := strconv.ParseInt(val, 10, 64); err != nil {
+						return fmt.Errorf("failed to parse integer enum value: %w", err)
+					} else {
+						rawValue = []byte(val)
+					}
```
a-hilaly: nit: `else` is not needed here :)
```suggestion
					}
					rawValue = []byte(val)
```

pkg/simpleschema/transform.go
```diff
@@ -205,15 +208,46 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+			if val, err := strconv.ParseFloat(marker.Value, 64); err != nil {
+				return fmt.Errorf("failed to parse minimum enum value: %w", err)
+			} else {
 				schema.Minimum = &val
 			}
 		case MarkerTypeMaximum:
-			if val, err := strconv.ParseFloat(marker.Value, 64); err == nil {
+			if val, err := strconv.ParseFloat(marker.Value, 64); err != nil {
+				return fmt.Errorf("failed to parse maximum enum value: %w", err)
+			} else {
```
a-hilaly: two nits:
- We can have this in a seperate PR since it's not related to enums
- We don't need an `else` statement

pkg/simpleschema/transform.go
```diff
@@ -205,15 +208,46 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
 				schema.Maximum = &val
 			}
+		case MarkerTypeEnum:
+			enumValues := strings.Split(marker.Value, ",")
+
+			var enumJSONValues []extv1.JSON
+
+			for _, val := range enumValues {
+				val = strings.TrimSpace(val)
+				var rawValue []byte
```
a-hilaly: Not sure how is this handling invalid enums such as `a,b,,,,c` - and i'm not quiet grokking how are unit tests passing here. Could be something wrong with the assert statements

pkg/simpleschema/transform.go
DhairyaMajmudar: Gotcha, let me remove this from here and segregate into an another PR

pkg/simpleschema/transform.go
DhairyaMajmudar: There's really a great edge case as in this case `string.Split(marker.Value, ",")` will make it `["a", "b", "", "", "", "c"]` and later loop and code further make this an empty string which is wrong. As in the case of integers it fails early when it hits the empty string and my test case checks the string that's why its failing. Let me add a if statement and a test case for this.

a-hilaly: Good catch, indeed `strconv.ParseInt` errors with empty strings

pkg/simpleschema/transform.go
```diff
@@ -212,8 +215,39 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+
+				if val == "" {
+					return fmt.Errorf("empty enum values are not allowed")
+				}
+
+				var rawValue []byte
+
+				switch schema.Type {
+				case "string":
+					rawValue = []byte(fmt.Sprintf("\"%s\"", val))
```
a-hilaly: nit: you can use %q to quote-format a string 
```suggestion
					rawValue = []byte(fmt.Sprintf("%q", val))
```

pkg/simpleschema/transform_test.go
```diff
@@ -279,6 +279,79 @@ func TestBuildOpenAPISchema(t *testing.T) {
...
+			obj: map[string]interface{}{
+				"threshold": "integer | enum=\"1,2,three\"",
+			},
+			want:    nil,
+			wantErr: true,
+		},
+		{
+			name: "empty enum integer type",
```
a-hilaly:
```suggestion
			name: "Invalid integer enum marker",
```

pkg/simpleschema/transform_test.go
```diff
@@ -279,6 +279,79 @@ func TestBuildOpenAPISchema(t *testing.T) {
...
+			obj: map[string]interface{}{
+				"errorCode": "integer | enum=\"1,,3\"",
+			},
+			want:    nil,
+			wantErr: true,
+		},
+		{
+			name: "string enum with empty values",
```
a-hilaly:
```suggestion
			name: "invalid string enum marker",
```

pkg/simpleschema/transform.go
```diff
@@ -212,8 +215,39 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+			for _, val := range enumValues {
+				val = strings.TrimSpace(val)
+
+				if val == "" {
+					return fmt.Errorf("empty enum values are not allowed")
+				}
+
+				var rawValue []byte
+
+				switch schema.Type {
```
a-hilaly: nit: while not specific to Go, it's a common practice to group related statements more closely together with fewer blank lines between them. Typically you'll see blank lines only to separate logical "paragraphs" of code
```suggestion
			var enumJSONValues []extv1.JSON

			enumValues := strings.Split(marker.Value, ",")
			for _, val := range enumValues {
				val = strings.TrimSpace(val)
				if val == "" {
					return fmt.Errorf("empty enum values are not allowed")
				}
				
				var rawValue []byte
				switch schema.Type {
```

DhairyaMajmudar: Done with all suggestions 👍🏻

pkg/simpleschema/transform_test.go
```diff
@@ -279,6 +279,79 @@ func TestBuildOpenAPISchema(t *testing.T) {
...
+			wantErr: true,
+		},
+		{
+			name: "Invalid integer enum marker",
+			obj: map[string]interface{}{
+				"errorCode": "integer | enum=\"1,,3\"",
+			},
+			want:    nil,
+			wantErr: true,
+		},
```
a-hilaly: Shall we split this into two test cases? one testing that `1,,3` is invalid asserting `empty enum values are not allowed` and another testing that `1,2,3,abc` is invalid asserting `invalid syntax` (which i believe is the error returned by `strconv.ParseInt`)

DhairyaMajmudar: Yep, let's do it

DhairyaMajmudar: Done 👍🏻

a-hilaly (approved):
Thanks a lot @DhairyaMajmudar !

matthchr (approved):

Issue #332, opened 2025-02-27
Add developer getting started and contributing guidelines to the website


jaiakash:
### Documentation Issue

Hi, I was reading about `kro`. 
I noticed that contributing guidelines are served as a markdown page from GitHub. Shouldn't it be better if we show it on the website? Most open source do serve the guidelines on website.

**Location**:
https://kro.run/docs/overview

**Current State**:
The current page shown is this https://github.com/kro-run/kro/blob/main/CONTRIBUTING.md

**Suggested Changes**:
Add above page and https://kro.run/docs/concepts/developer-getting-started to the website

Gauravsh-24:
Hello @jaiakash 
I would like to work on this issue . could you please assign to me.

shreyans413:
Hi, please assign it to me, planning to submit a PR soon..

AnushkaaBansal:
I have tried solving the issue.

jlbutler:
Hi folks, catching up here.

Thanks for the contribution. I don't believe we want to put contributor and dev docs into the project website. The website is primarily geared toward endusers, where the dev docs are best placed in the repo itself.

PR #425, opened 2025-03-19
Solved issue #332 by adding contributor guidelines

AnushkaaBansal:


docs/contributing.md
```diff
@@ -0,0 +1,73 @@
+# Contributing to Kro
```
matthchr, line 1: This looks like its duplicating the CONTRIBUTING content in multiple places.

I dont think we should be doing that. I also dont think we need 2x contributing.md, one at the top level and one in docs, so this instance of the file could I think be deleted.

website/docs/docs/community/contributing.md
```diff
@@ -0,0 +1,77 @@
+---
+sidebar_position: 1
+---
+
+# Contributing to Kro
```
matthchr, line 5: Instead of duplicating the contributing content here (which is I think what this has done), we could symlink the CONTRIBUTING.md here.

There are some downsides to that over time, but for _now_ it's I think easy to do any avoids having to maintain duplicate docs.

jlbutler:
Thanks for the contribution. As mentioned in the issue, we don't want to put contributor and dev docs into the project website. The website is primarily geared toward endusers, where the dev docs are best placed in the repo itself.

PR #426, opened 2025-03-19
Solved issue#423 by adding a security section

AnushkaaBansal:


a-hilaly (commented):
Hi @AnushkaaBansal - thanks for your PR!

I don't think the website is the right place for these CVE reports. If you have concerns about specific library vulnerabilities, the best approach would be to update the affected packages in our [go.mod](https://github.com/kro-run/kro/blob/main/go.mod) file instead.

Issue #418, opened 2025-03-18
Support test fixtures for creating ResourceGraphDefinitions


fsommar:
### Feature Description

**Problem Statement**:

Creating a new ResourceGraphDefinition for non-trivial resources requires a lot of iterative work, trying things out. The more conditionals are included, the harder it is to validate that each input creates valid output. Applying an RGD to the cluster, waiting for the status to show up and validate and then create an instance, only to delete them and redo it (until proper lifecycle support is in place) is a slow process.

**Proposed Solution**:

I'm going to [take inspiration from kubebuilder-declarative-pattern](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/d7eaf43a0e1a11fcad6a56157e2499cecef201da/examples/guestbook-operator/controllers/tests) and [Kyverno](https://kyverno.io/docs/kyverno-cli/usage/test/), and suggest support for test fixtures that takes an RGD instance and the expected expansion as input and validates it. I don't have strong opinions on the exact implementation of it, so what I'm suggesting below is very much up for discussion.

Given the definition of an RGD `rgd.yaml`:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: example
spec:
  schema:
    apiVersion: v1alpha1
    kind: Example
    spec:
      foo: string | required

  resources:
    - id: cm
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.metadata.name}-cm
        data:
          foo: ${schema.spec.foo}
    - id: cm2
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.metadata.name}-cm2
        data:
          bar: ${schema.spec.foo}
```

I want to be able to create a file `tests/happy.yaml`:

```yaml
apiVersion: tests.kro.run/v1alpha1
kind: Expectations
spec: {}
---
apiVersion: kro.run/v1alpha1
kind: Example
metadata:
  name: foobar
  namespace: sparkles
spec:
  foo: hello world!
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: foobar-cm
  namespace: sparkles
data:
  foo: hello world!
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: foobar-cm2
  namespace: sparkles
data:
  bar: hello world!
```

and a file `tests/sad.yaml`

```yaml
apiVersion: tests.kro.run/v1alpha1
kind: Expectations
[...]

PR #427, opened 2025-03-19
Solved issue#418 by adding test fixtures

AnushkaaBansal:


matthchr:
The added file just seems to be repeating what the issue already says? It's not clear to me what this PR is actually adding.

jlbutler:
@AnushkaaBansal can you please explain this PR? The issue seems reasonable, but we can't see how this proposed change addresses it.  Thanks!

a-hilaly (commented):
This PR appears to be adding the content of GitHub issue #418 directly to the website documentation, and that's not what we're looking for. If you'd like to contribute to addressing issue #418, can you please discuss the possible approaches in the GIthub issue first?

PR #432, opened 2025-03-19
Add CEL standard library to DefaultEnvironment

a-hilaly:
Fixes https://github.com/kro-run/kro/issues/80

Add `cel.StdLib()` to DefaultEnvironment to include standard type
conversion functions like `string()` and `int()` in our CEL environment

### DIFF
--- a/pkg/cel/environment.go
+++ b/pkg/cel/environment.go
@@ -57,6 +57,7 @@ func DefaultEnvironment(options ...EnvOption) (*cel.Env, error) {
 		// default stdlibs
 		ext.Lists(),
 		ext.Strings(),
+		cel.StdLib(),
 	}
 
 	for _, name := range opts.resourceIDs {


pkg/cel/environment.go
```diff
@@ -57,6 +57,7 @@ func DefaultEnvironment(options ...EnvOption) (*cel.Env, error) {
 		// default stdlibs
 		ext.Lists(),
 		ext.Strings(),
+		cel.StdLib(),
```
matthchr, line 60: It doesn't hurt to put this here, but isn't this implied by `cel.NewEnv`?

https://github.com/google/cel-go/blob/master/cel/env.go#L926

matthchr (approved):

pkg/cel/environment.go
a-hilaly: good point!

PR #437, opened 2025-03-20
devtools(1/many): Generate ATTRIBUTION.md with go generate

tomasaschan:
This is the first of a series of PRs to do what #382 does, but in ways that are easier to review and get merged.


a-hilaly (approved):
👍

matthchr (approved):

ATTRIBUTION.md
```diff
@@ -471,7 +501,7 @@ notice, this list of conditions and the following disclaimer.
 copyright notice, this list of conditions and the following disclaimer
 in the documentation and/or other materials provided with the
 distribution.
```
n3wscott, line 503: Google is an LLC as of 2017, I wonder if the lib we are using is real old?

n3wscott:
LGTM

ATTRIBUTION.md
a-hilaly: Yep looks like golang.org/x/exp just updated their license recently https://github.com/golang/exp/commit/1d5bc16f04a822c5d762aca81011815d8bb7eb4b

PR #448, opened 2025-03-23
Tests: adds tests for `cel/environment.go`

DhairyaMajmudar:
## Description

Adds tests for `pkg/cel/environment.go` file

## Test validation 

image

cc: @a-hilaly


a-hilaly (commented):
I'm hesitant here, I like the idea of increasing test coverage, but this PR is mainly testing constructors which I don't see tremendous value in, though I acknowledge they could catch regressions if someone makes changes to this package. With that said i think that it doesn't hurt to accept it.

a-hilaly (approved):

PR #452, opened 2025-03-24
Leverage the upstream k8s release-utils version strings and ldflags values.

n3wscott:
This uses the k8s release-utils package to define a place to land version of the software.

The PR renames the ldflags import paths from kro to k8s, and adds a helper method to remove `+` from releases in the binary version when local branches are dirty. 

Running locally results in the following:

```
(╯°□°)╯︵  kubectl get crd noops.kro.run -oyaml
metadata:
  creationTimestamp: "2025-03-24T21:31:02Z"
  generation: 1
  labels:
    kro.run/kro-version: v0.2.3-0.20250323205418-b9d1822983c4-dirty
    kro.run/owned: "true"
    kro.run/resource-graph-definition-id: cb241fb4-ab64-4b63-8804-ae89247d3d06
    kro.run/resource-graph-definition-name: noop
```

If the make entry point `make ko-apply` is used.

If `ko apply` is used directly, then the ld flags are ignored but also we are not bricked to do that dev loop locally. The upstream lib has defaults when no ldflags are passed.


helm/Chart.yaml
```diff
@@ -2,6 +2,6 @@ apiVersion: v2
 name: kro
 description: A Helm chart for kro
 type: application
-version: ${VERSION}
-appVersion: ${VERSION}
+version: 0.0.0-dev
```
n3wscott, line 5: whatever was using `${VERSION}` does not seem to be in-tree so I can't see or fix it.

a-hilaly: `${VERSION}` was a placeholder, nothing really interacts with it.

n3wscott:
cc @puerco, it continues!

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	k8s.io/client-go v0.31.0
 	k8s.io/kube-openapi v0.0.0-20240816214639-573285566f34
 	sigs.k8s.io/controller-runtime v0.19.0
+	sigs.k8s.io/release-utils v0.11.0
```
a-hilaly, line 25: w00t!

a-hilaly (approved):
👍

Issue #423, opened 2025-03-19
CVE-2025-22868 golang.org/x/oauth2


sachinbh95:
### Feature Description

**Problem Statement**:

Our vulnerability image scan found a high CVE-2025-22868 golang.org/x/oauth2 v0.21.0 in Kro release v0.2.2

**Proposed Solution**:

This is fixed in v0.27.0 version 

**Alternatives Considered**:

**Additional Context**:

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

AnushkaaBansal:
I have tried to solve this issue.

AnushkaaBansal:
Sir I have created a PR #454

PR #454, opened 2025-03-24
Update go.mod and go.sum as in #423

AnushkaaBansal:
Since the security issue has been fixed in a newer version of the library, specifically version v0.27.0. To resolve the problem, in go.mod file, under require(_), i first updated oauth2 to v0.27.0. Then as i ran, go mod tidy, two additional enteries got added to the go.sum file. Which is the reason why I commited code changes from both the files- go.mod and go.sum. 

Info on tidy as taken from a blog- 
Tidy makes sure go.mod matches the source code in the module. It adds any missing modules necessary to build the current module’s packages and dependencies, and it removes unused modules that don’t provide any relevant packages. It also adds any missing entries to go.sum and removes any unnecessary ones.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	github.com/x448/float16 v0.8.4 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/net v0.36.0 // indirect
-	golang.org/x/oauth2 v0.21.0 // indirect
+	golang.org/x/oauth2 v0.27.0 // indirect
 	golang.org/x/sys v0.30.0 // indirect
 	golang.org/x/term v0.29.0 // indirect
 	golang.org/x/text v0.22.0 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -66,7 +66,7 @@ require (
 	github.com/x448/float16 v0.8.4 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/net v0.36.0 // indirect
-	golang.org/x/oauth2 v0.21.0 // indirect
+	golang.org/x/oauth2 v0.27.0 // indirect
```
a-hilaly, line 69: I see that v0.28.0 is available - can we use that instead?

AnushkaaBansal: Yes 
Updation to v0.28.0 is made in PR #455

a-hilaly:
Thanks @AnushkaaBansal

PR #456, opened 2025-03-25
Remove demo scripts and "config files" from eks-cluster-mgmt example

allamand:
Just clean files that got pushed by mistakes


examples/aws/eks-cluster-mgmt/terraform/hub/kubeconfig
a-hilaly, line 1: ohhh shoot. :)

PR #459, opened 2025-03-26
Fork and update of Condition Sets from Knative, attributed

n3wscott:
As mentioned in https://github.com/kro-run/kro/issues/458, this starts to land some of the helper code that deals with "top-level" conditions and sub-conditions. 

This does not get used in the controller yet. Just landing the fork and updated code. 

Sorry it is so large, the code is only about 300 lines, the rest is tests.

### DIFF
--- a/api/v1alpha1/conditions.go
+++ b/api/v1alpha1/conditions.go
@@ -14,8 +14,9 @@
 package v1alpha1
 
 import (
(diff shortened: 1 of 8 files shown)


a-hilaly (approved):
Thanks @n3wscott !

PR #462, opened 2025-03-27
Do not store loggers in controller stucts

n3wscott:
Leverages the context to grab the correct logger for each method that uses a logger in ResourceGraphDefinition controller.

Old log lines:
```
kro-86944cdcf6-l5td8 kro 2025-03-24T20:58:13.410Z	DEBUG	controller.resourceGraphDefinition	updating resource graph definition status	{"resourcegraphdefinition": {"name":"noop"}, "state": "Inactive", "conditions": 3}
kro-5d77854d9f-gtqmd kro 2025-03-24T21:31:02.982Z	LEVEL(-5)	Reconcile successful	{"controller": "resourcegraphdefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition", "ResourceGraphDefinition": {"name":"deploymentservice"}, "namespace": "", "name": "deploymentservice", "reconcileID": "579ac0e9-0902-4dcc-b220-8d104df1e911"}
```

New log lines:
```
kro-db65c46d7-q9r6g kro 2025-03-27T21:37:27.670Z	DEBUG	rgd-controller	updating resource graph definition status	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition", "namespace": "", "name": "noop", "reconcileID": "28cd0c21-ac8b-4646-a1e4-d83b00d9b476", "state": "Active", "conditions": 3}
kro-db65c46d7-q9r6g kro 2025-03-27T21:37:27.681Z	LEVEL(-5)	rgd-controller	Reconcile successful	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition", "namespace": "", "name": "noop", "reconcileID": "28cd0c21-ac8b-4646-a1e4-d83b00d9b476"}
```

### DIFF
--- a/pkg/client/crd.go
+++ b/pkg/client/crd.go
@@ -18,13 +18,13 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/go-logr/logr"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	logr "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
@@ -51,7 +51,6 @@ type CRDClient interface {
 // CRDWrapper provides a simplified interface for CRD operations
 type CRDWrapper struct {
 	client       apiextensionsv1.CustomResourceDefinitionInterface
-	log          logr.Logger
 	pollInterval time.Duration
 	timeout      time.Duration
 }
@@ -59,7 +58,6 @@ type CRDWrapper struct {
 // CRDWrapperConfig contains configuration for the CRD wrapper
 type CRDWrapperConfig struct {
 	Client       *apiextensionsv1.ApiextensionsV1Client
-	Log          logr.Logger
 	PollInterval time.Duration
 	Timeout      time.Duration
 }
@@ -83,7 +81,6 @@ func newCRDWrapper(cfg CRDWrapperConfig) *CRDWrapper {
 
 	return &CRDWrapper{
 		client:       cfg.Client.CustomResourceDefinitions(),
-		log:          cfg.Log.WithName("crd-wrapper"),
 		pollInterval: cfg.PollInterval,
 		timeout:      cfg.Timeout,
 	}
@@ -95,18 +92,19 @@ func newCRDWrapper(cfg CRDWrapperConfig) *CRDWrapper {
 // The caller is responsible for ensuring the CRD, isn't introducing
 // breaking changes.
 func (w *CRDWrapper) Ensure(ctx context.Context, crd v1.CustomResourceDefinition) error {
+	log := logr.FromContext(ctx)
 	_, err := w.Get(ctx, crd.Name)
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return fmt.Errorf("failed to check for existing CRD: %w", err)
 		}
 
-		w.log.Info("Creating CRD", "name", crd.Name)
+		log.Info("Creating CRD", "name", crd.Name)
 		if err := w.create(ctx, crd); err != nil {
 			return fmt.Errorf("failed to create CRD: %w", err)
 		}
 	} else {
-		w.log.Info("Updating existing CRD", "name", crd.Name)
+		log.Info("Updating existing CRD", "name", crd.Name)
 		if err := w.patch(ctx, crd); err != nil {
 			return fmt.Errorf("failed to patch CRD: %w", err)
 		}
@@ -143,7 +141,8 @@ func (w *CRDWrapper) patch(ctx context.Context, newCRD v1.CustomResourceDefiniti
 
 // Delete removes a CRD if it exists
 func (w *CRDWrapper) Delete(ctx context.Context, name string) error {
-	w.log.Info("Deleting CRD", "name", name)
+	log := logr.FromContext(ctx)
+	log.Info("Deleting CRD", "name", name)
 
 	err := w.client.Delete(ctx, name, metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
@@ -154,7 +153,8 @@ func (w *CRDWrapper) Delete(ctx context.Context, name string) error {
 
 // waitForReady waits for a CRD to become ready
 func (w *CRDWrapper) waitForReady(ctx context.Context, name string) error {
-	w.log.Info("Waiting for CRD to become ready", "name", name)
+	log := logr.FromContext(ctx)
+	log.Info("Waiting for CRD to become ready", "name", name)
 
 	return wait.PollUntilContextTimeout(ctx, w.pollInterval, w.timeout, true,
 		func(ctx context.Context) (bool, error) {
--- a/pkg/controller/resourcegraphdefinition/controller.go
+++ b/pkg/controller/resourcegraphdefinition/controller.go
@@ -15,17 +15,19 @@ package resourcegraphdefinition
 
 import (
 	"context"
+
 	"github.com/go-logr/logr"
-	"github.com/kro-run/kro/api/v1alpha1"
-	kroclient "github.com/kro-run/kro/pkg/client"
-	"github.com/kro-run/kro/pkg/dynamiccontroller"
-	"github.com/kro-run/kro/pkg/graph"
-	"github.com/kro-run/kro/pkg/metadata"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlrtcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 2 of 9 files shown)


pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -37,79 +35,87 @@ import (
...
 func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	logConstructor := func(req *reconcile.Request) logr.Logger {
+		log := mgr.GetLogger().WithName("rgd-controller").WithValues(
+			"controller", "ResourceGraphDefinition",
+			"controllerGroup", v1alpha1.GroupVersion.Group,
+			"controllerKind", "ResourceGraphDefinition",
+		)
+		if req != nil {
+			log = log.WithValues(
+				"namespace", req.Namespace,
```
a-hilaly: RGDs are cluster scoped, should we omit the namespace value?

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -37,79 +35,87 @@ import (
...
 	}
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	logConstructor := func(req *reconcile.Request) logr.Logger {
+		log := mgr.GetLogger().WithName("rgd-controller").WithValues(
+			"controller", "ResourceGraphDefinition",
+			"controllerGroup", v1alpha1.GroupVersion.Group,
+			"controllerKind", "ResourceGraphDefinition",
```
a-hilaly, line 84: This seems like it's adding more verbosity to the logs, `rgd-controller` already insinuates that this is related RGDs?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -78,13 +78,14 @@ func (r *ResourceGraphDefinitionReconciler) setupLabeler(rgd *v1alpha1.ResourceG
...
 func (r *ResourceGraphDefinitionReconciler) setupMicroController(
+	ctx context.Context,
 	gvr schema.GroupVersionResource,
 	processedRGD *graph.Graph,
 	defaultSVCs map[string]string,
 	labeler metadata.Labeler,
 ) *instancectrl.Controller {
 
-	instanceLogger := r.rootLogger.WithName("controller." + gvr.Resource)
+	instanceLogger := ctrl.LoggerFrom(ctx).WithName(gvr.Resource)
```
a-hilaly: This causes the instance logger to share the same logger as the RGD reconciler, meaning that the instance reconciliation logs will contain unrelated RGD names/reconcile IDs. For example:

```
2025-03-27T21:52:13.884-0700	LEVEL(-2)	rgd-controller.dynamodbbackends	Using instance namespace	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition", "namespace": "", "name": "dynamodbbackend", "reconcileID": "757a4daa-919a-4c40-b302-5e1a563c6358", "namespace": "kro", "name": "todoapp5", "resourceID": "iamPolicy", "namespace": "kro"}
```

a-hilaly: This one of the reasons why we setup two different loggers, one that is passed to RGD reconciler, one that is passed to the instance reconcilers.

pkg/controller/resourcegraphdefinition/controller.go
n3wscott: good point!

pkg/controller/resourcegraphdefinition/controller.go
n3wscott: these are the defaults that come from the logger maker and if you want this reconciler to produce matching metrics for consistent logging across all things that used kubebuilder, I would recommend sticking with what they already use and their keys

pkg/controller/resourcegraphdefinition/controller_reconcile.go
n3wscott: I had a think about this while I was making the change and I like that the named logger tells you where that thread come from, I think this sideeffect is a feature, not a bug :D

n3wscott: > This causes the instance logger to share the same logger as the RGD reconciler

Not share, it is derived from, which is how these reconcilers are created.

n3wscott: >  will contain unrelated RGD names/reconcile IDs

oh, hmmm, let me think about this.

n3wscott: OK I see your point. How about we save an instance logger but we get it from the manager and don't pass it in the constructor

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -37,79 +34,89 @@ import (
...
 	clientSet  *kroclient.Set
 	crdManager kroclient.CRDClient
 
-	metadataLabeler   metadata.Labeler
-	rgBuilder         *graph.Builder
-	dynamicController *dynamiccontroller.DynamicController
+	metadataLabeler         metadata.Labeler
+	rgBuilder               *graph.Builder
+	dynamicController       *dynamiccontroller.DynamicController
+	MaxConcurrentReconciles int
```
a-hilaly: shall we make this private?

pkg/controller/resourcegraphdefinition/controller.go
a-hilaly: >if you want this reconciler to produce matching metrics for consistent logging across all things that used kubebuilder

Do we want to? i don't mind eventually diverging from that - I think these logs are unnecessarily repetitive and verbose.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: sounds good to me

pkg/controller/resourcegraphdefinition/controller.go
n3wscott: ack, but someone who uses this and installs it along side their other kubebuilder based things will make the same log patterns and their already build dashboards will work for this too

pkg/controller/resourcegraphdefinition/controller.go
n3wscott: I would rather make everything public and remove the java getter/setter/constructors, but for now I can make it private.

a-hilaly: > remove the java getter/setter/constructors

I can't believe i did this.

a-hilaly: I very rarelly do getters/setters, but `New` constructures are quiet common in go

a-hilaly (approved):
Awesome, thank you @n3wscott !

pkg/client/crd.go
```diff
@@ -83,7 +81,6 @@ func newCRDWrapper(cfg CRDWrapperConfig) *CRDWrapper {
 
 	return &CRDWrapper{
 		client:       cfg.Client.CustomResourceDefinitions(),
-		log:          cfg.Log.WithName("crd-wrapper"),
```
matthchr, line 86: we lost the `crd-wrapper` name here - do we want to re-include it down below on line 95?

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -15,20 +15,17 @@ package resourcegraphdefinition
 
 import (
 	"context"
-
 	"github.com/go-logr/logr"
-	"k8s.io/apimachinery/pkg/types"
```
matthchr, line 20: minor, style: why move these imports?

pkg/controller/resourcegraphdefinition/controller.go
matthchr: You're talking about this, yea?
https://github.com/kubernetes-sigs/controller-runtime/blob/d8b679322d5cc466d12301b8875c42040fd29e46/pkg/builder/controller.go#L434

I'm not sold that this is the optimal log-line format either, and (I think?) not even every kubebuilder project will use this format because if they override or set their own logger at any point, you don't get this default set of values.

With that said, there is some value in consistency and maybe that value outweighs the extra log verbosity...

matthchr (approved):

pkg/controller/resourcegraphdefinition/controller.go
n3wscott: the ide going ham... I can move it back

pkg/client/crd.go
n3wscott: Because this is not a thread or controller but just a lib, I would be inclined to remove the logging all together. As it is, it will get the name of the logger of the caller.

pkg/controller/resourcegraphdefinition/controller.go
n3wscott: yeah, I don't _really_ care about the log verbosity but having off the shelf tools show me the sorted log output with no customization seems like a nice thing if you accept what the ecosystem defaults are.

a-hilaly: I'm getting convinced here - in my mind compatibility with existing dashboards and tools outweighs the verbosity in logs

matthchr (approved):

a-hilaly (approved):
Thanks @n3wscott !

Issue #461, opened 2025-03-27
Nil pointer dereference during CEL evaluation when ResourceGraphDefinition includes CRD and CEL-using resources


Rakhmanov:
### Description

**Observed Behavior:**

When a ResourceGraphDefinition contains:

- a CRD resource without CEL expressions, and
- other resources that do use CEL expressions

KRO attempts to evaluate CEL expressions across all resources. This leads to a Go panic when accessing resource.emulatedObject on the CRD, which doesn’t support it.

Manifesting as:
`panic: runtime error: invalid memory address or nil pointer dereference`

**Expected Behavior:**

Resources without CEL expressions should be safely excluded from CEL validation logic.

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):

Apply the following ResourceGraphDefinition:

```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: example-graph
spec:
  schema:
    group: example.group
    apiVersion: v1
    kind: ExampleKind
    spec:
      name: string

  resources:
    - id: myCrd
      template:
        apiVersion: apiextensions.k8s.io/v1
        kind: CustomResourceDefinition
        metadata:
          name: examples.example.group
        spec:
          group: example.group
          names:
            plural: examples
            singular: example
            kind: Example
            listKind: ExampleList
          scope: Namespaced
          versions:
            - name: v1
              schema:
                openAPIV3Schema:
                  type: object
                  properties:
                    metadata:
                      type: object
                    spec:
                      type: object
                    status:
                      type: object
                      properties:
                        ready:
                          type: boolean
                        output:
                          type: object
                          properties:
                            state:
                              type: string
                              enum: ["InProgress", "Failed", "Successful"]
                            message:
                              type: string
              served: true
              storage: true
              subresources:
                status: {}
    - id: role
      template:
        apiVersion: rbac.authorization.k8s.io/v1
        kind: Role
        metadata:
          name: role
        rules:
          - apiGroups: ["example.example.group"]
            resources: ["mycrd", "mycrd/status"]
            verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
    - id: roleBinding
      template:
        apiVersion: rbac.authorization.k8s.io/v1
        kind: RoleBinding
        metadata:
          name: role-binding
        subjects:
          - kind: ServiceAccount
            name: default
            namespace: default
        roleRef:
          kind: Role
          name: ${role.metadata.name}
[...]

PR #463, opened 2025-03-27
Fix nil pointer panic when evaluating CEL expressions with CRDs

Rakhmanov:
The graph builder was attempting to access emulatedObject on resources without checking if it exists first. Since CRDs don't have emulated objects, this was causing a panic when any CEL expressions were present in the resource graph.

By adding conditional lookup on resource.emulatedObject . Avoiding null pointer dereference when working with types that dont have it: `apiextensions.k8s.io, v1, CustomResourceDefinition` in [builder.go#L286](https://github.com/kro-run/kro/blob/4a5b2a310b77fbd43f1babfaac53116d309e8980/pkg/graph/builder.go#L286)

Resolving #461

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -632,7 +632,9 @@ func dryRunExpression(env *cel.Env, expression string, resources map[string]*Res
 
 	context := map[string]interface{}{}
 	for resourceName, resource := range resources {
-		context[resourceName] = resource.emulatedObject.Object
+		if resource.emulatedObject != nil {
+			context[resourceName] = resource.emulatedObject.Object
+		}
 	}
 
 	output, _, err := program.Eval(context)
--- a/pkg/graph/builder_test.go
+++ b/pkg/graph/builder_test.go
@@ -1183,6 +1183,89 @@ func TestGraphBuilder_ExpressionParsing(t *testing.T) {
 				})
 			},
 		},
+		{
+			name: "crds not failing when cel is present in other resources",
+			resourceGraphDefinitionOpts: []generator.ResourceGraphDefinitionOption{
+				generator.WithSchema(
+					"Test", "v1alpha1",
+					map[string]interface{}{
+						"name": "string",
+					},
+					nil,
+				),
+				generator.WithResource("somecrd", map[string]interface{}{
+					"apiVersion": "apiextensions.k8s.io/v1",
+					"kind":       "CustomResourceDefinition",
+					"metadata": map[string]interface{}{
+						"name": "somecrd.ec2.services.k8s.aws",
+					},
+					"spec": map[string]interface{}{
+						"group":   "ec2.services.k8s.aws",
+						"version": "v1alpha1",
+						"names": map[string]interface{}{
+							"kind":     "SomeCRD",
+							"listKind": "SomeCRDList",
+							"singular": "SomeCRD",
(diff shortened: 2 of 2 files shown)


pkg/graph/builder.go
```diff
@@ -632,7 +632,9 @@ func dryRunExpression(env *cel.Env, expression string, resources map[string]*Res
 
 	context := map[string]interface{}{}
 	for resourceName, resource := range resources {
-		context[resourceName] = resource.emulatedObject.Object
+		if resource.emulatedObject != nil {
+			context[resourceName] = resource.emulatedObject.Object
```
n3wscott, line 636: should we return an error if the object is nil?

rakhmanov: Its expected to be nil for the CRD type. So no, because it does not have schema that can be parsed [here](https://github.com/kro-run/kro/blob/4a5b2a310b77fbd43f1babfaac53116d309e8980/pkg/graph/builder.go#L304C35-L304C48).

a-hilaly: Ohh good catch @Rakhmanov - i wonder why didn't we see these panics when we first tested these RGDs https://github.com/kro-run/kro/blob/main/examples/aws/ack-controller/crds/ecr-public-crd-group.yaml (i have those running in my test cluster w/ kro 0.2.2, no panic observed)

a-hilaly: cc @rushmash91

rakhmanov: @a-hilaly Those are precisely were my starting point and comparison, I could not figure it out why they worked and mine didn't, until ran it with debugger. Since in these CRD's there are 0 usage of cel in any of the resources, they never go down the path of access emulatedObject parameter. And once at least one resource does have usage, you get the panics. I've added the test that should fail on main, and pass on this branch.

a-hilaly: Nice sleuthing! Indeed, it's the CEL expression usage that triggers this behavior. The graph builder dry-run component assumed that all resources would have an emulated object, which is not the case for CRDs.

a-hilaly: Now I'm wondering, what would happen if a CEL expression tries to reference a CRD? It can't be validated since CRDs don't have emulated objects. I'm initially thinking about two options
- Allow references to the CRD metadata at least, by injecting a CRD with no versions/schemas.
- Do not allow CEL expressions to reference CRDs (leaning more towards this as a short term option, we can always expand later on)

a-hilaly (approved):
Thanks a lot @Rakhmanov! We can eventually address https://github.com/kro-run/kro/pull/463/files#r2019539364

pkg/graph/builder.go
rushmash91: Dang! Nice catch!

rakhmanov:
> Thanks a lot @Rakhmanov! We can eventually address https://github.com/kro-run/kro/pull/463/files#r2019539364

Accessing metadata makes sense, would be interesting how to make it obvious to the user the "uniqueness" of the CRD vs any other resource, but its better to have metadata values than nothing.

a-hilaly:
@Rakhmanov would like to take a stab at that? basically for CRD we can inject "dummy" emulated CRDs that only contain metadata and basic version fields, without setting any openapisec.

PR #474, opened 2025-04-02
feat: enqueue instances on rgd update

fabianburth:
**Enhancement**
Currently, when applying a modified `ResourceGraphDefinition`, the corresponding instance objects will not be enqueued for reconciliation. This means, the changes in the RGD do not take effect until the next reconciliation of the instance. If the instances are not modified, the next scheduled reconciliation occurs after the resync period of the dynamic controller.
This PR proposes a change to allow for a rather immediate reconciliation of corresponding instance objects.

This is also mentioned in this [issue](https://github.com/kro-run/kro/issues/188) and work on that has already been started in   [PR](https://github.com/kro-run/kro/pull/203).

TODO: @a-hilaly I'd like to revisit the idea of the update annotation.

### DIFF
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -110,13 +110,13 @@ const (
 	// InstanceUpdatePolicy is the annotation key for the instance update policy
 	InstanceUpdatePolicy = "kro.run/instance-update-policy"
 
-	// InstanceUpdatePolicyOnResourceGraphUpdate represents the update policy that leads to
+	// InstanceUpdatePolicyOnRGDUpdate represents the update policy that leads to
 	// a reconciliation of the instance after the corresponding resource graph is updated
-	InstanceUpdatePolicyOnResourceGraphUpdate = "on-resource-graph-update"
+	InstanceUpdatePolicyOnRGDUpdate = "on-rgd-update"
 
-	// InstanceUpdatePolicyIgnoreResourceGraphUpdate represents the update policy that explicitly
+	// InstanceUpdatePolicyIgnoreRGDUpdate represents the update policy that explicitly
 	// avoids reconciliation of the instance after the corresponding resource graph is updated
-	InstanceUpdatePolicyIgnoreResourceGraphUpdate = "ignore-resource-graph-update"
+	InstanceUpdatePolicyIgnoreRGDUpdate = "ignore-rgd-update"
 )
 
 // ResourceGraphDefinitionStatus defines the observed state of ResourceGraphDefinition
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -15,7 +15,7 @@ package v1alpha1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


fabianburth:
With regards to the update annotation - currently, I feel like the update probably should not be conditional at all. Primarily because even if an instance is annotated with `ignore-rgd-update`, the rgd update would still be part of the next reconciliation of that instance. 
I think with an actual `ignore-rgd-update` kind of annotation, the expectation would rather be that the instance handler would not be updated based on that annotation. But I guess, then we'd have to track the handlers on a per instance basis. What do you think?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -427,8 +429,26 @@ func (dc *DynamicController) StartServingGVK(ctx context.Context, gvr schema.Gro
 	_, exists := dc.informers.Load(gvr)
 	if exists {
 		// Even thought the informer is already registered, we should still
-		// still update the handler, as it might have changed.
+		// update the handler, as it might have changed.
 		dc.handlers.Store(gvr, handler)
+		// trigger reconciliation of the corresponding gvr's
+		objs, err := dc.kubeClient.Resource(gvr).Namespace("").List(ctx, metav1.ListOptions{})
```
a-hilaly, line 434: Does this list objects in all namespaces or just in the default namespace?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -62,8 +62,10 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/kro-run/kro/api/v1alpha1"
```
a-hilaly: can you please move this import to the L80 import group, checkout this commit message https://github.com/golang/tools/commit/ed69e84b1518b5857a9f4e01d1f9cefdcc45246e

api/v1alpha1/types.go
```diff
@@ -106,6 +106,17 @@ const (
...
+	// InstanceUpdatePolicy is the annotation key for the instance update policy
+	InstanceUpdatePolicy = "kro.run/instance-update-policy"
+
+	// InstanceUpdatePolicyOnRGDUpdate represents the update policy that leads to
+	// a reconciliation of the instance after the corresponding resource graph is updated
+	InstanceUpdatePolicyOnRGDUpdate = "on-rgd-update"
+
+	// InstanceUpdatePolicyIgnoreRGDUpdate represents the update policy that explicitly
+	// avoids reconciliation of the instance after the corresponding resource graph is updated
+	InstanceUpdatePolicyIgnoreRGDUpdate = "ignore-rgd-update"
```
a-hilaly: I'm thinking that we should first implement the core functionality of updating instances on RGD updates as the default behavior. I'm concerned about introducing annotations at this stage since they represent a potential breaking change and we probablyy should discuss the naming conventions and semantics carefully. maybe we could address the annotation mechanism/names in a separate PR?

a-hilaly (commented):
Awesome, thank you @fabianburth !! left a few thoughts/questions below

pkg/dynamiccontroller/dynamic_controller.go
fabianburth: Apparently, it does list objects in all namespaces. But since I actually had to go check after you asked, I included an object with a non-default namespace in the test.

pkg/dynamiccontroller/dynamic_controller.go
fabianburth: Resolved itself (due to the removal of the annotation constants)

api/v1alpha1/types.go
fabianburth: Sure! 
As mentioned in my comment, I feel very unsure about the annotations and their expected / intended behaviour anyway.

a-hilaly (approved):
LGTM, danke schoen!

PR #477, opened 2025-04-02
pkg simpleschema: CEL validation support

DhairyaMajmudar:
## Description

1. Adds CEL validation support
2. Adds test cases corresponding to the changes

fixes: https://github.com/kro-run/kro/issues/347

cc: @a-hilaly

### DIFF
--- a/pkg/simpleschema/transform_test.go
+++ b/pkg/simpleschema/transform_test.go
@@ -360,6 +360,103 @@ func TestBuildOpenAPISchema(t *testing.T) {
 			want:    nil,
 			wantErr: true,
 		},
+		{
+			name: "Schema with CEL validation - immutable filed",
+			obj: map[string]interface{}{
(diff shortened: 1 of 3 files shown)


pkg/simpleschema/transform_test.go
```diff
@@ -360,6 +360,103 @@ func TestBuildOpenAPISchema(t *testing.T) {
 			want:    nil,
 			wantErr: true,
 		},
+		{
+			name: "Schema with CEL validation - immutable filed",
+			obj: map[string]interface{}{
+				"immutableFoo": "string | validation={\"rule\": \"self == oldSelf\", \"message\": \"can't be changed\"}",
```
a-hilaly, line 366: I was thinking that maybe we can also have a simpler support for validation, like name: `string | validation="self.name != 'something'"`. And honestly i'm not really convinced that smashing jsons in the simple schema is a good thing, idk just sharing my thoughts :) - let's keep for now and see what other folks think.

a-hilaly:
Closing in favor of #484

PR #478, opened 2025-04-03
Custom additionalPrinterColumns for created CRDs

### LINKED ISSUE #476: Custom additionalPrinterColumns on created CRDs
### Feature Description

**Problem Statement**:
Hey, is there a way to set custom `AdditionalPrinterColumns` for a CRD that gets created from an RGD? I see there are [default printer columns](https://github.com/petar-cvit/kro/blob/main/pkg/graph/crd/crd.go#L64), but it would be nice if those could be defined in the RGD and passed down to the CRD.

**Proposed Solution**:
Add an additional printer columns field to the RGD, which will then be passed to the CRD and applied to the cluster.

**Additional Context**:
If other tools depend on the additional printer columns, it might make the integrations easier and give developers a better experience.

petar-cvit:
closes #476 

## Description
Adds `additionalPrinterColumns` to RGD and maps the columns to the created CRD in the RGD reconciler.

cc: @a-hilaly


a-hilaly (commented):
Thank you for your PR @petar-cvit ! The PR implementation for custom additionalPrinterColumns looks correct overall.

However i wanted to explore an alternative approach with you, what if we enhanced the SimpleSchema instead of introducing a new top level field? users could potentially do something like:

```yaml
schema:
  spec:
    name: string | required=true print=true # or kubectl_print=true
```
I recognize this wouldn't be sufficient for complex printer columns with additional attributes.

I also just came across KEP 4602 (https://github.com/kubernetes/enhancements/pull/4602/files) which seems relevant. We might want to explore this direction in parallel as it could influence our approach here.

petar-cvit:
Hey @a-hilaly, thanks for the alternative approach! If we want to add the printer columns to the simple schema, we can easily get the JSONPath for a specific field. Also, this seems easier for users to set up.

I was also thinking about whether `additionalPrinterColumns` should be top-level with `schema` and `resources`, but it seems like we are losing a lot of flexibility for users in the long run if those are defined in the simple schema, especially with the CEL enhancement proposal.

In the case of CEL columns (as described in the KEP), tieing a CEL expression to a field using simple schema doesn't seem like the right fit, primarily because CEL expressions work with the whole CRs and can get data from multiple fields at once (KEP example `expression: self.status.replicas == self.spec.replicas ? "READY" : "WAITING"`).

My point is that if we want to design it with CEL columns in mind, simple schema is probably not the way to go, but not sure how much should a KEP influence design decisions.

One argument for having a dedicated `additionalPrinterColumns` is that it follows how those are configured for CRDs, and users are familiar on how to use them.

Let me know what you think about the points I mentioned. Happy to discuss further

a-hilaly:
Yeah, I totally agree with everything you said @petar-cvit . Keeping `additionalPrinterColumns` separate makes a lot of sense, especially when thinking about where the KEP is heading. Maybe supporting the simple schema another day could provide a faster way to inject `additionalPrinterColumns`, just as a way to expose data directly into the kubectl output. with that said i think we should go with this PR :)

a-hilaly:
@petar-cvit BTW there's an ongoing [thread in our Slack](https://kubernetes.slack.com/archives/C081TMY9D6Y/p1742575757867799) about introducing simple schema custom types and whether they should live directly in `spec` or in `spec.schema`. Regardless of where we land on that one, I think `additionalPrinterColumns` and these type definitions should probably live under the same parent field for consistency. Mind dropping your thoughts in that thread if you get a chance?

pkg/graph/crd/crd.go
```diff
@@ -101,3 +101,11 @@ func newCRDSchema(spec, status extv1.JSONSchemaProps, statusFieldsOverride bool)
...
 	}
 }
+
+func newCRDAdditionalPrinterColumns(additionalPrinterColumns []extv1.CustomResourceColumnDefinition) []extv1.CustomResourceColumnDefinition {
+	if len(additionalPrinterColumns) == 0 {
+		return defaultAdditionalPrinterColumns
+	}
+
+	return additionalPrinterColumns
+}
```
a-hilaly, line 112: Thought: what are your thoughts on injecting specific default `CustomResourceColumnDefinition` objects, when they aren't specfied by users? that way we ensure that `STATE`/`SYNCED`/`AGE` are always included?

petar-cvit: My reasoning was that users might not need the default columns and might want to remove them. If they are already customizing printer columns, my guess is that they want to show a specific state of their CRs.

However, I don't have such a strong opinion on this, and I can merge the defined additionalPrinterColumns with the default ones based on the `name.` Let me know if you agree with the point above or if we should add the default columns, and I can add them.

petar-cvit:
Thanks for the link to the discussion @a-hilaly. I left a comment with a different approach in the thread that might allow you to reuse types in multiple RGDs. Let me know if it makes sense or if it's overkill for this problem

a-hilaly (approved):
Neat, thanks @petar-cvit !

api/v1alpha1/types.go
```diff
@@ -43,6 +44,11 @@ type ResourceGraphDefinitionSpec struct {
 	//
 	// +kubebuilder:validation:Optional
 	DefaultServiceAccounts map[string]string `json:"defaultServiceAccounts,omitempty"`
+	// AdditionalPrinterColumns defines additional printer columns that will
```
matthchr: Worth calling out that setting this overrides the default printer columns, and if the user wants to retain the default columns they will need to specify them here explicitly?

pkg/graph/crd/crd.go
matthchr: I think allowing users to override our default printer columns seems reasonable. If they want to maintain the existing ones, they should be able to.

I mentioned above we should call that out in the docs though

matthchr (approved):

api/v1alpha1/types.go
petar-cvit: @matthchr, I just added the comment you proposed. Let me know if there is anything else

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -60,6 +60,50 @@ spec:
             description: ResourceGraphDefinitionSpec defines the desired state of
               ResourceGraphDefinition
             properties:
+              additionalPrinterColumns:
+                description: |-
+                  AdditionalPrinterColumns defines additional printer columns that will
```
matthchr: Need to regenerate the YAML now, as you've changed the comment but it doesn't look like it's propagated to here yet?

matthchr (dismissed):
LGTM, but need to regenerate CRD yaml I think

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
petar-cvit: My bad, I just regenerated the CRD and pushed it

matthchr (dismissed):

matthchr:
Spoke with @a-hilaly and he pointed out this might make more sense on `schema` rather than at the top level like you have it. I agree, temporarily dropping my approval as I think we should push this to `schema`

matthchr (commented):
Move to schema (as mentioned in other comment)

joshuabaird:
Thanks for this @petar-cvit -- very useful feature!  I wonder if we could/should document `additionalPrinterColumns` somewhere?  If not in the official docs, maybe an example in the `example/` directory?

petar-cvit:
Moved `additionalPrinterColumns` down to `schema`.

> If not in the official docs, maybe an example in the example/ directory?

@joshuabaird, thanks for the feedback! I saw you opened an issue about the docs - #543. I'd be happy to add an example of this to the repo, but if there will be a generic solution for the docs, maybe we can leave it out for now. @a-hilaly, what do you think?

a-hilaly (approved):
Thank you very much @petar-cvit ! 👍

a-hilaly:
I think this should ideally be documented in https://kro.run/docs/concepts/simple-schema - so far we don't have any automation to generate that page. Happy to take a seperate PR for it :) @petar-cvit would like to take a stab a this? or maybe add an example in /examples directory?

petar-cvit:
Added examples of additionalPrinterColumns to /examples and to the website as well in a separate PR #547

cc @a-hilaly

PR #493, opened 2025-04-08
Fix infinite loop in fieldPath parser for single-letter field names

gfrey:
Field names with just a single letter, say 'data.A' for a data field in a configmap would result in an infinite loop.


a-hilaly (approved):
Thank you Gereon! :)

Issue #355, opened 2025-03-05
Document CEL for Kro


csantanapr:
### Documentation Issue

Add to the website CEL documentation for Kro

Some useful topics to document:
- Which CEL functions are available (strings, list, etc)
- How to debug/troubleshoot CEL functions CEL [playground](https://playcel.undistro.io/)

utsab818:
Since no one has moved forward to working on this, is it possible to assign me this issue. I have been looking at KRO cel package from past week to understand CEL, and hopefully I will be able to document a good one.

a-hilaly:
duplicate of https://github.com/kubernetes-sigs/kro/issues/503

PR #497, opened 2025-04-08
docs:added cel expressions from codebase for kro

Rudra-Sankha-Sinhamahapatra:
Related to #355 

Added cel expression docs for users in kro 

Examples are taken from codebase 

some wrong implementation i took from internet to show example 

Screenshot 2025-04-09 at 4 30 28 AM

cc: @a-hilaly


matthchr:
We agree that we need CEL documentation for kro, but we would prefer to see something more like a link to the CEL upstream documentation followed by a small set of examples showing how to use it in kro. This PR gets a bit too into describing upstream functions, which are already documented in the upstream docs.

We have an open issue tracking this item here: #355. closing this - let's discuss what the docs should look like on the issue and then we can progress from there.

PR #498, opened 2025-04-09
feat: Add support for (custom) types

### LINKED ISSUE #144: Add support for complex data types
### Feature Description

**Problem Statement**:
For complex objects, such as ACK DyanmoDB Table definition, there's a need to define a complex data-types that will be used as an input for the ResourceGroup.
As an example, ACK DynamoDB Table gets as an input, a list of the following struct:
```
 attributeDefinitions:
    - attributeName: id
      attributeType: S
    - attributeName: name
      attributeType: S
```

Currently there's no way to define the following:
```
tableElement: map[string][string]
attributeDefinition: []tableElement
```

**Proposed Solution**:
support complex data types to be able to pass those as an input to the ResourceGroup spec values.

**Alternatives Considered**:
none

**Additional Context**:
none

gfrey:
This is the first implementation that was discussed in the slack channel. The other idea of having the `types` field in the `spec` part of the schema, seems to be more complicated to implement.

Fixes #144.

### DIFF
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -15,21 +15,31 @@ package simpleschema
 
 import (
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
+// A predefined type is a type that is predefined in the schema.
+// It is used to resolve references in the schema, while capturing the fact
+// whether the type has the required marker set (this information would
+// otherwise be lost in the parsing process).
+type predefinedType struct {
+	Schema   extv1.JSONSchemaProps
+	Required bool
+}
+
 // transformer is a transformer for OpenAPI schemas
 type transformer struct {
-	preDefinedTypes map[string]extv1.JSONSchemaProps
+	preDefinedTypes map[string]predefinedType
 }
 
 // newTransformer creates a new transformer
 func newTransformer() *transformer {
 	return &transformer{
-		preDefinedTypes: make(map[string]extv1.JSONSchemaProps),
+		preDefinedTypes: make(map[string]predefinedType),
 	}
 }
 
@@ -39,7 +49,7 @@ func newTransformer() *transformer {
 // As of today, kro doesn't support custom types in the schema - do
 // not use this function.
 func (t *transformer) loadPreDefinedTypes(obj map[string]interface{}) error {
-	t.preDefinedTypes = make(map[string]extv1.JSONSchemaProps)
+	t.preDefinedTypes = make(map[string]predefinedType)
 
 	jsonSchemaProps, err := t.buildOpenAPISchema(obj)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


website/docs/docs/concepts/10-simple-schema.md
```diff
@@ -109,6 +115,23 @@ labels: "map[string]string"
...
+Example:
+
+```yaml
+schema:
+  types:
+    Person:
+      name: string
+      age: integer
+  spec:
+    people: '[]person | required=true`
```
csantanapr: Does it need to match case sensitive it doesn't have to be `[]Person` upper case P?

gfrey: It does! Fixed.

pkg/simpleschema/transform.go
```diff
@@ -15,21 +15,31 @@ package simpleschema
...
 	"strings"
 
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
+// A predefined type is a type that is predefined in the schema.
+// It is used to resolve references in the schema, while capturing the fact
+// whether the type has the required marker set (this information would
+// otherwise be lost in the parsing process).
+type predefinedType struct {
```
a-hilaly, line 29: Good call! initially i wasn't sure about this, but i just remembered that required fields always live parent-type-schema-definition.

pkg/simpleschema/transform_test.go
```diff
@@ -20,28 +20,15 @@ import (
...
-		},
-		"Person": map[string]interface{}{
-			"name": "string",
-			"age":  "integer",
-		},
-	})
-	if err != nil {
-		t.Fatalf("Failed to load pre-defined types: %v", err)
-	}
+func floatPtr(f float64) *float64 {
```
a-hilaly: looks like this function isn't used, shall we remove it?

a-hilaly (commented):
Thanks @gfrey! left two tiny comments

pkg/simpleschema/transform_test.go
gfrey: Absolutely!

pkg/simpleschema/transform.go
gfrey: I just noticed that adding this broke the test for the predefined types. So, I fixed that.

flake.lock
a-hilaly, line 1: can you please exclude nix files from this PR?

gfrey: Yes, I have force pushed the commits that contained them. Sorry for that!

barney-s: I would love to see the nix files setup for kro. Would you share it in a gist please ?

gfrey: @barney-s https://gist.github.com/gfrey/2101b155e1d1d9f56517bfd00fa06dcb is what I'm using.

a-hilaly (approved):
Neat, thank you @gfrey !!

barney-s:
Have a top level Q.
Since this is schema level changes how is `spec` and `status` different from custom types ? 

Instead of this: 
```yaml
schema:
  types:
    Person:
      name: string
      age: integer
  spec:
    people: '[]Person | required=true`
```

How does this look like ? 
```yaml
schema:
   Person:
     name: string
     age: integer
  spec:
    people: '[]Person | required=true`
``` 

??

barney-s:
Another alternative is:

```
schema:
  spec:
    people:
    - _properties: "array | required=true"
      name: string
      age: integer

```

for maps:

```
schema:
  spec:
    people: `[]string | required=true`
    peeps:
    - _properties: "array | required=true"
      name: string
      age: integer
```

PR #501, opened 2025-04-10
kro cli: rootCmd

DhairyaMajmudar:
# Description

1. Adds the rootCmd - `kro` which will be the main command of the kro-cli.
2. A very basic implementation of validation command (Currently, this command doesn't validate anything; it just prints on the terminal)

## Usage

-> To build & run the cli run the command `make-cli`. This commands builds the cli binary and stores it into the local bin folder. As of which we can use kro command globally.

## Availabe commands
1. kro --help or kro -h => prints the help texts and all the available commands in the cli. 
2. kro validate rgd [file_name] => validate command is under development it currently only prints the success message on the terminal.

## Demo

[Screencast from 2025-04-09 19-11-43.webm](https://github.com/user-attachments/assets/5c2beda1-b58e-4849-91c2-6e48fd20d738)


cmd/kro/main.go
a-hilaly, line 1: Ideally we need to keep the kro main here, we want to allow users to simply do `go install github.com/kro-run/kro/cmd/kro@latest` and automatically bring a binary called kro to their PATH. If we put the `main.go` a `cli` directory, the binary will be called `cli`.

With that said, I understand that `cli` makes more sense since we a `controller` binary... maybe something we can address another time?

cmd/cli/main.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Global flags
+	cmd.PersistentFlags().String("kubeconfig", kubeConfigPath, "Path to kubeconfig file")
+	cmd.PersistentFlags().String("context", "", "Kubernetes context to use")
+	cmd.PersistentFlags().Bool("verbose", false, "Enable verbose logging")
+
+	// TODO: Command groups
+	commands.AddValidateCommands(cmd)
+	return cmd
+}
```
a-hilaly: I would highly recommend to put the structure definitions outside of the function. just easier to scale once the code base gets bigger :)

cmd/cli/commands/validate.go
```diff
@@ -0,0 +1,43 @@
...
+
+package commands
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func AddValidateCommands(rootCmd *cobra.Command) {
```
a-hilaly: Same comment

a-hilaly (commented):
Thank you @DhairyaMajmudar! i left a few comments in line. Also can you please rebase and run a `make` before submit? we did add something new in https://github.com/kro-run/kro/pull/507

cmd/kro/main.go
DhairyaMajmudar: valid point, let me change the folder name to `cmd/kro` back

cmd/cli/main.go
DhairyaMajmudar: Gotcha! Took out the struct definations from the function.

cmd/cli/commands/validate.go
DhairyaMajmudar: Actually I added this file for testing purposes only, the actual validation implementation code is not yet added (I'm planning to open a separate PR for validation command). There, I'll surely make the structs as you suggested. But currently, here I don't feel like it needed something like that, Pls. correct if wrong.

cmd/kro/main.go
a-hilaly, line 1: please make sure to run `go generate` which will fix this license.

cmd/kro/commands/validate.go
```diff
@@ -0,0 +1,43 @@
...
+		Use:   "rgd [FILE]",
+		Short: "Validate a ResourceGraphDefinition file",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// TODO(DhairyaMajmudar): Implement the logic to validate the ResourceGraphDefinition file
+
+			fmt.Println("Validation successful! The ResourceGraphDefinition is valid.")
+			return nil
+		},
+	}
```
a-hilaly: nit: can we get these structs out of the function?

cmd/kro/main.go
```diff
@@ -1,36 +1,64 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use:   "kro",
+		Short: "kro- Kube Resource Orchestrator CLI",
+		Long: `kro CLI helps developers and administrators manage 
+ResourceGraphDefinitions (RGDs) and their instances in Kubernetes clusters.`,
+		SilenceErrors: true,
+		SilenceUsage:  true,
+	}
```
a-hilaly: same for the root command

a-hilaly (commented):
Thanks @DhairyaMajmudar - left few comments in line

cmd/kro/main.go
DhairyaMajmudar: Done 👍🏻

cmd/kro/commands/validate.go
DhairyaMajmudar: Done 👍🏻

cmd/kro/commands/validate.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+var validateRGDCmd = &cobra.Command{
+	Use:   "rgd [FILE]",
+	Short: "Validate a ResourceGraphDefinition file",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// TODO(DhairyaMajmudar): Implement the logic to validate the ResourceGraphDefinition file
+
+		fmt.Println("Validation successful! The ResourceGraphDefinition is valid.")
```
barney-s: Let's change the string to.
"Validation not implemented yet"

cmd/kro/main.go
```diff
@@ -15,22 +15,52 @@
...
+	KubeConfigPath string
+	Context        string
+	Verbose        bool
+}
+
+var cmd = &cobra.Command{
+	Use:   "kro",
+	Short: "kro- Kube Resource Orchestrator CLI",
+	Long: `kro CLI helps developers and administrators manage 
+ResourceGraphDefinitions (RGDs) and their instances in Kubernetes clusters.`,
```
barney-s: s/clusters/cluster

cmd/kro/commands/validate.go
DhairyaMajmudar: Done 👍🏻

cmd/kro/main.go
DhairyaMajmudar: Done 👍🏻

cmd/kro/main.go
DhairyaMajmudar: Done 👍🏻

barney-s (approved):
LGTM

PR #502, opened 2025-04-10
Fix field schema parsing

gfrey:
The CEL expressions in a validation marker can use `||` for 'or' expressions. Therefore, the test for at most two elements would fail in these cases. Fix is to just split once.

Related to #347.


a-hilaly (approved):
Thank you very much sir!

PR #506, opened 2025-04-14
feat: Adds random.seededString() CEL function.

### LINKED ISSUE #499: Introduce CEL function for generating random string
### Feature Description

**Problem Statement**:
It would be useful to be able to inject a random string into a _ResourceGraphDefinition_'s `resources` to enforce unique naming for resources -- eg, IAM roles.  This is useful for users who want to provision multiple account-scoped resources (AWS IAM roles) to the same account because these resource names need to be unique.

**Proposed Solution**:
A CEL function that can generate a random string and be used similar to this:

```
resources:
- id: iamRoleResource
   template:
     apiVersion: iam.services.k8s.aws/v1alpha1
     kind: Role
     metadata:
       name: ${schema.spec.name}-cert-${randomString}
```

joshuabaird:
Fixes #499.

Intoduces a new CEL function -- `random.seededString()` that generates a random (yet deterministic) alphanumeric string.  The `random.seededString()` function takes two parameters:

* the length of the string to generate
* the seed value used to make the string deterministic

In this PR, we also expose `schema.metadata.uid` on an instance because it's useful as a seed value to `random.seededString()` since it provides a deterministic value that is unique per cluster and it's immutable. 

The use-case driving this feature is the ability to inject random strings into resource names to make them unique.


joshuabaird:
@a-hilaly @michaelhtm Re-factored a bit to more closely follow the CEL library [pattern](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apiserver/pkg/cel/library) used in upstream Kubernetes.

pkg/graph/emulator/emulator.go
```diff
@@ -72,6 +72,7 @@ func (e *Emulator) GenerateDummyCR(gvk schema.GroupVersionKind,
 	cr.SetName(fmt.Sprintf("%s-sample", strings.ToLower(gvk.Kind)))
 	cr.SetNamespace("default")
 	cr.SetResourceVersion(fmt.Sprintf("%d", e.rand.Intn(1000)))
+	cr.SetUID("dummy-uid")
```
joshuabaird, line 75: @a-hilaly Is this the correct way to expose `schema.metadata.uid`?  It _does_ work, but you may have thoughts on this.

a-hilaly: During the emulation, yes that's correct! although we will eventually get rid of the emulation process

pkg/cel/environment.go
```diff
@@ -16,6 +16,7 @@ package cel
 import (
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/ext"
+	"github.com/kro-run/kro/pkg/cel/library"
 )
```
a-hilaly: nit:
```suggestion
import (
	"github.com/google/cel-go/cel"
	"github.com/google/cel-go/ext"
	
	"github.com/kro-run/kro/pkg/cel/library"
)
```

pkg/cel/library/random_string.go
```diff
@@ -0,0 +1,95 @@
...
+
+// RandomString returns a CEL function that generates deterministic random strings
+func RandomString() cel.EnvOption {
+	return cel.Function("randomString",
+		cel.Overload("randomString_int_string",
+			[]*cel.Type{cel.IntType, cel.StringType},
+			cel.StringType,
+			cel.BinaryBinding(generateDeterministicString),
+		),
+	)
```
a-hilaly: non-blocking: instead making this a function, can we overload a library called `random` ? i'm thinking that eventually we will probably want `random.string`, `random.integer`, `random.float`?

pkg/cel/ast/inspector.go
```diff
@@ -99,6 +99,11 @@ type Inspector struct {
 	loopVars map[string]struct{}
 }
 
+// knownFunctions contains the list of all CEL functions that are supported
+var knownFunctions = []string{
+	"randomString",
+}
```
a-hilaly: note for the future: we need some sort of mechanism to automatically detect all the "Overloaded" functions.

pkg/cel/ast/inspector.go
```diff
@@ -533,6 +537,9 @@ func isInternalFunction(name string) bool {
 		"all":        true,
 		"exists":     true,
 		"exists_one": true,
+
+		// Custom Functions
+		"randomString": true,
```
a-hilaly: Same note, i know i'm the original culprit here, couldn't find a way to list all the builtin functions, so i hardcoded them as a workaround.

gfrey: `isInternalFunction` is only used in a method of `Inspector`. Why not replace it with the [`HasFunction` method](https://pkg.go.dev/github.com/google/cel-go/cel#Env.HasFunction) of `cel.Env` (property of the `Inspector`)? Am I missing something?

barney-s:
I know this is a point solution to support Random strings as part of RG evaluation.

Few Qs just thinking aloud:
1. Can't we take a random number as an input from the user of the RGD as a parameter in schema. Whats the advantage of including this in the RGD defn ? Trying to understand the motivation.
2. Randomness may result in continuous drift correction if a proper seed is not used. 

That being said. From a library perspective, i have a few Qs and may be suggestions:
1. pkg/cel/library << would this be importable from any other project ? Preferred if yes.

Would you think have a separate repo for CEL libraries be preferred ? 
Allows us to collaborate with other projects like k8s. I could not find such library.

joshuabaird:
> 1. Can't we take a random number as an input from the user of the RGD as a parameter in schema. Whats the advantage of including this in the RGD defn ? Trying to understand the motivation.

The controller needs to generate a random string for this feature to be useful (at least for the use case described in #499).  

Having the controller generate the random string is useful if a user wants to deploy a set of manifests that contain account-scoped resources (eg, `roles.iam.services.aws`) into multiple namespaces or multiple clusters, since the names need to be unique for each. 

> 2. Randomness may result in continuous drift correction if a proper seed is not used.

Yeah, this is correct.  I felt it was best to allow a user to choose their own seed for their specific use-case.

> That being said. From a library perspective, i have a few Qs and may be suggestions:
> 
> 1. pkg/cel/library << would this be importable from any other project ? Preferred if yes.
> 
> Would you think have a separate repo for CEL libraries be preferred ? Allows us to collaborate with other projects like k8s. I could not find such library.

Are you suggesting that CEL functions that we author specifically for Kro should be able to be used in other projects?  That isn't something that I have considered in this PR, but maybe something that should be discussed as a long-term strategy for the project?  I honestly don't know what that would look like or have a strong opinion about it (yet).

pkg/cel/ast/inspector.go
joshuabaird: See [c27b86e](https://github.com/kro-run/kro/pull/506/commits/c27b86ead3c6df4191c17c05ae4eb39585fba487)!

pkg/cel/library/random_string.go
joshuabaird: See [c27b86e](https://github.com/kro-run/kro/pull/506/commits/c27b86ead3c6df4191c17c05ae4eb39585fba487).

barney-s:
> Yeah, this is correct. I felt it was best to allow a user to choose their own seed for their specific use-case.

I was wondering if there is a way for us to seed it in a deterministic way instead of the user providing the deterministic seed (constant or .metadata.uid). 
This way we reduce the cognitive burden. 

Also may be add another function that takes validChars (hardcoded in the function) and startingChar as well. 

Thinking along these lines:
```
randomString(len) // auto-seeded 
randomSeededString(len, seed) // explicit seeding
randomCustomString(len, validChars, firstChar) // custom character space with optional custom first character
```

joshuabaird:
I certainly see value in some of the functions that you described, but I personally feel that adding them all in this PR may introduce a fair amount of scope creep.   I think that this PR sets the stage for additional `random` functions (eg `random.customString()`, etc) in future PR's since we're introducing a `random` CEL library.  

I thought that I may be able to access `schema.metadata.*` from within the CEL function (instead of passing it as an argument) so that I could make the second (seed) value of `random.string()` optional, but I'm not having much luck doing so.  I'm not sure there is any value in providing a constant as a default seed.

Ultimately, while I agree that asking a user to provide a seed does introduce a bit more cognitive burden, it does allow more flexibility into how a user implements `random.string()`.

pkg/cel/ast/inspector.go
a-hilaly: >Why not replace it with the [HasFunction method](https://pkg.go.dev/github.com/google/cel-go/cel#Env.HasFunction) of cel.Env (property of the Inspector)? Am I missing something?

I didn't know that `env.HasFunction` existed back in the day. You're right we need to use that :)

barney-s:
Agreed with Scope creep. Just to be backward compatible would it make sense to rename the current function to `random.SetingWithSeed()` or `random.SeededString()` so that we can later introduce `random.String()` which is auto-seeded ?

joshuabaird:
> Agreed with Scope creep. Just to be backward compatible would it make sense to rename the current function to `random.SetingWithSeed()` or `random.SeededString()` so that we can later introduce `random.String()` which is auto-seeded ?

Yep - that makes sense!  [4e71aeb](https://github.com/kro-run/kro/pull/506/commits/4e71aeb416f2fb90fb82f449af9635d516e5c98e) renamed the function from `random.string` to `random.seededString()`.

PR #507, opened 2025-04-15
devtools(2/many): Generate license headers wth go generate

tomasaschan:
This configures [nwa] to generate license headers for all files in the repo as part of running `go generate ./...`.

It is also possible to configure it to check that license headers are present as part of CI, but this has not been implemented (yet).

This PR follows up on #437 and #382.

[nwa]: https://github.com/B1NARY-GR0UP/nwa


a-hilaly (dismissed):
Sweet, thanks a lot!

test/integration/suites/ackekscluster/generator.go
```diff
@@ -1,16 +1,16 @@
...
+// You may obtain a copy of the License at
 //
-//	http://www.apache.org/licenses/LICENSE-2.0
+//     http://www.apache.org/licenses/LICENSE-2.0
 //
-// or in the "license" file accompanying this file. This file is distributed
-// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
-// express or implied. See the License for the specific language governing
-// permissions and limitations under the License.
-package ackekscluster_test
```
a-hilaly, line 13: looks like some files lost their `package` directives - cc @tomasaschan

tomasaschan: Oooh, I recognize this from back when doing #382; I think the problem is that there is an existing header without a blank line before the package directive. I'll go over and add them back manually.

a-hilaly:
@tomasaschan  It looks like `/home/runner/work/kro/kro/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."` is failing too. Maybe we can remove that step?we no longer generate the controller go files using kubebuilder.

tomasaschan:
Good catch; that's because I thought the boilerplate file wouldn't be needed. Now that we use this to generate headers instead, I guess it's just a matter of ensuring correct ordering of `//go:generate` comments (once `controller-gen` is also configured to run through `go generate`).

For now, is it OK to just change the `controller-gen` setup to not generate headers, and leave properly configuring that for a follow-up PR? #382 included some pretty significant changes for how, and to where, manifests were generated, so I'd like to keep that separate from the header generation in the interest of keeping things contained and reviewable.

a-hilaly:
>For now, is it OK to just change the controller-gen setup to not generate headers, and leave properly configuring that for a follow-up PR?

Sure let's do that!

a-hilaly (approved):
👍

PR #508, opened 2025-04-16
fix: change dynamic controller resync period unit from hours to seconds

a-hilaly:
change the dynamic controller resync period unit from hours to seconds, allowing users to
configure much more frequent resyncs when needed for sensitive applications.


rushmash91 (approved):
Excited for `v0.2.3` 🚀

PR #512, opened 2025-04-16
fix: evaluate includeWhen before static expressions in runtime

michaelhtm:
Description of changes:
These changes address the failures that happen when the
controller attempts to evaluate a static expression that the
user does not want to include, with includeWhen expression 
evaluating to false.


a-hilaly (dismissed):
Dankeshoen herr Michael!

a-hilaly (dismissed):

a-hilaly (approved):
Thanks @michaelhtm !

PR #515, opened 2025-04-17
Implement validating admission policies

gfrey:
This uses the prepared `Validation` field of the RGD to enable validating admission policies. Changed the type from `[]string` to `[]Validation` to support custom messages.

NOTE: Documentation is missing, but I'm unsure where to put it.

### DIFF
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -78,8 +78,7 @@ type Schema struct {
 	Status runtime.RawExtension `json:"status,omitempty"`
 	// Validation is a list of validation rules that are applied to the
 	// resourcegraphdefinition.
-	// Not implemented yet.
-	Validation []string `json:"validation,omitempty"`
+	Validation []Validation `json:"validation,omitempty"`
 }
 
 type Validation struct {
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -236,7 +236,7 @@ func (in *Schema) DeepCopyInto(out *Schema) {
 	in.Status.DeepCopyInto(&out.Status)
 	if in.Validation != nil {
 		in, out := &in.Validation, &out.Validation
-		*out = make([]string, len(*in))
+		*out = make([]Validation, len(*in))
 		copy(*out, *in)
 	}
 }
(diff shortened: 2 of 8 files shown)


api/v1alpha1/types.go
```diff
@@ -78,8 +78,7 @@ type Schema struct {
 	Status runtime.RawExtension `json:"status,omitempty"`
 	// Validation is a list of validation rules that are applied to the
 	// resourcegraphdefinition.
-	// Not implemented yet.
-	Validation []string `json:"validation,omitempty"`
+	Validation []Validation `json:"validation,omitempty"`
```
a-hilaly, line 81: Even thought this is technically a breaking change, this field was never documented nor used for anything in the past. So i'm ok with making this change to move fast.

thoughts @matthchr @barney-s ?

barney-s: Agreed

barney-s:
LGTM.
@gfrey Would it be possible to add a few tests. 
And for doc we can add it here: 
https://github.com/kro-run/kro/blob/main/website/docs/docs/concepts/00-resource-group-definitions.md#understanding-the-schema

gfrey:
@barney-s I added a test and at least an example on how to use it.
I really struggled with the test. The current implementation makes sure that the policy is properly captured in the schema. But that's it. I was wondering whether this policy can be validated itself. Would be nice to get an error when pushing the RGD, if an unknown field is used in such a policy. But I would need some guidance on how to implement something like this.

barney-s:
> I really struggled with the test.

Do you think the e2e-tests outlined in this PR #290 help define more robust tests for this use-case ? 
If so leave a comment in that PR.

barney-s (approved):

barney-s:
@gfrey  - did you not want to merge this change ?

gfrey:
@barney-s I'm not sure what happened here. The commits are on the main branch. Apparently [merged by you](https://github.com/kro-run/kro/commit/62d43461c64d038c63bfcf34715a0a2dc9cd62ed). :shrug: Thank you!

gfrey:
> > I really struggled with the test.
> 
> Do you think the e2e-tests outlined in this PR #290 help define more robust tests for this use-case ? If so leave a comment in that PR.

@barney-s I think the test is good enough for the feature I added. It makes sure that the policy is added in the right location.  The only thing I would change is the test location. Should be on a separate function. Would you think we should have an integration test to make sure K8s doesn't break the way these are handled?
I will then create a separate PR moving the test and potentially adding a e2e test, if deemed necessary.

PR #517, opened 2025-04-18
Fix required marker value handling

gfrey:
It will now check the values and handle them accordingly. The truthy values are "true", "yes", and "1". The supported falsey values are "false" and "no".

Any other value will return an error.

### DIFF
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -206,11 +206,21 @@ func (tf *transformer) handleSliceType(key, fieldType string) (*extv1.JSONSchema
 }
 
 func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Marker, key string, parentSchema *extv1.JSONSchemaProps) error {
+	truthyValues := []string{"yes", "true", "1"}
+	falseyValues := []string{"no", "false"}
+
 	for _, marker := range markers {
 		switch marker.MarkerType {
 		case MarkerTypeRequired:
-			if parentSchema != nil {
+			switch {
+			case parentSchema == nil:
+				return fmt.Errorf("required marker can't be applied; parent schema is nil")
+			case slices.Contains(truthyValues, marker.Value):
 				parentSchema.Required = append(parentSchema.Required, key)
+			case slices.Contains(falseyValues, marker.Value):
+				// ignore
+			default:
+				return fmt.Errorf("unsupported required marker value: %s", marker.Value)
 			}
 		case MarkerTypeDefault:
 			var defaultValue []byte
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -16,6 +16,7 @@ package simpleschema
 
 import (
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/simpleschema/transform.go
```diff
@@ -189,11 +190,21 @@ func (tf *transformer) handleSliceType(key, fieldType string) (*extv1.JSONSchema
 }
 
 func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Marker, key string, parentSchema *extv1.JSONSchemaProps) error {
+	truthyValues := []string{"yes", "true", "1"}
+	falseyValues := []string{"no", "false"}
```
a-hilaly: How about using `strconv.HasBool` https://pkg.go.dev/strconv#ParseBool

a-hilaly (commented):
Thanks Gereon, i left a comment below

pkg/simpleschema/transform.go
gfrey: I like; prevents all discussion about new values.

gfrey:
@a-hilaly Looking at the failed CI run above, I noticed that it fails, because the expected and gotten lists of required values have different order. Should I move the test to a separate function that is not using DeepEqual, or is there a different way to make the test stable?

gfrey:
I've added a dedicated test for the required marker value handling.

a-hilaly (approved):
Thank you!

michaelhtm:
I have a question: Should required just be a standalone value? if it's missing it implies the field is not required, if present we assume it is required, without necessarily needing to have `true` or `false`...

a-hilaly:
>I have a question: Should required just be a standalone value? if it's missing it implies the field is not required, if present we assume it is required, without necessarily needing to have true or false...

@michaelhtm  Currently, markers are expected to be in the form `a=b` where `a` is the marker name and `b` is the marker value. We could consider standlone value where users could do something like `string | required default=1`, however this makes the parsing much more complex, and will probably need to token/syntax-aware

gfrey:
> > I have a question: Should required just be a standalone value? if it's missing it implies the field is not required, if present we assume it is required, without necessarily needing to have true or false...
> 
> @michaelhtm Currently, markers are expected to be in the form `a=b` where `a` is the marker name and `b` is the marker value. We could consider standlone value where users could do something like `string | required default=1`, however this makes the parsing much more complex, and will probably need to token/syntax-aware

At some point I had modified the parser to make that happen. The change wasn't too crazy. I dismissed the approach as I wasn't sure how well the subsequent code would handle Markers with empty string values. The original code was lost, but I recreated it from what I remembered and pushed it to https://github.com/gfrey/kro/commit/3d40980d7682c3b05c6bd04da2fa06a17311b3ca. Would certainly need polish.

gfrey:
@a-hilaly @michaelhtm What do you think? Is this something to add to this PR?

gfrey:
Fixed a merge conflict with the force push.

gfrey:
Fixes #569

barney-s:
would you please rebase and fix the presubmit issue @gfrey

gfrey:
@barney-s done; should've caught that one on last push. Sorry!

a-hilaly:
Thank you @gfrey !

a-hilaly (approved):
Thank you 👍

PR #525, opened 2025-04-25
feat: KRO optionals

fabianburth:
**What this PR does**
Currently, currently it is not possible to reference status fields that are part of dynamic values in cel expressions. Details in the slack discussion [here](https://kubernetes.slack.com/archives/C081TMY9D6Y/p1743696445370789). This PR resolves this by introducing CEL optionals into KRO.


jakobmoellerdev:
@a-hilaly I think you mentioned that the implications on how we are now waiting for optionals should be mentioned in docs. If you can give a hint here on where this could go, I can add proper documentation.

jakobmoellerdev:
looks like int tests are failing, looking at this early next week again.

Edit: Ran the tests locally a few times and they ran through, either we just hit the timeout by chance or theres a flake in here. (maybe in the random selection of generated values from the emulator, gotta see in detail)

jakobmoellerdev:
@a-hilaly We would need a reapprove of the tests here, just doing a rebase for now

jakobmoellerdev:
Found the flake. an array without any type definition gets rejected by the emulator. That got me to think that its probably not a good idea to generate an explicit type definition. In JSON Scheme we can declare a type that is allowing anything as an empty JSON Scheme item. We can return that in the conversion and then return a random value that is generated in the emulator and will always pass.

pkg/cel/ast/inspector_test.go
```diff
@@ -300,6 +300,15 @@ func TestInspector_InspectionResults(t *testing.T) {
 				{Name: "validate"},
 			},
 		},
+		{
+			name:       "simple optional check",
+			resources:  []string{"bucket"},
+			expression: `bucket.?spec.name == "my-bucket"`,
```
jpbetz, line 306: I was expecting `bucket.?spec.name == optional.Of("mybucket")`

pkg/cel/conversions.go
```diff
@@ -45,6 +45,15 @@ func GoNativeType(v ref.Val) (interface{}, error) {
...
 		return v.ConvertToNative(reflect.TypeOf(map[string]interface{}{}))
+	case types.OptionalType:
+		opt := v.(*types.Optional)
+		if opt.HasValue() {
+			return GoNativeType(opt.GetValue())
+		}
+		// this looks weird, but it is the easiest way to get any generic value
+		// that conforms to interface
+		var i interface{}
+		return &i, nil
```
jpbetz: It looks like `GoNativeType` is quite similar to `v.Value()` (this addition is similar to https://github.com/google/cel-go/blob/8890f56dd657d3f4746ad1c53f55b65574457d29/common/types/optional.go#L114)

Can `GoNativeType` leverage or be replaced by `v.Value()` ?

pkg/cel/environment.go
```diff
@@ -54,6 +54,7 @@ func DefaultEnvironment(options ...EnvOption) (*cel.Env, error) {
 	declarations := []cel.EnvOption{
 		ext.Lists(),
 		ext.Strings(),
+		cel.OptionalTypes(),
```
jpbetz, line 57: Just FYI that Kubernetes has a well defined and versioned, environment that includes options settings and libraries:

xref: https://github.com/kubernetes/kubernetes/blob/9549613361f1c6f82b133a24f9badc70d8caff9d/staging/src/k8s.io/apiserver/pkg/cel/environment/base.go#L232, https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/cel/environment/base.go#L62-L186

jpbetz: If you like our libraries and CEL option choices, you could leverage this and avoid having to maintain it yourselves.

pkg/cel/ast/inspector.go
```diff
@@ -522,6 +522,7 @@ func isInternalIdentifier(name string) bool {
 
 func isInternalFunction(name string) bool {
 	internalFunctions := map[string]bool{
+		"_?._":    true,
```
jpbetz: Is this to ensure CEL doesn't leak new operations in without your knowledge?

pkg/cel/conversions.go
jpbetz: Loosely related-

I realize that his conversion code is to get values OUT of CEL.  But I wanted to mention that  Kubernetes has `UnstructuredToVal` and `TypedToVal` utilities under https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/apiserver/pkg/cel/common to make it easier to get Kuberentes flavored objects IN to CEL.

jpbetz (approved):

jakobmoellerdev:
I suggest this doesnt get merged without a dedicated integration test. WDYT guys?

pkg/cel/conversions.go
jakobmoellerdev: I will check on if we can replace the value method. On first glance it looks it could be reused but there are some special cases such as `Null`:

```
// Value implements ref.Val.Value.
func (n Null) Value() any {
	return structpb.NullValue_NULL_VALUE
}
```

this looks very different to how we handle it. Not saying it couldnt be used, but that would likely lead to a much larger changeset. Open to if we want to change this here.

pkg/cel/environment.go
jakobmoellerdev: While in general we can take a look at this, ultimately KRO has opinionation on the CEL syntax it supports so I would hope that we have a controlled subset (not all options) to enable our lib. 

For now I didn't feel the need to extend this but maybe using the k8s version from upstream is eventually beneficial? Im torn on this, happy to hear other thoughts

pkg/cel/ast/inspector.go
jakobmoellerdev: AFAIK yes. This is mainly because certain functions have implications on the dependency graph (such as optionals only being available at runtime). However there were attempts to get rid of this already: https://github.com/kro-run/kro/pull/528

pkg/cel/ast/inspector_test.go
jakobmoellerdev: isn't the `?` a specifier for an optional value while "my-bucket" could be anything? I would expect that because we have an equality comparison, the "right side" does not matter in how its terminated. Basically here Im checking:

"Is (value or null) equal to my-bucket", and youre suggesting "Is (value or null) equal to (my-bucket or null)". I believe for this Test I always have the value, so that wouldnt be needed

pkg/cel/conversions.go
jakobmoellerdev: On a side note, is there any way we can return a "any type" in kubernetes JSON Schema? Ideally i would like to return a nil value that resolves to an empty json scheme property. but this gets rejected with

```
CustomResourceDefinition.apiextensions.k8s.io "deploymentservices.kro.run" is invalid: spec.validation.openAPIV3Schema.properties[status].properties[optionalUnavailable].type: Required value: must not be empty for specified object fields
```

I would like to be able to do this because for optional values, KRO / we are not able to determine a correct type so it must be a catch all definition that can then be used in the CRD

jakobmoellerdev: I changed to v.Value() now for this instance as I believe I made it backwards compatible, but there are other usages of GoNativeType that I didnt touch. @a-hilaly what is our stance on GoNativeType in general?

jakobmoellerdev:
I now did the change corresponding to what I think should happen: The schema now generates an empty type definition with XPreserveUnknownFields that has no type defined so it can either be a primitive or an object. Additionally, I changed the emulator to return a nil value instead of randomizing gibberish. Last but not least I now added 2 fields in the integration test for deployment controller to proof the working optional fields.

jakobmoellerdev:
@a-hilaly I believe this one is ready for a review on the approach. Since we went over a few ideas, I would kindly ask you to read a little into the conversation above as well. 

Thanks again!

barney-s:
It would help to clarify the behavioral goals of optional fields. 

If the intent is to allow using dynamic (not well defined static types in CRD) path in RGD and WAIT/BLOCK on that field, to the best of my knowledge Optional would not cut it. It allows you to use dynamic paths, but would not WAIT for it. 

If the intent is to allow using dynamic paths and dont depend on the values it returns (null) then this is ok. 

I do see value in optionals, but it seems like it might trip up novices. 

Related Bug: https://github.com/kro-run/kro/issues/535

jakobmoellerdev:
IMO Optionals in CEL allow:
- Schemaless Access to fields or subobjects within a CRD that is otherwise strongly typed, under the condition that it is not required for a Schema to be resolved. Think of an "x-embedded-resource" case in which the schema is not yet known in advance. Of course as you mentioned this is not WAIT/BLOCK, because why should it block if its optional? For me the semantic meaning is sound.
- Dynamically specifying status fields that are not a MUST for a successful instance reconciliation (think of a status field that has purely informational character and is not always present)

> If the intent is to allow using dynamic paths and dont depend on the values it returns (null) then this is ok.

I think this is acceptable considering that and optional field should specify a zero-value that is reasonably representative of "not having a value" => null.

The fact that most novices won't know the entirety of CEL will likely make the point about fresh users a bit strange, as it is very rare to get someone that understands how a CEL optional works and can use it but will get surprised on the effects in the RGD.

Happy to hear your thoughts!

jakobmoellerdev:
@barney-s whats your stance on the above? Do you mind giving a quick update here?

a-hilaly:
thank you @fabianburth @jakobmoellerdev and @gfrey !

Ashish7812:
@fabianburth and @jakobmoellerdev  Can you add some documentation on the usage in the https://kro.run/ website? May be an example.

I created an issue https://github.com/kro-run/kro/issues/568 and Can I use the CEL optionals to fix the issue?

jakobmoellerdev:
of course, ill create a small doc addition as PR

PR #528, opened 2025-04-26
Use CEL env's HasFunction method

gfrey:
Instead of maintaining a list of known functions, let's ask the env.

The test contains a copy of the previous list, just to make sure, this doesn't break anything. Note that the collection functions are not functions but macros.

### DIFF
--- a/pkg/cel/ast/inspector.go
+++ b/pkg/cel/ast/inspector.go
@@ -256,7 +256,7 @@ func (a *Inspector) inspectCall(call *exprpb.Expr_Call, currentPath string) Expr
 		inspection.FunctionCalls = append(inspection.FunctionCalls, FunctionCall{
 			Name: fmt.Sprintf("%s.%s", a.exprToString(call.Target), call.Function),
 		})
-	} else if !isInternalFunction(call.Function) {
+	} else if !a.env.HasFunction(call.Function) {
 		// This is an unknown function, but not an internal one
 		inspection.UnknownFunctions = append(inspection.UnknownFunctions, UnknownFunction{Name: call.Function})
 	}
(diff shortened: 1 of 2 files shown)


pkg/cel/ast/inspector.go
```diff
@@ -256,7 +256,7 @@ func (a *Inspector) inspectCall(call *exprpb.Expr_Call, currentPath string) Expr
 		inspection.FunctionCalls = append(inspection.FunctionCalls, FunctionCall{
 			Name: fmt.Sprintf("%s.%s", a.exprToString(call.Target), call.Function),
 		})
-	} else if !isInternalFunction(call.Function) {
+	} else if !a.env.HasFunction(call.Function) {
```
barney-s, line 259: Would this also cover the function added via #506 ?

gfrey: Yes! The test below has `random.seededString` in the list of tested functions.

jakobmoellerdev, line 259: This should imo not be merged unless we have support for optionals as it can lead to side effects for functions which are not properly interpreted with kros cel support. Example is in https://github.com/kro-run/kro/pull/525

matthchr (approved):

pkg/cel/ast/inspector.go
a-hilaly: Are you concerned about backward compatibility. FYI, we are planning to add optionals support :)

a-hilaly:
@jakobmoellerdev we merged this if you still think this is gonna be an issue can you please reach out on slack or file a github issue :)

pkg/cel/ast/inspector.go
jakobmoellerdev: I was concerned that now Optional functions are allowed but Optional Support is not yet present. I think its fine if we add optionals support as discussed :)

PR #529, opened 2025-04-26
Move validating admission policy test

gfrey:
This sub-test (created in #515 doesn't belong to the DependencyValidation test. As there is no better place, move it to a separate function for testing the created OpenAPISchema.


barney-s (approved):

a-hilaly (approved):
👍

PR #534, opened 2025-05-01
chore: wrap RGD compile errors

michaelhtm:
**Description of changes:**
These changes introduce error wrapping during RGD building. 
Currently if there is any error in RGD, it is returned immediately. 
When we introduce a CLI, we want to be able to see all the errors 
in the RGD at once, instead of needing to run the `validate` command 
multiple times for every single error in the RGD.


barney-s:
@DhairyaMajmudar  Please review this PR. It seems to be similar to the one you are proposing in #563

a-hilaly:
@michaelhtm can you please take alook at https://github.com/kubernetes-sigs/kro/pull/654 and share if it's solving the same problem you tried to address?

PR #537, opened 2025-05-05
GCP Gemma TPU Inference example

barney-s:
Adding a GCP Gemma on TPU Example. This uses Project as external reference

This example depends on : #273


a-hilaly (approved):
👍

PR #540, opened 2025-05-07
example: show and example using Job (set no ttl)

barney-s:
Add an example showing how to create Jobs using KRO. Note we need to make sure the job stays by not setting any TTL


barney-s:
related issue discovered: #541

a-hilaly (approved):
👍

Issue #458, opened 2025-03-25
KRO Status Conditions


n3wscott:
### Feature Description

**Problem Statement**:

The conditions defined in the schema of a CRD from an RGD should align with the “single top-level condition” pattern. By not needing to know the implementation details of the controller reconciling the resource and only looking at the “top-level” condition, we enable a form of duck-typing on readiness that allows the ecosystem to create safer interactions.

We will adhere to the advice from upstream [Kubernetes API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties), except we will strongly prefer normal-true (positive polarity) conditions.

Since a Resource Graph will continue to update the downstream resources based on the RGD template, it will always have a Ready condition as its top-level condition. Based on the implementation of the controller in KRO, we will have sub-conditions that describe the major work of the Resource Graph reconciler.

**Proposed Solution**:
[KRO Status Conditions](https://docs.google.com/document/d/1-Wc8m9LH6wq0URIPqzdpHc_MfQzHfScP5PSzzFCLaxw/edit?usp=sharing)

**Alternatives Considered**:
None.

**Additional Context**:
See document.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

n3wscott:
I am happy to do this work, or help someone do the work with me.

fsommar:
Hey, I created a similar issue too https://github.com/kro-run/kro/issues/415

n3wscott:
I landed the RGD conditions but I would like to do a pass on Instance conditions. It could be for 0.5.

n3wscott:
I think https://github.com/kro-run/kro/issues/602 is a dup of this.

barney-s:
related: https://github.com/kubernetes-sigs/kro/pull/622

PR #544, opened 2025-05-09
Rewrite conditions for ResourceGraphDefinition leveraging ConditionSet

n3wscott:
Related to #458

This updates ResourceGraphDefinition to use the following conditions structure:

```
Ready
  ├─ ResourceGraphAccepted - This controller has accepted the spec.schema and spec.resources.
  ├─ KindReady - The CRD status created on behalf of this RGD.
  └─ ControllerReady - The status of the controller thread reconciling this resource.
```

And we can see this happening when running the demos,

NoOp:
```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  creationTimestamp: "2025-05-16T18:17:48Z"
  finalizers:
  - kro.run/finalizer
  generation: 2
  name: noop
  resourceVersion: "727591"
  uid: fae3f9d5-08dc-4c7e-b26f-27305e51cb0f
spec:
  resources: []
  schema:
    apiVersion: v1alpha1
    group: no.op
    kind: NoOp
    spec: {}
    status: {}
status:
  conditions:
  - lastTransitionTime: "2025-05-16T18:17:48Z"
    message: resource graph and schema is valid
    observedGeneration: 2
    reason: Valid
    status: "True"
    type: ResourceGraphAccepted
  - lastTransitionTime: "2025-05-16T18:17:48Z"
    message: controller is running
    observedGeneration: 2
    reason: Running
    status: "True"
    type: ControllerReady
  - lastTransitionTime: "2025-05-16T18:17:48Z"
    message: kind NoOp has been accepted and ready
    observedGeneration: 2
    reason: Ready
    status: "True"
    type: KindReady
  - lastTransitionTime: "2025-05-16T18:17:48Z"
    message: ""
    observedGeneration: 2
    reason: Ready
    status: "True"
    type: Ready
  state: Active
```

And we can see what it looks like when a resource kind has been typoed:

```
...
spec:
  resources:
  - id: deployment
    template:
      apiVersion: apps/v1
      kind: Deploymentsasda
...
status:
  conditions:
  - lastTransitionTime: "2025-05-16T18:35:37Z"
    message: kind DeploymentService has been accepted and ready
    observedGeneration: 1
    reason: Ready
    status: "True"
    type: KindReady
  - lastTransitionTime: "2025-05-16T18:35:37Z"
    message: controller is running
    observedGeneration: 1
    reason: Running
    status: "True"
    type: ControllerReady
  - lastTransitionTime: "2025-05-16T18:37:18Z"
    message: 'failed to build resource "deployment": failed to get schema for resource
      deployment: cannot resolve group version kind "apps/v1, Kind=Deploymentsasda":
      schema not found'
    observedGeneration: 2
    reason: InvalidResourceGraph
    status: "False"
    type: ResourceGraphAccepted
  - lastTransitionTime: "2025-05-16T18:37:18Z"
    message: 'failed to build resource "deployment": failed to get schema for resource
      deployment: cannot resolve group version kind "apps/v1, Kind=Deploymentsasda":
      schema not found'
    observedGeneration: 2
    reason: InvalidResourceGraph
    status: "False"
    type: Ready
  state: Inactive
```


helm/crds/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -207,15 +207,32 @@ spec:
...
+                      - Unknown
                       type: string
                     type:
-                      description: Type is the type of the Condition
+                      description: type of condition in CamelCase or in foo.example.com/CamelCase.
+                      maxLength: 316
+                      pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$
                       type: string
                   required:
+                  - lastTransitionTime
```
n3wscott: @a-hilaly this might be a deal breaker or not, what do you think of using metav1.Conditions and their kubebuilder annotations.

matthchr: Is the reason you think this might be a dealbreaker because it's net-new required fields?

It doesn't seem that problematic to me because this is readonly anyway, so required here is "really" just saying what we're guaranteeing to return - which I think we're already doing?

matthchr: Discussed on the community call. We are _not_ always setting these today, which is the problem.

n3wscott:
@a-hilaly Can I get a review please?

api/v1alpha1/conditions.go
```diff
@@ -77,17 +79,50 @@ type Condition struct {
...
+	return c.Status
+}
+
+// Conditions is a list of conditions.
+type Conditions []Condition
+
+// Set sets the provided condition into the conditions list, if it exists already the condition is replaced.
+func (conditions Conditions) Set(condition Condition) []Condition {
 	for i, c := range conditions {
 		if c.Type == condition.Type {
```
matthchr, line 102: Should we sort conditions to preserve order? [CAPI does](https://github.com/kubernetes-sigs/cluster-api/blob/main/util/conditions/setter.go#L66). I'm not sure if it's a best practice or not, but could imagine it might be worth doing to prevent races in condition write ordering causing ordering differences that make looking at conditions at first glace via describe (or some other UI) awkward, as they're not in a consistent order so it's harder to find what you want quickly.

api/v1alpha1/conditions.go
```diff
@@ -97,7 +132,8 @@ func SetCondition(conditions []Condition, condition Condition) []Condition {
 	return append(conditions, condition)
 }
 
-func HasCondition(conditions []Condition, t ConditionType) bool {
+// Has returns true if the conditions list contains the given condition type.
+func (conditions Conditions) Has(t ConditionType) bool {
```
matthchr, line 111: minor: Do we need a get-by-type too?

pkg/apis/condition.go
```diff
@@ -18,22 +18,16 @@
...
 )
 
-// Condition aliases the upstream type and adds additional helper methods
-type Condition metav1.Condition
-
 type Object interface {
 	client.Object
-	GetConditions() []Condition
-	SetConditions([]Condition)
+	GetConditions() []v1alpha1.Condition
```
matthchr, line 27: Intentionally not using `v1alpha1.Conditions` here?

api/v1alpha1/conditions.go
matthchr: Ah, you do sort on the ConditionSet types, so I assume the intent is that that's taken care of at that level.

Though I guess that begs the question, what is this method for/who uses it?

pkg/apis/condition_set_test.go
```diff
@@ -63,35 +66,64 @@ func (tr *TestResource) DeepCopyObject() runtime.Object {
...
 
-var ignoreFields = cmpopts.IgnoreFields(Condition{}, "LastTransitionTime")
+// Custom comparer for pointer fields
+var conditionComparer = cmp.Comparer(func(x, y v1alpha1.Condition) bool {
+	if x.Type != y.Type || x.Status != y.Status || x.ObservedGeneration != y.ObservedGeneration {
+		return false
+	}
+
+	// Compare Reason pointers
+	if (x.Reason == nil) != (y.Reason == nil) {
```
matthchr: minor, style: Can you not use `ptr.Equal` to make this simpler?

```
if !ptr.Equal(x.Reason, y.Reason) {
    return false
}
if !ptr.Equal(x.Message, y.Message) {
    return false
}

return true
```

pkg/controller/resourcegraphdefinition/controller_status.go
```diff
@@ -16,119 +16,50 @@ package resourcegraphdefinition
...
+	o *v1alpha1.ResourceGraphDefinition,
 	topologicalOrder []string,
 	resources []v1alpha1.ResourceInformation,
-	reconcileErr error,
 ) error {
 	log, _ := logr.FromContext(ctx)
 	log.V(1).Info("calculating resource graph definition status and conditions")
 
-	processor := NewStatusProcessor()
+	cs := rgdConditionTypes.For(o)
```
matthchr: minor, style, feel free to ignore: 
```suggestion
	conds := rgdConditionTypes.For(o)
```

(or conditions)?

api/v1alpha1/conditions.go
```diff
@@ -22,6 +22,8 @@ import (
 // ConditionType is a type of condition for a resource.
 type ConditionType string
 
+func (c ConditionType) String() string { return string(c) }
+
 const (
 	// ResourceGraphDefinitionConditionTypeGraphVerified indicates the state of the directed
```
matthchr, line 28: Are these condition types even used anymore? I think no, right?

Can we remove?

pkg/controller/resourcegraphdefinition/controller_status.go
```diff
@@ -171,14 +107,66 @@ func (r *ResourceGraphDefinitionReconciler) setUnmanaged(ctx context.Context, rg
...
+}
+
+// A ConditionsMarker provides an API to mark conditions onto a ResourceGraphDefinition as the controller does work.
+type ConditionsMarker struct {
+	cs apis.ConditionSet
+}
+
+// ResourceGraphValid signals the rgd.spec.schema and rgd.spec.resources fields have been accepted.
+func (m *ConditionsMarker) ResourceGraphValid() {
+	m.cs.SetTrueWithReason(ResourceGraphAccepted, "Valid", "resource graph and schema is valid")
```
matthchr: minor
```suggestion
	m.cs.SetTrueWithReason(ResourceGraphAccepted, "Valid", "resource graph and schema are valid")
```

pkg/apis/condition.go
n3wscott: yeah it is not part of that package and would create a circular dependency if the api pulls in the condition manager

api/v1alpha1/conditions.go
n3wscott: String() gets used in `fmt.Printf("%s or %v")`

n3wscott: oh sorry missed the point of the comment. Let me check.

n3wscott: I think so because the RGD instances use them. I have only changed RGD types

pkg/apis/condition_set_test.go
n3wscott: good idea! though it is just testing code.

api/v1alpha1/conditions.go
n3wscott: if we need it later we can add it.

pkg/controller/resourcegraphdefinition/controller_status.go
n3wscott: lgtm

api/v1alpha1/conditions.go
n3wscott: Deleted

api/v1alpha1/conditions.go
n3wscott: it is for the condition manager that knows about the api types.

pkg/apis/condition.go
n3wscott: I want to keep it this way to enable us to move this to upstream metav1.Condition. The array wrapper is an implementation detail more or less internal to the api package.

matthchr: Fine with this.

matthchr (approved):

matthchr:
Looks like some E2E tests are failing (I think newly added to CI checks). Can you take a look and update the chainsaw tests to the new condition output format?

n3wscott:
Will fix the e2e tests in a bit

PR #550, opened 2025-05-13
Update instance states documentation

7navyasa:
Updating the documentation with the current Instance states.


a-hilaly (approved):
Great catch, thank you @7navyasa !

PR #564, opened 2025-05-23
feat: Add support for custom CEL functions

### LINKED ISSUE #361: Add support for custom functions

fsommar:
This adds support for a new top-level field on the RGD that enables users to create their own functions that they can use in CEL expressions. It currently only supports primitive types like strings and ints, but can be extended for more complex types in the future. Adding support for other runtimes is possible, but is left for a future addition. One idea I have in mind is to refer to an external resource that can be used to execute the functions, but one could imagine embedding starlark and wasm runtimes in kro itself and run functions embedded verbatim in the RGD.

**breaking module API change**: I removed the Inspector constructor that did not take an environment as it wasn't used other than outside of tests and was signaling that an environment wasn't necessary when it created one on its own anyway.

I also updated the Inspector to look at all functions declared in the environment, not only those in the static allow list. This resulted in some tests needing to be updated as they were referring to functions added to the environment and not in the allow list (such as duration).

For environment, I updated it to store resourceIDs directly as custom declarations, to avoid having to duplicate logic in the new krocel.Extend function that extends a CEL environment instead of creating a new one. Extending the environment reduces some duplication in builder.go and I imagine a similar pattern can be used in the future with strongly typed resources being added to the environment (effectively having an environment builder that has an understanding of the schema and resource types).

As for the functions themselves, they are topologically sorted so that one function can refer to another, not caring about which order the functions are declared. Functions support overloads by input and output parameters.

Here's what an example function declaration looks like:

```yaml
  functions:
    - name: greeting
      inputs:
      - type: string
      celExpression: "'hello ' + _0 + '!'"
```

And then used as such: `${greeting("world")}`, resulting in the string "hello world!".

Closes #361.


barney-s:
@jpbetz Would you please take a look at this ?

matthchr:
First off: Thanks for the contribution and sorry for the long time without a reply here.

This is a pretty big change and as mentioned in the issue, there are at least a few possible ways to expose custom functions that would be good to consider. We've recently added a more [official process for discussing design proposals](https://github.com/kro-run/kro/pull/582).

Could you send a PR with a simple design proposal for this change, that way we have a more durable/official record of the plan and alternatives? You could discuss (at a high level) how this could be expanded in the future to support other custom function "engines", and also the pros/cons of doing function definition in the RGD versus in an external resource.

Ideally then we can generate consensus on the plan, via design doc, merge that, and then that should allow us to make forward progress on this PR once the plan has consensus.

Thank you!

fsommar:
Sounds like a good plan, will do! It might take a couple of weeks until I have something ready.

matthchr:
We're trying to go through and clean up/manage PRs some. Can you please take what you need from this PR by Sept 1 and we'll close it (for now) in favor of the design document discussed above. Based on that document and plan you may be able to reopen this at a future date.

a-hilaly:
closing as this probably deserves a supporting design doc, wanna open KREP-003 @fsommar ? :)

PR #567, opened 2025-05-24
Fix lint errors

DhairyaMajmudar:
## Description

While running the command `make lint` there showed up several linting errors were made, This PR is focused on resolving those errors.

Screenshot from 2025-05-24 09-40-26

### DIFF
--- a/pkg/controller/instance/controller_reconcile.go
+++ b/pkg/controller/instance/controller_reconcile.go
@@ -23,7 +23,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 
 	"github.com/kro-run/kro/pkg/controller/instance/delta"
@@ -141,7 +140,7 @@ func (igr *instanceGraphReconciler) reconcileInstance(ctx context.Context) error
 // setupInstance prepares an instance for reconciliation by setting up necessary
 // labels and managed state.
 func (igr *instanceGraphReconciler) setupInstance(ctx context.Context, instance *unstructured.Unstructured) error {
-	patched, err := igr.setManaged(ctx, instance, instance.GetUID())
+	patched, err := igr.setManaged(ctx, instance)
 	if err != nil {
 		return err
 	}
@@ -439,7 +438,7 @@ func (igr *instanceGraphReconciler) finalizeDeletion(ctx context.Context) error 
 }
 
 // setManaged ensures the instance has the necessary finalizer and labels.
-func (igr *instanceGraphReconciler) setManaged(ctx context.Context, obj *unstructured.Unstructured, uid types.UID) (*unstructured.Unstructured, error) {
+func (igr *instanceGraphReconciler) setManaged(ctx context.Context, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
 	if exist, _ := metadata.HasInstanceFinalizerUnstructured(obj); exist {
 		return obj, nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/controller/instance/controller_reconcile.go
```diff
@@ -439,7 +438,7 @@ func (igr *instanceGraphReconciler) finalizeDeletion(ctx context.Context) error
 }
 
 // setManaged ensures the instance has the necessary finalizer and labels.
-func (igr *instanceGraphReconciler) setManaged(ctx context.Context, obj *unstructured.Unstructured, uid types.UID) (*unstructured.Unstructured, error) {
+func (igr *instanceGraphReconciler) setManaged(ctx context.Context, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
 	if exist, _ := metadata.HasInstanceFinalizerUnstructured(obj); exist {
```
barney-s, line 443: Try doing _ := uid and see if it removes the lint error. That way we dont have to change the func signature.

pkg/simpleschema/transform.go
```diff
@@ -147,7 +147,7 @@ func (tf *transformer) handleMapType(key, fieldType string) (*extv1.JSONSchemaPr
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse map type for %s: %w", key, err)
 	}
-	if keyType != "string" {
```
barney-s, line 150: i would define a const keyTypeString = "string"

barney-s (commented):
minor comments

pkg/controller/instance/controller_reconcile.go
DhairyaMajmudar: Yeah doing this  `_ = uid` worked, Thanks! :)

pkg/simpleschema/transform.go
DhairyaMajmudar: Done Done, added a constant for "string"

barney-s (approved):
LGTM. Thanks for the PR.

pkg/controller/instance/controller_reconcile.go
```diff
@@ -440,6 +440,7 @@ func (igr *instanceGraphReconciler) finalizeDeletion(ctx context.Context) error
 
 // setManaged ensures the instance has the necessary finalizer and labels.
 func (igr *instanceGraphReconciler) setManaged(ctx context.Context, obj *unstructured.Unstructured, uid types.UID) (*unstructured.Unstructured, error) {
+	_ = uid
```
a-hilaly: you could also:
```suggestion
func (igr *instanceGraphReconciler) setManaged(ctx context.Context, obj *unstructured.Unstructured, _ types.UID) (*unstructured.Unstructured, error) {
```

pkg/simpleschema/transform.go
a-hilaly: i like `keyTypeString`, since OpenAPISchema and SimpleSchema are technically identical, how about using https://github.com/kro-run/kro/blob/main/pkg/simpleschema/atomic.go#L36-L37 ?

a-hilaly (commented):
thanks @DhairyaMajmudar ! left two comments in line

pkg/controller/instance/controller_reconcile.go
DhairyaMajmudar: Clean and simple, Thanks for this suggestion

pkg/simpleschema/transform.go
DhairyaMajmudar: Then I need to typecast like this `if keyType != string(AtomicTypeString)`

barney-s: I think that is ok. But we should also change the "integer" in the other case statement as well. 
We do introduce "number" which is not in openapi. So shrug. 

Lets define a set of consts which map to openapi
const (
keyTypeString = string(AtomicTypeString)
keyTypeInteger = string(AtomicTypeInteger)
keyTypeNumber = "number"
)

DhairyaMajmudar: Cool, added these new constants in my latest commit. 

> We do introduce "number" which is not in openapi

We have [AtomicTypeFloat](https://github.com/kro-run/kro/blob/main/pkg/simpleschema/atomic.go#L35) how about using it here in place of number?

barney-s: thank you

PR #575, opened 2025-05-30
kro cli: Create and dry run command

markoskandylis:
# Add Create Command with Dry-Run Functionality to KRO CLI

## Description
Adds the `create` command to the KRO CLI with dry-run functionality. This enhancement allows users to preview RGD (Resource Graph Definition) changes before applying them to the cluster.

## Features
- Added `create rgd` command
- Implemented `--dry-run` flag option
- Shows delta comparison of changes to be applied

## Implementation Details
- Command syntax: 
  - Basic usage: `kro create rgd [FILE]`
  - With dry-run: `kro create rgd [FILE] --dry-run`
- Shows detailed resource changes including:
  - Path modifications
  - Old vs New value comparisons
  - Resource definition updates

Screenshot 2025-05-30 at 10 19 20

Demo

https://github.com/user-attachments/assets/29f098b0-c197-4524-98d7-7d1a937a58f0


barney-s:
Kubectl has a diff. Would this add more functionality than what is available in kubectl ? 
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_diff/

Also we would need to rebase this to conform to the new cli directory structure. 

We see value in showing diff in the rendered objects for an RGD instances. Thats not what this PR seems to be doing.

markoskandylis:
Hi @barney-s, thanks for the feedback!
The end goal is to have something like a terraform plan built into the Kro CLI — a Kro-aware preview of all changes before apply.
This PR is just the first step (diffing the RGD definition)  before extending it to instance-level, rendered-object diffs.
From a DevOps perspective, I’d like this available in the CLI so it can be easily used in pipelines as part of the deployment process.
Since you also see value in showing diffs at the rendered object level, I can start working on the instance-level extension next — which will be particularly important when Kro orchestrates cloud resources (ACK/Config Connector/ASO) so users can see potential creates, updates, or destroys before they happen.

barney-s:
@markoskandylis - Would you be inclined to implement the terraform plan like capability instead of this PR ? You can do it in a new PR or replace this PR code with the plan like functionality.

a-hilaly:
Thank you for your PR @markoskandylis - i think this feature will need a KREP to discuss exactly what dry runs mean for the kro CLI.

Issue #17, opened 2024-09-21
Mega Feature: Implement collections (for loops) in RGDs


a-hilaly:
Add support for collections in ResourceGroups to allow dynamic creation of multiple similar resources.. This will probably need a design and a review. It also needs the garbage collection mechanisms to be in place.

- Implement iteration based on a range or list
- Add current iteration index reference within resource definitions
- Support specifying "deployment strategies" for collection types

Example:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: worker-pool
spec:
  schema:
    apiVersion: v1alpha1
    kind: WorkerPool
    spec:
      workers: "[]string"
      image: string

  resources:
    - id: workerPods
      forEach:
        - worker: ${schema.spec.workers}
      template:
        apiVersion: v1
        kind: Pod
        metadata:
          name: ${schema.metadata.name + '-' + worker}
        spec:
          containers:
            - name: app
              image: ${schema.spec.image}
```

gfrey:
Index reminds me of Terraform's count iteration concept. This had plenty of downsides when combined with lists of actual configuration values. Any change to the that list would result in modification on all subsequent list items. This was eventually solved with `for_each` based on sets or dicts. Then the item or key was used as part of the resource address. This made things a lot more stable. Possible solution could be to change the relevant part of the example above to:
```
collection:
  items: ${range(0, spec.replicas)}
  deploymentStrategy: RollingUpdates
  definition:
    metadata:
      name: pod-unit-${item}
```
An immediate application for this would be distribution of a VPC's subnets to availability-zones. Then the list of `items` would be the AZs (or maybe a dictionary from AZ to CIDR block).

A follow-up question would be how the resources of a collection can be queried? In the subnets application I would need to associate a NetworkACL with all these subnets. How could that be done?

greghaynes:
Would the above solutions be limited in that they cannot allow for multiple sources of iteration within a single resource? 

Here's a contrived example: Imagine we want to support a RG which supports defining multiple containers where some values are defaulted below each container.
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGroup
metadata:
  name: Application
spec:
  apiVersion: v1alpha1
  kind: Application
  definition:
    spec:
      name: string
    containers:
      - name: frontend
        image: nginx
      - name: backend
        image: nginx
```

[...]

PR #580, opened 2025-06-07
WIP: Add iterators to resourceGraphDefinition to allow dynamic creation of multiple similar resources

VladRico:
Hi there ! 
Here's a "for loop" implementation related to #17 
tested on kubernetes `v1.31.2`

## Overview
it adds a `iterators` to the schema:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: loop
spec:
  schema:
    apiVersion: v1alpha1
    kind: Loop
    spec:
      name: "string | required=true"
      hostIP: "string | required=true"
      # Use custom type 
      ports: "[]port | required=true"
    # Define port type
    types:
      port:
        name: string | required
        protocol: string | default="TCP"
        port: integer | required
    # new iterators feature
    iterators:
      - name: deploymentContainerPorts
        iterator: item
        input: schema.spec.ports
        render:
          - containerPort: ${item.port}
            name: ${item.name}
            protocol: ${item.protocol}
            hostIP: ${schema.spec.hostIP}

  resources:
    - id: deployment
      template:
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: ${schema.spec.name}
        spec:
          replicas: 1
          selector:
            matchLabels:
              type: ${schema.spec.name}-feature
          template:
            metadata:
              labels:
                type: ${schema.spec.name}-feature
            spec:
              containers:
                - name: dev
                  image: nginx
                  ports:
                    ${iterators.deploymentContainerPorts}
---
```
Used with an instance like:
```yaml
apiVersion: kro.run/v1alpha1
kind: Loop
metadata:
  name: loop-example
spec:
  name: dev-loop
  hostIP: "10.13.37.1"
  ports:
    - name: http
      protocol: TCP
      port: 8080
    - name: https
      protocol: TCP
      port: 443
```

It produces this result:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: dev-loop
spec:
  replicas: 1
  selector:
    matchLabels:
      type: dev-loop-feature
  template:
    metadata:
      labels:
        type: dev-loop-feature
    spec:
      containers:
        - name: dev
          image: nginx
          ports:
            - containerPort: 8080
              name: http
              protocol: TCP
              hostIP: "10.13.37.1"
            - containerPort: 443
              name: https
              protocol: TCP
              hostIP: "10.13.37.1"
```

You can find in `examples/iterators/*` the 3 files used above.
As it is just an implementation idea, **I haven't checked whether this opens any security flaws**. I'm thinking about various injection from the instance, something like [that](https://www.wiz.io/blog/ingress-nginx-kubernetes-vulnerabilities) for example, even if it's not really the same thing.

## Notes
As this is a fairly complex feature, because it can be implemented in so many different ways, I'm not really sure that it matches what maintainers have in mind.

I have a real need for this kind of feature, that's why I tried to meet requirements based on what peoples said in #17 
Feel free to suggest changes if you like the implementation, or close it if it's too far away from what you have in mind

Thanks for you time working on this project !


handicraftsman:
Feature wise looks interesting (haven't checked the code), but it is missing an important piece: what if I want to request e.g. several instances of Postgres, where each of them is created by instantiating a custom resource? (like in zalando operator)

As a note, I am not checking the code here, because I am both not a part of kro community and I am not a Go developer.

VladRico:
@handicraftsman I'm sorry but I didn't fully understand your point. Could you please provide an example of the result wanted ?

handicraftsman:
Basically, I would love to have an `iterator` expand into multiple items in `resources` array.
For example, I might want to have multiple `Service`s created, with different policies applied to each of them.

For example, one `Service` would be accessible from the Internet, whereas the other service would only be reachable through a VPN, and used for management purposes.

Since I mentioned management purposes, here's an example: in Java there's [JMX](https://en.wikipedia.org/wiki/Java_Management_Extensions), which allows connecting to a Java app remotely, and issuing one or more commands there. I know there are better ways to do so, but this is just a theoretical example.

Or, for example, we are working with a similar case in production, but instead of exposing JMX API, our services expose a separate API for managing the system through an internal console. This also allows resolving support cases for us, allowing for manual administrative actions that are only available under VPN.

VladRico:
Sorry for the late answer. Ok I may have understand your requirement @handicraftsman and I also have this needs. I'll try to implement the feature but I can't promise anything

matthchr:
There are multiple efforts in flight on this topic, and I think the consensus is that we should have a design first before we land on an implementation. @a-hilaly is leading that effort - can you engage with him and work together?

Marking this donotmerge in the meantime.

a-hilaly:
Hey @VladRico ! Thank you so much for taking a stab on this, i'm currently in the process of writing a design doc on collections (or iterators as you're calling it :) ) - and would love to get your feedback/comments on it. Happy to collaborate with you to get this one out!

/close

VladRico:
Hey @a-hilaly ! 
Feel free to ping me when you release the design doc on collections. I'm not sure my feedback is worth something as I'm not actually a skilled dev ^^'
I'll be very busy the next month but I may have some free time near the end of October

a-hilaly:
@VladRico I recently opened https://github.com/kubernetes-sigs/kro/pull/679 your feedback would be highly valuable! let's incorporate any ideas you had in mind. I'm also planning on having an implementation plan, and you're welcome to pick a workstream if you're interrested!

PR #583, opened 2025-06-13
kro cli generate crd command

DhairyaMajmudar:
# Description

1. generates CRDs from input RGD files
2. prints the generated output either on terminal or creates a new file on the basis of flags given 

## Usage

 `kro generate crd -f [FILE]`

Flags:

| Shorthand | Flag | Usage | 
| ---- | ----- | --- |
 | -F |  --format |   Output format (yaml/json) |
 |  -o | --output  |   Path to the output file |

## Demo

[Screencast from 2025-06-13 20-25-09.webm](https://github.com/user-attachments/assets/e0a2d8bb-d04c-4a43-b782-eb892dbb4a9d)

### DIFF
--- a/cmd/kro/commands/generate.go
+++ b/cmd/kro/commands/generate.go
@@ -19,17 +19,17 @@ type GenerateConfig struct {
 
 var config = &GenerateConfig{}
 
+func init() {
+	generateCmd.PersistentFlags().StringVarP(&config.resourceGroupDefinitionFile, "file", "f", "",
+		"Path to the ResourceGroupDefinition file")
+	generateCmd.PersistentFlags().StringVarP(&config.outputFormat, "format", "o", "yaml", "Output format (yaml|json)")
+}
+
 var generateCmd = &cobra.Command{
 	Use:   "generate",
 	Short: "Generate the CRD from a given ResourceGroupDefinition",
 	Long: "Generate the CRD from a given ResourceGroupDefinition." +
 		"This command generates a CustomResourceDefinition (CRD) based on the provided ResourceGroupDefinition file.",
-}
-
-func init() {
-	generateCRDCmd.PersistentFlags().StringVarP(&config.resourceGroupDefinitionFile, "file", "f", "",
-		"Path to the ResourceGroupDefinition file")
-	generateCRDCmd.PersistentFlags().StringVarP(&config.outputFormat, "format", "o", "yaml", "Output format (yaml|json)")
 }
 
 var generateCRDCmd = &cobra.Command{
--- a/cmd/kro/commands/generate.go
+++ b/cmd/kro/commands/generate.go
@@ -83,27 +83,12 @@ func generateCRD(rgd *v1alpha1.ResourceGraphDefinition) error {
 	crd := rgdGraph.Instance.GetCRD()
 	crd.SetAnnotations(map[string]string{"kro.run/version": "dev"})
 
-	b, err := marshalObject(crd, outputFormat)
+	b, err := marshalObject(crd, config.outputFormat)
 	if err != nil {
 		return fmt.Errorf("failed to marshal CRD: %w", err)
 	}
 
-	if outputFile != "" {
-		if err := writeFile(outputFile, b); err != nil {
-			return fmt.Errorf("failed to write output file: %w", err)
-		}
-	} else {
-		fmt.Println(string(b))
-	}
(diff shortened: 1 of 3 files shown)


cmd/kro/commands/generate.go
```diff
@@ -0,0 +1,140 @@
...
+	Short: "Generate the CRD from a given ResourceGroupDefinition",
+	Long: "Generate the CRD from a given ResourceGroupDefinition." +
+		"This command generates a CustomResourceDefinition (CRD) based on the provided ResourceGroupDefinition file.",
+}
+
+func init() {
+	generateCRDCmd.PersistentFlags().StringVarP(&resourceGroupDefinitionFile, "file", "f", "",
+		"Path to the ResourceGroupDefinition file")
+	generateCRDCmd.PersistentFlags().StringVarP(&outputFile, "output", "o", "", "Path to the output file")
+	generateCRDCmd.PersistentFlags().StringVarP(&outputFormat, "format", "F", "yaml", "Output format (yaml|json)")
```
michaelhtm: should we give the `-o` flag the same functionality as `kubectl get <kind> -o?` (-oyaml, -ojson, etc..)
If someone wants to output to a file they can just use `>` output redirection.
any thoughts?

cmd/kro/commands/generate.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/kro-run/kro/pkg/graph"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	resourceGroupDefinitionFile string
+	outputFile                  string
+	outputFormat                string
+)
```
michaelhtm: nit: should we put these in some sort of Config struct?

cmd/kro/commands/generate.go
```diff
@@ -0,0 +1,140 @@
...
+		if err != nil {
+			return nil, err
+		}
+	case "yaml":
+		b, err = yaml.Marshal(obj)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("unsupported output format: %s", outputFormat)
```
michaelhtm, line 125: why do we want the format flag to be required?

cmd/kro/commands/generate.go
DhairyaMajmudar: Yes! That would be much better, let's do this ? 

```golang
type GenerateConfig struct {
	resourceGroupDefinitionFile string
	outputFormat                string
}
```

cmd/kro/commands/generate.go
DhairyaMajmudar: Nope it's not required : )

It is defaulted to `yaml` and a user can pass `json` value through flag to change the format.
Here the default is a fallback suppose in place of json or yaml user gives a garbage value or any unsupported format.

cmd/kro/commands/generate.go
DhairyaMajmudar: yup, let's do it this way

DhairyaMajmudar:
Thank you @michaelhtm for review comments : )

I've added the suggested changes in my latest commits

cmd/kro/commands/generate.go
michaelhtm: Yes that looks good

cmd/kro/commands/generate.go
michaelhtm: ohh that makes sense. Thanks!

cmd/kro/commands/generate.go
```diff
@@ -0,0 +1,125 @@
...
+	outputFormat                string
+}
+
+var config = &GenerateConfig{}
+
+var generateCmd = &cobra.Command{
+	Use:   "generate",
+	Short: "Generate the CRD from a given ResourceGroupDefinition",
+	Long: "Generate the CRD from a given ResourceGroupDefinition." +
+		"This command generates a CustomResourceDefinition (CRD) based on the provided ResourceGroupDefinition file.",
```
michaelhtm, line 32: will we generate other things besides a CRD? just wondering if we should instead have a command called `generate-crd`

DhairyaMajmudar: Yes, we are going to generate rbac, docs, and diagrams using generate command

barney-s (approved):

PR #589, opened 2025-06-26
kro cli generate diagram

DhairyaMajmudar:
# Description

Command to generate the graph visualisation from the kro RGDs

## Usage

 `kro generate diagram -f [FILE]`

Flags:

| Shorthand | Flag | Usage | 
| ---- | ----- | --- |

## Demo

[Screencast from 2025-07-15 22-45-05.webm](https://github.com/user-attachments/assets/6ef3b29f-4ea4-419b-8f93-304b057ea12d)


n3wscott:
Do you plan to render edges too?

DhairyaMajmudar:
Yes edges are supported too for dependent resources

The test sample which I'm using was not having any dependent resources let me add a demo with a sample having dependent resources.

cmd/kro/commands/generate.go
```diff
@@ -1,10 +1,26 @@
...
 import (
 	"encoding/json"
 	"fmt"
 	"os"
 
+	"github.com/go-echarts/go-echarts/v2/charts"
+	"github.com/go-echarts/go-echarts/v2/opts"
 	"github.com/kro-run/kro/api/v1alpha1"
 	kroclient "github.com/kro-run/kro/pkg/client"
 	"github.com/kro-run/kro/pkg/graph"
```
a-hilaly, line 26: nit: there is an import ordering style for golang projects, typically it's follow this format:
```golang
import (
  // standard library imports

  // external imports

  // internal imports (under kro-run/kro/pkg)
)
```

a-hilaly (commented):
w00t! this is really cool!

cmd/kro/commands/generate.go
```diff
@@ -81,6 +126,83 @@ func generateCRD(rgd *v1alpha1.ResourceGraphDefinition) error {
...
+					Repulsion: 5000,
+				},
+				Roam:           opts.Bool(true),
+				EdgeSymbol:     "arrow",
+				EdgeSymbolSize: []int{0, 7},
+			},
+		),
+	)
+
+	return graph.Render(os.Stdout)
```
a-hilaly, line 242: should we also write the file to a flag `-o` | `--output`

cmd/kro/commands/generate.go
DhairyaMajmudar: The import ordering is auto corrected by the `goimport` linter here https://github.com/kro-run/kro/blob/main/.golangci.yml#L28

We can configure Go Linter to use the specified import ordering, but it will create changes in around 49 files of the codebase

Ashish7812:
@DhairyaMajmudar is it possible to add deployment order to each node of the graph? This way we can get to know the order of deployment visually also.  If any resources need to be deployed in parallel then it will have the same deployment order.

image

DhairyaMajmudar:
That's a cool idea, I beleive we can, let me give it a try

DhairyaMajmudar:
Here we go : )

I tried to get the text on the graph nodes w/o performing the hover, but it seems like the lib. used doesn't not support it 

image

Ashish7812:
@DhairyaMajmudar In a RGD there can be resources defined under template / externalRef.

Can you put a different colour for externalRef? This way we can get to know different resource flavours referred in the RGD.
If possible you can add a legend also.

image

Ashish7812:
> Here we go : )
> 
> I tried to get the text on the graph nodes w/o performing the hover, but it seems like the lib. used doesn't not support it
> 
> image

This is awesome :).

a-hilaly:
@Ashish7812 great ideas, love it!

DhairyaMajmudar:
> @DhairyaMajmudar maybe you want to address the features @Ashish7812 requested in a seperate PR?

Yup that would be a good idea

PR #590, opened 2025-06-27
Add e2e tests using chainsaw

barney-s:
We attempted to roll our own E2E framework in https://github.com/kro-run/kro/pull/290 
Based on community discussions we decided to try Chainsaw. 

The `[]steps{input,outputs}` model the DIY framework match is a direct match to the `[]steps{apply, assert}` model of chainsaw.

So we can port the tests from the other PR to this as is. The go-test code in the DIY framework would be translate to `chainsaw-test.yaml` for each test scenario (folder)

Ported over the TestAutoscaledDeploymentRGD to multiple-resource test here

[Discussion notes](https://docs.google.com/document/d/1GqeHcBlOw6ozo-qS4TLdXSi5qUn88QU6dwdq0GvxRz4/edit?tab=t.0#bookmark=id.87q8fnqwyu01)
Thanks Mariam Fahmy.


test/e2e/multiple-resources/step0/inputs/rgd.yaml
```diff
@@ -0,0 +1,87 @@
...
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: AutoscaledDeployment
+    spec:
+      name: string | "some-app"
+      image: string | "nginx:latest"
+      minReplicas: integer | 1
+      maxReplicas: integer | 10
+      targetCPUUtilization: integer | 80
```
a-hilaly: @barney-s i think you just caught a bug in kro :)
```suggestion
      minReplicas: integer | default=1
      maxReplicas: integer | default=10
      targetCPUUtilization: integer | default=80
```

Makefile
```diff
@@ -268,4 +280,14 @@ ko-apply: ko
...
+
+##@ E2E Tests
+
+.PHONY: test-e2e
+test-e2e: chainsaw ## Run e2e tests
+	$(CHAINSAW) test ./test/e2e/
+
+.PHONY: test-e2e-kind
+test-e2e-kind: deploy-kind
+	make test-e2e
```
matthchr, line 293: minor, nit, style: newline at EoF?

test/e2e/multiple-resources/chainsaw-test.yaml
```diff
@@ -0,0 +1,110 @@
...
+    - apply:
+        file: step3/inputs/instance.yaml
+      description: Update AutoscaledDeployment Instance
+    - assert:
+        file: step3/outputs/deployment.yaml
+      description: Verify Deployment state
+    - assert:
+        file: step3/outputs/hpa.yaml
+      description: Verify HPA state
+    catch:
```
matthchr: minor: As you mentioned in the call, do we want to test this catch logic to ensure we actually get the logs?

Better to test it now and make sure it works before we copy it into a bunch of other tests?

Makefile
matthchr: Applies throughout

test/e2e/multiple-resources/step3/scenario.txt
```diff
@@ -0,0 +1 @@
+Update the instance to increase replicas to 3 and adjust HPA settings (min: 2, max: 5)
```
matthchr: Remove scenario.txt as it's not used/needed anymore?

test/e2e/multiple-resources/chainsaw-test.yaml
```diff
@@ -0,0 +1,110 @@
...
+    - description: kro controller pod logs collector
+      podLogs:
+        selector: app.kubernetes.io/instance=kro
+    finally:
+    - description: sleep operation
+      sleep:
+        duration: 5s
+
```
matthchr: minor: Excess lines at the eof here, not needed?

test/e2e/multiple-resources/chainsaw-test.yaml
barney-s: fixed and verified.

test/e2e/simple-deployment/chainsaw-test.yaml
```diff
@@ -0,0 +1,75 @@
...
+  creationTimestamp: null
+  name: simple-deployment
+spec:
+  description: | 
+    SimpleDeployment tests a ResourceGraphDefinition that creates a single Deployment
+    The test verifies
+    1. Creating RGD installs the Instance SimpleDeployment CRD
+    2. Creating SimpleDeployment instance creates a Deployment
+    3. Changing the replicas in the SimpleDeployment instance changes the Deployment
+    4. Deleting the SimpleDeployment instance deletes the Deployment
```
MariamFahmy98: I can't see where we delete the `SimpleDeployment` and check if the instance gets deleted.

You can add the following:
```
  - name: delete the rgd
    try:
    - delete:
        ref:
          apiVersion: kro.run/v1alpha1
          kind: ResourceGraphDefinition
          name: simple-deployment-rgd
  - name: check that the deployment is deleted
     try:
      - error:
          file: deployment-error.yaml
```

The file `deployment-error` should be the manifest that you expect to be deleted.

MariamFahmy98 (commented):
I'm really pleased to see that Chainsaw works well for your end-to-end use cases — great work overall! I just had a minor comment regarding the test directory structure.

Currently, it’s a bit hard to follow the flow of each test, and I think structuring them around individual use cases might improve clarity. I’d suggest having a separate Chainsaw test focused on each scenario.

For example, you might want to verify the following:
1. Creating an RGD installs the Instance SimpleDeployment CRD.
2. Creating a SimpleDeployment instance creates a Deployment.
3. Updating the replicas in the SimpleDeployment instance updates the Deployment.
4. Deleting the SimpleDeployment instance deletes the Deployment.

It might be cleaner to group these into three focused Chainsaw tests:

1. One for creation (steps 1 and 2),
2. One for update (step 3),
3. And one for deletion (step 4).

Each test can live in its own directory with a brief README summarizing what it covers and what is expected. That would improve readability and help contributors quickly understand the intent behind each test.

Also, I don’t think we need too many nested directories — one directory per test should be enough.

For example, a directory like `check-rgd-creation` could include:
1. `chainsaw-test.yaml`
2. `rgd.yaml`
3. `rgd-assert.yaml` – verifies the RGD status
4. `rgd-crd-assert.yaml` – asserts the CRD creation
5. `rgd-instance.yaml` – defines the instance to be created
6. `deployment-assert.yaml` – expected result after instance creation
7. `rgd-instance-assert.yaml` – confirms the instance status
8. `rgd-instance-update.yaml` – applies an update
9. `updated-deployment-assert.yaml` – expected result of the update

Let me know what you think! Happy to help iterate further if needed.

n3wscott:
To compare, do you plan to do similar work with the kube e2e test as well?

barney-s:
> To compare, do you plan to do similar work with the kube e2e test as well?

Can we have both ?
I did a cursory look at https://github.com/kubernetes-sigs/e2e-framework?tab=readme-ov-file 

my quick takeaways comparing to chainsaw:
1. go based vs yaml based
2. has cluster, namespace creation support vs automated namespace creation in chainsaw
3. go-code assertions for complex test scenarios vs cel/JMESPath in chainsaw
4. granular control flow in tests vs high level scenario based flow in chainsaw
..

My takeaway is, approximately the 80:20 rule could apply to chainsaw vs kube-e2e tests.
80% could be easily implemented in chainsaw 
20% in go-lang based tests (kube-e2e)

Happy to hear your thoughts @n3wscott @MariamFahmy98

barney-s:
> 1. One for creation (steps 1 and 2),
> 2. One for update (step 3),
> 3. And one for deletion (step 4).

If testcase 2 requires 1 should we copy over inputs for 1 in 2 ?
Same for 3: should we copy over inputs for 1,2 in 3 ? 

Can we make tests depend on a previous test ? 

We would need to create separate RGD and instance (CRD) names so that the tests case run in parallel.
I was kind of trying to avoid it.

barney-s:
@MariamFahmy98 - I have restructured the tests. Please take a look.

n3wscott:
I like the 80/20 plan, seems like a great option to me

MariamFahmy98:
> If testcase 2 requires 1 should we copy over inputs for 1 in 2 ?
> Same for 3: should we copy over inputs for 1,2 in 3 ?

Yes, you should copy inputs for testcase 1 in testcase 2.

An alternative approach is to define a StepTemplate. You can create a _step-templates directory to store various step templates. For instance, since each test creates an RGD resource, we can define a StepTemplate for that purpose, like this:

```
apiVersion: chainsaw.kyverno.io/v1alpha1
kind: StepTemplate
metadata:
  name: resource-graph-definition
spec:
  try:
  - create:
      resource:
        apiVersion: kro.run/v1alpha1
        kind: ResourceGraphDefinition
        metadata:
          name: ($name)
        spec:
          schema:
            apiVersion: v1alpha1
            kind: CheckInstanceCreationSimpleDeployment
            spec:
              replicas: integer | default=1
              image: string | default="nginx:latest"
          resources:
            - id: deployment
            .....
```

And, then in each chainsaw test, you can use this template as follows:
```
  - name: create-rgd
    use:
      template: ../../_step-templates/create-rgd-template.yaml
      with:
        bindings:
        - name: name
          value: rgd
```

>Can we make tests depend on a previous test ?

Unfortunately, no. Chainsaw creates a dedicated namespace internally to run the test, and once the test completes, it automatically deletes all resources within that namespace, along with the namespace itself.

test/e2e/chainsaw/check-rgd-deletion/ignore-chainsaw-test.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: IgnoreTest
```
MariamFahmy98:
```suggestion
kind: Test
```

test/e2e/chainsaw/check-rgd-deletion/ignore-chainsaw-test.yaml
```diff
@@ -0,0 +1,44 @@
...
+          apiVersion: kro.run/v1alpha1
+          kind: ResourceGraphDefinition
+          name: simple-deployment-for-deletion-rgd
+          namespace: default
+      description: Delete RGD
+    catch:
+    - description: kro controller pod logs collector
+      podLogs:
+        selector: app.kubernetes.io/instance=kro
+        namespace: kro-system
```
MariamFahmy98: This step deletes the RGD. Now, we need to check if its CRD is deleted as well, so we need to add another step as follows:
```
  - name: check-that-crd-is-deleted
    try:
    - error:
        file: instancecrd-assert.yaml
```

test/e2e/chainsaw/check-instance-creation/chainsaw-test.yaml
```diff
@@ -0,0 +1,50 @@
...
+      description: Verify Deployment state
+    catch:
+    - description: kro controller pod logs collector
+      podLogs:
+        selector: app.kubernetes.io/instance=kro
+        namespace: kro-system
+    finally:
+    - description: sleep operation
+      sleep:
+        duration: 5s
```
MariamFahmy98: I don't think we need to add a sleep here. What do you think?

MariamFahmy98 (commented):
I added minor comments. Let me know your thoughts!

test/e2e/chainsaw/check-rgd-deletion/ignore-chainsaw-test.yaml
barney-s: what does this step do ? I was trying to see how we can detect absence

barney-s:
> I added minor comments. Let me know your thoughts!

@MariamFahmy98 - Addressed the review comments. PTAL.

MariamFahmy98 (approved):
LGTM.

a-hilaly (approved):
Really nice, thank you @barney-s and @MariamFahmy98 !

matthchr (approved):

PR #593, opened 2025-06-30
feat: share http client

fabianburth:
Currently, all clients create their own http client. 
Sharing a single http client allows these clients to share, among other things, a connection pool which poses a slight performance and memory optimisation.

### DIFF
--- a/pkg/client/set.go
+++ b/pkg/client/set.go
@@ -80,10 +80,6 @@ func NewSet(cfg Config) (*Set, error) {
 func (c *Set) init() error {
 	var err error
 
-	if c.config.UserAgent == "" {
-		c.config.UserAgent = rest.DefaultKubernetesUserAgent()
-	}
-
 	// share http client between all k8s clients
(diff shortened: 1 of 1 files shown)


matthchr (approved):
This LGTM

pkg/client/set.go
```diff
@@ -80,17 +80,27 @@ func NewSet(cfg Config) (*Set, error) {
 func (c *Set) init() error {
 	var err error
 
-	c.kubernetes, err = kubernetes.NewForConfig(c.config)
+	if c.config.UserAgent == "" {
+		c.config.UserAgent = rest.DefaultKubernetesUserAgent()
```
michaelhtm: User agent would not be empty. It would be set here before calling init()
https://github.com/kro-run/kro/blob/e6c7fd38cdfe6cf0e95e92bb095bb27d7d5a30a6/pkg/client/set.go#L70-L73

Issue #595, opened 2025-07-07
[helm chart] Arguments passed to Kro via the helm chart don't seem to take effect


shabbskagalwala:
### Description

**Observed Behavior**: Using the Helm chart provided by the Kro team - i observed a weird behavior when running Kro with replicas set to 2. 

I observed logs in the pods along the lines 

```
the object has been modified; please apply your changes to the latest version and try again
```

Which led me to investigate further and I noticed that both pods were running as active controllers trying to modify the same RGD that was being created/updated/deleted.

Further digging into the code what I noticed was the flag for leader-elect is a boolean. So either the flag is specified on the args (to enable leader election or not to disable leader election). Setting the flag `--leader-elect=false` has no effect.

I further tried 
```
          args:
            - --leader-elect
```

Which worked as expected but then the RBAC was missing permissions to create the leases (see relevant log lines)

```
E0707 07:05:38.499547       1 leaderelection.go:436] error retrieving resource lock kro-run/6f0f64a5.kro.run: leases.coordination.k8s.io "6f0f64a5.kro.run" is forbidden: User "system:serviceaccount:kro-run:kro-run" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "kro-run"
```

The same behavior applies to `--allow-crd-deletion` as well - the chart as default turns on CRD deletion (found it out the hard way)

**Expected Behavior**:
Leader election, and CRD deletion to work as expected. Kro should not delete the CRD when disabled via env vars or perform leader election when set to true via the configs.

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):

Set the replicas in the helm chart to 2 and turn on leader election (set var `config.enableLeaderElection: true`)

Set var `config.allowCRDDeletion: true`

An example from Kubebuilder on how boolean (leader-elect) should be set https://github.com/kubernetes-sigs/kubebuilder/blob/5837b2d00288fe7e13bda8330372cb8920c30134/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/values.yaml#L8

[Slack thread](https://kubernetes.slack.com/archives/C081TMY9D6Y/p1751869840155329)

**Versions**:
- kro version: 0.3.0
- Kubernetes Version (`kubectl version`): 1.31

**Involved Controllers**:
- Controller URLs and Versions (if applicable):

**Error Logs** (if applicable)**:
```
Paste any relevant logs here
```

[...]

PR #597, opened 2025-07-08
[helm] fix: conditionally provide bool args and corrected RBAC for aggregation mode

ryan-atkins:
For boolean cmd flags, they should only be set when the desired functionality is enabled, otherwise they should be omitted. 
The fix here conditionally sets the below two boolean flags based on the Values provided via helm.
- --allow-crd-deletion
- --leader-elect

**_Note:_** left the environment variable setup unchanged per conversation here where they were added. https://github.com/kro-run/kro/pull/275#discussion_r1951759915

With corrected enabling of leader-election, an RBAC issue arises when running with `mode: aggregation` which is also addressed here. Adds the rules from config [leader_election_role](https://github.com/kro-run/kro/blob/3b37c6b462940a979b2c221ca51f83e3fdf4763a/config/rbac/leader_election_role.yaml#L13) to the helm chart.

Address helm issue: #595

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -70,8 +70,9 @@ spec:
             - name: KRO_LEADER_ELECTION
               value: {{ .Values.config.enableLeaderElection | quote }}
           args:
+            {{- if .Values.config.allowCRDDeletion }}
             - --allow-crd-deletion
-            - "$(KRO_ALLOW_CRD_DELETION)"
+            {{- end }}
             - --metrics-bind-address
             - "$(KRO_METRICS_BIND_ADDRESS)"
             - --health-probe-bind-address
@@ -92,8 +93,9 @@ spec:
             - "$(KRO_CLIENT_QPS)"
             - --client-burst
             - "$(KRO_CLIENT_BURST)"
+            {{- if .Values.config.enableLeaderElection }}
             - --leader-elect
-            - "$(KRO_LEADER_ELECTION)"
+            {{- end }}
           livenessProbe:
             httpGet:
               path: /healthz
--- a/helm/templates/cluster-role.yaml
+++ b/helm/templates/cluster-role.yaml
@@ -9,7 +9,7 @@ metadata:
       This ClusterRole grants access for the kro controller.
 
       The kro installation comes with enough for the controller to run, and start watching the relevant resources
-      (e.g. ResourceGraphDefinitions), but in order to allow kro to create and manage other types of resoruces you
+      (e.g. ResourceGraphDefinitions), but in order to allow kro to create and manage other types of resources you
       must manually grant access to those separately.
 
       The easiest way to accomplish this is to create a new ClusterRole with the label `rbac.kro.run/aggregate-to-controller: "true"`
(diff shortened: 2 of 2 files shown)


n3wscott (approved):
makes sense to me!

shabbskagalwala:
Thank you @ryan-atkins this really helps! 

Kro by default runs in "unrestricted" mode but for using the "aggregation" mode - the clusterrole used by Kro will need to updated to add the following permissions

https://github.com/kubernetes-sigs/kubebuilder/blob/5837b2d00288fe7e13bda8330372cb8920c30134/docs/book/src/cronjob-tutorial/testdata/project/dist/chart/templates/rbac/leader_election_role.yaml#L10

Here: 
https://github.com/kro-run/kro/blob/3b37c6b462940a979b2c221ca51f83e3fdf4763a/helm/templates/cluster-role.yaml#L29

helm/templates/deployment.yaml
```diff
@@ -70,8 +70,9 @@ spec:
             - name: KRO_LEADER_ELECTION
               value: {{ .Values.config.enableLeaderElection | quote }}
           args:
+            {{- if .Values.config.allowCRDDeletion }}
             - --allow-crd-deletion
-            - "$(KRO_ALLOW_CRD_DELETION)"
```
a-hilaly, line 74: should we also remove https://github.com/kro-run/kro/blob/3b37c6b462940a979b2c221ca51f83e3fdf4763a/helm/templates/deployment.yaml#L48-L49

helm/templates/deployment.yaml
```diff
@@ -92,8 +93,9 @@ spec:
             - "$(KRO_CLIENT_QPS)"
             - --client-burst
             - "$(KRO_CLIENT_BURST)"
+            {{- if .Values.config.enableLeaderElection }}
             - --leader-elect
-            - "$(KRO_LEADER_ELECTION)"
```
a-hilaly, line 96: same for https://github.com/kro-run/kro/blob/3b37c6b462940a979b2c221ca51f83e3fdf4763a/helm/templates/deployment.yaml#L70-L71

helm/templates/deployment.yaml
ryan-atkins: chose to leave them unchanged per the previous conversation here https://github.com/kro-run/kro/pull/275#discussion_r1944041236 about that being a common pattern.

ryan-atkins: although now looking at it , flag.Parse() does not look like it supports environment variable reading so I'm not sure those env vars are ever actually used so maybe it was forward thinking to enable future support.. ?

a-hilaly: It is indeed a common pattern (allowing users to leverage `kubectl set env`, but since you're switching to conditionally setting the flags without args (just `--allow-crd-deletion` instead of `--allow-crd-deletion=true`), we can omit those env variables. Unless if you want to keep them and use them as arguments to the flags :)

a-hilaly: and yes flag.Parse doesn't lookup the environment variables, they are only evaluated when a pod is starting

PR #599, opened 2025-07-09
Add base64 encoder to cel expression evaluation

Damy3:
This allows to use base64.decode and base64.encode in CEL expressions, useful since reading secrets with externalRef doesn't decode their contents, e.g.:
```yaml
spec:
 resources:  
  - id: readSecret
    externalRef:
      apiVersion: v1
      kind: Secret
      metadata:
        name: mySecret
        namespace: myNs
  - id: res1
    template:
      apiVersion: v1
      kind: ConfigMap
      metadata:
        name: myConfigMap
      data:
        encoded: ${readSecret.data.something}
        decoded: ${string(base64.decode(readSecret.data.something))}
        somethingElseEncoded: ${base64.encode(b'Hello')}
```

### DIFF
--- a/pkg/cel/ast/inspector.go
+++ b/pkg/cel/ast/inspector.go
@@ -103,6 +103,8 @@ type Inspector struct {
 // knownFunctions contains the list of all CEL functions that are supported
 var knownFunctions = []string{
 	"random.seededString",
+	"base64.decode",
+	"base64.encode",
 }
 
 // DefaultInspector creates a new Inspector instance with the given resources and functions.
--- a/pkg/cel/environment.go
+++ b/pkg/cel/environment.go
@@ -54,6 +54,7 @@ func DefaultEnvironment(options ...EnvOption) (*cel.Env, error) {
 	declarations := []cel.EnvOption{
 		ext.Lists(),
 		ext.Strings(),
+		ext.Encoders(),
 		library.Random(),
 	}
 


dominik-lekse-cel (approved):

pkg/cel/ast/inspector.go
```diff
@@ -103,6 +103,8 @@ type Inspector struct {
 // knownFunctions contains the list of all CEL functions that are supported
 var knownFunctions = []string{
 	"random.seededString",
+	"base64.decode",
+	"base64.encode",
```
a-hilaly, line 107: Thanks a lot @Damy3 ! qq: do we need to manually hard code library functions as well? AFAIK this is not needed

a-hilaly:
cc @fabianburth

pkg/cel/ast/inspector.go
Damy3: When I tested without this I got an error for `base64` and `decode` being unrecognized, after I added them here it worked

a-hilaly (approved):
Thank you! @Damy3

farhaan-shamsee:
Hi,
How can I encode a value when creating a secret?

```yaml
- id: myClusterSecret
  template:
    apiVersion: v1
    kind: Secret
    metadata:
      name: my-secret
      namespace: default
    type: Opaque
    data:
      name: ${base64.encode(resource.metadata.name)}
      server: ${base64.encode(resource.status.endpoint)}
```

I tried above, but it does not work.

I also tried
```yaml
name: ${base64.encode(b'resource.metadata.name')}
```

But this just encodes the tcpEksCluster.metadata.name and not the actual value.

Damy3:
hi @farhaan-shamsee in my case it worked fine with `string(base64.decode(vpcOutputs.data.vpc_id))` for example, where vpcOutputs is an external reference to a `kind: Secret`. The only things that comes to mind is that your references to `resource` could be wrong, as far as I know you can only reference spec and status for created resources, and data for secrets and configmaps, in no case you can use metadata with kro syntax, but please prove me wrong

PR #601, opened 2025-07-10
Fix parsing of array and scalar types for expected type 'any'

Damy3:
This allows to pass arrays and integers when the expected type specified in the CRD evaluates to "any" (e.g.: using `x-kubernetes-preserve-unknown-fields`)

Also enriched error message in case of unrecognized type for scalar type input to improve debug experience.

Reproduction steps:

1. Have a CRD that accepts any type of input for a specific field (I was using tofu-controller [vars.*.value](https://github.com/flux-iac/tofu-controller/blob/main/charts/tofu-controller/crds/crds.yaml#L4898-L4906))
2. Create a KRO RGD that passes an array or a number to that input field
3. Get error `expected array type for path spec.vars[0].value, got [map...` when passing an array
4. Get error `unexpected type for path spec.vars[0].value: int64` when passing a number (also misleading, since the current error doesn't provide the expected type)

### DIFF
--- a/pkg/cel/ast/inspector.go
+++ b/pkg/cel/ast/inspector.go
@@ -519,3 +519,4 @@ func (a *Inspector) structExpressionToString(s *exprpb.Expr_CreateStruct) string
 func isInternalIdentifier(name string) bool {
 	return name == "@result" || strings.HasPrefix(name, "$$")
 }
+


pkg/cel/ast/inspector.go
```diff
@@ -519,4 +519,3 @@ func (a *Inspector) structExpressionToString(s *exprpb.Expr_CreateStruct) string
 func isInternalIdentifier(name string) bool {
 	return name == "@result" || strings.HasPrefix(name, "$$")
 }
-
```
a-hilaly: can you please omit this change from your commit? @Damy3

Damy3: I run make fmt and this was generated, previously I had this error: https://github.com/kro-run/kro/actions/runs/16188076975/job/45700333488#step:4:41 but sure, I'll remove it

PR #603, opened 2025-07-15
kro cli generate instance

DhairyaMajmudar:
# Description

Command to generate kro RGDs dummy instances

## Usage

 `kro generate instance -f [FILE]`

Flags:

| Shorthand | Flag | Usage | 
| ---- | ----- | --- |

## Demo

[Screencast from 2025-07-15 23-44-11.webm](https://github.com/user-attachments/assets/d582b582-f0f4-465c-92c3-947c615c127e)

### DIFF
--- a/cmd/kro/commands/generate.go
+++ b/cmd/kro/commands/generate.go
@@ -13,45 +13,45 @@ import (
 )
 
 type GenerateConfig struct {
-	resourceGroupDefinitionFile string
+	resourceGraphDefinitionFile string
 	outputFormat                string
 }
 
 var config = &GenerateConfig{}
 
 func init() {
-	generateCmd.PersistentFlags().StringVarP(&config.resourceGroupDefinitionFile, "file", "f", "",
-		"Path to the ResourceGroupDefinition file")
+	generateCmd.PersistentFlags().StringVarP(&config.resourceGraphDefinitionFile, "file", "f", "",
+		"Path to the ResourceGraphDefinition file")
 	generateCmd.PersistentFlags().StringVarP(&config.outputFormat, "format", "o", "yaml", "Output format (yaml|json)")
 }
 
 var generateCmd = &cobra.Command{
 	Use:   "generate",
-	Short: "Generate the CRD from a given ResourceGroupDefinition",
-	Long: "Generate the CRD from a given ResourceGroupDefinition." +
-		"This command generates a CustomResourceDefinition (CRD) based on the provided ResourceGroupDefinition file.",
+	Short: "Generate the CRD from a given ResourceGraphDefinition",
+	Long: "Generate the CRD from a given ResourceGraphDefinition." +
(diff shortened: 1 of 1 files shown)


cmd/kro/commands/generate.go
```diff
@@ -62,6 +62,35 @@ var generateCRDCmd = &cobra.Command{
 	},
 }
 
+var generateInstanceCmd = &cobra.Command{
+	Use:   "instance",
+	Short: "Generate a CustomResourceDefinition (CRD) instance",
+	Long: "Generate a CustomResourceDefinition (CRD) instance from a " +
```
a-hilaly: to avoid confusion with `kro generate crd` let's call say `Generate an RGD instance`?

cmd/kro/commands/generate.go
```diff
@@ -81,6 +110,28 @@ func generateCRD(rgd *v1alpha1.ResourceGraphDefinition) error {
...
 }
 
+func generateInstance(rgd *v1alpha1.ResourceGraphDefinition) error {
+	rgdGraph, err := createGraphBuilder(rgd)
+	if err != nil {
+		return fmt.Errorf("failed to create resource graph definition: %w", err)
+	}
+
+	emulatedObj := rgdGraph.Instance.GetEmulatedObject()
+	emulatedObj.SetAnnotations(map[string]string{"kro.run/version": "dev"})
```
a-hilaly, line 120: just thoughts: i like the `dev` placeholder for now, i'm thinking we'll probably need CLI versioning seperately, maybe not...

cmd/kro/commands/generate.go
```diff
@@ -81,6 +110,28 @@ func generateCRD(rgd *v1alpha1.ResourceGraphDefinition) error {
...
+
+	emulatedObj.Object["status"] = nil
+	delete(emulatedObj.Object, "status")
```
a-hilaly: would the `delete` is enough here?

cmd/kro/commands/generate.go
DhairyaMajmudar: True, let's do it this way

DhairyaMajmudar: This helped me to get to know about a typo which I made, instead of using `ResourceGraphDefinition` I was using `ResourceGroupDefinition` in the whole file 🫠 

I've fixed it now

cmd/kro/commands/generate.go
DhairyaMajmudar: How about doing versioning the same as of kro ?

cmd/kro/commands/generate.go
DhairyaMajmudar: Right, I've removed the redundant line

cmd/kro/commands/generate.go
a-hilaly: probably the best approach :)

PR #605, opened 2025-07-18
chore(docs): add optionals example documentation

jakobmoellerdev:
This adds a documentation to document Conditional CEL Expression / Optional support based on https://github.com/kro-run/kro/pull/525.

It adds a new example as well as a sufficiently annotated section in the concept section.

relates to documentation requested in https://github.com/kro-run/kro/pull/525#issuecomment-3086231530


jakobmoellerdev:
@a-hilaly PTAL

website/docs/examples/basic/optionals.md
```diff
@@ -0,0 +1,60 @@
...
+              labels:
+                app: deployment
+            spec:
+              containers:
+                - name: ${schema.spec.name}-busybox
+                  image: busybox
+                  command: ["sh", "-c", "echo $MY_VALUE && sleep 3600"]
+                  env:
+                    - name: MY_VALUE
+                      value: ${input.data.?ECHO_VALUE}
```
Ashish7812, line 59: @jakobmoellerdev and @fabianburth Lets say I have added a **includeWhen** clause for the "input" resource id. This way the resource (ConfigMap) will be included based on a user input condition.

Modified resource block

```yaml
id: input
includeWhen:
 - ${schema.spec.configmap.enabled}
externalRef / template:
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: demo
    namespace: default
```

what will be the behavior of the **${input.data.?ECHO_VALUE}** present in the Deployment resource? Or is this scenario supported?

Ashish7812: In the above scenario the user has sent the value for **schema.spec.configmap.enabled** as false.

fabianburth: Since only the part after the question mark (so here `ECHO_VALUE`) is optional, I believe this would not affect that scenario at all. So, the order KRO would process this would still be `configmap -> deployment`. 
In case of `externalRef`, KRO would wait until the "demo" configmap is created externally, in the case of template, I believe KRO would wait indefinitely.

Ashish7812: @fabianburth I simulated the case where template is used and includeWhen statement is present.
Here is the sample RGD

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: conditional
spec:
  schema:
    apiVersion: v1alpha1
    kind: Conditional
    spec:
      name: string
      enabled: string

  # Define the resources this API will manage.
  resources:
    - id: pod-a
      includeWhen:
        - ${schema.spec.enabled == "true"}
      template:
        apiVersion: v1
        kind: Pod
        metadata:
          name: ${schema.spec.name}-pod
        spec:
          containers:
          - name: nginx-container
            image: nginx:latest
            ports:
            - containerPort: 80
            
    - id: pod-b
      template:
        apiVersion: v1
        kind: Pod
        metadata:
          name: ${schema.spec.name}-pod2
        spec:
          containers:
          - name: nginx-container
            image: nginx:latest
            ports:
            - containerPort: 80


    - id: config
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.spec.name}-cm # Use the name provided by user
        data:
          firstresourcename: ${string(pod-a.metadata.name)}
          secondresourcename: ${string(pod-b.metadata.name)}
```

When I created the instance of the RGD by passing the enabled to false the "pod-a" is not created and in the configmap also I did not see the key **firstresourcename**. This indicates that if you are dependent on a conditional resource and that resource is not part of the runtime rgd then corresponding keys are also not used. In the configmap only the second key was present i.e **secondresourcename**.

After the optionals support, I was thinking if the resource is not there then ut will fallback to the optional value.
so in the config map I will have both values for **firstresourcename (with optional value)** and **secondresourcename (with actual value sourced from pod-b)**.

This way I will have somevalue (either optional or actual) for my key. is my understanding correct related to the usage of optionals?

Ashish7812: On this comment "In case of externalRef, KRO would wait until the "demo" configmap is created externally"

Right now KRO does not wait for the external resource to be available rather it fails immediately. I have created a bug for this https://github.com/kro-run/kro/issues/598

fabianburth: I think you have a misunderstanding about what optionals are / how optionals work.
CEL optionals are not like a default value. 
So, in your example with **${input.data.?ECHO_VALUE}**, the logic would not be:

[pseudocode]
```
if input.data != nil {
   return input.data
} else {
   return ECHO_VALUE
}
```

Here is a code snippet from the `cel-go` libraries optional implementation:
```
type Optional struct {
	value ref.Val
}

func (o *Optional) Value() any {
	if o.value == nil {
		return nil
	}
	return o.value.Value()
}
```

So, what optional actually does is make a CEL expression successfully evaluate even though the variable after the question mark does not even exist.

Here is a more detailed explanation of what effects optionals have in KRO: https://github.com/kro-run/kro/pull/525#issuecomment-2857330868

jakobmoellerdev: Resolving this conversation as off-topic for now as we don't really talk about the PR here anymore

a-hilaly (approved):
👍

website/docs/docs/concepts/00-resource-group-definitions.md
```diff
@@ -188,13 +188,15 @@ Users can specify more controls in resources in `.spec.resources[]`
 ```yaml
 spec:
   resources:
-    - id:
-      template || externalRef:
-      readyWhen:
-      includeWhen:
+    - id: {}
```
matthchr: Can we fix this even further to maybe use a real string for id rather than just `{}`?

website/docs/docs/concepts/00-resource-group-definitions.md
```diff
@@ -188,13 +188,15 @@ Users can specify more controls in resources in `.spec.resources[]`
...
+      template || externalRef: {}
+      readyWhen: {}
```
matthchr: Do we want a more realistic example here, or at the very least "instructions"?

```suggestion
      readyWhen:
        - "CEL ready expression here"        
```

And maybe similar for the others?

jakobmoellerdev: added some better examples!

PR #607, opened 2025-07-18
chore: improve optional type handling and error reporting in GoNativeType

fabianburth:
**Why we need this**
The changes to `GoNativeType()` in https://github.com/kro-run/kro/pull/525 lead to optional values not underlying the same type restrictions as non-optional values. For an optional, a value of type `map[int]interface{}` would be returned, while for a non-optional, it would return an error (see added test cases).
This PR aligns the behaviour of `GoNativeType()` for optional values with non-optional values.

> @a-hilaly is there a particular reason to restrict the allowed types to this set? Otherwise, we might want to think about completely removing `GoNativeTypes()` and replace its calls with `v.Value()` (essentially, only replacing https://github.com/kro-run/kro/blob/6255087b2c49ac2f1542d0366e39d13d9acf054a/pkg/runtime/runtime.go#L586 with `return val.Value(), nil`).

`NullType` values are significantly different (see: https://github.com/kro-run/kro/pull/525#discussion_r2072925272)

### DIFF
--- a/pkg/cel/conversions.go
+++ b/pkg/cel/conversions.go
@@ -17,7 +17,6 @@ package cel
 import (
 	"errors"
 	"fmt"
-	"reflect"
 
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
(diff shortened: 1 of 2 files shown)


pkg/cel/conversions.go
```diff
@@ -30,28 +29,11 @@ var (
...
-		return v.ConvertToNative(reflect.TypeOf([]interface{}{}))
-	case types.MapType:
-		return v.ConvertToNative(reflect.TypeOf(map[string]interface{}{}))
-	case types.OptionalType:
-		return v.Value(), nil
-	case types.NullType:
-		return nil, nil
+	switch value := v.Value().(type) {
+	case bool, int64, uint64, float64, string, []interface{}, map[string]interface{}, nil:
+		return value, nil
```
a-hilaly: TIL `value := v.Value().(type) `, thank you!

a-hilaly:
>is there a particular reason to restrict the allowed types to this set?

@fabianburth no particular reason, when i originally wrote that part i didn't know about optional types and their type semantics.

a-hilaly (approved):
Dankeshoen!

a-hilaly:
Looks like intergrations tests aren't happy @fabianburth

jakobmoellerdev:
@fabianburth I tested your branch and noticed that the regression disappears if I do a more conservative change:

```
case types.OptionalType:
		opt := v.(*types.Optional)
		if !opt.HasValue() {
			return nil, nil
		}
		return GoNativeType(opt.GetValue())
```

with this the tests still run but the e2e test also succeeds. I believe we have some hidden behavior change in here (maybe related to the list / map type parsing that used ConvertToNative before). We can check next week!

fabianburth:
The Null values are the issue, actually. And I hate to admit it, but looking back at the optionals PR, someone predicted this without running into it first https://github.com/kro-run/kro/pull/525#discussion_r2072925272. 🙃

Thought this would be a small and clean solution, but obviously didn't have a thorough enough understanding yet. I'll adjust it to what @jakobmoellerdev proposed (which is what we started with anyway, I think). 

Still, appreciate the quick reviews!

PR #203, opened 2025-01-15
wip: dynamic controller enhancements

a-hilaly:
[Not ready for review]

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #611, opened 2025-07-23
feat: child object watch and resource reconciliation based on informers

jakobmoellerdev:
Watch child objects and trigger instance reconciliation on change.

Alternative implementation proposal draft as discussed in KRO community call (23. Jul)

Compares to https://github.com/kro-run/kro/pull/576 by:

- replacing watchset with dynamic registrations on the shared informers used by KRO
- ~using the native factory (as of now) for handler registration tracking~ the shared informer factory does not allow for efficient dynamic shutdown and restart of informers
- reusing our existing informer setup as much as possible
- taking over the existing approach for tracking gvrs for the graphs resource list
- adding a feature flag based on a label in the RGD ( we might want to enable by default, open for opinions, community call feedback was mixed on this )
- testing that this flag actually works

based on reviews:

- metadata informer based dynamic informers now instead of unstructured

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -116,8 +116,7 @@ func main() {
 
 	// reconciler parameters
 	flag.IntVar(&resyncPeriod, "dynamic-controller-default-resync-period", 36000,
-		"interval at which the controller will re list resources even with no changes, in seconds. "+
-			"By default resync is disabled.")
+		"interval at which the controller will re list resources even with no changes, in seconds.")
 	flag.IntVar(&queueMaxRetries, "dynamic-controller-default-queue-max-retries", 20,
 		"maximum number of retries for an item in the queue will be retried before being dropped")
 	// log level flags
--- a/pkg/dynamiccontroller/dynamic_controller.go
+++ b/pkg/dynamiccontroller/dynamic_controller.go
@@ -59,6 +59,7 @@ package dynamiccontroller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -79,6 +80,15 @@ import (
 	"github.com/kubernetes-sigs/kro/pkg/dynamiccontroller/internal"
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
 	"github.com/kubernetes-sigs/kro/pkg/requeue"
+)
+
+const (
+	// eventTypeAdd is emitted to the queue when a new object was created
+	eventTypeAdd = "add"
+	// eventTypeUpdate is emitted to the queue when an existing object was updated
+	eventTypeUpdate = "update"
+	// eventTypeDelete is emitted to the queue when an existing object was deleted
+	eventTypeDelete = "delete"
 )
 
 // Config holds the configuration for DynamicController
@@ -106,6 +116,8 @@ type Config struct {
 	QueueShutdownTimeout time.Duration
 }
 
+// Handler is used to actually perform the reconciliation logic for an instance GVR and will operate
+// on a single instance of the resource received from the queue
 type Handler func(ctx context.Context, req ctrl.Request) error
 
 // ObjectIdentifiers holds the key and GVR of the object to reconcile.
@@ -146,30 +158,47 @@ type registration struct {
 //     (handler map). This allows concurrent updates to
 //     different GVRs without blocking each other.
 type DynamicController struct {
-	config Config
-	log    logr.Logger
+	// Parent run context, inherited by all informer stop contexts.
+	// It is set by Start and meant to be used to register the controller context with a global handler
+	// such as the controller-runtime manager.
+	ctx context.Context
 
-	client k8smetadata.Interface
-	mapper meta.RESTMapper
-
-	// Map of active informers per GVR.
-	// Guarded by mu.
-	watches map[schema.GroupVersionResource]*internal.LazyInformer
-	// Map of parent registrations per GVR.
-	// Guarded by mu.
-	registrations map[schema.GroupVersionResource]*registration
-	// Global mutex protecting watches and registrations.
+	// mu is a global mutex protecting watches and registrations.
 	// Required because StartServingGVK and StopServiceGVK may run concurrently,
 	// and because Run or gracefulShutdown may also traverse these maps.
 	mu sync.Mutex
 
-	// Latest handler for each parent GVR used by syncFunc.
-	handlers sync.Map // map[schema.GroupVersionResource]Handler (thread-safe on its own)
-	queue    workqueue.TypedRateLimitingInterface[ObjectIdentifiers]
+	// config is the controller configuration used to steer detailed enqueue behavior when events
+	// are received by the dynamic controllers watches and need to be propagated to the handlers.
+	config Config
 
-	// Parent run context, inherited by all informer stop contexts.
-	// set by Run
-	ctx context.Context
+	log logr.Logger
+
+	// client is the Kubernetes client used to create informers and list resources.
+	// DynamicController only operates on metadata information, because all actual reconciliations
+	// happen in the instance handlers that are managed with Register and Deregister.
+	client k8smetadata.Interface
+	// mapper is used to translate from GVRs (REST Resources that can be watched) to GVKs to properly
+	// propagate instance events to the handlers.
+	mapper meta.RESTMapper
+
+	// Map of active informers per GVR.
+	// No matter if an instance resource or a child resource is being watched,
+	// eventually, they will be handled by an internal.LazyInformer registered here.
+	// Guarded by mu.
+	watches map[schema.GroupVersionResource]*internal.LazyInformer
+	// Map of parent registrations per GVR.
+	// A registration is created for each parent GVR that is being watched.
+	// Each registration contains a list of child GVRs that are being watched for the parent.
+	// Any event that is received for a child GVR will be propagated through a labelled reference to the parent.
+	// Guarded by mu.
+	registrations map[schema.GroupVersionResource]*registration
+
+	// handlers is a Handler collection for each parent GVR, invoked for queued objects.
+	handlers sync.Map // map[schema.GroupVersionResource]Handler (thread-safe on its own)
+	// queue is the work queue used to process items received via watches.
+	// The queue is shared between all informers and is used to propagate events to the handlers.
+	queue workqueue.TypedRateLimitingInterface[ObjectIdentifiers]
 }
 
 // NewDynamicController creates a new DynamicController.
@@ -246,7 +275,7 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
 		return true
 	}
 
-	gvrKey := item.GVR.String()
+	gvrKey := keyFromGVR(item.GVR)
 
 	switch typedErr := err.(type) {
 	case *requeue.NoRequeue:
@@ -276,7 +305,7 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
 }
 
 func (dc *DynamicController) syncFunc(ctx context.Context, oi ObjectIdentifiers, handler Handler) error {
-	gvrKey := oi.GVR.String()
+	gvrKey := keyFromGVR(oi.GVR)
 	dc.log.V(1).Info("Syncing object", "gvr", gvrKey, "key", oi.NamespacedName)
 
 	startTime := time.Now()
@@ -313,14 +342,14 @@ func (dc *DynamicController) enqueueParent(parentGVR schema.GroupVersionResource
 }
 
 func (dc *DynamicController) updateFunc(parentGVR schema.GroupVersionResource, oldObj, newObj interface{}) {
-	newMeta, ok := newObj.(*metav1.PartialObjectMetadata)
-	if !ok {
-		dc.log.Error(nil, "failed to cast new object to PartialObjectMetadata")
+	newMeta, err := meta.Accessor(newObj)
+	if err != nil {
+		dc.log.Error(err, "failed to access new object meta")
 		return
 	}
-	oldMeta, ok := oldObj.(*metav1.PartialObjectMetadata)
-	if !ok {
-		dc.log.Error(nil, "failed to cast old object to PartialObjectMetadata")
+	oldMeta, err := meta.Accessor(oldObj)
+	if err != nil {
+		dc.log.Error(err, "failed to access old object meta")
 		return
 	}
 	if newMeta.GetGeneration() == oldMeta.GetGeneration() {
@@ -328,7 +357,7 @@ func (dc *DynamicController) updateFunc(parentGVR schema.GroupVersionResource, o
 			"name", newMeta.GetName(), "namespace", newMeta.GetNamespace(), "generation", newMeta.GetGeneration())
 		return
 	}
-	dc.enqueueParent(parentGVR, newObj, "update")
+	dc.enqueueParent(parentGVR, newObj, eventTypeUpdate)
 }
 
 // Register registers parent and children via reconciliation.
@@ -366,7 +395,7 @@ func (dc *DynamicController) Register(
 		}
 	}
 
-	dc.log.V(1).Info("Successfully registered GVR", "gvr", parent)
+	dc.log.V(1).Info("Successfully registered GVR", "gvr", keyFromGVR(parent))
 	return nil
 }
 
@@ -380,16 +409,18 @@ func (dc *DynamicController) Deregister(_ context.Context, parent schema.GroupVe
 		return nil
 	}
 
+	gvrKey := keyFromGVR(parent)
+
 	if err := dc.reconcileChildrenLocked(parent, nil, reg); err != nil {
-		dc.log.Error(err, "failed to detach children", "parent", parent)
+		dc.log.Error(err, "failed to detach children", "parent", gvrKey)
 	}
 	if err := dc.reconcileParentLocked(parent, nil, reg); err != nil {
-		dc.log.Error(err, "failed to detach parent", "parent", parent)
+		dc.log.Error(err, "failed to detach parent", "parent", gvrKey)
 	}
 
 	delete(dc.registrations, parent)
 
-	dc.log.V(1).Info("Successfully unregistered GVR", "gvr", parent)
+	dc.log.V(1).Info("Successfully unregistered GVR", "gvr", gvrKey)
 	return nil
 }
 
@@ -417,22 +448,20 @@ func (dc *DynamicController) reconcileParentLocked(
 	reg *registration,
 ) error {
 	if instanceHandler == nil {
-		// remove parent handler if present
-		if reg.parentHandlerID != "" {
-			if w, ok := dc.watches[parent]; ok {
-				stopped, err := w.RemoveHandler(reg.parentHandlerID)
-				if err != nil {
-					return fmt.Errorf("remove parent handler %s: %w", parent, err)
-				}
-				if stopped {
-					delete(dc.watches, parent)
-				}
-			}
-			reg.parentHandlerID = ""
-			dc.handlers.Delete(parent)
-			gvrCount.Dec()
-			dc.log.V(1).Info("Detached parent", "gvr", parent)
+		if reg.parentHandlerID == "" {
+			return nil
 		}
+		// remove parent handler if present
+		if err := dc.removeHandlerLocked(parent, reg.parentHandlerID); err != nil {
+			return fmt.Errorf("removing parent handler %s: %w", parent, err)
+		}
+		reg.parentHandlerID = ""
+		dc.handlers.Delete(parent)
+
+		gvrCount.Dec()
+		handlerDetachTotal.WithLabelValues("parent").Inc()
+		handlerCount.WithLabelValues("parent").Dec()
+		dc.log.V(1).Info("Detached parent", "gvr", parent)
 		return nil
 	}
 
@@ -441,16 +470,19 @@ func (dc *DynamicController) reconcileParentLocked(
 
 	// create handler if missing
 	if reg.parentHandlerID == "" {
-		handlerID := "parent:" + parent.String()
-		if err := w.AddHandler(dc.ctx, handlerID, cache.ResourceEventHandlerFuncs{
-			AddFunc:    func(obj interface{}) { dc.enqueueParent(parent, obj, "add") },
+		parentHandlerID := parentHandlerID(parent)
+		if err := w.AddHandler(dc.ctx, parentHandlerID, cache.ResourceEventHandlerFuncs{
+			AddFunc:    func(obj interface{}) { dc.enqueueParent(parent, obj, eventTypeAdd) },
 			UpdateFunc: func(oldObj, newObj interface{}) { dc.updateFunc(parent, oldObj, newObj) },
-			DeleteFunc: func(obj interface{}) { dc.enqueueParent(parent, obj, "delete") },
+			DeleteFunc: func(obj interface{}) { dc.enqueueParent(parent, obj, eventTypeDelete) },
 		}); err != nil {
 			return fmt.Errorf("add parent handler %s: %w", parent, err)
 		}
-		reg.parentHandlerID = handlerID
+		reg.parentHandlerID = parentHandlerID
+
 		gvrCount.Inc()
+		handlerAttachTotal.WithLabelValues("parent").Inc()
+		handlerCount.WithLabelValues("parent").Inc()
 		dc.log.V(1).Info("Attached parent", "gvr", parent)
 	}
 
@@ -472,21 +504,22 @@ func (dc *DynamicController) reconcileChildrenLocked(
 		desiredSet[g] = struct{}{}
 	}
 
+	parentGVRKey := keyFromGVR(parent)
+
 	// remove obsolete
-	for child, id := range reg.childHandlerIDs {
-		if _, keep := desiredSet[child]; !keep {
-			if w, ok := dc.watches[child]; ok {
-				stopped, err := w.RemoveHandler(id)
-				if err != nil {
-					return fmt.Errorf("remove child handler %s: %w", child, err)
-				}
-				if stopped {
-					delete(dc.watches, child)
-				}
-			}
-			delete(reg.childHandlerIDs, child)
-			dc.log.V(1).Info("Detached child", "parent", parent, "gvr", child)
+	for child, childHandlerID := range reg.childHandlerIDs {
+		if _, keep := desiredSet[child]; keep {
+			continue
 		}
+		if err := dc.removeHandlerLocked(child, childHandlerID); err != nil {
+			return fmt.Errorf("removing child handler %s: %w", child, err)
+		}
+		delete(reg.childHandlerIDs, child)
+
+		childGVRKey := keyFromGVR(child)
+		handlerDetachTotal.WithLabelValues("child").Inc()
+		handlerCount.WithLabelValues("child").Dec()
+		dc.log.V(1).Info("Detached child", "parent", parentGVRKey, "gvr", childGVRKey)
 	}
 
 	// add missing
@@ -495,15 +528,88 @@ func (dc *DynamicController) reconcileChildrenLocked(
 			continue
 		}
 		w := dc.ensureWatchLocked(child)
-		handlerID := "child:" + parent.String() + "->" + child.String()
-		if err := w.AddHandler(dc.ctx, handlerID, dc.handlerForChildGVR(parent, child)); err != nil {
+		childHandlerID := childHandlerID(parent, child)
+		if err := w.AddHandler(dc.ctx, childHandlerID, dc.handlerForChildGVR(parent, child)); err != nil {
 			return fmt.Errorf("add child handler %s: %w", child, err)
 		}
-		reg.childHandlerIDs[child] = handlerID
-		dc.log.V(1).Info("Attached child", "parent", parent, "gvr", child)
+		reg.childHandlerIDs[child] = childHandlerID
+
+		childGVRKey := keyFromGVR(child)
+		handlerAttachTotal.WithLabelValues("child").Inc()
+		handlerCount.WithLabelValues("child").Inc()
+		dc.log.V(1).Info("Attached child", "parent", parentGVRKey, "gvr", childGVRKey)
 	}
 
 	return nil
+}
+
+// removeHandlerLocked removes a handler with the specified ID for a given GroupVersionResource watch.
+// If the handler removal stops the watch, the corresponding GroupVersionResource entry is deleted from the watches map.
+// Returns an error if the handler removal fails.
+func (dc *DynamicController) removeHandlerLocked(gvr schema.GroupVersionResource, handlerID string) error {
+	w, ok := dc.watches[gvr]
+	if !ok {
+		// no watch means there cannot be a handler
+		return nil
+	}
+	stopped, err := w.RemoveHandler(handlerID)
+	if err != nil {
+		return fmt.Errorf("removing handler %s for watch on %s failed: %w", handlerID, gvr, err)
+	}
+	if stopped {
+		delete(dc.watches, gvr)
+	}
+	return nil
+}
+
+func (dc *DynamicController) handlerForChildGVR(parent, child schema.GroupVersionResource) cache.ResourceEventHandler {
+	parentGVRKey, childGVRKey := keyFromGVR(parent), keyFromGVR(child)
+	handle := func(obj interface{}, eventType string) {
+		objMeta, err := meta.Accessor(obj)
+		if err != nil {
+			dc.log.Error(err, "failed to get metadata accessor for object", "eventType", eventType)
+			return
+		}
+		lbls := objMeta.GetLabels()
+		owned, ok := lbls[metadata.OwnedLabel]
+		if !ok || owned != "true" {
+			return
+		}
+		name, ok := lbls[metadata.InstanceLabel]
+		if !ok {
+			return
+		}
+		namespace, ok := lbls[metadata.InstanceNamespaceLabel]
+		if !ok {
+			return
+		}
+
+		parentGVK, err := dc.mapper.KindFor(parent)
+		if err != nil {
+			dc.log.Error(err, "failed to get parent GVK", "parent", parentGVRKey)
+			return
+		}
+		pom := &metav1.PartialObjectMetadata{}
+		pom.SetGroupVersionKind(parentGVK)
+		pom.SetName(name)
+		pom.SetNamespace(namespace)
+
+		dc.log.V(1).Info("Child triggered parent reconciliation",
+			"parent", parentGVRKey,
+			"child", childGVRKey,
+			"eventType", eventType,
+			"childName", objMeta.GetName(),
+			"childNamespace", objMeta.GetNamespace(),
+			"targetName", name,
+			"targetNamespace", namespace,
+		)
+		dc.enqueueParent(parent, pom, eventType)
+	}
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { handle(obj, eventTypeAdd) },
+		UpdateFunc: func(oldObj, newObj interface{}) { handle(newObj, eventTypeUpdate) },
+		DeleteFunc: func(obj interface{}) { handle(obj, eventTypeDelete) },
+	}
 }
 
 func (dc *DynamicController) gracefulShutdown() error {
@@ -511,7 +617,7 @@ func (dc *DynamicController) gracefulShutdown() error {
 
 	dc.mu.Lock()
 	for gvr, w := range dc.watches {
-		dc.log.V(1).Info("Stopping watch", "gvr", gvr)
+		dc.log.V(1).Info("Stopping watch", "gvr", keyFromGVR(gvr))
 		w.Shutdown()
 	}
 	dc.mu.Unlock()
@@ -539,51 +645,31 @@ func (dc *DynamicController) gracefulShutdown() error {
 	}
 }
 
-func (dc *DynamicController) handlerForChildGVR(parent, child schema.GroupVersionResource) cache.ResourceEventHandler {
-	handle := func(obj interface{}, eventType string) {
-		objMeta, err := meta.Accessor(obj)
-		if err != nil {
-			dc.log.Error(err, "failed to get metadata accessor for object", "eventType", eventType)
-			return
-		}
-		lbls := objMeta.GetLabels()
-		owned, ok := lbls[metadata.OwnedLabel]
-		if !ok || owned != "true" {
-			return
-		}
-		name, ok := lbls[metadata.InstanceLabel]
-		if !ok {
-			return
-		}
-		namespace, ok := lbls[metadata.InstanceNamespaceLabel]
-		if !ok {
-			return
-		}
-
-		parentGVK, err := dc.mapper.KindFor(parent)
-		if err != nil {
-			dc.log.Error(err, "failed to get parent GVK", "parent", parent)
-			return
-		}
-		pom := &metav1.PartialObjectMetadata{}
-		pom.SetGroupVersionKind(parentGVK)
-		pom.SetName(name)
-		pom.SetNamespace(namespace)
-
-		dc.log.V(1).Info("Child triggered parent reconciliation",
-			"parent", parent.String(),
-			"child", child.String(),
-			"eventType", eventType,
-			"childName", objMeta.GetName(),
-			"childNamespace", objMeta.GetNamespace(),
-			"targetName", name,
-			"targetNamespace", namespace,
-		)
-		dc.enqueueParent(parent, pom, eventType)
+// keyFromGVR returns a compact, allocation-efficient string key for the given
+// GroupVersionResource in the canonical "group/version/resource" format.
+// Unlike [schema.GroupVersionResource.String], it omits labels and avoids extra
+// allocations, making it suitable for use as map keys, metrics labels, and cache identifiers.
+func keyFromGVR(gvr schema.GroupVersionResource) string {
+	var b strings.Builder
+	if gvr.Group != "" {
+		b.WriteString(gvr.Group)
 	}
-	return cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { handle(obj, "add") },
-		UpdateFunc: func(oldObj, newObj interface{}) { handle(newObj, "update") },
-		DeleteFunc: func(obj interface{}) { handle(obj, "delete") },
+	if gvr.Version != "" {
+		b.WriteRune('/')
+		b.WriteString(gvr.Version)
 	}
+	if gvr.Resource != "" {
+		b.WriteRune('/')
+		b.WriteString(gvr.Resource)
+	}
+	return b.String()
+}
+
+func childHandlerID(parent schema.GroupVersionResource, child schema.GroupVersionResource) string {
+	return "child:" + keyFromGVR(parent) + "->" + keyFromGVR(child)
+}
+
+func parentHandlerID(parent schema.GroupVersionResource) string {
+	handlerID := "parent:" + keyFromGVR(parent)
+	return handlerID
 }
--- a/pkg/dynamiccontroller/metrics.go
+++ b/pkg/dynamiccontroller/metrics.go
@@ -28,6 +28,9 @@ func init() {
 		reconcileDuration,
 		gvrCount,
 		queueLength,
+		handlerCount,
+		handlerAttachTotal,
+		handlerDetachTotal,
 		handlerErrorsTotal,
 		informerSyncDuration,
 		informerEventsTotal,
@@ -63,7 +66,7 @@ var (
 	gvrCount = prometheus.NewGauge(
 		prometheus.GaugeOpts{
 			Name: "dynamic_controller_gvr_count",
-			Help: "Number of GVRs currently managed by the controller",
+			Help: "Number of Instance GVRs currently managed by the controller",
 		},
 	)
 	queueLength = prometheus.NewGauge(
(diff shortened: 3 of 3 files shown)


test/integration/suites/core/crd_test.go
jakobmoellerdev, line 1: Note to Reviewer: I noticed that all of our suites lack proper context propagation so I decided to add it for one I was debugging while coding. We might want to split this off

test/integration/environment/setup.go
```diff
@@ -133,8 +143,14 @@ func (e *Environment) initializeClients() error {
 }
 
 func (e *Environment) setupController() error {
+	var logger logr.Logger
```
jakobmoellerdev: Note to Reviewer: I noticed we discard all controller logs in tests right now, currently I changed one occurrence to use the GinkgoLogger which defers logs until a failure occurs

pkg/metadata/labels.go
```diff
@@ -45,6 +45,8 @@ const (
 	ResourceGraphDefinitionNameLabel      = LabelKROPrefix + "resource-graph-definition-name"
 	ResourceGraphDefinitionNamespaceLabel = LabelKROPrefix + "resource-graph-definition-namespace"
 	ResourceGraphDefinitionVersionLabel   = LabelKROPrefix + "resource-graph-definition-version"
+
+	InstanceWatchResourcesLabel = LabelKROPrefix + "instance-watch-resources"
```
jakobmoellerdev: Note to reviewer: This is basically a feature flag by RGD. based on discussion results we might want to enable this by default but since its so impactful I figured we gate it

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -422,111 +430,195 @@ func (dc *DynamicController) enqueueObject(obj interface{}, eventType string) {
...
 	}
-	dc.handlers.Store(gvr, handler)
 
-	informerContext := context.Background()
-	cancelableContext, cancel := context.WithCancel(informerContext)
-	// Start the informer
-	go func() {
-		dc.log.V(1).Info("Starting informer", "gvr", gvr)
-		// time.Sleep(5 * time.Millisecond)
-		informer.Run(cancelableContext.Done())
```
jakobmoellerdev, line 507: Note to Reviewer: This was always buggy, and switching the informer Run to the factory run caused a regression. As it turns out we tried waiting for this informer before in `gracefulShutdown` (https://github.com/kro-run/kro/pull/611/files#diff-ca0d44b610e2a74f9fafd4ea251defaa98c3099a8cb93467efaf88f923ed184eL336) but since we didnt do it through the factory this actually was a regression

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -422,111 +430,195 @@ func (dc *DynamicController) enqueueObject(obj interface{}, eventType string) {
...
 	}
 	if err := informer.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
 		dc.log.Error(err, "Watch error", "gvr", gvr)
 	}); err != nil {
 		dc.log.Error(err, "Failed to set watch error handler", "gvr", gvr)
-		return fmt.Errorf("failed to set watch error handler for GVR %s: %w", gvr, err)
 	}
-	dc.handlers.Store(gvr, handler)
 
-	informerContext := context.Background()
```
jakobmoellerdev, line 501: Using context.Background here hid a bug in which our informers were not properly stopped. Since the context we pass here is actually the Manager start context it is safe to use the parent context here

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -329,16 +332,21 @@ func (dc *DynamicController) gracefulShutdown(timeout time.Duration) error {
 	defer cancel()
 
 	var wg sync.WaitGroup
-	dc.informers.Range(func(key, value interface{}) bool {
+	dc.handlers.Range(func(key, value interface{}) bool {
```
jakobmoellerdev: Iterating through the handlers is slightly safer than iterating through the factories because `StopServiceGVK` is actually doing more than just shutting down the informers. This helps bring the controller back to a "zero" state as if it was never started

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -422,111 +430,195 @@ func (dc *DynamicController) enqueueObject(obj interface{}, eventType string) {
...
+		// update the instanceHandler, as it might have changed.
+		dc.handlers.Store(gvr, instanceHandler)
+		// trigger reconciliation of the corresponding resourceGVR's
+		objs, err := dc.kubeClient.Resource(gvr).Namespace("").List(fw.ctx, metav1.ListOptions{})
 		if err != nil {
 			return fmt.Errorf("failed to list objects for GVR %s: %w", gvr, err)
 		}
 		for _, obj := range objs.Items {
-			dc.enqueueObject(&obj, "update")
+			dc.EnqueueObject(&obj, "update")
```
jakobmoellerdev: here we use EnqueueObject so we need to export the method.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -86,6 +97,61 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(ctx
...
+	update := func(obj interface{}, eventType string) {
+		gvr := processedRGD.Instance.GetGroupVersionResource()
+		obju, ok := obj.(*unstructured.Unstructured)
+		if !ok {
+			return
+		}
+
+		// if there is no kro instance label set, we assume the object is irrelevant for the instance.
+		labels := obju.GetLabels()
+		owned, ok := labels[metadata.OwnedLabel]
```
jakobmoellerdev: Note to Reviewer: We might want to change this owns check from a client side to a server side check, by applying it to the list filter when calling a watch while building the informer. See https://github.com/kro-run/kro/pull/611/files#diff-ca0d44b610e2a74f9fafd4ea251defaa98c3099a8cb93467efaf88f923ed184eR470

test/integration/suites/core/setup_test.go
```diff
@@ -33,12 +32,13 @@ func TestCore(t *testing.T) {
 	RegisterFailHandler(Fail)
 	BeforeSuite(func() {
 		var err error
-		env, err = environment.New(
+		env, err = environment.NewWithContext(t.Context(),
```
jakobmoellerdev: passing t.Context() is safe because that means the env will definetly be cancelled if the test stops!

fabianburth:
@a-hilaly, we found out why the tests were failing and fixed that up!

a-hilaly:
solves https://github.com/kro-run/kro/issues/323

pkg/metadata/labels.go
a-hilaly: Can we document this in the variables? (it's on me for not the documenting the ones above). side note how about: 
```
kro.run/watch-mode: "true"`
```

a-hilaly: since the controller doesn't reconcile on metadata changes (labels/annotations) due to [predicate usage](https://github.com/kubernetes-sigs/kro/blob/main/pkg/controller/resourcegraphdefinition/controller.go#L102), we might want to consider making this configurable at the spec level rather than just a label-based feature flag.

I was thinking about this in relation to #203 where users could configure the reconcile resync time. Zooming out, it might make sense to have a more comprehensive reconciliation configuration:
```yaml
spec:
  reconcile:
    resyncPeriod: 5m
    mode: "periodic" | "watch" # or maybe "reactive"?
```

pkg/metadata/labels.go
```diff
@@ -63,6 +65,11 @@ func SetKROUnowned(meta metav1.ObjectMeta) {
 	setLabel(&meta, OwnedLabel, stringFromBoolean(false))
 }
 
+func GetInstanceWatchResources(meta metav1.ObjectMeta) bool {
```
a-hilaly: nit: godoc string

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -86,6 +97,61 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(ctx
 	return processedRGD.TopologicalOrder, resourcesInfo, nil
 }
 
+func (r *ResourceGraphDefinitionReconciler) getResourceGraphResourceHandlers(ctx context.Context, processedRGD *graph.Graph) map[schema.GroupVersionResource]cache.ResourceEventHandlerFuncs {
```
a-hilaly: does this function need to live here? this strikes me as something we can abstract in dynamic controller and only pass the list of GVKs from the RGD controller package.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -86,6 +97,61 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(ctx
...
+			return
+		}
+		name, ok := labels[metadata.InstanceLabel]
+		if !ok {
+			return
+		}
+		namespace, ok := labels[metadata.InstanceNamespaceLabel]
+		if !ok {
+			return
+		}
```
a-hilaly: nit: if it's managed by kro but missing an important label do we want to log something?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -86,6 +97,61 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(ctx
...
+		if err == nil {
+			r.instanceLogger.Info("update from resource",
+				"name", name,
+				"namespace", namespace,
+				"gvr", gvr,
+				"resourceName", obju.GetName(),
+				"resourceNamespace", obju.GetNamespace(),
+				"eventType", eventType)
+			r.dynamicController.EnqueueObject(instance, eventType)
+		}
```
a-hilaly: Do we need to /GET the object everytime there is an event? can we just trigger a instance reonciliation without a request to the api server? the /GET is already happening here https://github.com/kubernetes-sigs/kro/blob/main/pkg/controller/instance/controller.go#L121-L129

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -160,8 +226,13 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionCRD(
...
 // reconcileResourceGraphDefinitionMicroController starts the microcontroller for handling the resources
-func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionMicroController(ctx context.Context, gvr *schema.GroupVersionResource, handler dynamiccontroller.Handler) error {
-	err := r.dynamicController.StartServingGVK(ctx, *gvr, handler)
+func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionMicroController(
+	ctx context.Context,
+	gvr *schema.GroupVersionResource,
+	handler dynamiccontroller.Handler,
+	resourceHandlers map[schema.GroupVersionResource]cache.ResourceEventHandlerFuncs,
+) error {
+	err := r.dynamicController.StartServingGVK(ctx, *gvr, handler, resourceHandlers)
```
a-hilaly: ditto: we could just pass a list of GVKs instead of resourceHandlers

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -422,111 +430,195 @@ func (dc *DynamicController) enqueueObject(obj interface{}, eventType string) {
...
+	if !ok {
+		return nil, fmt.Errorf("unexpected type %T, expected *unstructured.Unstructured", obj)
+	}
+	for key := range u.Object {
+		if _, allowed := allowedFieldsForPartialObjectMetadata[key]; !allowed {
+			delete(u.Object, key)
+		}
+	}
+	return u, nil
+}
```
a-hilaly: How about using metadata watches? which streams only the partial metadata: [github.com/kubernetes/client-go/tree/master/metadata](https://github.com/kubernetes/client-go/tree/master/metadata)

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -422,111 +430,195 @@ func (dc *DynamicController) enqueueObject(obj interface{}, eventType string) {
...
+		if err := dc.refreshResourceHandlers(resourceHandlers, fw); err != nil {
+			return fmt.Errorf("failed to refresh resource handlers for instance gvr %s: %w", gvr, err)
+		}
+		fw.factory.Start(fw.ctx.Done())
+
+		// Even thought the fw is already registered, we should still
+		// update the instanceHandler, as it might have changed.
+		dc.handlers.Store(gvr, instanceHandler)
+		// trigger reconciliation of the corresponding resourceGVR's
+		objs, err := dc.kubeClient.Resource(gvr).Namespace("").List(fw.ctx, metav1.ListOptions{})
```
a-hilaly: @jakobmoellerdev I'm wondering if it would be reasonable to implement informer reuse across RGDs? 

now if two RGDs are both managing Pods, we create separate Pod informers for each RGD. This leads to duplicate /LIST,/WATCH operations against the API server. 

What are your thoughts on using a shared informer approach where:
- A single GVK (say pod) informer is shared across all RGDs that need it
- Use reference counting to track which RGDs are using each informer
- Only shutdown the informer when no RGDs need it anymore
- Use label selectors to only track resources with `kro.run/owned=true` label to restrict the stream to kro concerned resources.

a-hilaly: Another thought - what about skipping the initial LIST operation entirely by providing a noop `ListFunc` to the informer? we could potentially 1/  still receive an signal to re-reconcile an instance 2/ use the existing informer/handler pattern which works nicely with dynamic controller 3/ as stated above, use a metadata/partialObjectWatch watch only

a-hilaly: More thoughts: I'm a little worried about what happens when the controller restarts with a bunch of existing instances. Seems like we could get into a infinite loop - the instances reconcile, update their resources, then those updates trigger the watchers to reconcile the instances again and again. We can't really tell the difference between updates we just made ourselves vs actual changes that need attention. Maybe we need some kind of cooldown period after startup?

a-hilaly (commented):
@jakobmoellerdev Solid work - thanks a lot! I really like the informer based implementation - it feel to like the right architectural choice. +1 for the feature flag as annotation although i'm down to make it a spec field considering other things stated below.

I've dropped a few comments below - got some thoughts on the startup reconciliation behavior and wondering if we could potentially share informers between RGDs to reduce API server load. Curious to hear your thoughts on these, but overall this is heading in a great direction!

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: I think this is a filter per instance, so it can be set just for a different managing instance right? isnt there potential for log spam here?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: i can try to pass through the instance object, will have to see if that works 👍

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: we have to be careful around our naming: GVK!=GVR.

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: Forget anything I wrote down there. You are completely right I can use a metadata informer factory for the correct header based / server side filtering. good catch!

~I am using metadata watches already. It would not make any practical difference to use the different struct (instead of unstructured) and i even had patch for this. can adopt to your preference. WDYT? (i think most of our code works with unstructured so I thought it would be a little more useful to us)~

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: This thing is named GVK but in reality its a GVR passing.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: Will experiment here

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: I really do not want to complicate this PR. I think in theory the watch can be shared but there were a few reasons why we opted to not go for this:

1. The instance watch is affected by the service account and privilege of the impersonation, which in turn affects the watch. Currently I don't know how we can unify watches efficiently and at the same time keep Service Account Impersonation (happy to hear ideas)
2. The label selector is something I wanted to implement, but that will break informers picking up objects for adoption (as in freshly created / recreated but not with the label). This can lead to extremely weird side-effects so I opted against it. (see my comment here https://github.com/kubernetes-sigs/kro/pull/611#discussion_r2228971694)

jakobmoellerdev: TLDR: We need to think about which level we want to unify watching under and then have unified permission management for that level. Otherwise sharing watches is IMHO problematic because of conflicts between potential service account permissions.

Regarding Cooldown Period: I would think it can be useful, but I would like to implement this as a follow-up to keep the PR reviewable here. (not opposed)

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: One big problem im facing here is that the event trigger here is completely detached from the instance controller right now. I would love to reuse an informer cache here instead of the dynamic client to not push against API server but currently im not sure how to achieve that? (Instance / Micro Controller you linked is started after this code here)

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: Still I agree :)

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: Completely migrated dynamic controller to metadata client.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: fixed up, only passing GVRs now

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: Completely got rid of this by only needing to access GVR information which I dont need the client for.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: got rid of this function completely and made the Enqueue private again

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: I have prepared a diff for you to look over that uses singleton informers across all watches managed by one dynamic controller instance: https://github.com/jakobmoellerdev/kro/compare/replace-watchset-with-informers...jakobmoellerdev:kro:replace-watchset-with-informers-singleton-factory

if you like this I can patch it in, but it would extend the changes made in the controller

jakobmoellerdev: resolving this as I believe I tackled your concern

pkg/metadata/labels.go
jakobmoellerdev: I have included a spec field now with a policy.

jakobmoellerdev:
/hold

I believe I spotted a bug on restart that is not properly handling watches. This might be due to the factory handler. have to investigate some more

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -46,6 +48,35 @@ type ResourceGraphDefinitionSpec struct {
...
+	// When set to "Reactive", the instances will be reconciled when any changes are detected
+	// on the resources owned by the instance. This change detection is achieved by watching
+	// the resources owned by the instance and enqueueing a reconcile request for the controller
+	// responsible for the instance.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Periodic
+	// +kubebuilder:validation:Enum=Periodic;Reactive
+	InstancePolicy ResourceGraphDefinitionInstancePolicy `json:"instancePolicy,omitempty"`
+}
```
a-hilaly: side thread: do we want a field for enabling/disabling applyset?

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,124 @@
...
+	defer w.mu.Unlock()
+
+	reg, ok := w.handlers[id]
+	if !ok {
+		return false, nil
+	}
+	delete(w.handlers, id)
+	if err := w.informer.RemoveEventHandler(reg); err != nil {
+		return false, err
+	}
```
a-hilaly: The handler is already deleted before, if `informer.RemoveEventHandler` fails, you will have inconsistent state. Let's swap them?

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,124 @@
...
+	if w.informer != nil {
+		return
+	}
+	inf := metadatainformer.NewFilteredMetadataInformer(
+		w.client, w.gvr, metav1.NamespaceAll, w.resync,
+		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
+		w.tweak,
+	).Informer()
+	_ = inf.SetWatchErrorHandler(func(_ *cache.Reflector, err error) {
+		fmt.Printf("Watch error for %s: %v\n", w.gvr, err)
```
a-hilaly: let's use the dc logger?

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (w *LazyInformer) ensureInformer() {
+	if w.informer != nil {
+		return
+	}
+	inf := metadatainformer.NewFilteredMetadataInformer(
+		w.client, w.gvr, metav1.NamespaceAll, w.resync,
+		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
+		w.tweak,
+	).Informer()
```
a-hilaly, line 83: Do we want to only watch resources label by `kro.run/owned=true` ? (i'm looking at externalReferences)

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,124 @@
...
+// AddHandler registers a new event handler and starts the informer if needed.
+func (w *LazyInformer) AddHandler(id string, h cache.ResourceEventHandler) error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.ensureInformer()
+	reg, err := w.informer.AddEventHandler(h)
+	if err != nil {
+		return err
+	}
```
a-hilaly, line 108: do we want to clear the informer if `len(w.handlers) == 1`?

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,124 @@
...
+
+	done   <-chan struct{}
+	cancel context.CancelFunc
+}
+
+func NewLazyInformer(
+	ctx context.Context,
+	client metadata.Interface,
+	gvr schema.GroupVersionResource,
+	resync time.Duration,
```
a-hilaly, line 53: late night thoughts: should we just set the resyncPeriod to 0 by default? i don't think we should periodically list everything..

pkg/dynamiccontroller/internal/gvr_watch.go
a-hilaly, line 1: can be addressed at the end: we probably want some sort of basic unit tests here

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -316,246 +219,291 @@ func (dc *DynamicController) syncFunc(ctx context.Context, oi ObjectIdentifiers)
...
+		objMeta, err := meta.Accessor(obj)
+		if err != nil {
+			dc.log.Error(err, "failed to get metadata accessor for object", "eventType", eventType)
+			return
+		}
+		lbls := objMeta.GetLabels()
+		owned, ok := lbls[metadata.OwnedLabel]
+		if !ok || owned != "true" {
+			return
+		}
```
a-hilaly, line 577: how about externalRefs? i'm okkey with ignoring them for now. If we do so we can configure metadatalister to only select objects with kro owned label

pkg/dynamiccontroller/internal/gvr_watch.go
jakobmoellerdev: done, adjusted handler setting to trigger dynamic informer recreation

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: discussed async, basically if I do this we would have to switch watch behavior to track separately for parent and children because parents cannot have informer tracking limited to the owned label, where children can. I believe its technically feasible but I wanna keep the change simple for now. I can address this in an immediate follow up PR that can be reviewed separately

pkg/dynamiccontroller/internal/gvr_watch.go
jakobmoellerdev: I agree on this, I defaulted resync period on 0

pkg/dynamiccontroller/internal/gvr_watch.go
jakobmoellerdev: see https://github.com/kubernetes-sigs/kro/pull/611#discussion_r2421431273

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: yes we do, tackled in https://github.com/kubernetes-sigs/kro/pull/713

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -33,6 +35,35 @@ type ResourceGraphDefinitionSpec struct {
...
+	// When set to "Periodic", or not set, the instances will be reconciled periodically.
+	// When set to "Reactive", the instances will be reconciled when any changes are detected
+	// on the resources owned by the instance. This change detection is achieved by watching
+	// the resources owned by the instance and enqueueing a reconcile request for the controller
+	// responsible for the instance.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Periodic
+	// +kubebuilder:validation:Enum=Periodic;Reactive
+	InstancePolicy ResourceGraphDefinitionInstancePolicy `json:"instancePolicy,omitempty"`
```
tjamet: I think this shouldn't be exposed.

There are at least 2 cases which makes the current behaviour unpredictable. Either when the KRO pod is restarted or when the RGD is updated. Both cases trigger a reconciliation of each instance and hence override any changes on the children.

Making KRO always reactive helps bring predictability in the system as any manual change on the children would be overridden almost immediately, just as soon as the child is updated.
With this, I would expect the person doing the manual change to wonder why the manual changes were overridden rather than taking the manual change for done.

tjamet (commented):
Great to see reconciliation based on child objects.
I think the code can get easier to reason about, specially removing many locks and hence risks of deadlocks, but the feature is worth giving it a try

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: idk about this. I think this policy is not about the instance itself but resources it manages. Periodic behaves more like traditional gitops systems such as flux/argo with a resync interval, whereas reactive is based on dynamic watches purely.

jakobmoellerdev: still i agree that default should probably be reactive.

pkg/dynamiccontroller/internal/gvr_watch.go
jakobmoellerdev: added tests

api/v1alpha1/resourcegraphdefinition_types.go
a-hilaly: The more i think about it, the more i'm leaning towards making "Reactive" the default behaviour, and not expose this configuration at all.

cmd/controller/main.go
```diff
@@ -115,8 +115,9 @@ func main() {
 		"Burst size of events for the dynamic controller rate limiter.")
 
 	// reconciler parameters
-	flag.IntVar(&resyncPeriod, "dynamic-controller-default-resync-period", 36000,
-		"interval at which the controller will re list resources even with no changes, in seconds")
+	flag.IntVar(&resyncPeriod, "dynamic-controller-default-resync-period", 0,
```
a-hilaly: I would still keep this 10hours, just in case we miss events, or encounter watch errors.

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,171 @@
...
+		delete(w.handlers, id)
+	}
+
+	if len(w.handlers) == 0 {
+		w.cancel()
+		<-w.done
+		w.cancel = nil
+		w.informer = nil
+		return true, nil
+	}
```
a-hilaly: Would this cause a panic if the w.Shutdown() is called when the informer has 0 handlers?

api/v1alpha1/resourcegraphdefinition_types.go
tjamet: > Periodic behaves more like traditional gitops systems such as flux/argo with a resync interval, whereas reactive is based on dynamic watches purely.

I get this. but I their context is different. gitops handles external sources and hardly have a way to get notified when the source changes. That's the main reason why they have to go for resyncs.

I may be mistaking, but I don't think this is a problem we have with KRO. In the KRO case, all resources are k8s resources and should be watchable.

jakobmoellerdev:
/unhold
PR should be fine now.

Probably only thing left is if we want the spec extension or not. Considering what I received as feedback and our choice on applysets, I will remove the spec field again.

jakobmoellerdev:
@a-hilaly interestingely the fix from https://github.com/kubernetes-sigs/kro/pull/764 was necessary to make integration tests run, I believe the test was giving false positives before dynamic watching which is why we didnt catch it.

Also still have to figure out what happened to the E2E tests now that i moved to reactive reconcile by default.

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: Moved to reactive by default, resolving and fixing tests separately

pkg/dynamiccontroller/internal/gvr_watch.go
jakobmoellerdev: changed the check on the handlers so that it only runs this if we had a running handler before.

cmd/controller/main.go
```diff
@@ -116,7 +116,8 @@ func main() {
 
 	// reconciler parameters
 	flag.IntVar(&resyncPeriod, "dynamic-controller-default-resync-period", 36000,
-		"interval at which the controller will re list resources even with no changes, in seconds")
+		"interval at which the controller will re list resources even with no changes, in seconds. "+
+			"By default resync is disabled.")
```
michaelhtm: why do we default this value to 36000 if it's turned off by default?

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -60,6 +60,25 @@ spec:
             description: ResourceGraphDefinitionSpec defines the desired state of
               ResourceGraphDefinition
             properties:
+              reconcile:
+                description: Reconcile defines the desired reconciliation settings
+                  for a ResourceGraphDefinition.
+                properties:
+                  instancePolicy:
+                    default: Reactive
+                    description: |-
```
michaelhtm: is this field affected by the `dynamic-controller-default-resync-period` flag? Should we allow users to define the instance resync period with a field in RGD as well? with a fallback to the flag

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -89,6 +85,14 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(
...
 }
 
+func (r *ResourceGraphDefinitionReconciler) getResourceGVRsToWatchForRGD(processedRGD *graph.Graph) []schema.GroupVersionResource {
+	resourceHandlers := make(map[schema.GroupVersionResource]struct{}, len(processedRGD.Resources))
+	for _, resource := range processedRGD.Resources {
+		resourceHandlers[resource.GetGroupVersionResource()] = struct{}{}
+	}
+	return slices.Collect(maps.Keys(resourceHandlers))
+}
+
```
michaelhtm, line 95: should this be an opt in feature? maybe via flag/RGD field

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -116,167 +134,120 @@ type Config struct {
...
+	// Map of active informers per GVR.
+	// Guarded by mu.
+	watches map[schema.GroupVersionResource]*internal.LazyInformer
+	// Map of parent registrations per GVR.
+	// Guarded by mu.
+	registrations map[schema.GroupVersionResource]*registration
+	// Global mutex protecting watches and registrations.
+	// Required because StartServingGVK and StopServiceGVK may run concurrently,
+	// and because Run or gracefulShutdown may also traverse these maps.
+	mu sync.Mutex
```
michaelhtm: nit: should we have the mutex as the first field?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -116,167 +134,120 @@ type Config struct {
...
+
+	// Latest handler for each parent GVR used by syncFunc.
+	handlers sync.Map // map[schema.GroupVersionResource]Handler (thread-safe on its own)
+	queue    workqueue.TypedRateLimitingInterface[ObjectIdentifiers]
+
+	// Parent run context, inherited by all informer stop contexts.
+	// set by Run
+	ctx context.Context
```
michaelhtm: nit: maybe this could go to the top as well

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -116,167 +134,120 @@ type Config struct {
...
+	}
+
+	err := dc.syncFunc(ctx, item, handler.(Handler))
 	if err == nil || apierrors.IsNotFound(err) {
 		dc.queue.Forget(item)
 		return true
 	}
 
-	gvrKey := fmt.Sprintf("%s/%s/%s", item.GVR.Group, item.GVR.Version, item.GVR.Resource)
+	gvrKey := item.GVR.String()
```
michaelhtm: nit: why are we changing the GVR string format?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +275,315 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
+	case <-queueShutdownDone:
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("timeout waiting for queue to shutdown: %w", ctx.Err())
+	}
+}
 
-	// Unregister the handler if any
-	dc.handlers.Delete(gvr)
+func (dc *DynamicController) handlerForChildGVR(parent, child schema.GroupVersionResource) cache.ResourceEventHandler {
```
michaelhtm, line 565: nit: should we put this above the shutdown function? It would be nice to have shutdown at the end of the file

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,172 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/metadata"
+	"k8s.io/client-go/metadata/metadatainformer"
+	"k8s.io/client-go/tools/cache"
+)
+
+// LazyInformer manages a SharedIndexInformer per GVR with multiple handlers.
+// It lazily starts when the first handler is added and stops when the last is removed.
+// It can restart again after a full shutdown.
+type LazyInformer struct {
```
michaelhtm, line 34: I like the name

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +275,315 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
+	return nil
+}
+
+// reconcileChildrenLocked ensures that reg.childHandlerIDs matches the desired set.
+// Must be called with dc.mu held.
+func (dc *DynamicController) reconcileChildrenLocked(
+	parent schema.GroupVersionResource,
+	desired []schema.GroupVersionResource,
+	reg *registration,
+) error {
```
michaelhtm, line 501: TIL having `Locked` as function postfix means the function should only be called with a held lock

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +275,315 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
+	// ensure watch
+	w := dc.ensureWatchLocked(parent)
+
+	// create handler if missing
+	if reg.parentHandlerID == "" {
+		handlerID := "parent:" + parent.String()
+		if err := w.AddHandler(dc.ctx, handlerID, cache.ResourceEventHandlerFuncs{
+			AddFunc:    func(obj interface{}) { dc.enqueueParent(parent, obj, "add") },
+			UpdateFunc: func(oldObj, newObj interface{}) { dc.updateFunc(parent, oldObj, newObj) },
+			DeleteFunc: func(obj interface{}) { dc.enqueueParent(parent, obj, "delete") },
```
michaelhtm: nit: should we make the `add`, `delete` eventTypes enum/const?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +275,315 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
-	if !synced {
-		cancel()
-		return fmt.Errorf("failed to sync informer cache for GVR %s", gvr)
+	// add missing
+	for child := range desiredSet {
+		if _, exists := reg.childHandlerIDs[child]; exists {
+			continue
+		}
+		w := dc.ensureWatchLocked(child)
+		handlerID := "child:" + parent.String() + "->" + child.String()
```
michaelhtm: nit: can we make the ID creation into a helper function?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +275,315 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
+			if w, ok := dc.watches[child]; ok {
+				stopped, err := w.RemoveHandler(id)
+				if err != nil {
+					return fmt.Errorf("remove child handler %s: %w", child, err)
+				}
+				if stopped {
+					delete(dc.watches, child)
+				}
+			}
+			delete(reg.childHandlerIDs, child)
```
michaelhtm: nit: can we make a function for watches removals? we can reuse it in parent as well

cmd/controller/main.go
a-hilaly: see https://github.com/kubernetes-sigs/kro/pull/611#discussion_r2476961932

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: https://github.com/kubernetes-sigs/kro/pull/611#discussion_r2452593503

cmd/controller/main.go
michaelhtm: we should probably update the arg description then..

a-hilaly:
```suggestion
		"interval at which the controller will re list resources even with no changes, in seconds. ")
```

jakobmoellerdev: agree, leftover. thanks!

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
jakobmoellerdev: we decided in the PR to remove the instance policy, so this field will disappear in favor of watches always being enabled.

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: actually youre right this would break labelling. moving back to old format

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: extracted removal func

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: done

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: done

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: moved to top

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: moved all id gens to the old keying mechanism and used string builder, avoids tons of allocs

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: moved above

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: acked, moved as first

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: decided to stick permanently with watches.

tjamet (approved):
I think it makes sense to move forward with this.
Ideally, we could have a more flexible interface similar to

```
dc.Register(ctx, dynamiccontroller.For(rgdGVR).Owns(childGVRs....), reconcileFunc)
```

This can definitely come in future iterations

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,172 @@
...
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	return w.informer
+}
+
+// Shutdown stops the informer and clears state.
+func (w *LazyInformer) Shutdown() {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	w.cancel()
```
matthchr, line 169: Not setting cancel to nil here?

pkg/dynamiccontroller/internal/gvr_watch.go
```diff
@@ -0,0 +1,172 @@
+// Copyright 2025 The Kube Resource Orchestrator Authors
```
matthchr, line 1: minor, style, feel free to ignore: Might be nice to just name this `lazy_informer.go` given that's all it contains?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +304,372 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
-		close(done)
-	}()
-
-	select {
-	case <-done:
-		dc.log.Info("All informers shut down successfully")
-	case <-ctx.Done():
-		dc.log.Error(ctx.Err(), "Timeout waiting for informers to shut down")
-		return ctx.Err()
+func (dc *DynamicController) enqueueParent(parentGVR schema.GroupVersionResource, obj interface{}, eventType string) {
```
matthchr, line 327: minor, style: 
```suggestion
func (dc *DynamicController) enqueueParent(parentGVR schema.GroupVersionResource, obj iany, eventType string) {
```

and elsewhere.

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +304,372 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
+	}
+	return nil
+}
 
-	wrapper, ok := informerObj.(*informerWrapper)
-	if !ok {
-		return fmt.Errorf("invalid informer type for GVR: %s", gvr)
+func (dc *DynamicController) handlerForChildGVR(parent, child schema.GroupVersionResource) cache.ResourceEventHandler {
+	parentGVRKey, childGVRKey := keyFromGVR(parent), keyFromGVR(child)
+	handle := func(obj interface{}, eventType string) {
```
matthchr, line 567: minor, style, feel free to ignore: Could this be a method on `DynamicController` too rather than inline here?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -306,265 +304,372 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
-	if !ok {
-		dc.log.Error(nil, "failed to cast old object to unstructured")
+	oldMeta, err := meta.Accessor(oldObj)
+	if err != nil {
+		dc.log.Error(err, "failed to access old object meta")
 		return
 	}
-
-	if newObj.GetGeneration() == oldObj.GetGeneration() {
+	if newMeta.GetGeneration() == oldMeta.GetGeneration() {
```
matthchr, line 355: Is this always correct? Are there cases where annotation or labels change in a way that will result in us wanting to reconcile parent, but that doesn't trigger generation update?

matthchr (commented):
Late to the game as this merged last night, but had a few minor comments.

If you want to address them, can be done in a separate PR.

pkg/dynamiccontroller/dynamic_controller.go
matthchr: Spoke with @a-hilaly about this and realized that this is only for the parent and the children are handled differently in a way that (should) trigger reconcile if annotations/labels change.

Not reconciling the graph when we change the instance labels/annotations seems OK/correct. Maybe one day we'll have some annotations/labels that are feature-controlling in which case we might want to reconcile when those change, but we could list them there explicitly rather than doing it on any change, so this seems correct to me.

jakobmoellerdev:
Will run over this with another PR to address these follow ups. Thanks @matthchr!

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: I believe that we have 2 cases I know of where we can  have a problem with this, mainly config maps and secrets. pretty much all other resources that have a spec subresource will be able to use this filter on generation.

pkg/dynamiccontroller/internal/gvr_watch.go
jakobmoellerdev: Actually edited: https://github.com/kubernetes-sigs/kro/pull/611/files#diff-2d9495fa1bbee5c492736c35d73bbca277d5b4abeb1d7fc76d77db80f8876d92R98 I think this can cause misbehavior and I need to fix that. good catch!

PR #616, opened 2025-07-31
fix!: properly shutdown controller manager and dynamic controller

jakobmoellerdev:
previously the informer queue of the dynamic controller and the manager shut-down immediately (got PID killed). 

This is problematic because it means that a reconcile loop can potentially be interrupted in the middle of a run even though we have a normally graceful shutdown.

Fortunately controller runtime provides a nice way to fix this: by adding the dynamic controller into the manager runnables instead of leaking a goroutine as well as fixing how the manager is actually stopped (previously the manager was run in a goroutine).

Additionally we now properly shutdown the queue of the dynamic controller and wait for the workers to finish processing.

Note that if this process still takes longer than the graceful shutdown period, we forcefully kill the manager and the binary.

Because the shutdown timeout is now respected globally, I have replaced 

- `--dynamic-controller-default-shutdown-timeout` with `--graceful-shutdown-timeout`, and
- `KRO_DYNAMIC_CONTROLLER_DEFAULT_SHUTDOWN_TIMEOUT` with `KRO_GRACEFUL_SHUTDOWN_TIMEOUT` 

This is because the shutdown timeout is now handled globally within the manager and not individually just for the dynamic controller. A nice side-effect is that this shutdown timeout now applies globally to other controllers as well and not just the dynamic controller

An example log now when the process receives a shutdown signal:

```
2025-07-31T15:34:25.977+0200	INFO	Stopping and waiting for non leader election runnables
2025-07-31T15:34:25.977+0200	INFO	Stopping and waiting for leader election runnables
2025-07-31T15:34:25.977+0200	INFO	rgd-controller	Shutdown signal received, waiting for all workers to finish	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition"}
2025-07-31T15:34:25.977+0200	INFO	dynamic-controller	Received shutdown signal, shutting down dynamic controller queue
2025-07-31T15:34:25.977+0200	INFO	rgd-controller	All workers finished	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition"}
2025-07-31T15:34:25.977+0200	INFO	dynamic-controller	Dynamic controller worker received shutdown signal, stopping
2025-07-31T15:34:25.977+0200	INFO	dynamic-controller	All workers have stopped
2025-07-31T15:34:25.977+0200	INFO	dynamic-controller	Starting graceful shutdown
2025-07-31T15:34:25.977+0200	DEBUG	dynamic-controller	Shutting down informer	{"gvr": "kro.run/v1alpha1, Resource=deploymentservices"}
2025-07-31T15:34:25.977+0200	INFO	dynamic-controller	All informers shut down successfully
2025-07-31T15:34:25.977+0200	INFO	dynamic-controller	Shutting down dynamic controller
2025-07-31T15:34:25.977+0200	INFO	Stopping and waiting for caches
2025-07-31T15:34:25.977+0200	INFO	Stopping and waiting for webhooks
2025-07-31T15:34:25.977+0200	INFO	Stopping and waiting for HTTP servers
2025-07-31T15:34:25.977+0200	INFO	shutting down server	{"name": "health probe", "addr": "[::]:8079"}
2025-07-31T15:34:25.977+0200	INFO	controller-runtime.metrics	Shutting down metrics server with timeout of 1 minute
2025-07-31T15:34:25.977+0200	INFO	Wait completed, proceeding to shutdown the manager
```


cmd/controller/main.go
```diff
@@ -112,8 +113,6 @@ func main() {
 		"interval at which the controller will re list resources even with no changes, in seconds")
 	flag.IntVar(&queueMaxRetries, "dynamic-controller-default-queue-max-retries", 20,
 		"maximum number of retries for an item in the queue will be retried before being dropped")
-	flag.IntVar(&shutdownTimeout, "dynamic-controller-default-shutdown-timeout", 60,
```
matthchr, line 120: We should probably call this out in the release notes since it's breaking, especially given we've also changed the name of the parameter in the Helm chart.

@a-hilaly do we have a good process to flag something that should be highlighted in the release notes?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -224,17 +220,40 @@ func (dc *DynamicController) Run(ctx context.Context) error {
...
+		dc.queue.ShutDown()
+	}()
+
+	wg.Wait()
+	dc.log.Info("All workers have stopped")
 
-	<-ctx.Done()
-	return dc.gracefulShutdown(dc.config.ShutdownTimeout)
+	// when shutting down, the context given to Start is already closed,
+	// and the expectation is that we block until the graceful shutdown is complete.
```
matthchr, line 243: If graceful shutdown is hit, the manager runnable will just stop waiting and process will exit, correct? (That's why we can just block forever here).

jakobmoellerdev: Right, the manager runnable will run Start(ctx) with a graceful shutdown passed to the manager. If that period is exceeded, this Start method here will be aborted even though its still running at that point. Because the manager runnable context returns though, we kill the process!

cmd/controller/main.go
jakobmoellerdev: if we would use something like labels and conventional commits, we could make it so that something like `fix!: myfix` would mark the breaking change with a `!` (see https://www.conventionalcommits.org/en/v1.0.0/#summary). but idk if thats too restrictive. apart from that it would be good to mark it somehow manually with a label and then be able to generate release notes from that

a-hilaly: >@a-hilaly do we have a good process to flag something that should be highlighted in the release notes?

Not really, so far we manually flagged things. We do use https://github.com/softprops/action-gh-release for github releases and at first glance looks like they parse conventional commits.

jakobmoellerdev: I marked it right now and will pay attention on the next release, if the notes get generated correctly, all good, then it should just become part of the contributor guide

jakobmoellerdev:
@a-hilaly @matthchr anything left to do for me here before merge?

jakobmoellerdev:
@a-hilaly @matthchr friendly ping as this has been open for a few weeks now

matthchr:
As far as I am concerned this is good to merge. @a-hilaly if you want to take another look that's fine by me, but otherwise my vote is merge?

I'm out of office all next week but when I'm back if no more comments I'll merge this. Otherwise feel free to merge while I am out as well.

a-hilaly (approved):
Dankeshoen!
/lgtm

jakobmoellerdev:
FYI for anyone looking at this I will reintroduce controller shutdown timeouts that will limit max waiting time for shutdown for individually started instance controllers in a separate PR (the flag that was removed originally wanted to do this, but never did)

PR #617, opened 2025-08-01
kro cli package command

DhairyaMajmudar:
## Description 

The new kro cli package command is useful to convert a RGD file into an OCI image which can be further published into registries like AWS or Docker hub.

## Usage

 `kro package -f [FILE] -t [TAG]`

Flags:

| Shorthand | Flag | Usage | 
| ---- | ----- | --- |
 | -t |  --tag | Tag to the image, Default: Latest |

## Demo

[kro-package.webm](https://github.com/user-attachments/assets/995b9068-5b08-412b-a95e-20da0b5e5f74)

### DIFF
--- a/cmd/kro/commands/package/package.go
+++ b/cmd/kro/commands/package/package.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package commands
+package pkg
 
 import (
 	"archive/tar"


barney-s:
I would add commands related to packaging under `kro package`  similar to helm package? 

Maybe the PR description is wrong.

cmd/kro/commands/package.go
```diff
@@ -0,0 +1,130 @@
...
+	resourceGraphDefinitionFile string
+	outputFile                  string
+}
+
+var packageConfig = &PackageConfig{}
+
+func init() {
+	packageRGDCmd.PersistentFlags().StringVarP(&packageConfig.resourceGraphDefinitionFile, "file", "f", "",
+		"Path to the ResourceGraphDefinition file")
+	packageRGDCmd.PersistentFlags().StringVarP(&packageConfig.outputFile, "output", "o", "",
```
barney-s: Can this be made optional ? If not provided deduce name from file name OR RGD name ?

cmd/kro/commands/package.go
```diff
@@ -0,0 +1,130 @@
...
+		"Output file name for the OCI image tarball")
+}
+
+var packageRGDCmd = &cobra.Command{
+	Use:   "package",
+	Short: "Package ResourceGraphDefinition file into an OCI image",
```
barney-s: Suggest "Create an OCI Image packaging the ResourceGraphDefinition"

cmd/kro/commands/package.go
```diff
@@ -0,0 +1,130 @@
...
+	Long: "Package command packages the ResourceGraphDefinition" +
```
barney-s: Creates an OCI image bundle ?

cmd/kro/commands/generate/generate.go
```diff
@@ -33,12 +33,13 @@ type GenerateConfig struct {
 	outputFormat                string
 }
 
-var config = &GenerateConfig{}
+var generateConfig = &GenerateConfig{}
```
jakobmoellerdev, line 26: nit: imo unneeded variable rename

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,130 @@
...
+	resourceGraphDefinitionFile string
+	outputFile                  string
+}
+
+var packageConfig = &PackageConfig{}
+
+func init() {
+	packageRGDCmd.PersistentFlags().StringVarP(&packageConfig.resourceGraphDefinitionFile, "file", "f", "",
+		"Path to the ResourceGraphDefinition file")
+	packageRGDCmd.PersistentFlags().StringVarP(&packageConfig.outputFile, "output", "o", "",
```
jakobmoellerdev: should this maybe default to a local file in current working dir?

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,130 @@
...
+
+type PackageConfig struct {
+	resourceGraphDefinitionFile string
+	outputFile                  string
+}
+
+var packageConfig = &PackageConfig{}
+
+func init() {
+	packageRGDCmd.PersistentFlags().StringVarP(&packageConfig.resourceGraphDefinitionFile, "file", "f", "",
```
jakobmoellerdev: similar, should this default to looking for a specific file name that we usually use with docs etc?

go.mod
```diff
@@ -4,14 +4,15 @@ go 1.24.0
 
 require (
 	github.com/go-echarts/go-echarts/v2 v2.6.1
-	github.com/go-logr/logr v1.4.2
+	github.com/go-logr/logr v1.4.3
 	github.com/gobuffalo/flect v1.0.2
 	github.com/google/cel-go v0.24.1
-	github.com/google/go-cmp v0.6.0
+	github.com/google/go-cmp v0.7.0
+	github.com/google/go-containerregistry v0.20.6
```
jakobmoellerdev: I know we are probably opinionated towards go-containerregistry, but I would argue that the more "open-sourcy" version right now is https://oras.land/

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,130 @@
...
+
+		fmt.Fprintf(os.Stderr, "Successfully packaged ResourceGraphDefinition to %s\n",
+			packageConfig.outputFile)
+
+		return nil
+	},
+}
+
+func packageRGD(data []byte, rgd *v1alpha1.ResourceGraphDefinition) error {
+	layer := static.NewLayer(data, types.MediaType("application/vnd.kro.resourcegraphdefinition.v1alpha1+yaml"))
```
jakobmoellerdev: IMO this should be a plain yaml layer, and the image should have the artifact type (https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions) `application/vnd.kro.resourcegraphdefinition`

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,130 @@
...
+		"kro.run/type": "resourcegraphdefinition",
+		"kro.run/name": rgd.Name,
+	}
+
+	img, err = mutate.ConfigFile(img, configFile)
+	if err != nil {
+		return fmt.Errorf("failed to update image config: %w", err)
+	}
+
+	ref, err := name.ParseReference(fmt.Sprintf("kro.run/rgd/%s:latest", rgd.Name))
```
jakobmoellerdev: I think latest should be configurable

cmd/kro/commands/package/package.go
jakobmoellerdev, line 1: im missing a test for this command

cmd/kro/commands/package.go
DhairyaMajmudar: That's helpful, thank you :)

cmd/kro/commands/generate/generate.go
DhairyaMajmudar: yup that varialbe rename is unrelated to the new command but I made it here to follow up the consistency across other commands.

In package command I'm using `packageConfig`, and same as in generate command it should be `generateConfig`.

cmd/kro/commands/package.go
DhairyaMajmudar: Thanks for this suggestion, I've now made the -o flag to be optional... suppose it is not provided, then the output tar folder would be named according to the rgd name available, and suppose there's no rgd name then it will take the file name provided.

go.mod
DhairyaMajmudar: Cool, ORAS is a CNCF project. Yes more open-sourcy project is oras here in this case. Let me check, if we're getting any extra benefits from using it or no in place of Google container registry.

cmd/kro/commands/package/package.go
DhairyaMajmudar: Right, added a flag "-t" to configure the tag : )

cmd/kro/commands/generate/generate.go
a-hilaly: let's do these in a seperate PR @DhairyaMajmudar

cmd/kro/commands/package/package.go
a-hilaly: should it be `rgd.yaml`?

a-hilaly: i know our examples our using different names for RGD files, happy to standardize this and rename them all.

go.mod
a-hilaly: TIL - i think it would be great to switch to a CNCF project, however given the GSoC timelines, please feel free to use own judgement here. We could always revisit.

cmd/kro/commands/package/package.go
jakobmoellerdev: +1 for rgd.yaml, probably the easiest one to use / remember

cmd/kro/commands/package/package.go
DhairyaMajmudar: Right, made it default to the local file

cmd/kro/commands/package/package.go
DhairyaMajmudar: Made it default to rgd.yaml : )

cmd/kro/commands/package.go
DhairyaMajmudar: done

go.mod
DhairyaMajmudar: Migrated to ORAS : )

cmd/kro/commands/package/package.go
DhairyaMajmudar: made suggested changes

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("failed to push RGD blob: %w", err)
+	}
+
+	layer := blobDesc
+	if layer.Annotations == nil {
+		layer.Annotations = map[string]string{}
+	}
+	layer.Annotations[ocispec.AnnotationTitle] = filepath.Base(packageConfig.resourceGraphDefinitionFile)
+
+	artifactType := "application/vnd.kro.resourcegraphdefinition"
```
jakobmoellerdev, line 118: can be a global const that people can depend on. 

Also probably needs a version (so we can change the format later)

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,213 @@
...
+	packOpts := oras.PackManifestOptions{
+		Layers: []ocispec.Descriptor{layer},
+		ManifestAnnotations: map[string]string{
+			"kro.run/type": "resourcegraphdefinition",
```
jakobmoellerdev, line 122: whats the use of this type manifest annotation?

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,213 @@
...
+			fmt.Println("failed to remove temp dir", tempDir, ":", err)
+		}
+	}()
+
+	store, err := oci.New(tempDir)
+	if err != nil {
+		return fmt.Errorf("failed to create OCI layout store: %w", err)
+	}
+
+	mediaType := "application/vnd.kro.resourcegraphdefinition"
```
jakobmoellerdev, line 106: this is not a valid media type. you store the rgd as yaml right? so the media type would be `application/x-yaml` or similar that is recognized by IANA.

cmd/kro/commands/package/package.go
```diff
@@ -0,0 +1,213 @@
...
+		if err := yaml.Unmarshal(data, &rgd); err != nil {
+			return fmt.Errorf("failed to unmarshal ResourceGraphDefinition: %w", err)
+		}
+
+		basename := filepath.Base(packageConfig.resourceGraphDefinitionFile)
+		ext := filepath.Ext(basename)
+		nameWithoutExt := basename[:len(basename)-len(ext)]
+		outputTar := nameWithoutExt + ".tar"
+
+		if err := packageRGD(outputTar, data, &rgd); err != nil {
```
jakobmoellerdev, line 79: I just noticed that you store the data here from the file. This can be a problem. Imagine if you have an unknown key inside your file. yaml.Unmarshal would work and then that key would be stored in the layer. Im not sure if thats intended. To maintain consistency with a stable media type, I would suggest that you marshal the rgd back into a byte stream and use that so you KNOW the format.

if you dont do this, then you need to at least understand what format the rgd file is in to suffix either `+json` or `+yaml` to the media type. Otherwise it will be hard to introspect later.

jakobmoellerdev:
@DhairyaMajmudar friendly reminder, are you still interested in working on this?

a-hilaly:
Closing for now - @DhairyaMajmudar we'll need a KREP doc on the CLI before we move forward.

PR #622, opened 2025-08-07
Update instance status sub-conditions.

n3wscott:
Updated instance status to:

```
  Ready
  ├─ GraphResolved - Runtime graph has been created and resources resolved
  ├─ ResourcesReady - All resources in the graph are created and ready
  └─ InstanceManaged - Instance finalizers and labels are properly set
```


Makefile
```diff
@@ -96,7 +96,7 @@ vet: ## Run go vet against code.
 .PHONY: test
 test: manifests generate fmt vet ## Run tests. Use WHAT=unit or WHAT=integration
 ifeq ($(WHAT),integration)
-	go test -v ./test/integration/suites/... -coverprofile integration-cover.out
+	go test -timeout=20m -v ./test/integration/suites/... -coverprofile integration-cover.out
```
n3wscott:
```suggestion
	go test -v ./test/integration/suites/... -coverprofile integration-cover.out
```

pkg/controller/instance/controller.go
```diff
@@ -269,3 +278,35 @@ func getServiceAccountUserName(namespace, serviceAccount string) (string, error)
...
+
+	status := map[string]interface{}{
+		"state": InstanceStateError,
+	}
+
+	if conditions := conditionSet.List(); len(conditions) > 0 {
+		conditionsInterface := make([]interface{}, len(conditions))
+		for i, condition := range conditions {
+			conditionsInterface[i] = map[string]interface{}{
+				"type":               string(condition.Type),
```
n3wscott: marshal this instead

pkg/controller/instance/controller_status.go
```diff
@@ -16,34 +16,151 @@ package instance
...
+// wrapInstance creates an adapter that allows unstructured instances to work with condition management
+func wrapInstance(instance *unstructured.Unstructured) *unstructuredConditionAdapter {
+	return &unstructuredConditionAdapter{Unstructured: instance}
+}
+
+// NewConditionsMarkerFor creates a marker to manage conditions and sub-conditions for instances.
+//
+// ```
+// Ready
+//
```
n3wscott, line 89:
```suggestion
```

docs/design/proposals/status-conditions.md
```diff
@@ -198,18 +198,84 @@ Things to note:
...
+    status: "True"
+    type: ResourcesReady
+```
+
+Things to note:
+
+- The messages have been customized to reflect the specific phase of reconciliation
+- ObservedGeneration is added and used in all conditions
+- The condition Types reflect the major work sections in the instance reconciler:
+  - InstanceSynced → InstanceManaged + GraphResolved + ResourcesReady structure
```
jakobmoellerdev, line 276: wondering where InstanceSynced is in your overview now? do you overwrite it with InstanceManaged?

In general I must admit that to me InstanceSynced is much more intuititive but I guess thats subjective. Thats because for me the resource being managed is not directly a correlation to it being correctly synced with the instance graph

n3wscott: being synced is just one small part of the sub-condition of `ResourcesReady`. ResourcedReady would be they are synced and ready, they can't be ready without being synced + the observed generation match generation.

n3wscott: I could be in convinced we need to break it apart into `InstancesSynced` + `InstancesReady` but the code did not seem well structured to make this split and would have to require a refactor.

docs/design/proposals/status-conditions.md
```diff
@@ -198,18 +198,84 @@ Things to note:
...
+    observedGeneration: 1
+    reason: "Resolved"
+    status: "True"
+    type: GraphResolved
+  - lastTransitionTime: "2025-08-07T21:31:02Z"
+    message: All resources are created and ready
+    observedGeneration: 1
+    reason: "AllResourcesReady"
+    status: "True"
+    type: ResourcesReady
```
a-hilaly, line 268: Do we want to also add a condition per managed resource? i'm thinking it might be noisy but helpful to debug what's wrong and at which step things are failing.

pkg/controller/instance/controller.go
```diff
@@ -135,6 +137,10 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) error {
 	// runtime object in it's fields.
 	rgRuntime, err := c.rgd.NewGraphRuntime(instance)
 	if err != nil {
+		// Mark graph resolution failure and update status before returning error
+		mark := NewConditionsMarkerFor(instance)
+		mark.GraphNotResolved("failed to create runtime resource graph definition: %v", err)
+		c.updateInstanceStatusOnError(ctx, instance)
 		return fmt.Errorf("failed to create runtime resource graph definition: %w", err)
 	}
```
a-hilaly: do we want to set GraphResolved right after this block?

docs/design/proposals/status-conditions.md
```diff
@@ -198,18 +198,84 @@ Things to note:
...
-To be able to use a condition manager, we would need a signal from the provided
-status shape to understand if the defined condition is a dependent condition or
-just informational.
+Based on analysis of the instance reconciler, the major sections of work can be
+broken down into the following sub-conditions that align with the reconciliation
+phases:
+
+```
+Ready
+├─ InstanceManaged - Instance finalizers and labels are properly set
```
a-hilaly, line 211: ~IIRC kro doesn't inject labels on instances - they are typically set by users~

pkg/controller/instance/controller.go
```diff
@@ -135,6 +137,10 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) error {
 	// runtime object in it's fields.
 	rgRuntime, err := c.rgd.NewGraphRuntime(instance)
 	if err != nil {
+		// Mark graph resolution failure and update status before returning error
+		mark := NewConditionsMarkerFor(instance)
+		mark.GraphNotResolved("failed to create runtime resource graph definition: %v", err)
+		c.updateInstanceStatusOnError(ctx, instance)
```
a-hilaly: shall we propagate the error into the condition here?

pkg/controller/instance/controller_reconcile.go
```diff
@@ -112,42 +112,61 @@ func (igr *instanceGraphReconciler) handleReconciliation(ctx context.Context, re
...
 	instance := igr.runtime.GetInstance()
+	mark := NewConditionsMarkerFor(instance)
 
 	// Set managed state and handle instance labels
 	if err := igr.setupInstance(ctx, instance); err != nil {
 		return fmt.Errorf("failed to setup instance: %w", err)
 	}
 
+	// Mark that the graph has been resolved (runtime is successfully created and resources are available)
+	mark.GraphResolved()
```
a-hilaly: runtime instiantiation happens outside this function, shall move logic outside?

pkg/controller/instance/controller_status.go
```diff
@@ -16,34 +16,150 @@ package instance
...
+	}
+
+	var conditionsInterface []interface{}
+	if err := json.Unmarshal(conditionsJSON, &conditionsInterface); err != nil {
+		return // Fail silently - could log this in the future
+	}
+
+	if err := unstructured.SetNestedSlice(u.Object, conditionsInterface, "status", "conditions"); err != nil {
+		return // Fail silently - could log this in the future
+	}
```
a-hilaly, line 78: a part of things that this is worthy of a panic. basically this should never happen, we own the full logic of conditions

docs/design/proposals/status-conditions.md
```diff
@@ -198,18 +198,84 @@ Things to note:
...
 
-To be able to use a condition manager, we would need a signal from the provided
-status shape to understand if the defined condition is a dependent condition or
-just informational.
+Based on analysis of the instance reconciler, the major sections of work can be
+broken down into the following sub-conditions that align with the reconciliation
+phases:
+
+```
+Ready
```
a-hilaly, line 210: i've been thinking about adding a feature that let's kro validate it's permissions to manage RGD instances (using [SelfSubjectRuleReview](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-rules-review-v1/)), maybe we need a condition "Authorized" (there is a probably a better name?) which could surface state describing client impersonation and subject rule reviews evaluation results?

docs/design/proposals/status-conditions.md
n3wscott: I really would like to, maybe it could be an option? I think it would help debug and understand but also it leaks a lot of details that you might not want to leak. For example, if a resource makes 100 children, not sure if we want to show 100 sub-conditions :D 

What do others think? We could start with one and do an experiment where we do verbose mode via an annotation and if that is interesting we could roll it into the real spec settings on RGD?

pkg/controller/instance/controller.go
n3wscott: good call, let me look at this

docs/design/proposals/status-conditions.md
n3wscott: I just added this feature to crossplane: https://github.com/crossplane/crossplane/pull/6572

docs/design/proposals/status-conditions.md
a-hilaly: fair point! let's start small and add a verbose mode via annotations (thinking at both RGD and Instance levels)

docs/design/proposals/status-conditions.md
jakobmoellerdev: @a-hilaly I think this deserves an own issue, right? do you already have something for that? if not I can create one and we can discuss further there

jakobmoellerdev: I took a jab at this in https://github.com/kro-run/kro/pull/646. Feel free to take a look anytime

a-hilaly (approved):
Thank you @n3wscott for putting this together! 👍 

@jakobmoellerdev @barney-s wanna take afinal look and merge this?

docs/design/proposals/status-conditions.md
jakobmoellerdev: I think this is something we can do separately as we agreed our only breaking point here should be Ready

pkg/controller/instance/controller_status.go
```diff
@@ -16,34 +16,150 @@ package instance
...
+
+// NewConditionsMarkerFor creates a marker to manage conditions and sub-conditions for instances.
+//
+// ```
+// Ready
+//	├─ InstanceManaged - Instance finalizers and labels are properly set
+//	├─ GraphResolved - Runtime graph created and all resources resolved
+//	└─ ResourcesReady - All resources are created and ready
+//
+// ```
```
n3wscott, line 94:
```suggestion
// NewConditionsMarkerFor creates a marker to manage conditions and sub-conditions for instances.
// ```
// Ready
//	├─ InstanceManaged - Instance finalizers and labels are properly set
//	├─ GraphResolved - Runtime graph created and all resources resolved
//	└─ ResourcesReady - All resources are created and ready
// ```
```

pkg/controller/instance/controller.go
n3wscott: the condition already has the error, updateInstanceStatusOnError just manages the state field and updates conditions based on what is already in the condition manager.

test/e2e/chainsaw/check-arbitary-objects/instance-create-assert.yaml
```diff
@@ -21,10 +21,24 @@ spec:
...
+  - message: all resources are created and ready
+    observedGeneration: 1
+    reason: AllResourcesReady
+    status: "True"
+    type: ResourcesReady
+  - message: ""
+    observedGeneration: 1
+    reason: Ready
+    status: "True"
+    type: Ready
```
a-hilaly, line 43: @n3wscott should the reason be `AllReady` according to your proposal?

n3wscott: the proposal was a copy/paste from knative and we impled slightly different

docs/design/proposals/status-conditions.md
matthchr: Im not sold that individual conditions per managed resource is actually all that useful, since as far as I can tell this condition already will give a reason `ResourcesNotReady` with a message `"failed to reconcile resource %s: %v", resourceID, err` (for example), which seems to cover at least the basic case.

Can you expand on your thinking for how users would use the individual conditions?

api/v1alpha1/condition_test.go
```diff
@@ -143,3 +146,161 @@ func TestCondition_IsUnknown(t *testing.T) {
...
+	originalCondition := Condition{
+		Type:               ConditionType("TestCondition"),
+		Status:             metav1.ConditionTrue,
+		Reason:             ptr.To("TestReason"),
+		Message:            ptr.To("Test message"),
+		LastTransitionTime: &now,
+		ObservedGeneration: 42,
+	}
+
+	t.Logf("Original condition: Type=%s, Status=%s, LastTransitionTime=%v, ObservedGeneration=%d",
```
matthchr, line 161: minor: You could add a `String` method to condition and then rely on that here (and possibly elsewhere in our codebase where we want to log the condition we're setting)?

pkg/controller/instance/controller_reconcile.go
```diff
@@ -427,10 +453,15 @@ func (igr *instanceGraphReconciler) finalizeDeletion(ctx context.Context) error
 		}
 	}
 
-	// Remove finalizer from instance
+	// All resources are deleted, mark as ready for finalization
 	instance := igr.runtime.GetInstance()
+	mark := NewConditionsMarkerFor(instance)
+	mark.ResourcesReady() // All resources have been successfully deleted
```
matthchr: This was somewhat unexpected to me. Insofar as the user will actually see it (because presumably it won't be visible for long given we're removing the finalizer at the same time as we actually write this). The fact the message here is: "all resources are created and ready" doesn't match reality?

pkg/graph/parser/parser_test.go
```diff
@@ -144,8 +144,8 @@ func TestParseResource(t *testing.T) {
 			{Path: "specialCharacters[\"doted.annotation.key\"]", Expressions: []string{"dotedannotationvalue"}, ExpectedTypes: []string{"string"}, StandaloneExpression: true},
 			{Path: "specialCharacters[\"\"]", Expressions: []string{"emptyannotation"}, ExpectedTypes: []string{"string"}, StandaloneExpression: true},
 			{Path: "specialCharacters[\"array.name.with.dots\"][0]", Expressions: []string{"value"}, ExpectedTypes: []string{"string"}, StandaloneExpression: true},
-			{Path: "schemalessField.something", Expressions: []string{"schemaless.value"}, ExpectedTypes: []string{"string"}, StandaloneExpression: true},
-			{Path: "schemalessField.nestedSomething.nested", Expressions: []string{"schemaless.nested.value"}, ExpectedTypes: []string{"string"}, StandaloneExpression: true},
+			{Path: "schemalessField.something", Expressions: []string{"schemaless.value"}, ExpectedTypes: []string{"any"}, StandaloneExpression: true},
```
matthchr, line 149: How is this change related to the rest of this PR?

a-hilaly: That's a great catch from Scott, i moved it to a seperate PR https://github.com/kubernetes-sigs/kro/pull/698.

docs/design/proposals/status-conditions.md
a-hilaly: @matthchr my thinking was that today kro follows a topological order, but I'd like to make it DFS or BFS based where multiple resources can be managed in parallel. In that world, individual conditions would help debug which resources failed concurrently. But yeah, for now it's not needed - we can iterate on this later. and +1 to your suggestion of having the error message point to the problematic resource

pkg/controller/instance/controller_reconcile.go
a-hilaly: @matthchr agreed, setting `ResourcesReady()` during deletion doesn't feel right

api/v1alpha1/condition_test.go
n3wscott: true, could be a nice followup

pkg/controller/instance/controller_reconcile.go
n3wscott: I had to move things around to help but in general the pattern I have preferred is the reconcile method is the only place you are really mutating the conditions and the helper methods relate to the mark methods. I find this makes the purpose of the helper method (setupInstance) tie to the marked conditions (GraphResolved) or that helper returns an error (mark.GraphNotResolved) giving the reconciler method a better flow for the reader. If it is all hidden in the setupInstance, it becomes unclear whos job it is to track changes to the resource under reconciliation. 

ideally we never pass the mark helper anywhere but to not refactor the code too badly, I had to pass it around a little. I would rather have refactored the code to keep it tight and clean.

docs/design/proposals/status-conditions.md
```diff
@@ -222,23 +288,29 @@ Graph Definition.
...
+  - [x] This adds ObservedGeneration to Conditions explicitly.
 - [x] [ResourceGraphDefinition] Watch and react to Resource Graph CRD changes,
       especially the “Accepted” status of a CRD.
-- [ ] [Instance] Document the large chunks of work that the reconciler does,
-      each of these will get a sub-resource.
+- [x] [Instance] Document the large chunks of work that the reconciler does,
+      each of these will get a sub-condition.
+  - [x] Identified three major phases: InstanceManaged, GraphResolved, ResourcesReady
+  - [x] Mapped reconciliation phases to specific sub-conditions
 - [ ] [Instance] Introduce a signal in the schema of provided conditions to
```
jakobmoellerdev, line 302: do we even still want this?

docs/design/proposals/status-conditions.md
```diff
@@ -222,23 +288,29 @@ Graph Definition.
...
       determine if a condition should be considered for Readiness or it is just
       informational.
   - [ ] We might also need to support defining the polarity of success, which is
```
jakobmoellerdev, line 305: similarly here, something we really want right now?

pkg/controller/instance/controller.go
jakobmoellerdev: I do not think it is smart to call update on the instance for the status and then not requeue immediately. This is because the enqueue will be triggered after this (Im basing that on the thought that we do not only react to generation change here)

pkg/controller/instance/controller_status.go
```diff
@@ -16,35 +16,151 @@ package instance
...
+var instanceConditionTypes = apis.NewReadyConditions(InstanceManaged, GraphResolved, ResourcesReady)
+
+// unstructuredConditionAdapter adapts an unstructured.Unstructured to implement apis.Object
+type unstructuredConditionAdapter struct {
+	*unstructured.Unstructured
+}
+
+// GetConditions implements apis.Object interface
+func (u *unstructuredConditionAdapter) GetConditions() []v1alpha1.Condition {
+	if conditions, found, err := unstructured.NestedSlice(u.Object, "status", "conditions"); err == nil && found {
```
jakobmoellerdev, line 47: this is missing an error return that should signal if we expected conditions but found none imho. silent fail is never nice, and just returning empty is probably unwanted right?

test/integration/suites/core/instance_conditions_test.go
```diff
@@ -0,0 +1,296 @@
...
+)
+
+var _ = Describe("Instance Conditions", func() {
+	var (
+		ctx       context.Context
+		namespace string
+	)
+
+	BeforeEach(func() {
+		ctx = context.Background()
```
jakobmoellerdev: please do not use context.Background but reuse SpecContext (see https://github.com/kubernetes-sigs/kro/pull/700 for motivation and example of change)

docs/design/proposals/status-conditions.md
jakobmoellerdev: I spoke to @a-hilaly on [this](https://github.com/kubernetes-sigs/kro/pull/622#discussion_r2384829339) recently: What is described above is to implement a form of Levelling Algorithm which splits the RGDs resource graph into a set of concurrently appliable / checkable layers:

Assuming we have a graph like

```
1      2
| \    |
|  \   |
v   v  v
3   4  |
 \ / \ |
  v   vv
  5    6
   \  / \
    vv   v
    7    8
```

we can label and level them as per the algorithm.
This results in 

```
L1:  1 
     | \ 
     |  \ 
L2:  3   4   2
     |  /|  /
     | / | /
L3:  5   6
     |  /|
     | / |
L4:  7   8
```

which is equivalent to a concurrently appliable, bottom-up resource graph (L4 has to be appllied before L3 and so on).
The levels coincide with applysets by the way as each level can be translated to one set of resources managed in an applyset concurrently by SSA.

There are some algorithms out there that already do this quite well such as Coffman–Graham. 

I dont think that outputting conditions for each level is useful without understanding the "slicing" process underneath.

barney-s: If at all we want to add resource level or coffman-graham-level subcondition it should be for cases where we saw a failure in the resource. During normal operation we dont want them in subconditions. Also beware of the object size limitation in etcd for k8s. We dont want status updates to fail because we exceeded object size. 

TBH we are too early for this pattern. Lets follow up in a separate bug/feature req for this.

test/e2e/chainsaw/check-rgd-deletion/instancecrd-assert.yaml
```diff
@@ -19,10 +19,9 @@ spec:
       jsonPath: .status.state
       name: State
       type: string
-    - description: Whether a ResourceGraphDefinition instance have all it's subresources
-        ready
-      jsonPath: .status.conditions[?(@.type=="InstanceSynced")].status
-      name: Synced
+    - description: Whether a ResourceGraphDefinition instance is ready
```
barney-s, line 22: Should this be CRD instance ? or just instance. 
This is definitely not a RGD.

test/e2e/chainsaw/check-rgd-creation/instancecrd-assert.yaml
```diff
@@ -19,10 +19,9 @@ spec:
       jsonPath: .status.state
       name: State
       type: string
-    - description: Whether a ResourceGraphDefinition instance have all it's subresources
-        ready
-      jsonPath: .status.conditions[?(@.type=="InstanceSynced")].status
-      name: Synced
+    - description: Whether a ResourceGraphDefinition instance is ready
```
barney-s, line 22: s/ResourceGraphDefinition/checkrgdcreationsimpledeployments ?

test/e2e/chainsaw/check-multi-resource-rgd/instancecrd-assert.yaml
```diff
@@ -19,10 +19,9 @@ spec:
       jsonPath: .status.state
       name: State
       type: string
-    - description: Whether a ResourceGraphDefinition instance have all it's subresources
-        ready
-      jsonPath: .status.conditions[?(@.type=="InstanceSynced")].status
-      name: Synced
+    - description: Whether a ResourceGraphDefinition instance is ready
```
barney-s, line 22: s/ResourceGraphDefinition/<crd-name> ?

test/integration/suites/core/instance_conditions_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+			// Check that RGD has Ready condition
+			var readyCondition *krov1alpha1.Condition
+			for _, cond := range rgd.Status.Conditions {
+				if cond.Type == "Ready" {
+					readyCondition = &cond
+					break
+				}
+			}
+			g.Expect(readyCondition).ToNot(BeNil(), "Ready condition should exist")
```
barney-s, line 94: we dont explicitly test for sub conditions. Is this intentional ?

test/integration/suites/core/instance_conditions_test.go
```diff
@@ -0,0 +1,296 @@
...
+					condType, _ := cond["type"].(string)
+					if condType == ctrlinstance.Ready {
+						readyCondition = cond
+						break
+					}
+				}
+			}
+
+			// Validate Ready condition
+			g.Expect(readyCondition).ToNot(BeNil(), "Ready condition should be present")
```
barney-s, line 141: ditto for subcondition testing

test/integration/suites/deploymentservice/suite_test.go
```diff
@@ -193,6 +193,29 @@ var _ = Describe("DeploymentService", func() {
...
+					condType, _ := cond["type"].(string)
+					if condType == ctrlinstance.Ready {
+						readyCondition = cond
+						break
+					}
+				}
+			}
+
+			// Validate Ready condition
+			g.Expect(readyCondition).ToNot(BeNil(), "Ready condition should be present")
```
barney-s, line 213: ditto for subcondition testing.

test/e2e/chainsaw/check-instance-update/instancecrd-assert.yaml
```diff
@@ -19,10 +19,9 @@ spec:
       jsonPath: .status.state
       name: State
       type: string
-    - description: Whether a ResourceGraphDefinition instance have all it's subresources
-        ready
-      jsonPath: .status.conditions[?(@.type=="InstanceSynced")].status
-      name: Synced
+    - description: Whether a ResourceGraphDefinition instance is ready
```
barney-s, line 22: s/ResourceGraphDefinition/<crd-name> ?

test/e2e/chainsaw/check-instance-creation/instancecrd-assert.yaml
```diff
@@ -19,10 +19,9 @@ spec:
       jsonPath: .status.state
       name: State
       type: string
-    - description: Whether a ResourceGraphDefinition instance have all it's subresources
-        ready
-      jsonPath: .status.conditions[?(@.type=="InstanceSynced")].status
-      name: Synced
+    - description: Whether a ResourceGraphDefinition instance is ready
```
barney-s, line 22: s/ResourceGraphDefinition/<crd-name> ?

test/e2e/chainsaw/check-arbitary-objects/instancecrd-assert.yaml
```diff
@@ -19,10 +19,9 @@ spec:
       jsonPath: .status.state
       name: State
       type: string
-    - description: Whether a ResourceGraphDefinition instance have all it's subresources
-        ready
-      jsonPath: .status.conditions[?(@.type=="InstanceSynced")].status
-      name: Synced
+    - description: Whether a ResourceGraphDefinition instance is ready
```
barney-s, line 22: s/ResourceGraphDefinition/<crd-name> ?

test/e2e/chainsaw/check-arbitary-objects/instance-create-assert.yaml
barney-s: Ready is a more conventional k8s condition.

pkg/graph/parser/parser_test.go
barney-s: should this change not appear now that #698 is merged ?

docs/design/proposals/status-conditions.md
```diff
@@ -198,18 +198,84 @@ Things to note:
...
-just informational.
+Based on analysis of the instance reconciler, the major sections of work can be
+broken down into the following sub-conditions that align with the reconciliation
+phases:
+
+```
+Ready
+├─ InstanceManaged - Instance finalizers and labels are properly set
+├─ GraphResolved - Runtime graph has been created and resources resolved
+└─ ResourcesReady - All resources in the graph are created and ready
```
barney-s, line 213: Havent checked the code completely. Are we checking for individual reasource readiness for this to be qualified as true ? Or just creation.

pkg/controller/instance/controller_status.go
```diff
@@ -16,35 +16,151 @@ package instance
...
+	cs apis.ConditionSet
+}
+
+// InstanceManaged signals the instance has proper finalizers and labels set.
+func (m *ConditionsMarker) InstanceManaged() {
+	m.cs.SetTrueWithReason(InstanceManaged, "Managed", "instance is properly managed with finalizers and labels")
+}
+
+// InstanceNotManaged signals there was an issue setting up the instance management.
+func (m *ConditionsMarker) InstanceNotManaged(format string, a ...any) {
```
barney-s, line 111: Not sure if this is convention. Do we want to rename this to InstanceNotManagedf() ? to indicate it takes a format string and vaargs ?

pkg/controller/instance/controller.go
```diff
@@ -272,3 +282,31 @@ func getServiceAccountUserName(namespace, serviceAccount string) (string, error)
...
+			var conditionsInterface []interface{}
+			if err := json.Unmarshal(conditionsJSON, &conditionsInterface); err == nil {
+				status["conditions"] = conditionsInterface
+			}
+		}
+	}
+
+	// Update status - ignore errors as this is best effort
+	instance.Object["status"] = status
+	_, _ = c.clientSet.Dynamic().Resource(c.gvr).
```
barney-s: log on error ?

pkg/controller/instance/controller_status.go
```diff
@@ -16,35 +16,151 @@ package instance
...
+
+var instanceConditionTypes = apis.NewReadyConditions(InstanceManaged, GraphResolved, ResourcesReady)
+
+// unstructuredConditionAdapter adapts an unstructured.Unstructured to implement apis.Object
+type unstructuredConditionAdapter struct {
+	*unstructured.Unstructured
+}
+
+// GetConditions implements apis.Object interface
+func (u *unstructuredConditionAdapter) GetConditions() []v1alpha1.Condition {
```
barney-s, line 46: Is GetConditions used for testing only. Should we unexport this ?

pkg/controller/instance/controller_status.go
```diff
@@ -16,35 +16,151 @@ package instance
...
+		if err := json.Unmarshal(conditionsJSON, &result); err != nil {
+			return []v1alpha1.Condition{}
+		}
+		return result
+	}
+	return []v1alpha1.Condition{}
+}
+
+// SetConditions implements apis.Object interface
+func (u *unstructuredConditionAdapter) SetConditions(conditions []v1alpha1.Condition) {
```
barney-s, line 64: Is SetConditions used for testing only. Should we unexport this ?

pkg/controller/instance/controller.go
matthchr: > I do not think it is smart to call update on the instance for the status and then not requeue immediately. This is because the enqueue will be triggered after this (Im basing that on the thought that we do not only react to generation change here)

We spoke about this in the community call and I had missed seeing this earlier but do generally agree with this as well. 

When I initially read this I was thinking that this was the only status write and then we would be immediately requeuing due to the error return, but now that I look more in-depth there's another unconditional status write at `pkg/controller/resourcegraphdefinition/controller.go`. Should we just rely on that write (`updateStatus` call one level up) to propagate this status change?

Though, possibly worth noting that:
> This is because the enqueue will be triggered after this (Im basing that on the thought that we do not only react to generation change here)

We do set `WithEventFilter(predicate.GenerationChangedPredicate{}).` so I don't think we'll be getting double-events anyway, but even so I think @jakobmoellerdev's point stands.

test/e2e/chainsaw/check-multi-resource-rgd/instancecrd-assert.yaml
n3wscott: i have no idea what this comment means.

test/e2e/chainsaw/check-rgd-creation/instancecrd-assert.yaml
n3wscott: no

test/e2e/chainsaw/check-rgd-deletion/instancecrd-assert.yaml
n3wscott: it i an instance of an RGD the Kind created from the CRD that is named RGD. It is confusing

test/integration/suites/core/instance_conditions_test.go
n3wscott: no, dont test sub-conditions in integration testing. Subconditions are a controller impl detail

pkg/controller/instance/controller_status.go
n3wscott: I don't think so

pkg/controller/instance/controller.go
n3wscott: maybe as a followup

pkg/controller/instance/controller_status.go
n3wscott: not just for testing

pkg/controller/instance/controller_status.go
n3wscott: not just for testing

pkg/controller/instance/controller.go
```diff
@@ -135,6 +141,17 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) error {
...
+	// TODO(a-hilaly): client caching
+	executionClient, err := c.getExecutionClient(namespace)
+	if err != nil {
+		// Mark instance as not managed due to execution client failure
+		mark := NewConditionsMarkerFor(instance)
+		mark.InstanceNotManaged("failed to create execution client: %v", err)
+		c.updateInstanceStatusOnError(ctx, instance)
+		return fmt.Errorf("failed to create execution client: %w", err)
+	}
+
```
n3wscott:
```suggestion

```

pkg/controller/instance/controller.go
n3wscott: it happens in the reconciler.

n3wscott:
PTAL, if possible, feel free to update the branch for nits. I am highly time constrained at the moment and am traveling.

pkg/controller/instance/controller.go
n3wscott: I think `GenerationChangedPredicate` is an anti-pattern.

docs/design/proposals/status-conditions.md
n3wscott: they all need to be ready

n3wscott:
Rebased and addressed the comments that made sense. Unit tests passed locally. PTAL

a-hilaly (approved):
@jakobmoellerdev and I discussed about this offline - we're happy to merge this as is and do some general refactoring in the instance reconciler to make one status patch update.

Thanks a lot @n3wscott ! Hoping this lands in 0.5.0!

PR #624, opened 2025-08-08
Prefer to use interfaces for kube types to allow for unit testing controllers without real kube clients.

n3wscott:
Working on https://github.com/kro-run/kro/pull/622 I get a wall where I need to unit test some things but found that kro is storing the real kubeclient structs in controllers rather than interfaces. 

Don't store real kube structs, only interface types. It makes testing possible.

### DIFF
--- a/pkg/client/fake/json_safe_client.go
+++ b/pkg/client/fake/json_safe_client.go
@@ -1,0 +1,210 @@
+// Copyright 2025 The Kube Resource Orchestrator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fake
+
+import (
+	"context"
+	"encoding/json"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
+)
+
+// JSONSafeDynamicClient wraps a fake dynamic client and performs JSON marshal/unmarshal
+// to sanitize data before passing it to the underlying client, preventing deep copy issues
+// with pointer fields
+type JSONSafeDynamicClient struct {
+	client dynamic.Interface
+}
+
+// NewJSONSafeDynamicClient creates a new JSON-safe wrapper around a dynamic client
+func NewJSONSafeDynamicClient(client dynamic.Interface) *JSONSafeDynamicClient {
+	return &JSONSafeDynamicClient{
+		client: client,
+	}
+}
+
+// Resource returns a resource interface for the given resource
+func (j *JSONSafeDynamicClient) Resource(resource schema.GroupVersionResource) dynamic.NamespaceableResourceInterface {
+	return &jsonSafeNamespaceableResourceInterface{
+		resource: j.client.Resource(resource),
+	}
+}
+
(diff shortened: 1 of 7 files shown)


a-hilaly (approved):
👍 

`pkg/client` might be one of the worst ~software design~ decisions i've made in my life

pkg/client/fake/json_safe_client.go
```diff
@@ -0,0 +1,210 @@
...
+	sanitized, err := sanitizeUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+	return j.resource.ApplyStatus(ctx, name, sanitized, options)
+}
+
+// sanitizeUnstructured performs JSON marshal/unmarshal to sanitize unstructured data
+// This removes pointer references and makes the data safe for fake client deep copying
+func sanitizeUnstructured(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
```
matthchr: This is a deep copy, right?

maybe that's a better name for it?

pkg/client/set.go
```diff
@@ -25,6 +25,28 @@ import (
...
+
+	// CRD returns a new CRDInterface instance
+	CRD(cfg CRDWrapperConfig) CRDInterface
+
+	// WithImpersonation returns a new client that impersonates the given user
+	WithImpersonation(user string) (SetInterface, error)
+}
+
 // Set provides a unified interface for different Kubernetes clients
 type Set struct {
```
barney-s, line 51: should this be made non-exportable ? 
`type set`

pkg/client/fake/json_safe_client.go
```diff
@@ -0,0 +1,210 @@
...
+
+// JSONSafeDynamicClient wraps a fake dynamic client and performs JSON marshal/unmarshal
+// to sanitize data before passing it to the underlying client, preventing deep copy issues
+// with pointer fields
+type JSONSafeDynamicClient struct {
+	client dynamic.Interface
+}
+
+// NewJSONSafeDynamicClient creates a new JSON-safe wrapper around a dynamic client
+func NewJSONSafeDynamicClient(client dynamic.Interface) *JSONSafeDynamicClient {
```
matthchr, line 37: My assumption is that you will pass an actual fake client impl here, right?

matthchr (approved):

pkg/client/set.go
n3wscott: no, it is fine to export, the constructor returns you a Set* but we store it in controllers as a SetInterface.

pkg/client/fake/json_safe_client.go
n3wscott: correct

pkg/client/fake/json_safe_client.go
n3wscott: it is doing a deepcopy but it is attempting to remove `string*` from the unstructured so it does a bit more. This only matters for the fake dynamic client, it does not understand how to deep copy a `string*` in the conditions

matthchr: it accomplishes that just be doing a full deepcopy, no?

Not sure I am following how there is "special `*string` logic here". I think what you're saying is that the actual fake dynamic client has logic that attempts to deep copy (but not like this), which misses ptr-string in the conditions.

Which TBH feels like an issue with that fake impl, but I'm open to having this here because presumably fixing upstream is nontrivial (are they aware of that problem and/or do you have a link to where that problem arises from?)

barney-s (approved):

pkg/client/fake/json_safe_client.go
n3wscott: note this only matters for the fake dynamic client.

n3wscott: as I understand it, the real client does not have this issue because the unstructured is converted to json. But the fake version is trying to keep an in-memory version so it tried to deep copy each field inside the unstructured and gets to `string*` and gets confused.

matthchr: Right that makes sense. My question was specifically for the fake impl - you'd think they would deepCopy in the fake impl to save us having to do this was my point.

n3wscott: renamed to `deepCopyUnstructured`

matthchr (approved):

pkg/client/fake/json_safe_client.go
n3wscott: I don't think they do because the unstructured object does not have the member structs that have deepcopy attached and they do a reflection based copy field by field.

n3wscott:
> `pkg/client` might be one of the worst ~software design~ decisions i've made in my life

*worse decisions you have made in your life, so far. 😉

PR #627, opened 2025-08-10
feat: Add support for passing unstructured data to CRDs managed by KRO

Moortiii:
~~This PR allows users to specify 'object' as a type when defining their spec, spec alongside a new marker type 'MarkerTypePreserve'. This annotates the key with  `x-kubernetes-preserve-unknown-fields: true` and allows us to pass unstructured data, such as values for Helm charts, to CRDs managed by KRO~~

This PR allows users to specify 'object' as a type when defining their spec. This annotates the key with  `x-kubernetes-preserve-unknown-fields: true` and allows us to pass unstructured data, such as values for Helm charts, to CRDs managed by KRO.

### DIFF
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/utils/ptr"
 )
 
 const (
@@ -28,6 +29,7 @@ const (
 	keyTypeInteger = string(AtomicTypeInteger)
 	keyTypeBoolean = string(AtomicTypeBool)
 	keyTypeNumber  = "number"
+	keyTypeObject  = "object"
 )
 
 // A predefinedType is a type that is predefined in the schema.
@@ -128,6 +130,9 @@ func (tf *transformer) parseFieldSchema(key, fieldValue string, parentSchema *ex
 
 	if isAtomicType(fieldType) {
 		fieldJSONSchemaProps.Type = fieldType
+	} else if fieldType == keyTypeObject {
+		fieldJSONSchemaProps.Type = fieldType
+		fieldJSONSchemaProps.XPreserveUnknownFields = ptr.To(true)
 	} else if isCollectionType(fieldType) {
 		if isMapType(fieldType) {
 			fieldJSONSchemaProps, err = tf.handleMapType(key, fieldType)
--- a/test/e2e/chainsaw/check-arbitary-objects/chainsaw-test.yaml
+++ b/test/e2e/chainsaw/check-arbitary-objects/chainsaw-test.yaml
@@ -1,0 +1,43 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: check-arbitrary-objects
+spec:
+  description: |
+    Tests if an instance with a key of type 'object' allows passing unstructured data
+  steps:
+  - name: install-rgd
+    try:
+    #description: Install the RGD that creates an Instance CRD
+    - apply:
+        file: rgd.yaml
+      description: Apply RGD
+    - assert:
+        file: rgd-assert.yaml
+      description: Verify RGD state
+    - assert:
+        file: instancecrd-assert.yaml
+      description: Verify Instance CRD state
+    catch:
+    - description: kro controller pod logs collector
+      podLogs:
+        selector: app.kubernetes.io/instance=kro
+        namespace: kro-system
+    finally:
+    - description: sleep operation
+      sleep:
+        duration: 5s
+  - name: create-instance
+    try:
+    #description: Create instance
+    - apply:
+        file: instance.yaml
+      description: Create ArbitraryObjectNoOp Instance
+    - assert:
+        file: instance-create-assert.yaml
+      description: Verify Instance state
+    catch:
+    - description: kro controller pod logs collector
+      podLogs:
+        selector: app.kubernetes.io/instance=kro
+        namespace: kro-system
--- a/test/e2e/chainsaw/check-arbitary-objects/rgd.yaml
+++ b/test/e2e/chainsaw/check-arbitary-objects/rgd.yaml
@@ -1,0 +1,12 @@
+apiVersion: kro.run/v1alpha1
+kind: ResourceGraphDefinition
+metadata:
+  name: check-arbitrary-objects
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: CheckArbitraryObjectsNoOp
+    spec:
+      values: object | required=true
+  resources: []
+
--- a/website/docs/docs/concepts/10-simple-schema.md
+++ b/website/docs/docs/concepts/10-simple-schema.md
@@ -23,6 +23,9 @@ spec:
       replicas: integer | default=1 minimum=1 maximum=100
       image: string | required=true
 
+      # Unstructured objects
+      values: object | required=true
+
       # Structured type
       ingress:
         enabled: boolean | default=false
@@ -86,6 +89,46 @@ user:
     street: string
     city: string
   contacts: "[]string" # Array of strings
+```
+
+### Unstructured Objects
+
+Unstructured objects are declared using `object` as a type.
+
+:::warning
+
+This disables the field-validation normally offered by kro, and forwards the values to your RGD as-is. This is generally discouraged and should therefore be used with caution. In most cases, using a structured object is a better approach.
+
+:::
+
+```yaml
+kind: ResourceGraphDefintion
+metadata: {}
+spec:
+  schema:
+    spec:
+      additionalHelmChartValues: object
+```
+
+This allows you to pass data to your CRDs directly in cases where the schema is not known in advance. This type supports any valid object, and can mix and match different primitives as well as structured types.
+
+```yaml
+apiVersion: kro.run/v1alpha1
+kind: CRDWithUnstructuredObjects
+metadata:
+  name: test-instance
+spec:
+  additionalHelmChartValues:
+    boolean-value: true
+    numeric-value: 42
+    structural-type:
+      with-additional:
+        nested: fields
+    string-value: my-string
+    mapping-value:
+      - item1
+      - item2
+      - item3
 ```
 
 ### Array Types
@@ -228,7 +271,7 @@ schema:
   spec:
     image: string | default="nginx"
   status:
-      availableReplicas: ${deployment.status.availableReplicas}
+    availableReplicas: ${deployment.status.availableReplicas}
   additionalPrinterColumns:
     - jsonPath: .status.availableReplicas
       name: Available replicas
--- a/pkg/simpleschema/transform_test.go
+++ b/pkg/simpleschema/transform_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/utils/ptr"
 )
 
 func TestBuildOpenAPISchema(t *testing.T) {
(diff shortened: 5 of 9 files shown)


pkg/simpleschema/markers.go
```diff
@@ -55,12 +55,14 @@ const (
 	MarkerTypeValidation MarkerType = "validation"
 	// MarkerTypeEnum represents the `enum` marker.
 	MarkerTypeEnum MarkerType = "enum"
+	// MarkerTypePreserve represents the 'x-kubernetes-preserve-unknown-fields' marker.
+	MarkerTypePreserve MarkerType = "x-kubernetes-preserve-unknown-fields"
```
a-hilaly: Does this this mean that users will have to author something like:
```yaml
spec:
  field: object | x-kubernetes-preserve-unknown-fields=true
```

a-hilaly: How about using `any` as a type instead?
```yaml
spec:
    field: any | required=true default={"a": "b"}
```

Moortiii: I'm open to that!

My initial idea was that users looking for this might be familiar with the CRD syntax. Let me know if you would prefer the use of `any` instead.

Moortiii:
@jakobmoellerdev Added a test case for both the happy path and the less-than-happy path now. Does that suffice, or is additional coverage needed?

Moortiii:
After a bit more experimentation, I think @a-hilaly was correct in that using `any` as a type feels more ergonomic, and is simpler to use in combination with other existing markers. I've modified the PR and updated the tests to accomodate this.

Let me know if you are satisfied, and I'll squash the commits into one. I'm keeping both for now in case you want me to revert back to the previous implementation.

pkg/simpleschema/markers.go
Moortiii: I agree. After some testing this seems much more ergonomic. The `object` type could be mistaken for a way to represent structured data, and made no sense on its own if a user forgot to add the preserve marker. Using `any` allows it to be combined with existing markers cleanly, and avoids all of this confusion.

jakobmoellerdev: one issue I have with any is that any really is "any object", because preserve unknown is only valid for objects. is any also valid for primitive types?

Moortiii: Since there is no schema available for the object, the API server would reject all CRs that use `object` as a type, unless explicitly instructed not to. What about removing the marker, sticking with `object` and simply enforcing it behind the scenes? I.e:

```go
values: object // valid
values: object | required=true // valid
values: object | x-kubernetes-preserve-unknown-fields // unknown marker
values: 'object | required=true default={"a": "b"}' // valid
```

Moortiii: The latest commit now uses `object` and no longer requires (or allows, even)  the user to specify that unknown fields should be preserved, as that is now the default behavior. Verified that the solution works with existing markers, such as `default` and `required`.

I wanted to add information to the documentation in the same PR, but it appears that that may be a different project altogether?

pkg/simpleschema/transform.go
```diff
@@ -128,6 +130,9 @@ func (tf *transformer) parseFieldSchema(key, fieldValue string, parentSchema *ex
 
 	if isAtomicType(fieldType) {
 		fieldJSONSchemaProps.Type = fieldType
+	} else if fieldType == keyTypeObject {
```
jakobmoellerdev, line 133: looking at this if statement again, im not sure if we should check for object equality before we check for predefined types. is it possible that preDefinedTypes contains object?

jakobmoellerdev (approved):
looking good now, i think with object we are more in line how kubernetes treats this case too. @a-hilaly would be cool if you could take another look now

pkg/simpleschema/markers.go
jakobmoellerdev: ~To second reviewer: Im not sure if doing unknown field preservation by default is a good idea. Looking for external opinions.~ see https://github.com/kro-run/kro/pull/627#issuecomment-3180938852

Moortiii:
Hmm, what do you mean by changing CRDs? Attempting to use 'object' as a type before this PR would throw an error about it being an unknown type. So no existing RGDs should be using it as of right now.

I'll look into writing an end-to-end test. This is my first meeting with Go, so I'm sort of stumbling my way through it for now.

pkg/simpleschema/transform.go
Moortiii: No, `object` is not currently part of the predefined types. The following snippet in the check for predefined types throws the error `unknown type: object` unless we explicitly check for objects beforehand:

```go
preDefinedType, ok := tf.preDefinedTypes[fieldType]
if !ok {
	return nil, fmt.Errorf("unknown type: %s", fieldType)
}
```

Moortiii:
@jakobmoellerdev Added an end-to-end test that verifies the creation of a noop instance whose spec contains a key of type object with arbitrary fields.

Ashish7812:
@Moortiii Can you add some documentation on this new field type here https://kro.run/docs/concepts/simple-schema?

Moortiii:
@Ashish7812 Wrote an initial docs draft. I was debating if we should include an example of an instance with unstructured objects or not, but opted to include it. Let me know what you think.

jakobmoellerdev:
@a-hilaly PTAL: I think this is ready for another pass by you

pkg/simpleschema/transform.go
a-hilaly: that's a good point, i think we shouldn't allow preDefinedTypes to declare any builtin types (int, string, object, map ...) - maybe in a seperate PR we can add some validation to `loadPredefinedTypes`.

website/docs/docs/concepts/10-simple-schema.md
```diff
@@ -88,6 +91,35 @@ user:
   contacts: "[]string" # Array of strings
 ```
 
+### Unstructured Objects
+
+Unstructured objects are declared using `object` as a type.
+
+```yaml
+additionalHelmChartValues: object
```
a-hilaly: nit:
```suggestion
kind: ResourceGraphDefintion
metadata: {}
spec:
  schema:
    spec:
      additionalHelmChartValues: object
```

website/docs/docs/concepts/10-simple-schema.md
```diff
@@ -228,7 +260,7 @@ schema:
   spec:
     image: string | default="nginx"
   status:
-      availableReplicas: ${deployment.status.availableReplicas}
+    availableReplicas: ${deployment.status.availableReplicas}
```
a-hilaly, line 274: 👍

Moortiii:
@a-hilaly and @jakobmoellerdev Introduced the requested nit, and squashed the changes into a single commit. I think we should be good to go now! :-)

test/e2e/chainsaw/check-arbitary-objects/rgd.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kro.run/v1alpha1
+kind: ResourceGraphDefinition
+metadata:
+  name: check-arbitrary-objects
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: CheckArbitraryObjectsNoOp
+    spec:
+      values: object | required=true
```
barney-s, line 10: How would the object be used in a resource. Could you provide an example  ?

would it be `spec.values` OR `spec.values.something` ?

Moortiii: You could do either, but I believe the former makes more sense than the latter in most cases. Accessing an attribute from the passed object is possible, but you would have no way to verify or guarantee that the user has provided the key you want to access. In that case, using a structured type with a predefined schema seems like a better solution.

To give a concrete example, I wanted to facilitate the creation of ephemeral environments in my organization, similar to what is described [by the Flux operator team here](https://fluxcd.control-plane.io/operator/resourcesets/github-pull-requests/). In my own implementation, this requires the configuration of 18 different resources. By leveraging kro, we could create an abstraction, i.e. `EphemeralEnvironment`, thereby reducing repetition, and accounting for the fact that we might need to configure additional or fewer resources in the future.

Unfortunately, one of the resource we need to configure is a [`HelmRelease`](https://fluxcd.io/flux/components/helm/helmreleases/). The HelmRelease resource can take in arbitrary inputs via the field `spec.values`.  Currently, kro does not provide a way to pass unstructured data to the RGD, and as a result I don't have a way to let my users specify a value for this field. Forcing the underlying Helm charts to use a shared schema for the values object is not realistic or desired.

This PR allows us to implement the following:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: ephemeral-environment
spec:
  schema:
    apiVersion: v1alpha1
    kind: EphemeralEnvironment
    spec:
      additionalHelmChartValues: object

  resources:
    - id: resourcesetinputprovider
      template:
        apiVersion: fluxcd.controlplane.io/v1
        kind: ResourceSetInputProvider
        ...

    - id: resourceset
      template:
        apiVersion: fluxcd.controlplane.io/v1
        kind: ResourceSet
        resources:
          - apiVersion: helm.toolkit.fluxcd.io/v2
            kind: HelmRelease
            spec:
              values: ${schema.spec.additionalHelmChartValues}
          ...
    ...
```

We can then create an instance as such:

```yaml
apiVersion: kro.run/v1alpha1
kind: EphemeralEnvironment
metadata:
  namespace: default
spec:
  additionalHelmChartValues:
    slug: << inputs.branch >> # <- This is accessing a field from the `ResourceSetInputProvider`
    tag: << inputs.sha >> # <- This is accessing a field from the `ResourceSetInputProvider`
    some:
      deeply:
        nested: value-required-by-the-chart
```

We have no idea what the schema for the underlying Helm chart is going to look like.  For example, some charts might use `{{ .Values.tag }} ` with a fixed image when defining a deployment. Another chart might make it possible to swap out the entire image using `{{ .Values.image }}`.

As a sidenote, the example looks a bit strange, because both `ResourceGraphDefinition` and `ResourceSet` has the ability to configure multiple resources. Inputs from the `ResourceSetInputProvider`  is accessed using the `<< >>` notation, as shown above and is necessary for this specific usecase.

test/e2e/chainsaw/check-arbitary-objects/chainsaw-test.yaml
```diff
@@ -0,0 +1,43 @@
...
+kind: Test
+metadata:
+  name: check-arbitrary-objects
+spec:
+  description: |
+    Tests if an instance with a key of type 'object' allows passing unstructured data
+  steps:
+  - name: install-rgd
+    try:
+    #description: Install the RGD that creates an Instance CRD
```
matthchr, line 11: Are these intentionally left as descriptions here, or were commented out in favor of the actual descriptions below and just not deleted?

Seems like it might be the latter?

website/docs/docs/concepts/10-simple-schema.md
```diff
@@ -88,6 +91,40 @@ user:
   contacts: "[]string" # Array of strings
 ```
 
+### Unstructured Objects
+
+Unstructured objects are declared using `object` as a type.
```
matthchr, line 96: Wondering if we want a warning somewhere in here that we don't recommend using unstructured objects unless you really have to, as doing so loses you some of the validation that you would otherwise get with kro

matthchr (approved):
LGTM -- had a few nits/questions.

website/docs/docs/concepts/10-simple-schema.md
barney-s: Plus 1 to this. For an user creating the instance, they need to know the RGD spec to understand what values need to be passed. Leaky abstraction. Can be generally discouraged.

Moortiii: I agree, it is a leaky abstraction and I would also generally advise against it. Unfortunately, for my specific use case and similar use cases, I don't see a way around it. What do you think about the following amendment?

```
### Unstructured Objects

Unstructured objects are declared using `object` as a type. This disables the field-validation normally offered by kro, and forwards the values to your RGD as-is. This is generally discouraged and should therefore be used with caution. In most cases, using a structured object is a better approach.
```

test/e2e/chainsaw/check-arbitary-objects/chainsaw-test.yaml
Moortiii: Ah, this is the result of a poor copy-paste job. I'd never written a chainsaw test, and therefore used the existing test `check-instance-creation` as a template for my own test.

I figured these descriptions were left there intentionally as an overarching description for the overall set of instructions carried out by each `step` in `spec.steps`. They can be removed, if desired.

website/docs/docs/concepts/10-simple-schema.md
matthchr: Yeah, something like that sounds good to me. Only note is we support `:::warning` and we could also put it in one of those. Reference [here](https://github.com/kro-run/kro/blob/1490e520efad1e89b46f00234d0f7e8577a8ce6d/website/docs/docs/concepts/00-resource-group-definitions.md?plain=1#L275C1-L275C11), which might make it a bit more obvious/pop more.

It renders like you see [here](https://kro.run/docs/concepts/resource-group-definitions#status-reporting)

Moortiii: The amendment is now represented as a warning, @matthcr in 233c7c7385a2d42a07249e49c6d2235d9d79414e. Need another workflow run, but should be good to go!

barney-s:
Thanks @Moortiii

Moortiii:
Thanks to everyone for helpful reviews and discussions as well!

PR #628, opened 2025-08-13
Implement kro CEL library for maps

gfrey:
Only function right now is merge. Potentially there are more. This implementation is inspired by the CEL lists extension.

### DIFF
--- a/pkg/cel/library/maps.go
+++ b/pkg/cel/library/maps.go
@@ -1,0 +1,106 @@
+package library
+
+import (
+	"math"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
+)
+
+// Maps returns a cel.EnvOption to configure extended functions for map manipulation.
+//
+// # Merge
+//
+// Merges two maps. Keys from the first map take precedence over keys in the second map.
+//
+//	map(string, int).merge(map(string, int)) -> map(string, int)
+//
+// Examples:
+//
+//	{}.merge({}) == {}
+//	{}.merge({'a': 1}) == {'a': 1}
+//	{'a': 1}.merge({}) == {'a': 1}
+//	{'a': 1}.merge({'b': 2}) == {'a': 1, 'b': 2}
+//	{'a': 1}.merge({'a': 2, 'b': 2}) == {'a': 1, 'b': 2}
+func Maps(options ...MapsOption) cel.EnvOption {
+	l := &mapsLib{version: math.MaxUint32}
+	for _, opt := range options {
+		opt(l)
+	}
+	return cel.Lib(l)
+}
+
+type mapsLib struct {
+	version uint32
+}
+
+type MapsOption func(*mapsLib) *mapsLib
+
+// LibraryName implements the cel.SingletonLibrary interface method.
+func (l *mapsLib) LibraryName() string {
+	return "cel.lib.ext.kro.maps"
+}
+
+// CompileOptions implements the cel.Library interface method.
+func (l *mapsLib) CompileOptions() []cel.EnvOption {
+	mapType := cel.MapType(cel.TypeParamType("K"), cel.DynType)
+	// mapDynType := cel.MapType(cel.DynType, cel.DynType)
+	opts := []cel.EnvOption{
+		cel.Function("merge",
+			cel.MemberOverload("map_merge",
+				[]*cel.Type{mapType, mapType},
+				mapType,
+				cel.BinaryBinding(func(arg1, arg2 ref.Val) ref.Val {
+					self, ok := arg1.(traits.Mapper)
+					if !ok {
+						return types.ValOrErr(arg1, "no such overload: %v.merge(%v)", arg1.Type(), arg2.Type())
+					}
+					other, ok := arg2.(traits.Mapper)
+					if !ok {
+						return types.ValOrErr(arg1, "no such overload: %v.merge(%v)", arg1.Type(), arg2.Type())
+					}
+					result, err := merge(self, other)
+					if err != nil {
+						return types.WrapErr(err)
+					}
+					return result
+				}),
+			),
+		),
+	}
+	return opts
+}
+
+// ProgramOptions implements the cel.Library interface method.
+func (l *mapsLib) ProgramOptions() []cel.ProgramOption {
+	return []cel.ProgramOption{}
+}
+
+// merge merges two maps. Keys from the first map take precedence over keys in
+// the second map.
+func merge(self traits.Mapper, other traits.Mapper) (traits.Mapper, error) {
+	var result traits.MutableMapper
+
+	if mapVal, ok := self.Value().(map[ref.Val]ref.Val); ok {
+		result = types.NewMutableMap(types.DefaultTypeAdapter, mapVal)
+	} else {
+		result = types.NewMutableMap(types.DefaultTypeAdapter, nil)
+		for i := self.Iterator(); i.HasNext().(types.Bool); {
+			k := i.Next()
+			v := self.Get(k)
+			result.Insert(k, v)
+		}
+	}
+
+	for i := other.Iterator(); i.HasNext().(types.Bool); {
+		k := i.Next()
+		if result.Contains(k).(types.Bool) {
+			continue
+		}
+		v := other.Get(k)
+		result.Insert(k, v)
+	}
+	return result.ToImmutableMap(), nil
+}
--- a/pkg/cel/library/maps_test.go
+++ b/pkg/cel/library/maps_test.go
@@ -1,0 +1,77 @@
+package library
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/google/cel-go/cel"
+)
+
+func TestMaps(t *testing.T) {
+	mapsTests := []struct {
+		expr string
+		err  string
+	}{
+		{expr: `{}.merge({}) == {}`},
+		{expr: `{}.merge({'a': 1}) == {'a': 1}`},
+		{expr: `{'a': 1}.merge({}) == {'a': 1}`},
+		{expr: `{'a': 1}.merge({'b': 2}) == {'a': 1, 'b': 2}`},
+		{expr: `{'a': 1}.merge({'a': 2, 'b': 2}) == {'a': 1, 'b': 2}`},
+		{expr: `{'a': 1}.merge({'b': 2.1}) == {'a': 1, 'b': 2.1}`},
+		{expr: `{'a': 1}.merge({'b': 'foo'}) == {'a': 1, 'b': 'foo'}`},
+		{expr: `{'a': 1}.merge({'b': {'c': 2}}) == {'a': 1, 'b': {'c': 2}}`},
+
+		// {expr: `{}.merge([])`, err: "ERROR: <input>:1:9: found no matching overload for 'merge' applied to 'map(dyn, dyn).(list(dyn))'"},
+	}
+
+	env := testMapsEnv(t)
+	for i, tc := range mapsTests {
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			var asts []*cel.Ast
+			pAst, iss := env.Parse(tc.expr)
+			if iss.Err() != nil {
+				t.Fatalf("env.Parse(%v) failed: %v", tc.expr, iss.Err())
+			}
+			asts = append(asts, pAst)
+			cAst, iss := env.Check(pAst)
+			if iss.Err() != nil {
+				t.Fatalf("env.Check(%v) failed: %v", tc.expr, iss.Err())
+			}
+			asts = append(asts, cAst)
+
+			for _, ast := range asts {
+				prg, err := env.Program(ast)
+				if err != nil {
+					t.Fatalf("env.Program() failed: %v", err)
+				}
+				out, _, err := prg.Eval(cel.NoVars())
+				if tc.err != "" {
+					if err == nil {
+						t.Fatalf("got value %v, wanted error %s for expr: %s",
+							out.Value(), tc.err, tc.expr)
+					}
+					if !strings.Contains(err.Error(), tc.err) {
+						t.Errorf("got error %v, wanted error %s for expr: %s", err, tc.err, tc.expr)
+					}
+				} else if err != nil {
+					t.Fatal(err)
+				} else if out.Value() != true {
+					t.Errorf("got %v, wanted true for expr: %s", out.Value(), tc.expr)
+				}
+			}
+		})
+	}
+}
+
+func testMapsEnv(t *testing.T, opts ...cel.EnvOption) *cel.Env {
+	t.Helper()
+	baseOpts := []cel.EnvOption{
+		Maps(),
+	}
+	env, err := cel.NewEnv(append(baseOpts, opts...)...)
+	if err != nil {
+		t.Fatalf("cel.NewEnv(Maps()) failed: %v", err)
+	}
+	return env
+}
--- a/pkg/cel/environment.go
+++ b/pkg/cel/environment.go
@@ -57,6 +57,7 @@ func DefaultEnvironment(options ...EnvOption) (*cel.Env, error) {
 		cel.OptionalTypes(),
 		ext.Encoders(),
 		library.Random(),
+		library.Maps(),
 	}
 
 	opts := &envOptions{}


a-hilaly:
@jakobmoellerdev i think that's a great point for our next community meeting conversaiton. My mental model is that we'll need both kubernetes upstream CEL env combined with some KRO specific CEL functions

gfrey:
@a-hilaly this lint error is weird. `merge` is implementing a function type for the `cel.Function`, so the "(error) is always nil (unparam)" problem is not something I think I can/want to change.
I'm a noob to GH actions, but the golangci-lint action seems to use v6 whilst v8 would be latest. Maybe that resolves the problem?

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/common/types/traits"
+)
+
+// Maps returns a cel.EnvOption to configure extended functions for map manipulation.
+//
+// # Merge
+//
+// Merges two maps. Keys from the first map take precedence over keys in the second map.
```
a-hilaly: `Keys from the first map take precedence over keys in the second map.` - is this a common behaviour? i have the impression that most languages/libraries do the opposite - where the second map's values overwrite the first one. i.e in python
```python
a = {'x':1}
b= {'x':2}
print(**a, **b) # prints {'x':2}
```

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,106 @@
...
+//
+//	map(string, int).merge(map(string, int)) -> map(string, int)
```
a-hilaly: does this mean that `.merges` won't work with `map(string, string)` types?

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,106 @@
...
+type mapsLib struct {
+	version uint32
+}
+
+type MapsOption func(*mapsLib) *mapsLib
+
+// LibraryName implements the cel.SingletonLibrary interface method.
+func (l *mapsLib) LibraryName() string {
+	return "cel.lib.ext.kro.maps"
+}
```
a-hilaly, line 59: shall we edit the library in name for `random` to match this one and similar patterns in cel-go? https://github.com/kro-run/kro/blob/main/pkg/cel/library/random.go#L53-L55 - https://github.com/google/cel-go/blob/master/ext/lists.go

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,106 @@
...
+
+// CompileOptions implements the cel.Library interface method.
+func (l *mapsLib) CompileOptions() []cel.EnvOption {
+	mapType := cel.MapType(cel.TypeParamType("K"), cel.DynType)
+	// mapDynType := cel.MapType(cel.DynType, cel.DynType)
```
a-hilaly:
```suggestion
	mapType := cel.MapType(cel.TypeParamType("K"), cel.TypeParamType("V")
```

I'm wondering if we should restrict this function to only act on maps of similar types? Like `map[string]int` only merging with `map[string]int`. Users could leverage macros to transform map types if needed. My highlevel toughts here is that i'd rather restrict this now and relax it later, than try to make a generic function and hit tricky type situations.

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,106 @@
...
+				mapType,
+				cel.BinaryBinding(func(arg1, arg2 ref.Val) ref.Val {
+					self, ok := arg1.(traits.Mapper)
+					if !ok {
+						return types.ValOrErr(arg1, "no such overload: %v.merge(%v)", arg1.Type(), arg2.Type())
+					}
+					other, ok := arg2.(traits.Mapper)
+					if !ok {
+						return types.ValOrErr(arg1, "no such overload: %v.merge(%v)", arg1.Type(), arg2.Type())
+					}
```
a-hilaly: i wonder if the `!ok` checks are redundant here - doesn't CEL already validate argument types?


PS: I see what we're doing in https://github.com/kro-run/kro/blob/main/pkg/cel/library/random.go#L53-L55 and i wonder if we should fix that too :)

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// ProgramOptions implements the cel.Library interface method.
+func (l *mapsLib) ProgramOptions() []cel.ProgramOption {
+	return []cel.ProgramOption{}
+}
+
+// merge merges two maps. Keys from the first map take precedence over keys in
+// the second map.
+func merge(self traits.Mapper, other traits.Mapper) (traits.Mapper, error) {
```
a-hilaly: golangcilint is complaining here because `merge` never returns a non-nil error (which makes it a redundant parameter)

pkg/cel/library/maps.go
gfrey: Yes, I was thinking the same, when picking this up again a few weeks back. It was motivated by the implementation, which it shouldn't be (cel map implementation doesn't seem to allow to replace values). I can fix that easily.

pkg/cel/library/maps.go
gfrey: No, it takes string keys, but values can be anything. I'll change `int` to `any`, unless you have a better idea.

pkg/cel/library/maps.go
gfrey: I've just matched the patterns ;). So, I think changing `random` would make sense from that perspective.

pkg/cel/library/maps.go
gfrey: It gets worse. It allows memory to grow, which undermines some of the core CEL assumptions, if I understand that correctly.
~~Despite all of this, my motivation is to get from a list of strings (say a list of availability zones to a list of subnet configurations that I can use for setting up a VPC. This would require this feature, but also the ability to merge maps of arbitrary types (not sure the suggested `cel.TypeParamType("V")` would allow for that).~~
Upon reflecting on that for some more, I agree. Let's start simple and extend if necessary. We know it is possible, but not need to go there immediately. I'll try to build it with pinned types.

pkg/cel/library/maps.go
gfrey: D'oh, my bad. Removed the unused error return value.

pkg/cel/library/maps.go
gfrey: I think you're right, at least in [the test added](https://github.com/kro-run/kro/pull/628/files#diff-4ecda9ea9392d19c864b41309dc68faebaa99054f334c8bead047920aa093473R36), the error is found prior to any execution in the `env.Check` call.

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,103 @@
+package library
```
a-hilaly, line 15: nit: missing license header

pkg/cel/library/maps_test.go
```diff
@@ -0,0 +1,76 @@
+package library
```
a-hilaly, line 15: ditto

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,103 @@
...
+
+	for i := self.Iterator(); i.HasNext().(types.Bool); {
+		k := i.Next()
+		if result.Contains(k).(types.Bool) {
+			continue
+		}
+		v := self.Get(k)
+		result.Insert(k, v)
+	}
+	return result.ToImmutableMap()
```
a-hilaly: i'm thinking, wouldn't something like this just work?
```go
    // populate with self map
    for it := self.Iterator(); it.HasNext() == types.True; {
        k := it.Next()
        result.Insert(k, self.Get(k))
    }
    
    // overwrides with other map
    for it := other.Iterator(); it.HasNext() == types.True; {
        k := it.Next()
        result.Insert(k, other.Get(k))
    }
```

a-hilaly:
Cc @jakobmoellerdev

pkg/cel/library/maps.go
jakobmoellerdev: @a-hilaly we should maybe write a linter for new files to avoid this. WDYT?

pkg/cel/library/maps.go
jakobmoellerdev: +1 for both of these. Let's change up the random lib name in a separate mini PR. Shouldnt block anything here though since its mostly for conflict resolution

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,103 @@
...
+// CompileOptions implements the cel.Library interface method.
+func (l *mapsLib) CompileOptions() []cel.EnvOption {
+	mapType := cel.MapType(cel.TypeParamType("K"), cel.TypeParamType("V"))
+	// mapDynType := cel.MapType(cel.DynType, cel.DynType)
+	opts := []cel.EnvOption{
+		cel.Function("merge",
+			cel.MemberOverload("map_merge",
+				[]*cel.Type{mapType, mapType},
+				mapType,
+				cel.BinaryBinding(func(arg1, arg2 ref.Val) ref.Val {
```
jakobmoellerdev: nit: expose this function separately instead of inlining, and then unit test this function as such instead of having to go through the environment

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// ProgramOptions implements the cel.Library interface method.
+func (l *mapsLib) ProgramOptions() []cel.ProgramOption {
+	return []cel.ProgramOption{}
+}
+
+// merge merges two maps. Keys from the first map take precedence over keys in
+// the second map.
+func merge(self traits.Mapper, other traits.Mapper) traits.Mapper {
```
jakobmoellerdev: nit: this looks very complex (similar to https://github.com/kro-run/kro/pull/628/files#r2283120971 comment I looked into this a bit). why not go for something like this:

```go
func mergeVals(lhs, rhs ref.Val) ref.Val {
	left, lok := lhs.(traits.Mapper)
	right, rok := rhs.(traits.Mapper)
	if !lok || !rok {
		return types.ValOrErr(lhs, "no such overload: %v.merge(%v)", lhs.Type(), rhs.Type())
	}
	return merge(left, right)
}

// merge returns a new map containing entries from both maps.
// Keys in 'other' overwrite keys in 'self'.
func merge(self, other traits.Mapper) traits.Mapper {
	result := mapperTraitToMutableMapper(other)
	for i := self.Iterator(); i.HasNext().(types.Bool); {
		k := i.Next()
		if !result.Contains(k).(types.Bool) {
			result.Insert(k, self.Get(k))
		}
	}
	return result.ToImmutableMap()
}

// mapperTraitToMutableMapper copies a traits.Mapper into a MutableMap.
func mapperTraitToMutableMapper(m traits.Mapper) traits.MutableMapper {
	vals := make(map[ref.Val]ref.Val, m.Size().(types.Int))
	for it := m.Iterator(); it.HasNext().(types.Bool); {
		k := it.Next()
		vals[k] = m.Get(k)
	}
	return types.NewMutableMap(types.DefaultTypeAdapter, vals)
}

```

Then you can call

```go
cel.Function("merge",
			cel.MemberOverload("map_merge",
				[]*cel.Type{mapType, mapType},
				mapType,
				cel.BinaryBinding(mergeVals),
			),
		)
```

pkg/cel/library/maps_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/google/cel-go/cel"
+)
+
+func TestMaps(t *testing.T) {
```
jakobmoellerdev, line 26: nit: using testify can make your life a lot easier here probably (we already use it in other tests in this package):


```go
package library

import (
	"fmt"
	"testing"

	"github.com/stretchr/testify/require"
)

func TestMaps(t *testing.T) {
	tests := []struct {
		expr string
		err  require.ErrorAssertionFunc
	}{
		{expr: `{}.merge({}) == {}`},
		{expr: `{}.merge({'a': 1}) == {'a': 1}`},
		{expr: `{}.merge({'a': 2.1}) == {'a': 2.1}`},
		{expr: `{}.merge({'a': 'foo'}) == {'a': 'foo'}`},
		{expr: `{'a': 1}.merge({}) == {'a': 1}`},
		{expr: `{'a': 1}.merge({'b': 2}) == {'a': 1, 'b': 2}`},
		{expr: `{'a': 1}.merge({'a': 2, 'b': 2}) == {'a': 2, 'b': 2}`},
		{expr: `{}.merge([])`, err: func(t require.TestingT, err error, i ...interface{}) {
			require.ErrorContains(t, err, "ERROR: <input>:1:9: found no matching overload for 'merge' applied to 'map(dyn, dyn).(list(dyn))'")
		}},
	}

	env := testMapsEnv(t)
	for i, tc := range tests {
		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
			r := require.New(t)
			ast, iss := env.Compile(tc.expr)
			if tc.err != nil {
				tc.err(t, iss.Err())
				return
			}
			r.NoError(iss.Err(), "compile failed for expr: %s", tc.expr)

			prg, err := env.Program(ast)
			r.NoError(err, "program failed for expr: %s", tc.expr)

			out, _, err := prg.Eval(cel.NoVars())
			if tc.err != nil {
				r.Error(err, "expected error for expr: %s", tc.expr)
				r.Contains(err.Error(), tc.err)
			} else {
				r.NoError(err, "unexpected error for expr: %s", tc.expr)
				r.IsType(cel.BoolType, out.Type(), "unexpected type for expr: %s", tc.expr)
				r.True(out.Value().(bool), "unexpected result for expr: %s", tc.expr)
			}
		})
	}
}

func testMapsEnv(t *testing.T, opts ...cel.EnvOption) *cel.Env {
	t.Helper()
	env, err := cel.NewEnv(append([]cel.EnvOption{Maps()}, opts...)...)
	require.NoError(t, err, "cel.NewEnv(Maps()) failed")
	return env
}
```

pkg/cel/library/maps_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}{
+		{expr: `{}.merge({}) == {}`},
+		{expr: `{}.merge({'a': 1}) == {'a': 1}`},
+		{expr: `{}.merge({'a': 2.1}) == {'a': 2.1}`},
+		{expr: `{}.merge({'a': 'foo'}) == {'a': 'foo'}`},
+		{expr: `{'a': 1}.merge({}) == {'a': 1}`},
+		{expr: `{'a': 1}.merge({'b': 2}) == {'a': 1, 'b': 2}`},
+		{expr: `{'a': 1}.merge({'a': 2, 'b': 2}) == {'a': 2, 'b': 2}`},
+
+		// {expr: `{}.merge([])`, err: "ERROR: <input>:1:9: found no matching overload for 'merge' applied to 'map(dyn, dyn).(list(dyn))'"},
```
jakobmoellerdev: nit: do we want dynamic support in here? If so, either uncomment the test, add todo for the comment if outstanding, or remove the comments

pkg/cel/library/maps_test.go
gfrey: Haha, I had started converting the tests to using testify (back when I originally wrote this code, I didn't see that testify was available). Thanks, learnt a few tricks from this (`r := require.New(t)` f.ex. and usage of ErrorAssertionFunc)

pkg/cel/library/maps.go
gfrey: this works in general (flip order so that other is written first, then fill in whatever keys aren't there from self); have picked suggestions from @jakobmoellerdev

pkg/cel/library/maps.go
a-hilaly: I thought we had one! :) `go generate` should get that done. cc @tomasaschan

a-hilaly (approved):
Thank you @gfrey !

matthchr (commented):
Before we merge this it might be good to check with the cel-go maintainers if they'd just accept this upstream as an `ext/maps.go` extension.

It doesn't really seem kro specific and I'd rather we:
1. Depend on shared logic rather than roll out own as much as possible.
2. Share these sorts of "generally useful CEL helpers with upstream"

If making that happen is going to be difficult or take a lot of time we can just merge here and upstream later, but it feels like we should check first.

a-hilaly:
pinged Joe on upstream slack, let's see what cel-go folks think.

barney-s:
Reached out to CEL team internally this week. They seem to be open. They were expressing concerns about merge semantics (overwrite, replace, combine etc) especially when the values are lists. Will keep working with them and get something part of standard extensions. Meanwhile we can merge this here in KRO or K8s (Joe Betz). 

My recommendation would be to indicate `experimental` in the library name or `alpha` ? 
"cel.lib.ext.kro.maps" -> "cel.lib.ext.kro.experimental.maps"  ?

pkg/cel/library/maps.go
barney-s: https://github.com/kro-run/kro/pull/640

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,112 @@
...
+//
+// Examples:
+//
+//	{}.merge({}) == {}
+//	{}.merge({'a': 1}) == {'a': 1}
+//	{'a': 1}.merge({}) == {'a': 1}
+//	{'a': 1}.merge({'b': 2}) == {'a': 1, 'b': 2}
+//	{'a': 1}.merge({'a': 2, 'b': 2}) == {'a': 1, 'b': 2}
+func Maps(options ...MapsOption) cel.EnvOption {
+	l := &mapsLib{version: math.MaxUint32}
```
barney-s, line 43: Is this common to version it to MAX_INT ?

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,112 @@
...
+
+type mapsLib struct {
+	version uint32
+}
+
+type MapsOption func(*mapsLib) *mapsLib
+
+// LibraryName implements the cel.SingletonLibrary interface method.
+func (l *mapsLib) LibraryName() string {
+	return "cel.lib.ext.kro.maps"
```
barney-s, line 58: is there a way to indicate this is still experimental/alpha. 

cel.lib.ext.experimental.maps ? 
or 
cel.lib.ext.kro.experimental.maps ?

pkg/cel/library/maps.go
barney-s: Wondering if we need to rename the folder from library to ext ? 
Even a separate repo under kro-run ?

pkg/cel/library/maps.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// CompileOptions implements the cel.Library interface method.
+func (l *mapsLib) CompileOptions() []cel.EnvOption {
+	mapType := cel.MapType(cel.TypeParamType("K"), cel.TypeParamType("V"))
```
barney-s, line 63: More of a Q: "K" and "V" are not used here. Are those more for documentation ?> 
If so can we rename to Key, Value ?

a-hilaly:
ping @gfrey - as Barni stated would you like to open a PR against CEL official std libs?

a-hilaly:
also can you please sign the CLA?

jakobmoellerdev:
@gfrey friendly ping for you to double check on this! Thanks!

jakobmoellerdev:
@gfrey we really want to have this, would be a bummer to loose out on your progress! let us know if we can support

gfrey:
> @gfrey we really want to have this, would be a bummer to loose out on your progress! let us know if we can support

Sorry for the delay, I was out for a while. For the CLA I'm trying to get into the CCLA, so work in progress. I'm going to start the same for the cel-go project, too. When done, I'll create an issue/PR.

gfrey:
@jakobmoellerdev Sorry for the delays. I've created https://github.com/gfrey/cel-go/commit/9301f551b8681488f4f68c848247d6179a38a333. Is that roughly what would be expected here? Can someone give a thumbs-up/down before I go ahead and create a PR?

kahirokunn:
> @jakobmoellerdev Sorry for the delays. I've created [gfrey/cel-go@9301f55](https://github.com/gfrey/cel-go/commit/9301f551b8681488f4f68c848247d6179a38a333). Is that roughly what would be expected here? Can someone give a thumbs-up/down before I go ahead and create a PR?


Thanks for tackling this! I took a look at the implementation and it looks really solid to me. Nice work on the approach here. 👍

gfrey:
Closing PR in favor of google/cel-go#1239. Thanks to everyone who reviewed and helped!

PR #631, opened 2025-08-14
Have separate instance labellers for different reconcile paths.

barney-s:
This is needed when we have RGD2 instance in RGD1 resources. Today both paths use the same label resulting in conflict.

path 1: RGD2.instance created as part of an RGD1 instance reconciliation
path 2: RGD2.instance reconciled by RGD2 reconciler

### DIFF
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -41,10 +41,17 @@ const (
 	InstanceLabel          = LabelKROPrefix + "instance-name"
 	InstanceNamespaceLabel = LabelKROPrefix + "instance-namespace"
 
+	// Parent Identifier Labels. These are set by the instance reconciler that creates/updates the resource.
 	ResourceGraphDefinitionIDLabel        = LabelKROPrefix + "resource-graph-definition-id"
 	ResourceGraphDefinitionNameLabel      = LabelKROPrefix + "resource-graph-definition-name"
 	ResourceGraphDefinitionNamespaceLabel = LabelKROPrefix + "resource-graph-definition-namespace"
 	ResourceGraphDefinitionVersionLabel   = LabelKROPrefix + "resource-graph-definition-version"
+
+	// Source Identifier Labels. These are set by the instance reconciler that reconciles the resource.
+	SourceResourceGraphDefinitionIDLabel        = LabelKROPrefix + "source-resource-graph-definition-id"
+	SourceResourceGraphDefinitionNameLabel      = LabelKROPrefix + "source-resource-graph-definition-name"
+	SourceResourceGraphDefinitionNamespaceLabel = LabelKROPrefix + "source-resource-graph-definition-namespace"
+	SourceResourceGraphDefinitionVersionLabel   = LabelKROPrefix + "source-resource-graph-definition-version"
 )
 
 // IsKROOwned returns true if the resource is owned by KRO.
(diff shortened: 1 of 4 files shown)


jakobmoellerdev (commented):
this is a good change but I think we should introduce a test case here to ensure this never breaks again. WDYT?

barney-s:
> this is a good change but I think we should introduce a test case here to ensure this never breaks again. WDYT?

I modified an existing test to check for the source label. Can add a separate test too.

jakobmoellerdev:
existing test is perfectly fine imo 👍 🙇

test/e2e/chainsaw/check-multi-resource-rgd/instance-create-assert.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   generation: 1
   labels:
     kro.run/owned: "true"
-    kro.run/resource-graph-definition-name: check-multi-resource-rgd
+    kro.run/source-resource-graph-definition-name: check-multi-resource-rgd
```
a-hilaly: I agree with @jakobmoellerdev here, let's have a test that shows path1 and path2 (i guess both rgd and source-rgd label will exist in those scenarios?)

barney-s: done

jakobmoellerdev (approved):

pkg/metadata/labels.go
```diff
@@ -41,10 +41,17 @@ const (
...
 	InstanceNamespaceLabel = LabelKROPrefix + "instance-namespace"
 
+	// Parent Identifier Labels. These are set by the instance reconciler that creates/updates the resource.
 	ResourceGraphDefinitionIDLabel        = LabelKROPrefix + "resource-graph-definition-id"
 	ResourceGraphDefinitionNameLabel      = LabelKROPrefix + "resource-graph-definition-name"
 	ResourceGraphDefinitionNamespaceLabel = LabelKROPrefix + "resource-graph-definition-namespace"
 	ResourceGraphDefinitionVersionLabel   = LabelKROPrefix + "resource-graph-definition-version"
+
+	// Source Identifier Labels. These are set by the instance reconciler that reconciles the resource.
+	SourceResourceGraphDefinitionIDLabel        = LabelKROPrefix + "source-resource-graph-definition-id"
```
matthchr: We talked about this and think we should change this label to something more clear to what it is:

- owned-by-resource-graph-definition-id
- reconciled-by-resource-graph-definition-id
- other options?

pkg/metadata/labels.go
```diff
@@ -41,10 +41,17 @@ const (
...
+	SourceResourceGraphDefinitionNameLabel      = LabelKROPrefix + "source-resource-graph-definition-name"
+	SourceResourceGraphDefinitionNamespaceLabel = LabelKROPrefix + "source-resource-graph-definition-namespace"
```
matthchr: Can we remove namespace and version (and above?)

because we're not using it AFAIK.

pkg/metadata/labels.go
```diff
@@ -41,10 +41,17 @@ const (
 	InstanceLabel          = LabelKROPrefix + "instance-name"
 	InstanceNamespaceLabel = LabelKROPrefix + "instance-namespace"
 
+	// Parent Identifier Labels. These are set by the instance reconciler that creates/updates the resource.
 	ResourceGraphDefinitionIDLabel        = LabelKROPrefix + "resource-graph-definition-id"
```
matthchr, line 63: Do we want this to be `created-by`? Possibly we can keep these (back-compat) and add new ones and remove the old later.

pkg/metadata/labels.go
barney-s: what about `defined-by-resource-graph-definition-id` ?

jakobmoellerdev: since these labels are exposed to users we should probably document them a little bit better. they will likely be looked up by technically versed users, and maybe we should even think about explaining them in the docs (doesn't have to be in this PR imo)

barney-s:
Added a proposal here: #639

pkg/metadata/labels.go
barney-s: made this PR track the proposal #639. Once we merge that proposal i will change the implementation here. Closing all comments for now.

pkg/metadata/labels.go
barney-s: done.

pkg/metadata/labels.go
barney-s: made this PR track the proposal #639. Once we merge that proposal i will change the implementation here. Closing all comments for now.

pkg/metadata/labels.go
barney-s: added the docs in this PR.

PR #632, opened 2025-08-14
Introduce collections support

a-hilaly:
This PR introduces collections support in RGDs through a new `forEach` field, enabling dynamic creation of multiple resources based on iterative expressions.

- Added `forEach` field to RGDs spec allowing iteration over both arrays and maps
- Implemented CEL based collection evaluation with support for filtering and mapping
- Added collection aware resource reconciliation in the controller
- Introduced `each` context variables (`item`, `index`, `length`, `key`, `value`) for template access

Example usage:
```yaml
resources:
  - id: replicaSet
    forEach: ${schema.spec.roles.filter(x, x.substring(0, 4) != 'test')}
    template:
      apiVersion: v1
      kind: Pod
      metadata:
        name: ${schema.spec.name + '-' + each.item}
      spec:
        containers:
          - name: ${schema.spec.name + '-' + each.item}
            image: ${schema.spec.image}
```


barney-s:
Do we want to start with a design doc first ?

brandonjbjelland:
This is at the top of my wish list for kro - thank you for this work and anxiously awaiting its arrival!

a-hilaly:
@brandonjbjelland we're excited as well! i did open a proposal describing the collection semantics and possible use cases, your feedback would be highly valuable! => https://github.com/kubernetes-sigs/kro/pull/679

vast-raymond-page:
Instead of a "simple" case, can we have forEach be an array of collections as the only supported mechanic? I really like the idea of an array of collections to allow multiple iterators running over every combination of collections like discussed here:
https://github.com/kubernetes-sigs/kro/pull/679#discussion_r2369433483

```yaml
  forEach:
    - collection: ${spec.regions}
      name: "region"
    - collection: ${spec.zones}
      name: "zone"
  template:
    kind: Pod
    metadata:
      name: ${'pod-' +region.entry+ '-' +  zone.entry}
```

PR #639, opened 2025-08-18
Adding a propsal for changing the KRO instance and resource labels

barney-s:
Related PR that triggered this doc: https://github.com/kro-run/kro/pull/631

### DIFF
--- a/docs/design/proposals/labels.md
+++ b/docs/design/proposals/labels.md
@@ -1,0 +1,109 @@
+# Proposal: Standardize KRO Labels
+
+## Problem statement
+
+The current labels in KRO are used to track ownership and metadata of resources managed by the orchestrator. While these labels provide essential information, they lack a clear and standardized way to represent the relationships between different KRO components, such as ResourceGraphDefinitions (RGDs), instances, and the resources they generate. This ambiguity can make it difficult to query for related objects and understand the provenance of resources within a cluster.
+
+Specifically, we need a clearer way to identify:
+1.  Which RGD is responsible for reconciling a given instance.
+2.  Which RGD a created resource originates from.
+3.  Which instance a created resource belongs to.
+
+This proposal aims to introduce new, more descriptive labels to clarify these relationships, aligning with Kubernetes best practices for resource management and observability.
+
+## Proposal
+
+We propose introducing a set of standardized labels to better represent the relationships between RGDs, instances, and the resources they manage. This will improve the ability to discover, manage, and debug resources created and managed by KRO.
+
+#### Overview
+
+The proposal is to add the following labels:
+-   A label on an instance to indicate which RGD is reconciling it.
+-   A label on resources created by an instance reconciliation to indicate which RGD they come from.
+-   Labels on resources created by an instance reconciliation to indicate which instance created them.
+
+This aligns with the common Kubernetes practice of using labels for discovery and organization of related resources, as seen in tools like Helm and Kustomize.
+
+#### Design details
+
+We will rename or introduce the following labels, which will be defined in `@pkg/metadata/labels.go`:
+
+1.  **Labels for mapping created resources back to the instance:**
+    - To provide a direct and queryable link from a created resource back to the instance that caused its creation. This set of labels uniquely identifies the instance.
+    - **Proposed Labels:**
+      - `kro.run/instance-gvk: mygroup.example.com/v1/MyKind`: The Group Version Kind (GVK) of the instance
+      - `kro.run/instance: default/my-instance`: The namespace and name of the instance.
+      - `app.kubernetes.io/instance: MyKind/default/my-instance`: Kind namespace and name of the instance
+    - **Existing Labels**
+      - `kro.run/instance-id: 6e6a1d95-4855-4062-b51b-8f288cb96365`
+      - `kro.run/instance-name: test-instance`
+      - `kro.run/instance-namespace: default`
+2.  **Label for mapping created resources back to the RGD:**
+    - To identify which RGD was used as a template to create a resource during the reconciliation of an instance. This label will be applied to all resources created by the instance controller.
+    - **Proposed Labels:**
+      - `kro.run/part-of: my-rgd-name` - The name of the ResourceGraphDefinition
+      - `app.kubernetes.io/part-of: my-rgd-name`
+    - **Existing Labels:**
+      - `kro.run/resource-graph-definition-id: 761e8fb7-14d5-4da1-b4f9-0a16fc334d6c`
+      - `kro.run/resource-graph-definition-name: my-rgd-name`
+3. **Label for nested RGD instance:**
+    - To differenciate the RGD that defines (and hence reconciles) the instance and the RGD that created it, we need to introduce a separate label.
+    - Currently we dont differentiate, it results in a conflict (https://github.com/kro-run/kro/pull/631)
+    - To indicate that an instance is being actively reconciled against a specific RGD. This label will be applied to the instance itself.
+    - **Proposed Labels:**
+      - `kro.run/managed-by-rgd: my-rgd-name` - name of the ResourceGraphDefinition that reconciles the instance.
+    - **Existing Labels:**
+      - `kro.run/resource-graph-definition-id: 761e8fb7-14d5-4da1-b4f9-0a16fc334d6c`
+      - `kro.run/resource-graph-definition-name: my-rgd-name`
+      - Conflicts with label used in creation path for nested RGD (https://github.com/kro-run/kro/pull/631)
+4. **Kubernetes Common Labels:**
+    - **Purpose:** Conforming to standard/common k8s labels allows easier integration across tools in k8s ecosystem.
+    - **Proposed Labels:**
+      - `app.kubernetes.io/managed-by: KRO` : Fixed value
+
+## Other solutions considered
+
+1. We could continue using the existing labels, but their purpose is not as explicit for relationship tracking, which can lead to confusion for users and client tools.
+2. We could also use annotations, but labels are better suited for this purpose as they are queryable via the Kubernetes API, which is a key requirement for observability and tooling.
+3. Use common k8s labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
+  * `app.kubernetes.io/name:	mysql`
+  * `app.kubernetes.io/instance: mysql-abcxyz`
+  * `app.kubernetes.io/version: 5.7.21`
+  * `app.kubernetes.io/component: database`
+  * `app.kubernetes.io/part-of: wordpress`
+  * `app.kubernetes.io/managed-by: Helm`
+
+
+## Scoping
+
+#### What is in scope for this proposal?
+
+-  Defining and adding the new labels to the KRO API.
+-  Updating the instance controller to apply these labels to instances and created resources during reconciliation.
+-  Updating the official KRO documentation to reflect the new labels and their usage.
+-  Adding docs reflecting the new labels. Old labels would not be documented.
+-  Deprecate old labels in next minor release or the one after that.
+
+#### What is not in scope?
+
+- Tracking UID od RGD, Instance in the labels. We need to handle recreation of parent resources where UID changes. Can be a future enhancement.
+- Supporting deeply nested RGDs. When RGD1 created RGD2 and that creates RGD3, do we add labels in objects of RGD3 linking it back to RGD2 and RGD1 ?
+- Changing how KRO uses annotations.
+- Changing these existing labels:
+  - kro.run/kro-version: v0.4.0
+  - kro.run/owned: "true"
+
+## Testing strategy
+
+#### Requirements
+
+-   A running Kubernetes cluster to deploy KRO and test the label application.
+
+#### Test plan
+
+-   **Unit tests:** Add unit tests for any new labeler functions in `pkg/metadata/labels.go` to ensure they construct the correct labels.
+-   **Integration tests:**
+    -   Create an RGD and an instance of that RGD.
+    -   Verify that the instance object has the `kro.run/reconciled-by` label pointing to the RGD.
+    -   Verify that all resources created by the instance reconciliation have the `kro.run/created-by` label, and the `kro.run/instance-group`, `kro.run/instance-kind`, `kro.run/instance-namespace`, and `kro.run/instance-name` labels with the correct values.
+    -   Update an instance to be reconciled by a different RGD and verify that the `kro.run/reconciled-by` label is updated accordingly.


docs/design/proposals/labels.md
```diff
@@ -0,0 +1,83 @@
...
+        - `kro.run/managed-by-instance-kind`: The kind of the instance (e.g., `MyKind`).
+        - `kro.run/managed-by-instance-namespace`: The namespace of the instance (e.g., `default`).
+        - `kro.run/managed-by-instance-name`: The name of the instance (e.g., `my-instance`).
+        - The `managed-by` label is a common pattern in the Kubernetes ecosystem.
+    -   **Purpose:** To provide a direct and queryable link from a created resource back to the instance that caused its creation. This set of labels uniquely identifies the instance.
+    -   **Alternative Name prefix:** `kro.run/owner-`, `kro.run/created-by-`, etc.
+
+## Other solutions considered
+
+We could continue using the existing labels, but their purpose is not as explicit for relationship tracking, which can lead to confusion for users and client tools. We could also use annotations, but labels are better suited for this purpose as they are queryable via the Kubernetes API, which is a key requirement for observability and tooling.
```
jakobmoellerdev: nit: The existing labels should maybe listed here but honestly its fine for this proposal

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,83 @@
...
+
+#### What is in scope for this proposal?
+
+-   Defining and adding the new labels to the KRO API.
+-   Updating the instance controller to apply these labels to instances and created resources during reconciliation.
+-   Updating the official KRO documentation to reflect the new labels and their usage.
+
+#### What is not in scope?
+
+-   Removing or deprecating the existing labels. This could be considered in a future proposal to maintain backward compatibility.
```
jakobmoellerdev: I think we SHOULD start deprecating old labels. That does not mean I want to remove them directly but I think their use is discouraged from this point on right?

jakobmoellerdev (commented):
Mostly LGTM. Just thinking we should really start unifying everything and not keep the old labels.

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,83 @@
...
+We will introduce the following new labels, which will be defined in `@pkg/metadata/labels.go`:
+
+1.  **Label for an instance being reconciled:**
+    -   **Proposed Label:** `kro.run/reconciled-by`
+    -   **Value:** The name of the ResourceGraphDefinition (e.g., `my-rgd-name`).
+    -   **Purpose:** To indicate that an instance is being actively reconciled against a specific RGD. This label will be applied to the instance itself.
+    -   **Alternative Names:** `kro.run/reconciling-rgd`, `kro.run/template`.
+
+2.  **Label for resources created during instance reconciliation:**
+    -   **Proposed Label:** `kro.run/created-by`
```
fabianburth: I don't have a strong opinion about any of the suggested variants, but I generally think having `rgd` in the label name improves clarity (e.g. `kro.run/created-by-rgd`), especially since it is an indirect relationship.

fabianburth (commented):
Looks good!

docs/design/proposals/labels.md
barney-s: will add it.

docs/design/proposals/labels.md
barney-s: sounds good.

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+1.  **Label for an instance being reconciled:**
+    - **Existing Labels:**
+      - kro.run/resource-graph-definition-id: 761e8fb7-14d5-4da1-b4f9-0a16fc334d6c
+      - kro.run/resource-graph-definition-name: check-instance-creation
+      - Conflicts with label used in creation path for nested RGD (https://github.com/kro-run/kro/pull/631)
+    - **Proposed Label:**
+      - `kro.run/reconciled-by-rgd`
+    - **Value:** The name of the ResourceGraphDefinition (e.g., `my-rgd-name`).
+    - **Purpose:** To indicate that an instance is being actively reconciled against a specific RGD. This label will be applied to the instance itself.
+    - **Alternative Names:**
```
jakobmoellerdev: are these alternative names set too or are these just suggestions? I would prefer to limit us to one.

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+    - **Existing Labels:**
+      - kro.run/resource-graph-definition-id: 761e8fb7-14d5-4da1-b4f9-0a16fc334d6c
+      - kro.run/resource-graph-definition-name: check-instance-creation
+    - **Proposed Label:**
+      - `kro.run/defined-by-rgd`
+    - **Value:** The name of the ResourceGraphDefinition (e.g., `my-rgd-name`).
+    - **Purpose:** To identify which RGD was used as a template to create a resource during the reconciliation of an instance. This label will be applied to all resources created by the instance controller.
+    - **Alternative Names:**
+      - kro.run/created-by-rgd
+      - kro.run/managed-by-rgd
```
jakobmoellerdev: I think managed-by probably fits best since we continuously work on it

jakobmoellerdev: I think if we have `kro.run/managed-by-rgd` it could be reused for the instance too. Is there a specific need to have two labels here?

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+      - kro.run/template-rgd
+
+3.  **Labels for resources to link back to the instance:**
+    - **Existing Label:**
+      - kro.run/instance-id: 6e6a1d95-4855-4062-b51b-8f288cb96365
+      - kro.run/instance-name: test-instance
+      - kro.run/instance-namespace: default
+    - **Proposed Labels:**
+      - `kro.run/managed-by-instance-group`: The API group of the instance (e.g., `mygroup.example.com`).
+      - `kro.run/managed-by-instance-kind`: The kind of the instance (e.g., `MyKind`).
```
jakobmoellerdev: if you define a group+kind, don't you also need to define a version?

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+      - `kro.run/managed-by-instance-namespace`: The namespace of the instance (e.g., `default`).
+      - `kro.run/managed-by-instance-name`: The name of the instance (e.g., `my-instance`).
+      - The `managed-by` label is a common pattern in the Kubernetes ecosystem.
+    - **Purpose:** To provide a direct and queryable link from a created resource back to the instance that caused its creation. This set of labels uniquely identifies the instance.
+    - **Alternative Name prefix:**
+      - `kro.run/owner-`
+      - `kro.run/created-by-`
+
+## Other solutions considered
```
jakobmoellerdev, line 64: Im wondering if it doesn't make sense to reuse the [kubernetes](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) labels or at least co-apply them. Im sure you thought about this too so it would be good to have a thought about this in the proposal

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+
+This aligns with the common Kubernetes practice of using labels for discovery and organization of related resources, as seen in tools like Helm and Kustomize.
+
+#### Design details
+
+We will introduce the following new labels, which will be defined in `@pkg/metadata/labels.go`:
+
+1.  **Label for an instance being reconciled:**
+    - **Existing Labels:**
+      - kro.run/resource-graph-definition-id: 761e8fb7-14d5-4da1-b4f9-0a16fc334d6c
```
a-hilaly: I wanted to add some context about why I originally introduced the RGD ID label. The goal was to help the controller detect when RGDs are deleted and recreated, which creates completely new RGDs and can trigger unwanted cleanup. I was thinking we could have an RGD annotation like `kro.run/allow-recreation` (or maybe `kro.run/allow-rgd-replacement`?) that users would need to explicitly set. Without this annotation, kro would reject the recreation as a conflict to prevent accidental data loss. cc @matthchr @jakobmoellerdev @n3wscott

a-hilaly: This might influence how we think about the labeling scheme - we still need some way to track RGD identity beyond just the name. but i fully agree the current implementation could be clearer

docs/design/proposals/labels.md
a-hilaly: How about:
```yaml
kro.run/instance: production/my-web-app  # namespace/name format
kro.run/instance-gvk: myapp.io/v1/WebService  # optional for additional context ?
```

a-hilaly: another approach would be to just include an RGD ref?

docs/design/proposals/labels.md
a-hilaly: +1 here i would love for us to adopt k8s standard labels. at least the `app.kuberentes.io/managed-by` and `app.kubernets.io/part-of`

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+#### Overview
+
+The proposal is to add the following labels:
+-   A label on an instance to indicate which RGD is reconciling it.
+-   A label on resources created by an instance reconciliation to indicate which RGD they come from.
+-   Labels on resources created by an instance reconciliation to indicate which instance created them.
+
+This aligns with the common Kubernetes practice of using labels for discovery and organization of related resources, as seen in tools like Helm and Kustomize.
+
+#### Design details
```
a-hilaly, line 27: should we have  a section addressing deeply nested RGDs (3+ levels)? thinking what happens when RGD1 creates an RGD2 instance, which creates an RGD3 instance, which creates an RGD4 instance ? With the two label scheme, we lose the intermediate relationships e.g an RGD4 instance would only show it was defined by RGD3 and controlled by RGD4, with no trace of RGD1 or RGD2. this means queries like "show me all resources that originated from RGD1, including those created indirectly through nested RGDs" won't be possible - which is fine, but worth noting.

a-hilaly (commented):
Thanks for putting this together @barney-s ! you've identified an important issue - the current labeling scheme is definitely confusing with both `resource-graph-definition-id` and `resource-graph-definition-name` without clear semantics.

docs/design/proposals/labels.md
jakobmoellerdev: I think the Kubernetes UID field is sufficient for this purpose because it is effectively an ID that gets recreated when the object is recreated. Any specific reason why you would want to roll your own?

docs/design/proposals/labels.md
jakobmoellerdev: We could get around this by having a root RGD, although this starts to get reaaaally complicated by then :D

barney-s: That is true. We can recursively enumerate and get the nested resources. But not with a single list query.

Question would be what do you do with the listed result ? 

And this is a current pattern in k8s as well. Deployment -> Replicaset -> Pods. 
There are ways to allow traceability by injecting labels downstream.

docs/design/proposals/labels.md
barney-s: just alternative names in the design. we will only use proposed label. Just capturing alternatives in the design. Can remove them.

docs/design/proposals/labels.md
barney-s: ah thats the whole point of the proposal. 
I guess the problem statement did not do a good job. Let me know how i can update the problem statement to make the need for 2 labels more apparent.

docs/design/proposals/labels.md
barney-s: removed the alternatives.

docs/design/proposals/labels.md
barney-s: I think the value is the UID. Having UID forces us to make a decision when this label mismatches. Do we replace the label or error out ? Deletion may be a bad thing. 

At this point in time having name is durable (across instance re-creates) and any leftover resources are adopted (if not deleted already).

docs/design/proposals/labels.md
barney-s: The scope is not to identify all parents. Just the immediate ones. Similar to how label/refs work in current k8s objects.

barney-s: called this out as not part of scope in the design

docs/design/proposals/labels.md
barney-s: called UID part out of scope in the design proposal. GVK NN is a good user and programmatic tracking filters. For UID lets do it in a separate PR.

docs/design/proposals/labels.md
barney-s: what do you mean by RGD Ref ? in the ownerrefs ?

barney-s:
@jakobmoellerdev @a-hilaly @matthchr  - Please Take a look at this updated proposal.

docs/design/proposals/labels.md
barney-s: updated proposal to include those

jakobmoellerdev (commented):
For me this is looking good to go. Thanks a lot for the investment @barney-s !

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
+# Proposal: Standardize KRO Labels
+
+## Problem statement
+
+The current labels in KRO are used to track ownership and metadata of resources managed by the orchestrator. While these labels provide essential information, they lack a clear and standardized way to represent the relationships between different KRO components, such as ResourceGraphDefinitions (RGDs), instances, and the resources they generate. This ambiguity can make it difficult to query for related objects and understand the provenance of resources within a cluster.
```
tjamet, line 5: It would be interesting to remind about the current set of annotations/labels to better understand the problem and how it gets fixed

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+#### Design details
+
+We will rename or introduce the following labels, which will be defined in `@pkg/metadata/labels.go`:
+
+1.  **Labels for mapping created resources back to the instance:**
+    - To provide a direct and queryable link from a created resource back to the instance that caused its creation. This set of labels uniquely identifies the instance.
+    - **Proposed Labels:**
+      - `kro.run/instance-gvk: mygroup.example.com/v1/MyKind`: The Group Version Kind (GVK) of the instance
+      - `kro.run/instance: default/my-instance`: The namespace and name of the instance.
+      - `app.kubernetes.io/instance: MyKind/default/my-instance`: Kind namespace and name of the instance
```
tjamet, line 36: Labels are extremely useful to select resources `kubectl -n my-namespace get X -l my=label` and identify them in a human readable manner.

Given that we refer to 2 owning objects, would it make sense to rather use dns notation:

```
instance.kro.run/apiVersion: mygroup.example.com/v1
instance.kro.run/kind: MyKind
instance.kro.run/name: my-instance
instance.kro.run/namespace: my-namespace
rgd.kro.run/name: some-rgd
rgd.kro.run/apiVersion: kro.run/v1alpha1
app.kubernetes.io/instance: MyKind/default/my-instance
```

that way we can easily perform different queries `kubectl -n my-namespace get X -l instance.kro.run/kind` to get all namespaced resources deployed by a certain kind, `kubectl -n my-namespace get -A X -l instance.kro.run/name=my-instance`

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+    - **Existing Labels**
+      - `kro.run/instance-id: 6e6a1d95-4855-4062-b51b-8f288cb96365`
+      - `kro.run/instance-name: test-instance`
+      - `kro.run/instance-namespace: default`
+2.  **Label for mapping created resources back to the RGD:**
+    - To identify which RGD was used as a template to create a resource during the reconciliation of an instance. This label will be applied to all resources created by the instance controller.
+    - **Proposed Labels:**
+      - `kro.run/part-of: my-rgd-name` - The name of the ResourceGraphDefinition
+      - `app.kubernetes.io/part-of: my-rgd-name`
```
tjamet, line 45: Do we really need both?

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+    - **Existing Labels:**
+      - `kro.run/resource-graph-definition-id: 761e8fb7-14d5-4da1-b4f9-0a16fc334d6c`
+      - `kro.run/resource-graph-definition-name: my-rgd-name`
+3. **Label for nested RGD instance:**
+    - To differenciate the RGD that defines (and hence reconciles) the instance and the RGD that created it, we need to introduce a separate label.
+    - Currently we dont differentiate, it results in a conflict (https://github.com/kro-run/kro/pull/631)
+    - To indicate that an instance is being actively reconciled against a specific RGD. This label will be applied to the instance itself.
+    - **Proposed Labels:**
+      - `kro.run/managed-by-rgd: my-rgd-name` - name of the ResourceGraphDefinition that reconciles the instance.
```
tjamet, line 54: What about

```
instance.kro.run/managed-by: my-rgd-name
```

Could it help reveal the hierarchy for humans?

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+      - Conflicts with label used in creation path for nested RGD (https://github.com/kro-run/kro/pull/631)
+4. **Kubernetes Common Labels:**
+    - **Purpose:** Conforming to standard/common k8s labels allows easier integration across tools in k8s ecosystem.
+    - **Proposed Labels:**
+      - `app.kubernetes.io/managed-by: KRO` : Fixed value
+
+## Other solutions considered
+
+1. We could continue using the existing labels, but their purpose is not as explicit for relationship tracking, which can lead to confusion for users and client tools.
+2. We could also use annotations, but labels are better suited for this purpose as they are queryable via the Kubernetes API, which is a key requirement for observability and tooling.
```
tjamet, line 67: I would suggest 1 or 2 annotations for programmatic access that include IDs

for example


```
instance.kro.run/reference: {"apiVersion": "kro.run/v1alpha1", "kind": "MyKind", "resource": "mykinds", "namespace": "default", "name": "my-instance", "id": "$id"}
rgd.kro.run/reference: {"apiVersion": "kro.run/v1", "name": "my-rgd", "id": "${id}"}
```

ID would help reconcilers detect that the instance (or RGD) was fully replaced between 2 reconciliation loops

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Scoping
+
+#### What is in scope for this proposal?
+
+-  Defining and adding the new labels to the KRO API.
+-  Updating the instance controller to apply these labels to instances and created resources during reconciliation.
+-  Updating the official KRO documentation to reflect the new labels and their usage.
+-  Adding docs reflecting the new labels. Old labels would not be documented.
+-  Deprecate old labels in next minor release or the one after that.
```
tjamet, line 85: isn't 1-2 versions a bit short?
Are we confident most users will upgrade to migrate all their labels?

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+## Testing strategy
+
+#### Requirements
+
+-   A running Kubernetes cluster to deploy KRO and test the label application.
+
+#### Test plan
+
+-   **Unit tests:** Add unit tests for any new labeler functions in `pkg/metadata/labels.go` to ensure they construct the correct labels.
+-   **Integration tests:**
```
tjamet, line 105: we may want a migration test, to ensure we don't mess up deployed objects with the new version.

We may also want to ensure that in case users jumps directly from the current labels layout to the proposed one (i.e. skip the transition versions) kro keeps running as expected

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+This aligns with the common Kubernetes practice of using labels for discovery and organization of related resources, as seen in tools like Helm and Kustomize.
+
+#### Design details
+
+We will rename or introduce the following labels, which will be defined in `@pkg/metadata/labels.go`:
+
+1.  **Labels for mapping created resources back to the instance:**
+    - To provide a direct and queryable link from a created resource back to the instance that caused its creation. This set of labels uniquely identifies the instance.
+    - **Proposed Labels:**
+      - `kro.run/instance-gvk: mygroup.example.com/v1/MyKind`: The Group Version Kind (GVK) of the instance
```
chrisdoherty4, line 34: In practice it may be moot, but won't this violate the maximum label length for kinds bumping up against the 63 character limit?

docs/design/proposals/labels.md
chrisdoherty4, line 36: Similar length question here: both kind and instance names can be 63 characters long which is double the maximum of a label value.

docs/design/proposals/labels.md
```diff
@@ -0,0 +1,109 @@
...
+-   A label on resources created by an instance reconciliation to indicate which RGD they come from.
+-   Labels on resources created by an instance reconciliation to indicate which instance created them.
+
+This aligns with the common Kubernetes practice of using labels for discovery and organization of related resources, as seen in tools like Helm and Kustomize.
+
+#### Design details
+
+We will rename or introduce the following labels, which will be defined in `@pkg/metadata/labels.go`:
+
+1.  **Labels for mapping created resources back to the instance:**
```
chrisdoherty4, line 31: Does the ID still have some value? If a resource is recreated with the same spec the ID is different: perhaps that's useful - though admittedly I'm spitballing? This is also a 2-way door.

What was the motivation for dropping IDs?

chrisdoherty4 (commented):
I love the intent. A few questions on technicalities.

docs/design/proposals/labels.md
a-hilaly: Good point. Dropping these here for reviewers:
on instances:
```yaml
labels:
  kro.run/resource-graph-definition-id: <rgd-uid>
  kro.run/resource-graph-definition-name: <rgd-name>
  kro.run/owned: "true"
  kro.run/kro-version: v0.x.x
  applyset.kubernetes.io/id: <applyset-hash>
annotations:
  applyset.kubernetes.io/tooling: kro/<version>
  applyset.kubernetes.io/contains-group-kinds: <gk-list>
  applyset.kubernetes.io/additional-namespaces: <ns-list>
```

on child resources:
```yaml
labels:
  # Instance identity (6 labels)
  kro.run/instance-id: <instance-uid>
  kro.run/instance-name: <name>
  kro.run/instance-namespace: <namespace>
  kro.run/instance-group: <group>
  kro.run/instance-version: <version>
  kro.run/instance-kind: <kind>
  # RGD identity (2 labels)
  kro.run/resource-graph-definition-id: <rgd-uid>
  kro.run/resource-graph-definition-name: <rgd-name>
  # Node/ownership (4 labels)
  kro.run/node-id: <node-id>
  kro.run/owned: "true"
  kro.run/kro-version: v0.x.x
  app.kubernetes.io/managed-by: kro
  # ApplySet membership
  applyset.kubernetes.io/part-of: <applyset-hash>
  # Collection only (if forEach)
  kro.run/collection-index: <index>
  kro.run/collection-size: <size>

```

PR #646, opened 2025-08-22
feat: eager service account `SelfSubjectAccessReview` support when evaluating `ResourceGraphDefinition`

jakobmoellerdev:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  # ...
spec:
  defaultServiceAccounts:
    '*': configmap-creator
  resources:
  - id: configmap
    template:
      apiVersion: v1
      data:
        key: value
      kind: ConfigMap
      metadata:
        name: ${schema.spec.name}-configmap
  schema:
    apiVersion: v1alpha1
    group: kro.run
    kind: ConfigMapService
    spec:
      name: string
status:
  conditions:
  # ...
  - lastTransitionTime: "2025-08-22T21:22:15Z"
    message: |-
      specified service account system:serviceaccount:default:configmap-creator does not have permission to delete "configmap" (/v1, Resource=configmaps) in namespace default
      specified service account system:serviceaccount:default:configmap-creator does not have permission to patch "configmap" (/v1, Resource=configmaps) in namespace default
      specified service account system:serviceaccount:default:configmap-creator does not have permission to update "configmap" (/v1, Resource=configmaps) in namespace default
      specified service account system:serviceaccount:default:configmap-creator does not have permission to watch "configmap" (/v1, Resource=configmaps) in namespace default
    observedGeneration: 1
    reason: InvalidResourceGraph
    status: "False"
    type: Ready
  state: Inactive
```

Added preflight RBAC validation for every resource in a ResourceGraphDefinition (RGD).

- Selects the service account per resource based on spec.defaultServiceAccounts (namespace-specific override first, then * catch-all).
- Performs SelfSubjectAccessReview calls (create, update, delete, patch, watch) using impersonation of the selected service account.
- Runs these SARs concurrently and aggregates denials into a single error.
- Treats permission failures as “invalid resource graph,” marking the RGD Ready=False and surfacing a precise, multi-line message listing missing verbs per resource.
- Wired the authorization client and impersonation into the shared client set (pkg/client/set.go) and extended the fake client to support it for tests.

Added E2E tests: Namespaced and cluster-scoped permission checks that assert Ready=False with exact denial messages. A deletion flow test that applies an RGD, verifies, deletes it, and confirms the object is gone.

## Why it’s useful

- Fast, explicit failure: Users get immediate, actionable feedback when service accounts lack required verbs, before any workloads are created.
- Least-privilege safety: Encourages narrowly scoped RBAC by showing exactly which verbs are needed for each resource.
- Better UX and debuggability: Aggregated, deterministic messages show all missing permissions in one place and propagate to the Ready condition.

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller_reconcile.go
+++ b/pkg/controller/resourcegraphdefinition/controller_reconcile.go
@@ -16,11 +16,19 @@ package resourcegraphdefinition
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"maps"
+	"slices"
+	"sync"
 	"time"
 
+	"github.com/kro-run/kro/pkg/client"
+	authv1 "k8s.io/api/authorization/v1"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	authclient "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/kro-run/kro/api/v1alpha1"
@@ -122,7 +130,7 @@ func (r *ResourceGraphDefinitionReconciler) setupMicroController(
 
 // reconcileResourceGraphDefinitionGraph processes the resource graph definition to build a dependency graph
 // and extract resource information
-func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionGraph(_ context.Context, rgd *v1alpha1.ResourceGraphDefinition) (*graph.Graph, []v1alpha1.ResourceInformation, error) {
+func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionGraph(ctx context.Context, rgd *v1alpha1.ResourceGraphDefinition) (*graph.Graph, []v1alpha1.ResourceInformation, error) {
 	processedRGD, err := r.rgBuilder.NewResourceGraphDefinition(rgd)
 	if err != nil {
 		return nil, nil, newGraphError(err)
@@ -134,6 +142,11 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionGrap
 		if len(deps) > 0 {
 			resourcesInfo = append(resourcesInfo, buildResourceInfo(name, deps))
 		}
+	}
+
+	// make sure the service accounts configured in the resource graph definition have the required permissions.
+	if err := r.reconcileResourceGraphDefinitionPermissions(ctx, processedRGD.Resources, rgd.Spec.DefaultServiceAccounts); err != nil {
+		return nil, nil, err
 	}
 
 	return processedRGD, resourcesInfo, nil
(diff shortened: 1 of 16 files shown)


pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -189,3 +202,137 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
 func newMicroControllerError(err error) error { return &microControllerError{err} }
+
+func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionPermissions(ctx context.Context, resources map[string]*graph.Resource, accounts map[string]string) error {
+	var errs []error
+	for _, resource := range resources {
+		selectedServiceAccount := selectServiceAccountForResource(accounts, resource)
+		if err := verifyResourcePermissions(ctx, r.clientSet, selectedServiceAccount, resource); err != nil {
+			errs = append(errs, err)
+		}
+	}
```
fabianburth, line 213: Is there a reason not to run this concurrently as well?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -189,3 +202,137 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return results, nil
+}
+
+// getServiceAccountUserName builds the impersonate service account user name.
+// The format of the user name is "system:serviceaccount:<serviceaccount>"
```
fabianburth, line 332:
```suggestion
// The format of the user name is "system:serviceaccount:<namespace>:<serviceaccount>"
```
nit

fabianburth (commented):
I really like the idea and the usability gain introduced with this feature! 

I suppose you decided against `SelfSubjectRulesReview` since
- we wanted to actually fail in case of insufficient authorization (so, we cannot affort to have incomplete results, because otherwise we might wrongfully fail)
- we can handle a slight delay as we are not a UI

I figure with several large rgds, we'd quickly be doing quite a bunch of requests - especially at ramp up. Wouldn't we have to be concerned about rate limiting here (or is this less of a concern because each rgd might use a different user identity)?

jakobmoellerdev:
The main reason to not use `SelfSubjectRulesReview` is because it is not supposed to be used as such. https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/self-subject-rules-review-v1/#SelfSubjectRulesReview:

> The returned list of actions may be incomplete depending on the server's authorization mode, and any errors experienced during the evaluation. SelfSubjectRulesReview should be used by UIs to show/hide actions, or to quickly let an end user reason about their permissions. It should NOT Be used by external systems to drive authorization decisions as this raises confused deputy, cache lifetime/revocation, and correctness concerns. SubjectAccessReview, and LocalAccessReview are the correct way to defer authorization decisions to the API server.

The PR is still in draft because it is a proof of concept created based on another issue discussion and it should be optimized much further before being considered for performance reasons or similar IMO

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -136,6 +144,11 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionGrap
 		}
 	}
 
+	// make sure the service accounts configured in the resource graph definition have the required permissions.
+	if err := r.reconcileResourceGraphDefinitionPermissions(ctx, processedRGD.Resources, rgd.Spec.DefaultServiceAccounts); err != nil {
+		return nil, nil, err
+	}
```
a-hilaly, line 150: Do we want a special condition for this? i'm thinking that it shouldn't necessarily be Terminal (as in marking the RGD not Ready) - but more of an advisory condition informing the admins that one of their permissions set isn't configured correctly.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -189,3 +202,137 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+				Spec: authv1.SelfSubjectAccessReviewSpec{
+					ResourceAttributes: &authv1.ResourceAttributes{
+						Namespace: namespace,
+						Verb:      verb,
+						Group:     gvr.Group,
+						Version:   gvr.Version,
+						Resource:  gvr.Resource,
+					},
+				},
+			}
```
a-hilaly, line 304: I have some concerns about the practical applicability of this permission checking approach. The implementation assumes static namespace values, but in reality, some/maybe-majority of kro use cases leverage CEL expressions for dynamic namespace resolution. Users commonly write templates with expressions like `${schema.spec.something + "-test"}`, `${resource.configMap.metadata.namespace}`, or even conditional logic that resolves to different namespaces based on runtime conditions, thinking `${somethingBoolean? a: b}`.

I'm wondering if we should consider limiting this validation to resources that don't specify a namespace field at all, since those inherit the "instance" namespace by default. At least for that subset of resources, we could provide meaningful validation. Alternatively, we could delegate permission checking to the instance reconcilers, but this introduces its own challenges. Sometimes we'll only be able to evaluate permissions at certain points in the DAG execution (dynamic variables point to other resources status?), and by the time we discover permission issues, we may have already created some resources. This somewhat defeats the purpose of "eager" validation.

I'm questioning whether this feature provides enough value in its current form. Perhaps we should either constrain this to a smaller well defined scope where it can actually work reliably, or rethink the approach entirely to handle dynamic values appropriately for namespaces (as not allowing dynamic namespace settings)? would love to hear your thoughts @fabianburth @jakobmoellerdev

a-hilaly (commented):
Thank you @jakobmoellerdev for taking a stab at this, i left a few "philosophical" questions below.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: sure thing I think thats possible

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: Hmm im wondering about this. Wouldn't the emulated namespace for the resource be the only thing that we care about? as in, whatever `namespace` on the resource resolves to? And that is currenty the value im using here (im using `processedRGD.Resources`).
But maybe im just wrong on this :D My assumption currently is that the namespace value im using here is the pre-evaluated one from the graph analysis.

> Sometimes we'll only be able to evaluate permissions at certain points in the DAG execution (dynamic variables point to other resources status?), and by the time we discover permission issues, we may have already created some resources. This somewhat defeats the purpose of "eager" validation.

This is only true if KRO moves fully away from its graph evaluation. I see it like CEL itself. Sure there are some checks you could only do at runtime, but this should not be the norm. I think KRO should default to eager validation and to only fallback to runtime checks if nothing else works.

In an ideal world, KRO uses its graph to divide a topology into applicable "layers" in which each layer is validatable when all previous layers have been considered valid and applied as well.

jakobmoellerdev:
closing this as I believe we should build back the service account management spec. We need a better access model that is more sufficiently suited towards the dynamic integrations we plan on with kro

PR #649, opened 2025-08-28
Switching to the Kubernetes code of conduct

bridgetkromhout:
Switching to the K8s code of conduct


barney-s (approved):

barney-s:
image

We are asserting too many fields after update. The diff is exposed to any changes in defaulting behavior across k8s version. No need to block this PR. Will send a fix for the failure separately.

PR #651, opened 2025-08-30
docs: Update FAQ with instructions for using KRO with ArgoCD

RafPe:
As first step to enable easier adoption for KRO amongst ArgoCD  enabled organizations/engineers , this PR brings simple note in the FAQ section on how to handle the proper resource tracking.


RafPe:
Referencing original PR that inspired this  https://github.com/kro-run/kro/pull/366

website/docs/docs/faq.md
```diff
@@ -59,7 +59,26 @@ sidebar_position: 100
...
+
+   ```yaml
+   annotations:
+     argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"]}
+   ```
+
+   This annotation allows ArgoCD to properly track and manage the resources
+   created by KRO instances. 
+   
+   Note that this approach only works when using ArgoCD's default resource tracking
```
jakobmoellerdev: I think this works with `annotation` and `annotation+label`. Even `app.kubernetes.io/instance` should work if you set this label right? Would be cool if you could confirm this!

website/docs/docs/faq.md
```diff
@@ -59,7 +59,26 @@ sidebar_position: 100
...
+   to all templated resources in your ResourceGraphDefinition. 
+   
+   The following code needs to be added to each templated resource:
+
+   ```yaml
+   annotations:
+     argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"]}
+   ```
+
+   This annotation allows ArgoCD to properly track and manage the resources
```
jakobmoellerdev, line 82: nit: not strictly necessary, but it would be cool to insert a screenshot of the working tracking or the likes

jakobmoellerdev (commented):
Thank you for the contribution, this is great ❤️

I only think we should confirm exact behavior once with all tracking modes in Argo, apart from this, LGTM from me

website/docs/docs/faq.md
RafPe: It kind a does... With `annotation+label` the setup is that label is used more for infotmative purposes to different apps

> label is not used for tracking purposes, and the value is still truncated if longer than 63 characters. The annotation argocd.argoproj.io/tracking-id is used instead to track application resources

For just `label` its well documented `Argo CD uses the app.kubernetes.io/instance label` 

I have expanded the text to highlight the choice and re-point people to the documentation

website/docs/docs/faq.md
RafPe: Done

website/docs/docs/faq.md
```diff
@@ -59,7 +59,30 @@ sidebar_position: 100
...
+5. **How do I use KRO resources with ArgoCD?**
+
+   To use KRO resources with ArgoCD, you need to add a specific tracking annotation
+   to all templated resources in your ResourceGraphDefinition. 
+   
+   The following code needs to be added to each templated resource:
+
+   ```yaml
+   annotations:
+     argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"]}
```
RemyFV: if you're not creating resources that have an automatic explicit parent children relationship, it's also useful to define it in the RGD:

```
        metadata:
          ownerReferences:
            - apiVersion: kro.run/v1alpha1
              kind: ${schema.kind}
              name: ${schema.metadata.name}
              uid: ${schema.metadata.uid}
              blockOwnerDeletion: true
              controller: false
```

that way you get the neat parent -> children arrows and layout for things that otherwise would just be dropped at the same level as the RGD.

RemyFV: and since argoCD is [kind of in the middle of switching between labels and annotations](https://argo-cd.readthedocs.io/en/stable/user-guide/resource_tracking/), some people are going to want to do

```
          labels:
            app.kubernetes.io/instance: ${schema.metadata.?labels["app.kubernetes.io/instance"]}
```

instead of `annotations: - argocd.argoproj.io/tracking-id`, worth mentioning and adding a link to the doc maybe?

RemyFV: adding this info to the documentation will definitely help adoption, that's basically a hard requirement for us and I was pleasantly surprised to see how easy it was.

RafPe: @RemyFV Regarding the label+annotation and just labels - these are non default configurations and I have created the note in the doc sending the users to the source https://github.com/kro-run/kro/pull/651/files#diff-abb2bdd6923c61f13af63414af33fdc860e8be0b7ea5b456c23c5f9b204b669aR83 

I think it does not make sense to go with every possible scenario if we show one example being really good which can be easily adapted

RafPe: I have also added ownerRef into the example. This will in deed create appropiate child relationship in ArgoCD

RafPe:
@RemyFV @jakobmoellerdev @a-hilaly Could you guys let me know if there are any further changes I should make before we could merge this ? Happy to work these out :)

jakobmoellerdev:
in general I think we will merge this but the KRO Repo/Org is currently under transition to CNCF so we are very cautious with merges until we have that figured out (sorry for the delay)!

RafPe:
> in general I think we will merge this but the KRO Repo/Org is currently under transition to CNCF so we are very cautious with merges until we have that figured out (sorry for the delay)!

No probs! Just wanted to make sure the form of the PR and outstanding feedback is 👍

a-hilaly:
Hey @Rafal P! Thanks for pushing this forward. I wanted to share some context on the owner references situation:
We actually used to set owner references in kro, but had to stop because it doesn't play well with cross-namespace resources or combinations of cluster-scoped and namespaced resources - this causes unwanted behavior with the Kubernetes garbage collector. Do you know if Argo has any ways to support tracking in these scenarios?

I do like the idea of letting users set owner references themselves, though it does leak some implementation details that we'd ideally abstract away.

side note: I've been thinking about making the DAG parser smarter - when the stars align (all resources in the same namespace and all are namespaced (not cluster scoped)), kro could automatically start setting owner references. This would give us the best of both worlds in the common case. cc @barney-s @jakobmoellerdev @matthchr 

@RafPe Since we're still figuring out these details, maybe your PR would be better suited for the examples directory or the docs website rather than the front README page? That way we can document the workaround while keeping the main docs focused on stable patterns.

a-hilaly (commented):
Hey @Rafal P! Thanks for pushing this forward. I wanted to share some context on the owner references situation:
We actually used to set owner references in kro, but had to stop because it doesn't play well with cross-namespace resources or combinations of cluster-scoped and namespaced resources - this causes unwanted behavior with the Kubernetes garbage collector. Do you know if Argo has any ways to support tracking in these scenarios?

I do like the idea of letting users set owner references themselves, though it does leak some implementation details that we'd ideally abstract away.

side note: I've been thinking about making the DAG parser smarter - when the stars align (all resources in the same namespace and all are namespaced (not cluster scoped)), kro could automatically start setting owner references. This would give us the best of both worlds in the common case. cc @barney-s @jakobmoellerdev @matthchr 

@RafPe Since we're still figuring out these details, maybe your PR would be better suited for the examples directory or the docs website rather than the front README page? That way we can document the workaround while keeping the main docs focused on stable patterns.

RemyFV:
> Do you know if Argo has any ways to support tracking in these scenarios?

since that part is from me...

the tracking annotation or label puts the resources in the same argocd "application" as the RGD instance, the `ownerReference` puts the children under their parent (the arrow relationship you see in the PR screenshot).

I can see how this could be a problem if the RGD creates resources in a different namespace from the instance, I'll have to try that scenario to see how argocd behaves in this case. I'll test tomorrow and let you know.

ultimately both are what argocd users expect and `ownerReference` is a standard kubernetes pattern if I'm not mistaken - if a resource is created by something it should both be in the same app AND under its parents so if kro could do that automatically that would be nice.

RafPe:
> @RafPe Since we're still figuring out these details, maybe your PR would be better suited for the examples directory or the docs website rather than the front README page? That way we can document the workaround while keeping the main docs focused on stable patterns.

@a-hilaly  I would be more than happy to change and put this into the examples directory since ( hopefully ) its just a "workaround" until we can get a proper solution in. Just let me know that we take that route 


> I can see how this could be a problem if the RGD creates resources in a different namespace from the instance, I'll have to try that scenario to see how argocd behaves in this case. I'll test tomorrow and let you know.
> 

@RemyFV Could you share some examples for this scenario ? I am interested to get more context about the challenges with the ownerRef/CrossNamespaces and other edge scenarios. 

> ultimately both are what argocd users expect and `ownerReference` is a standard kubernetes pattern if I'm not mistaken - if a resource is created by something it should both be in the same app AND under its parents so if kro could do that automatically that would be nice.

Yes - that would be perfect situation - KRO adds the `ownerRef` into the defined resources and thanks to that we have the appropriate relationships configured. 




Out of interest how does Crossplane is handling this ? They must have had across the same problems as we are facing ?

a-hilaly:
>I would be more than happy to change and put this into the examples directory since ( hopefully ) its just a "workaround" until we can get a proper solution in. Just let me know that we take that route

@RafPe  This would be great, thank you! :)

>Could you share some examples for this scenario ? I am interested to get more context about the challenges with the ownerRef/CrossNamespaces and other edge scenarios.

@RafPe Here's a concrete example of the cross-namespace challenge:

```yaml
spec:
  schema:
    ns1: string
    ns2: string
  resources:
  - id: pod1
    template:
      kind: Pod
      metadata:
        namespace: ${schema.spec.ns1}
  - id: pod2
    template:
      kind: Pod
      metadata:
        namespace: ${schema.spec.ns2}
```

If you create an instance in `namespace-a` and it creates `pod1` in `namespace-b` and `pod2` in `namespace-c`, Kubernetes owner references break down. When KRO sets an owner reference from `pod1` to the instance, the K8s garbage collector looks for the owner in the same namespace as the owned resource (`namespace-b`). Since the instance is actually in `namespace-a`, the garbage collector thinks the owner doesn't exist and immediately deletes `pod1`. KRO then recreates it, and we get stuck in an infinite deletion/recreation loop.

This is why we had to remove automatic owner references. The same issue happens with cluster-scoped resources - you can't set an owner reference from a cluster-scoped resource to a namespaced one.

a-hilaly:
One wild idea I've been thinking about - what if kro created "shadow instances" when it detects resources spanning multiple namespaces? 

The concept would be:
- KRO detects that resources will be created in namespaces different from the instance
- It automatically creates lightweight, non-reconciling shadow instances in those namespaces
- These shadows would only exist to serve as owner reference targets
- The real instance would still do all the actual reconciliation

So for the example above, if you create an instance in `namespace-a` that deploys to `namespace-b` and `namespace-c`, kro would:
1. watch the real instance in `namespace-a`
2. Create shadow instances in `namespace-b` and `namespace-c`
3. Set owner references from resources to their namespace-local shadow
4. However the two shadow instances cannot point to the main instance for the same reasons we've been talking about :)

What do you all think? Too magical/complex or worth exploring? cc @jakobmoellerdev @fabianburth @barney-s

RemyFV:
so, results of the test:

```
   ownerReferences:
         - apiVersion: kro.run/v1alpha1
           kind: ${schema.kind}
           name: ${schema.metadata.name}
           uid: ${schema.metadata.uid}
           blockOwnerDeletion: true
           controller: false
```
the `kind` has to be hardcoded: `kind: <parent RGD kind>` because `${schema.kind}` returns

>failed to validate resource CEL expressions: failed to ensure resource
>configMap1 expressions: failed to dry-run expression schema.kind: failed
>to dry-run expression schema.kind: failed to evaluate expression: no such
>key: kind

I swear it worked with `${schema.kind}` at some point on v0.4.1 <.<
it would be nice if kro surfaced an event for this instead of a transition message, because it looks like the RGD is deployed properly unless you go and check the manifest in cluster.

it's also a weird error because `${schema.kind}` does exist.

for multiple namespaces:

if the RGD instance creates objects they HAVE to be either unscoped or in the same namespace. deploying something in `default` from an instance in `tests-remy` never worked at all (even though kro is deployed with cluster-wide permissions).
from the ArgoCD interface the object in the other namespace never appears and the only info you get is in the instance's manifest that says "error".

in the case of an ArgoCD deployment, leaving `namespace` unset is perfectly fine (and probably the recommended approach) as everything will just inherit the parent's namespace (application -> instance -> children).

RafPe:
Now most of the deployments/configuration scenarios for me does not span multiple resources. Interestingly if you look at ArgoCD it does have the namespaace destination for your resources which essentially create the resources in a given namespace - want other namespace then create other app ? 

So question ( that must have for sure be answered ) are we not trying to overengineer this one with multinamespaces support ?

RafPe:
> One wild idea I've been thinking about - what if kro created "shadow instances" when it detects resources spanning multiple namespaces?
> 
> The concept would be:
> 
>     * KRO detects that resources will be created in namespaces different from the instance
> 
>     * It automatically creates lightweight, non-reconciling shadow instances in those namespaces
> 
>     * These shadows would only exist to serve as owner reference targets
> 
>     * The real instance would still do all the actual reconciliation
> 
> 
> So for the example above, if you create an instance in `namespace-a` that deploys to `namespace-b` and `namespace-c`, kro would:
> 
>     1. watch the real instance in `namespace-a`
> 
>     2. Create shadow instances in `namespace-b` and `namespace-c`
> 
>     3. Set owner references from resources to their namespace-local shadow
> 
>     4. However the two shadow instances cannot point to the main instance for the same reasons we've been talking about :)
> 
> 
> What do you all think? Too magical/complex or worth exploring? cc @jakobmoellerdev @fabianburth @barney-s

For a side way observer this feels very much complex

barney-s:
@RafPe Please move the image into the docs folder. 
The doc build is failing. 
May be here: https://github.com/kubernetes-sigs/kro/tree/main/website/static/img

RafPe:
> @RafPe Please move the image into the docs folder. The doc build is failing. May be here: https://github.com/kubernetes-sigs/kro/tree/main/website/static/img

@barney-s I have moved the image and updated the reference. Let me know if this works :)

a-hilaly:
Hey @RafPe looks like the docs build isn't passing, can you PTAL?

RafPe:
> Hey @RafPe looks like the docs build isn't passing, can you PTAL?

On it

a-hilaly:
it's live https://kro.run/docs/faq 🎉

PR #653, opened 2025-08-30
api: add AdditionalPrinterColumnPolicy and merge logic for additional…

yashpal2104:
… printer columns

### API change

###What the policy does (simple)

###Backwards compatibility

2. For columns with neither Name nor JSONPath the current code uses a provisional fallback key (pointer-based placeholder). This is not stable across runs; recommend choosing one:
use the default slice index as a stable fallback key,

### DIFF
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -19,10 +19,14 @@ import (
 	runtime "k8s.io/apimachinery/pkg/runtime"
 )
 
+type AdditionalPrinterColumnPolicy string
+
 const (
 	// DefaultServiceAccountKey is the key to use for the default service account
 	// in the serviceAccounts map.
 	DefaultServiceAccountKey = "*"
+	AdditionalPrinterColumnPolicyReplace AdditionalPrinterColumnPolicy = "Replace"
+	AdditionalPrinterColumnPolicyAdd     AdditionalPrinterColumnPolicy = "Add"
 )
 
 // ResourceGraphDefinitionSpec defines the desired state of ResourceGraphDefinition
@@ -86,11 +90,24 @@ type Schema struct {
 	// Validation is a list of validation rules that are applied to the
 	// resourcegraphdefinition.
 	Validation []Validation `json:"validation,omitempty"`
-	// AdditionalPrinterColumns defines additional printer columns
-	// that will be passed down to the created CRD. If set, no
-	// default printer columns will be added to the created CRD,
-	// and if default printer columns need to be retained, they
-	// need to be added explicitly.
+	// AdditionalPrinterColumnPolicy defines additional printer columns
+	// that will be passed down to the created CRD. The Default tag is "Replace", that only uses the provided columns, "Add:" tag Merge provided columns with defaults.
+	//
+	// AdditionalPrinterColumnPolicy controls how provided additional printer
+	// columns are applied to the generated CRD.
+	//
+	// - "Replace": only use the provided columns; if none are provided, fall
+	//   back to the operator defaults (backwards compatible default).
+	// - "Add": merge the provided columns with the defaults, with user
+	//   definitions overriding defaults when they share a identifying key
+	//   (Name preferred, fallback to JSONPath).
+	//
+	// +kubebuilder:validation:Enum=Replace;Add
+	// +kubebuilder:default=Replace
+	AdditionalPrinterColumnPolicy AdditionalPrinterColumnPolicy `json:"additionalPrinterColumnPolicy,omitempty"`
+
+	// AdditionalPrinterColumns is the list of user-specified printer columns
+	// to apply according to AdditionalPrinterColumnPolicy.
 	//
 	// +kubebuilder:validation:Optional
 	AdditionalPrinterColumns []extv1.CustomResourceColumnDefinition `json:"additionalPrinterColumns,omitempty"`
--- a/pkg/graph/crd/crd.go
+++ b/pkg/graph/crd/crd.go
@@ -27,12 +27,13 @@ import (
 
 // SynthesizeCRD generates a CustomResourceDefinition for a given API version and kind
 // with the provided spec and status schemas~
-func SynthesizeCRD(group, apiVersion, kind string, spec, status extv1.JSONSchemaProps, statusFieldsOverride bool, additionalPrinterColumns []extv1.CustomResourceColumnDefinition) *extv1.CustomResourceDefinition {
+func SynthesizeCRD(group, apiVersion, kind string, spec, status extv1.JSONSchemaProps, statusFieldsOverride bool, policy v1alpha1.AdditionalPrinterColumnPolicy, additionalPrinterColumns []extv1.CustomResourceColumnDefinition) *extv1.CustomResourceDefinition {
 	crdGroup := group
 	if crdGroup == "" {
 		crdGroup = v1alpha1.KRODomainName
 	}
-	return newCRD(crdGroup, apiVersion, kind, newCRDSchema(spec, status, statusFieldsOverride), additionalPrinterColumns)
+	merged := newCRDAdditionalPrinterColumns(policy, additionalPrinterColumns)
+	return newCRD(crdGroup, apiVersion, kind, newCRDSchema(spec, status, statusFieldsOverride), merged)
 }
 
 func newCRD(group, apiVersion, kind string, schema *extv1.JSONSchemaProps, additionalPrinterColumns []extv1.CustomResourceColumnDefinition) *extv1.CustomResourceDefinition {
@@ -62,7 +63,7 @@ func newCRD(group, apiVersion, kind string, schema *extv1.JSONSchemaProps, addit
 					Subresources: &extv1.CustomResourceSubresources{
 						Status: &extv1.CustomResourceSubresourceStatus{},
 					},
-					AdditionalPrinterColumns: newCRDAdditionalPrinterColumns(additionalPrinterColumns),
+					AdditionalPrinterColumns: newCRDAdditionalPrinterColumns(v1alpha1.AdditionalPrinterColumnPolicyReplace, additionalPrinterColumns),
 				},
 			},
 		},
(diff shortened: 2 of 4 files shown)


yashpal2104:
@jakobmoellerdev Do let me know about your thoughts on this PR. Also adding code review on the existing logic of using the merge key in the `crd.go` file

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,11 +90,24 @@ type Schema struct {
 	// Validation is a list of validation rules that are applied to the
 	// resourcegraphdefinition.
 	Validation []Validation `json:"validation,omitempty"`
-	// AdditionalPrinterColumns defines additional printer columns
-	// that will be passed down to the created CRD. If set, no
-	// default printer columns will be added to the created CRD,
-	// and if default printer columns need to be retained, they
-	// need to be added explicitly.
+	// AdditionalPrinterColumnPolicy defines additional printer columns
+	// that will be passed down to the created CRD. The Default tag is "Replace", that only uses the provided columns, "Add:" tag Merge provided columns with defaults.
```
jakobmoellerdev:
```suggestion
	// that will be passed down to the created CRD. The default policy is "Replace".
```

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,11 +90,24 @@ type Schema struct {
...
+	//
+	// AdditionalPrinterColumnPolicy controls how provided additional printer
+	// columns are applied to the generated CRD.
```
jakobmoellerdev:
```suggestion
	// columns via AdditionalPrinterColumns are applied to the generated CRD.
```

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,11 +90,24 @@ type Schema struct {
...
+	//
+	// - "Replace": only use the provided columns; if none are provided, fall
```
jakobmoellerdev:
```suggestion
	// - "Replace": uses the provided columns as is; if no or empty AdditionalPrinterColumns are provided, fall
```

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,11 +90,24 @@ type Schema struct {
...
+	//   back to the operator defaults (backwards compatible default).
```
jakobmoellerdev:
```suggestion
	//   back to the default columns.
```

Also: Please insert the default columns here so they are documented

pkg/graph/crd/crd.go
```diff
@@ -27,12 +27,13 @@ import (
...
 // SynthesizeCRD generates a CustomResourceDefinition for a given API version and kind
 // with the provided spec and status schemas~
-func SynthesizeCRD(group, apiVersion, kind string, spec, status extv1.JSONSchemaProps, statusFieldsOverride bool, additionalPrinterColumns []extv1.CustomResourceColumnDefinition) *extv1.CustomResourceDefinition {
+func SynthesizeCRD(group, apiVersion, kind string, spec, status extv1.JSONSchemaProps, statusFieldsOverride bool, policy v1alpha1.AdditionalPrinterColumnPolicy, additionalPrinterColumns []extv1.CustomResourceColumnDefinition) *extv1.CustomResourceDefinition {
 	crdGroup := group
 	if crdGroup == "" {
 		crdGroup = v1alpha1.KRODomainName
 	}
-	return newCRD(crdGroup, apiVersion, kind, newCRDSchema(spec, status, statusFieldsOverride), additionalPrinterColumns)
+	merged := newCRDAdditionalPrinterColumns(policy, additionalPrinterColumns)
```
jakobmoellerdev: nit:
```suggestion
	printerColumns := newCRDAdditionalPrinterColumns(policy, additionalPrinterColumns)
```
merged is a misnomer, because they are not always merged.

pkg/graph/crd/crd.go
```diff
@@ -62,7 +63,7 @@ func newCRD(group, apiVersion, kind string, schema *extv1.JSONSchemaProps, addit
 					Subresources: &extv1.CustomResourceSubresources{
 						Status: &extv1.CustomResourceSubresourceStatus{},
 					},
-					AdditionalPrinterColumns: newCRDAdditionalPrinterColumns(additionalPrinterColumns),
+					AdditionalPrinterColumns: newCRDAdditionalPrinterColumns(v1alpha1.AdditionalPrinterColumnPolicyReplace, additionalPrinterColumns),
```
jakobmoellerdev: arent the additional printer columns here already the finished ones? why do you need to reuse the method here? You already set the final columns in https://github.com/kro-run/kro/pull/653/files#diff-4be45bed5efa573c9cfcf7c8eec372cb4de316a24e6709f1bbda6e60722685a0R35 no?

pkg/graph/crd/crd.go
```diff
@@ -103,10 +104,42 @@ func newCRDSchema(spec, status extv1.JSONSchemaProps, statusFieldsOverride bool)
 	}
 }
 
-func newCRDAdditionalPrinterColumns(additionalPrinterColumns []extv1.CustomResourceColumnDefinition) []extv1.CustomResourceColumnDefinition {
-	if len(additionalPrinterColumns) == 0 {
-		return defaultAdditionalPrinterColumns
+func newCRDAdditionalPrinterColumns(policy v1alpha1.AdditionalPrinterColumnPolicy, userCols []extv1.CustomResourceColumnDefinition) []extv1.CustomResourceColumnDefinition {
```
jakobmoellerdev: Nit: There are a few logic paths here that imo you can simply omit because they are very weird (but maybe im just oversimplifying, happy to hear your thoughts):

1. You do NOT need to deduplicate based on JSONPath, because two columns with the same JSONPath are allowed (even though theyre weird, we shouldnt opinionate to override them)
2. You DO need to override based on Name, because a unique name is part of the contract of the columns.
3. I believe you can assume that the policy is always set to AdditionalPrinterColumnPolicyReplace because you default it with kubebuilder
4. You can short-circuit to defaultAdditionalPrinterColumns if userCols are empty for both add  and replace

So you can drop the keyOf inner function and simplify to

```go
func newCRDAdditionalPrinterColumns(
	policy v1alpha1.AdditionalPrinterColumnPolicy,
	userCols []extv1.CustomResourceColumnDefinition,
) []extv1.CustomResourceColumnDefinition {
	if len(userCols) == 0 {
		return defaultAdditionalPrinterColumns
	}

	if policy == v1alpha1.AdditionalPrinterColumnPolicyReplace {
		return userCols
	}

	merged := make([]extv1.CustomResourceColumnDefinition, len(defaultAdditionalPrinterColumns), len(defaultAdditionalPrinterColumns)+len(userCols))
	copy(merged, defaultAdditionalPrinterColumns)

	index := make(map[string]int, len(merged))
	for i, c := range merged {
		index[c.Name] = i
	}

	for _, u := range userCols {
		if pos, ok := index[u.Name]; ok {
			merged[pos] = u
		} else {
			merged = append(merged, u)
		}
	}

	return merged
}
```

jakobmoellerdev: No matter what you implement here, I am missing unit tests for this!

jakobmoellerdev:
Generally PR is sound and you can finalize it. There are some improvements in the code and the wording but the general idea is great 🥳

yashpal2104:
@jakobmoellerdev Do let me know if there is something else I can help you with 😊!

jakobmoellerdev (changes requested):
you are missing tests and a regeneration of the CRD here!

yashpal2104:
Do let me know if there is anything else I may be missing or need clarification!

jakobmoellerdev (commented):
For me this is LGTM. PTAL @a-hilaly when you got a spare minute 🙇

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -19,10 +19,14 @@ import (
...
 )
 
+type AdditionalPrinterColumnPolicy string
+
 const (
 	// DefaultServiceAccountKey is the key to use for the default service account
 	// in the serviceAccounts map.
 	DefaultServiceAccountKey = "*"
+	AdditionalPrinterColumnPolicyReplace AdditionalPrinterColumnPolicy = "Replace"
+	AdditionalPrinterColumnPolicyAdd     AdditionalPrinterColumnPolicy = "Add"
```
a-hilaly: Please document the types and the constants.

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,11 +90,23 @@ type Schema struct {
...
+	// AdditionalPrinterColumnPolicy defines additional printer columns
+	// that will be passed down to the created CRD. The default policy is "Replace".
+	//
+	// AdditionalPrinterColumnPolicy controls how provided additional printer
+	// columns via AdditionalPrinterColumns are applied to the generated CRD.
+	//
+	// - "Replace": uses the provided columns as is; if no or empty AdditionalPrinterColumns are provided, fallback to the default columns.
+	// - "Add": merge the provided columns with the defaults, with user
+	//   definitions overriding defaults when they share a identifying key
+	//   (Name preferred, fallback to JSONPath).
```
a-hilaly: How about `Append/Prepend` instead?

pkg/graph/crd/crd.go
```diff
@@ -103,10 +111,46 @@ func newCRDSchema(spec, status extv1.JSONSchemaProps, statusFieldsOverride bool)
...
+) []extv1.CustomResourceColumnDefinition {
+	// Return defaults if no user columns are provided
+	if len(userCols) == 0 {
 		return defaultAdditionalPrinterColumns
 	}
 
+	// Default to Replace policy if empty
+	if policy == "" {
+		policy = v1alpha1.AdditionalPrinterColumnPolicyReplace
+	}
```
a-hilaly: This can be configured at the RGD CRD level via kubebuilder markers

yashpal2104:
I have added chainsaw-tests too for this

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -19,10 +19,19 @@ import (
...
+// AdditionalPrinterColumnPolicy defines how additional printer columns are applied to the generated CustomResourceDefinition (CRD). This policy controls whether user-provided columns replace or merge with the default columns.
+type AdditionalPrinterColumnPolicy string
+
 const (
 	// DefaultServiceAccountKey is the key to use for the default service account
 	// in the serviceAccounts map.
 	DefaultServiceAccountKey = "*"
+	
+	// AdditionalPrinterColumnPolicyReplace specifies that the provided additional printer columns should completely replace the default columns. If no AdditionalPrinterColumns are provided or the list is empty, the system will fall back to using the default columns.
+	AdditionalPrinterColumnPolicyReplace AdditionalPrinterColumnPolicy = "Replace"
```
a-hilaly: please use multi line string for comments e.g

```suggestion
	// AdditionalPrinterColumnPolicyReplace specifies that the provided additional printer columns should
	// completely replace the default columns. If no AdditionalPrinterColumns are provided or the list is
	// empty, the system will fall back to using the default columns.
	AdditionalPrinterColumnPolicyReplace AdditionalPrinterColumnPolicy = "Replace"

	// AdditionalPrinterColumnPolicyAdd specifies that the provided additional printer columns should be appended to the default columns. When there are conflicts (columns with the same Name or JSONPath), the user-provided columns take precedence over the defaults.
```

pkg/graph/crd/crd_test.go
```diff
@@ -289,3 +289,232 @@ func TestNewCRDSchema(t *testing.T) {
...
+			userCols: []extv1.CustomResourceColumnDefinition{
+				{Name: "State", JSONPath: ".status.customState", Type: "string"}, // override default State
+			},
+			want: func() []extv1.CustomResourceColumnDefinition {
+				// Create expected result with State overridden
+				result := make([]extv1.CustomResourceColumnDefinition, len(defaultAdditionalPrinterColumns))
+				copy(result, defaultAdditionalPrinterColumns)
+				// Find and override State column
+				for i, col := range result {
+					if col.Name == "State" {
```
a-hilaly: The golint is complaining that "State" is hardcoded in the code 3 times in this file, just declare it as a constant on top of the file and reuse it here.

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,11 +95,23 @@ type Schema struct {
...
+	//
+	// - "Replace": uses the provided columns as is; if no or empty AdditionalPrinterColumns are provided, fallback to the default columns.
+	// - "Add": appends the provided columns to the defaults, with user
+	//   definitions overriding defaults when they share a identifying key
+	//   (Name preferred, fallback to JSONPath).
+	//
+	// +kubebuilder:validation:Enum=Replace;Add
+	// +kubebuilder:default=Replace
+	AdditionalPrinterColumnPolicy AdditionalPrinterColumnPolicy `json:"additionalPrinterColumnPolicy,omitempty"`
+
```
a-hilaly, line 111:
```suggestion
```

api/v1alpha1/resourcegraphdefinition_types.go
a-hilaly, line 1: this file needs to be formatted by `gofmt`, most editors will automatically format the file, on save, if you install a go extension

pkg/graph/crd/crd_test.go
a-hilaly, line 1: same, formatting issues

yashpal2104:
I have added all the necessary changes to this PR and tested the changes locally everything seems ok to me

a-hilaly:
@yashpal2104 looks like some tests are failing, mind taking a look?
/ok-to-test

a-hilaly:
@yashpal2104 looks like this branch needs a rebase and tests are still failing PTAL

yashpal2104:
opened a new PR #790

PR #654, opened 2025-08-30
Multi Error Validation

HeeManSu:
Currently, Kro validation returns only single error. But, for the LSP with VS Code as the extension client. We need to show all errors at once. 

This PR includes multiple errors validation with extra context. So, LSP can show diagnostics at the correct positions.

Currently, I have implemented multiple errors for; 

- [ ] naming conventions errors  (validateResourceGraphDefinitionNamingConventions)
- [ ] build resource errors  (buildRGResource)
- [ ] build instance errors (buildInstanceResource)
- [ ] CEL expression errors (validateResourceCELExpressions)
- [ ] build dependecy graph errors (buildDependencyGraph)

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -106,6 +106,9 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
 	// original object.
 	rgd := originalCR.DeepCopy()
 
+	// Collect all the errors from ALL validation stages
+	allErrors := []error{}
+
 	// There are a few steps to build a resource graph definition:
 	// 1. Validate the naming convention of the resource graph definition and its resources.
 	//    kro leverages CEL expressions to allow users to define new types and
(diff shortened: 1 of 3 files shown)


pkg/graph/builder.go
```diff
@@ -269,42 +300,105 @@ func (b *Builder) buildExternalRefResource(
...
+				"",
+				fmt.Sprintf("resources.%s", rgResource.ID),
+				fmt.Errorf("exactly one of template or externalRef must be provided"),
+			),
+		)
+	}
+
+	// If we couldn't get a valid resource object, return early with errors
+	if len(errors) > 0 {
+		return nil, &MultiError{Errors: errors}
```
jakobmoellerdev, line 344: idk about these error returns (and a few below here) becausee you can basically just return it immediately in the error branch instead of doing this length check, because up until this point you cannot have multiple errors at once

a-hilaly:
@jakobmoellerdev I agree with your suggestion - let's use `errors.Join()/Unwrap()` instead of the custom `MultiError` type. 

What do you think about how we should expose these errors in the status conditions? i'm thinking that we probably want to provide useful information without overwhelming the error messages. my initial thoughts here is that we could have a summary condition (GraphVerified) like "2 resources failed validation: myResource1 and myResource2" and then separate conditions for each failure like "resource myResource1 failed: invalid CEL expression in readyWhen" and "resource myResource2 failed: invalid template structure". Still not sure what's the right balance but I would love to explore this path.

jakobmoellerdev:
/close

please feel free to reopen if you wanna revisit this

PR #655, opened 2025-08-31
kro cli publish command

DhairyaMajmudar:
## Description 

The new kro cli publish command is useful to publish an OCI image into container registries like docker hub, aws, gcr, etc....

## Usage

 `kro package -f [FILE] -r [REGISTRY_URL] -u [USER_NAME] -p [PASSWORD]`

Flags:

| Shorthand | Flag | Usage | 
| ---- | ----- | --- |
 | -r |  --ref | Remote reference to publish the OCI image to (e.g., 'ghcr.io/user/repo:tag') |
 | -u  | --username | Username for the remote registry (Optional) |
 | -p | --password |  Password for the remote registry (Optional) |

**Note:** If username or password is not provided via flags the cli will prompt the user to provide them in further steps.

## Demo

[kro-publish.webm](https://github.com/user-attachments/assets/d5ff44c6-c68d-42ce-9d0f-815dbe9a56db)

### DIFF
--- a/cmd/kro/commands/publish/publish.go
+++ b/cmd/kro/commands/publish/publish.go
@@ -1,0 +1,244 @@
+// Copyright 2025 The Kube Resource Orchestrator Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"archive/tar"
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content/oci"
+	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/retry"
+)
+
+type PublishConfig struct {
+	ociTarballPath string
+	remoteRef      string
+	username       string
+	password       string
+}
+
+var publishConfig = &PublishConfig{}
+
+func init() {
+	publishCmd.PersistentFlags().StringVarP(&publishConfig.ociTarballPath,
+		"file", "f", "",
+		"Path to the OCI image tarball created by the 'package' command",
+	)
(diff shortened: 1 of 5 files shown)


go.mod
a-hilaly, line 1: @DhairyaMajmudar Can we have a seperate go.mod file for the CLI? looks like our dependencies are growing with each command that we're adding

a-hilaly: same comment as in #656

cmd/kro/commands/root.go
```diff
@@ -18,10 +18,12 @@ import (
 	"github.com/spf13/cobra"
 
 	generate "github.com/kro-run/kro/cmd/kro/commands/generate"
+	publish "github.com/kro-run/kro/cmd/kro/commands/publish"
 	validate "github.com/kro-run/kro/cmd/kro/commands/validate"
```
a-hilaly, line 22: ditto

cmd/kro/commands/publish/publish.go
```diff
@@ -0,0 +1,235 @@
...
+
+		if publishConfig.password == "" {
+			fmt.Print("Password: ")
+			bytePassword, err := term.ReadPassword(int(os.Stdin.Fd()))
+			fmt.Println()
+			if err != nil {
+				return fmt.Errorf("failed to read password: %w", err)
+			}
+			publishConfig.password = string(bytePassword)
+		}
```
a-hilaly, line 98: Users have to pass the password each time they need to publish their files?

cmd/kro/commands/publish/publish.go
```diff
@@ -0,0 +1,235 @@
...
+
+		ctx := context.Background()
```
a-hilaly, line 100: nit: context not used yet.
also let's have a timeout context instead?

cmd/kro/commands/publish/publish.go
```diff
@@ -0,0 +1,235 @@
...
+		fmt.Println("Found manifest to push:", rootDesc.Digest)
+
+		fmt.Println("Publishing to", publishConfig.remoteRef, "...")
+		_, err = oras.Copy(ctx, store, rootDesc.Digest.String(), repo, publishConfig.remoteRef, oras.DefaultCopyOptions)
+		if err != nil {
+			return fmt.Errorf("failed to publish OCI image: %w", err)
+		}
+
+		fmt.Println("Successfully published OCI image to", publishConfig.remoteRef)
+		return nil
```
a-hilaly, line 156: I think we're missing the `kro package` command, i wouldn't expect `kro publish` to do any package/extracting here. It should just get a tarball file and sent it to OCI

tjamet:
hi @DhairyaMajmudar I'm curious to understand how you integrate OCIs with kro
That would help understand how we could integrate OCI packages push in KRO CLI

tjamet:
It seems this, together with #617 and #656, aims to make the kro CLI an alternative to kustomize or helm for building, publishing and deploying manifests.
Is deploying RGDs a goal the Kro CLI should pursue?
There are already plenty of options for this (kustomize, argocd, helm, oras, ....). I doubt Kro can add significant value by entering this field.
The Kro CLI should instead be about providing tools to validate, debug, manage, and troubleshoot RGDs and instances in live clusters or simulate its own behaviour (generate CRD, render RGD instance , ...) locally

a-hilaly:
Closing for now - @DhairyaMajmudar we'll need a KREP doc on the CLI before we move forward.

PR #661, opened 2025-09-06
feat(simpleschema): add support for pattern, minLength, maxLengt, uniqueItems, maxItems and minItems markers

simonfuhrer:
Hi

I noticed that simpleschema currently doesn’t support some common string-related JSON Schema markers such as pattern, minLength, and maxLength.
Since these are widely used in k8s, I was a bit surprised they weren’t already supported.

In this PR i added new marker types: pattern, minLength, maxLength and uniqueItems

Question for the maintainers:
Is there a specific reason why these markers were not supported so far (e.g., intentionally left out), or was it more a case of “not yet implemented”?

I opened this PR mainly to start the discussion. If these markers are welcome, I’d be happy to extend the PR with docs and tests.

Thanks for the feedback 🙏

### DIFF
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -16,6 +16,7 @@ package simpleschema
 
 import (
 	"fmt"
+	"regexp"
 	"slices"
 	"strconv"
 	"strings"
@@ -30,6 +31,7 @@ const (
 	keyTypeBoolean = string(AtomicTypeBool)
 	keyTypeNumber  = "number"
 	keyTypeObject  = "object"
+	keyTypeArray   = "array"
 )
 
 // A predefinedType is a type that is predefined in the schema.
@@ -202,7 +204,7 @@ func (tf *transformer) handleSliceType(key, fieldType string) (*extv1.JSONSchema
 	}
 
 	fieldJSONSchemaProps := &extv1.JSONSchemaProps{
-		Type: "array",
+		Type: keyTypeArray,
 		Items: &extv1.JSONSchemaPropsOrArray{
 			Schema: &extv1.JSONSchemaProps{},
 		},
@@ -225,6 +227,7 @@ func (tf *transformer) handleSliceType(key, fieldType string) (*extv1.JSONSchema
 	return fieldJSONSchemaProps, nil
 }
 
+//nolint:gocyclo
 func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Marker, key string, parentSchema *extv1.JSONSchemaProps) error {
 	for _, marker := range markers {
 		switch marker.MarkerType {
@@ -265,7 +268,7 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
 			}
 			schema.Maximum = &val
 		case MarkerTypeValidation:
-			if marker.Value == "" {
+			if strings.TrimSpace(marker.Value) == "" {
 				return fmt.Errorf("validation failed")
 			}
 			validation := []extv1.ValidationRule{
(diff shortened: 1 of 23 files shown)


pkg/simpleschema/transform.go
```diff
@@ -265,8 +266,8 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
 			}
 			schema.Maximum = &val
 		case MarkerTypeValidation:
-			if marker.Value == "" {
-				return fmt.Errorf("validation failed")
+			if strings.TrimSpace(marker.Value) == "" {
```
jakobmoellerdev, line 271: why do you need a trim space here?

simonfuhrer: I added TrimSpace because the existing Empty validation test treats whitespace-only values as invalid, which would otherwise slip through.

simonfuhrer:
Thanks for the feedback! I’ve now added the tests and also updated the docs. Additionally, I’ve added another marker: uniqueItems. Does this look good, or is there anything else I should do?

a-hilaly:
This is very cool, thank you @simonfuhrer ! maybe we can add some chainsaw e2e to validate the CRD open api schema when all of these new markers are used?

simonfuhrer:
Hi @a-hilaly 

I've implemented the chainsaw e2e tests to validate the CRD OpenAPI schema generation with all the new markers.

The e2e tests were incredibly valuable - they immediately caught a real issue with the uniqueItems marker! Kubernetes was throwing a "complexity error" because of it. After some research, I found that Kubernetes recommends using `x-kubernetes-list-type: "set"` instead of "uniqueItems: true" for CRDs to avoid runtime complexity issues.

**Changes made:**

- Added comprehensive chainsaw e2e tests in test/e2e/chainsaw/check-validation-markers
- Tests validate complete pipeline: RGD → CRD generation → instance validation -> configMap validation
- Fixed uniqueItems implementation to use x-kubernetes-list-type: "set"
- Tests cover both valid instances and validation error cases

**E2E test coverage:**

- CRD schema generation with all validation constraints
- Valid instance creation with compliant data
- Invalid instance rejection for constraint violations (empty arrays, oversized arrays, etc.)

The e2e tests now pass cleanly and validate that the OpenAPI schema is correctly generated with all the new validation markers working as expected in a real Kubernetes environment!

**Marker naming:** 
I kept the uniqueItems marker name since it's semantically correct and familiar to developers, but the implementation now properly uses x-kubernetes-list-type: "set" under the hood as recommended by Kubernetes.

what do you think about the naming of this marker?

pkg/simpleschema/transform.go
```diff
@@ -302,6 +303,79 @@ func (tf *transformer) applyMarkers(schema *extv1.JSONSchemaProps, markers []*Ma
...
+				return fmt.Errorf("invalid pattern regex: %w", err)
+			}
+			schema.Pattern = marker.Value
+		case MarkerTypeUniqueItems:
+			// UniqueItems is only valid for array types
+			switch isUnique, err := strconv.ParseBool(marker.Value); {
+			case err != nil:
+				return fmt.Errorf("failed to parse uniqueItems marker value: %w", err)
+			case parentSchema == nil:
+				return fmt.Errorf("uniqueItems marker can't be applied; parent schema is nil")
```
a-hilaly: You don't need this block in here. The reason why we are doing that for `case MarkerTypeRequired` is that required fields can only be configure at the parentSchema (`parentSchema.Required = append(parentSchema.Required, key)`)

test/e2e/chainsaw/check-validation-markers/instancecrd-assert.yaml
a-hilaly, line 1: this is definetely what I thought about for e2e tests. however validating whether CRD validation layer works or not feels slightly out of scope for us (thinking of `test/e2e/chainsaw/check-validation-markers/invalid-instance-priorities-max.yaml`) - but i'm very happy to keep it. thoughts @jakobmoellerdev ?

website/docs/docs/concepts/10-simple-schema.md
a-hilaly, line 1: 👍

pkg/simpleschema/transform_test.go
```diff
@@ -527,6 +527,350 @@ func TestBuildOpenAPISchema(t *testing.T) {
...
+				Type:        "object",
+				UniqueItems: true,
+				Properties: map[string]extv1.JSONSchemaProps{
+					"tags": {
+						Type: "array",
+						Items: &extv1.JSONSchemaPropsOrArray{
+							Schema: &extv1.JSONSchemaProps{Type: "string"},
+						},
+					},
+				},
```
a-hilaly: Do you need to x-validation-set in here?

simonfuhrer: Good point — I’ve added it here and also in a few other places where it applies.

test/e2e/chainsaw/check-validation-markers/instancecrd-assert.yaml
jakobmoellerdev: when I saw this I was thinking its nice someone does this but I agree this isnt something we need to strictly test against. If you want to test against the CRD you should only test against the fields that are truly relevant, (i.e. acceptance and the spec fields). I will say it is quite valuable to determine breaking changes for rendering in our schema logic in the future though!

jakobmoellerdev: TLDR: +1 to keeping it, but I wouldnt expect it for every contribution

simonfuhrer:
linting issues fixed.

PR #664, opened 2025-09-09
expose leader election namespace in args

michaelhtm:
* expose leader election namespace in args
* add configurations in helm chart

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -91,6 +91,10 @@ spec:
             - "$(KRO_CLIENT_BURST)"
             {{- if .Values.config.enableLeaderElection }}
             - --leader-elect
+            {{- if ne .Values.config.leaderElectionNamespace "" }}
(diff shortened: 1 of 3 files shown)


helm/templates/deployment.yaml
```diff
@@ -91,6 +93,8 @@ spec:
             - "$(KRO_CLIENT_BURST)"
             {{- if .Values.config.enableLeaderElection }}
             - --leader-elect
+            - --leader-election-namespace
```
jakobmoellerdev, line 95: nit: im thinking its probably safer to no set this arg at all if empty because otherwise the value would be incorrectly parsed when there is a follow up arg. WDYT?

michaelhtm: I believe the value of the arg would be an empty string, as defaulted in `values.yaml`. This is the current default behavior (controller-runtime will default it to the namespace of the service account the controller uses). 

Or is this about a case where the value would not exist?

PR #666, opened 2025-09-10
feat: add schema-based resource quantity detection

yashpal2104:
Replace hardcoded path-based resource quantity detection with intelligent 
schema-aware comparison that leverages OpenAPI schema Format field to 
identify quantity fields.

This improvement enables more accurate and maintainable semantic equivalence 
checking for Kubernetes resource quantities (CPU, memory, storage) by using 
the actual schema metadata instead of maintaining hardcoded path patterns.

**Key Changes:**

* **Enhanced Delta Comparison API**
  - Add `CompareWithSchema()` function accepting optional `*spec.Schema`
  - Maintain backward compatibility with existing `Compare()` function
  - Implement schema-aware traversal functions for maps, slices, and objects

* **Schema-Based Resource Quantity Detection**
  - Prioritize schema Format field (`Format: "quantity"`) over path patterns
  - Fall back to existing path-based detection for backward compatibility
  - Remove dependency on maintaining hardcoded resource path suffixes

* **Controller Integration**
  - Extract resource schema from runtime during reconciliation
  - Pass schema context to delta comparison for enhanced semantic checking
  - Leverage existing OpenAPI schema infrastructure in Kro

* **Comprehensive Testing**
  - Add test cases verifying schema-based quantity detection
  - Test semantic equivalence (e.g., "100m" == "0.1" for CPU quantities)
  - Ensure non-quantity fields are properly differentiated

**Benefits:**
- **Maintainability**: Eliminates need to maintain hardcoded paths for every resource type
- **Accuracy**: Uses actual schema metadata for precise quantity field identification  
- **Extensibility**: Automatically supports custom resources using quantity format
- **Performance**: More efficient than string suffix pattern matching
- **Compatibility**: Full backward compatibility with existing path-based detection

Resolves the limitation where new Kubernetes resource types with quantity fields
required manual path pattern updates, making the system more robust and
self-maintaining.

### DIFF
--- a/pkg/controller/instance/controller_reconcile.go
+++ b/pkg/controller/instance/controller_reconcile.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/kube-openapi/pkg/validation/spec"
 
 	"github.com/kro-run/kro/pkg/controller/instance/delta"
 	"github.com/kro-run/kro/pkg/metadata"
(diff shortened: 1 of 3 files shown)


barney-s:
we should be switching to server side apply. This would not work if we have mutating webhooks in between. 
Please look at #561  - which removed the delta folders.

yashpal2104:
Yeah I feel I should close this PR since it is already being taken care of with apply sets

PR #674, opened 2025-09-14
chore: remove GitHub Actions stale bot workflow in favor of Prow

a-hilaly:
delete stale-bot.yaml - switching to Prow for stale issues


jakobmoellerdev:
/lgtm

(currently looking for a second sponsor, then this will work)

PR #679, opened 2025-09-17
design(kro): add declarative resource collections support

a-hilaly:
Introduces declarative resource collections to kro, enabling dynamic creation of multiple resources based on runtime data. This addresses the fundamental limitation where each RGD resource block could only create a single Kubernetes resource, requiring static definition of all resources at design time.


docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+> NOTE(a-hilaly): The dry run validation is currently limiting and often too
+> strict, i do have plans to completely get rid of it in favor of better type
+> checking and CEL validation.
+
+Importantly, the dependency analysis system requires no changes. Collections are
+treated as single nodes in the DAG regardless of how many resources they might
+create at runtime. If resource B depends on a collection resource A, the
+dependency is at the collection level - B waits for all of A's resources to be
+ready based on A's readyWhen condition. This maintains KRO's existing DAG
+validation and cycle detection without modification.
```
jakobmoellerdev: Does that mean it is not possible to wait for individual elements created by a collection as a result?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+- id: targetNamespaces
+  externalRef:
+    apiVersion: v1
+    kind: Namespace
+    selector:
+      matchLabels:
+        team: frontend
+```
+
+#### Discover resources using forEach semantics
```
jakobmoellerdev: is extertnalRef thus exclusive with forEach as well? I am assuming forEach would only be usable together with template right?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+  these may be added in future iterations
+- Dealing with cross-namespace resource creation
+
+### Non-Goals
+
+This proposal is not attempting to:
+
+- Turn KRO into a general purpose programming language
+- Replace specialized controllers that need complex state machines
+- Provide workflow orchestration with conditional branching
```
jakobmoellerdev, line 392: We already have conditional branching with `if` like logic and I think its in theory also possible to define conditional runs by checking if forEach evaluates to no elements right? Are you referring to more complex conditions?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+the resources - creating new ones for added items and deleting resources for
+removed items. Each resource maintains a stable identity based on its position
+or key in the collection.
+
+### Iteration Context Variables
+
+When a resource uses forEach, special variables become available within CEL
+expressions in the template:
+
+- `each.item` - The current element being iterated over
```
jakobmoellerdev: im wondering if we will keep adding more "introspection" fields on ones own structured data in a resource. if so, maybe it could make sense to structure this under `self` instead of `each`, because self can be used for other fields as well. WDYT?

```
- id: workers
  forEach: ${spec.workers} # ["alice", "bob", "charlie"]
  template:
    kind: Pod
    metadata:
      name: ${'worker-' + self.item} # worker-alice, worker-bob, worker-charlie
      labels:
        worker-name: ${self.item} # alice, bob, charlie
        worker-index: ${string(self.index)} # "0", "1", "2"
        total-workers: ${string(self.length)} # "3"
```

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+- `each.index` - The index (for arrays)
+- `each.key` - The key (for maps)
+- `each.value` - The value (for maps)
+- `each.length` - Total size of the collection
```
jakobmoellerdev: length breaks with the pattern you establish here because it is the only value you do not define on the individual item, but the collection, maybe worth separating out

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+
+## Proposal
+
+### Overview
+
+This KREP introduces the `forEach` field for resource definitions, enabling a
+single resource block to create multiple resources based on collections. The
+forEach field can iterate over:
+
+- Arrays from the custom resource spec
```
ellistarn, line 37: Would we ever iterate over maps, or even over structs? Can always be added later, but this struck me as a common pattern in other PLs.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+  forEach: ${spec.workers}
+  # 'workers' node is ready when ALL worker pods are running
+  readyWhen: ${workers.all(w, w.status.phase == 'Running')}
+  template:
+    kind: Pod
+    # ...
+```
+
+The resource collection is considered ready only when ALL worker pods are
+running. You can also use:
```
ellistarn: sugg: easier to read if all of the new aggregation functions are defined in a list. i.e. add `all` here.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+
+- `workers.any(w, condition)` - Ready when at least one instance matches
+- `workers.filter(w, condition)` - Filter instances based on a condition
+- Custom aggregations across the collection
```
ellistarn: what does this look like?

docs/design/proposals/collections.md
ellistarn: Or perhaps all of the new cel functions we're supporting. Right now it's a bit difficult to distinguish functions that already exist from net-new ones.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+    apiVersion: v1alpha1
+    kind: Application
+    spec:
+      workers: []string
+    status:
+      # Reference collections directly in status
+      totalWorkers: ${size(workerPods.items)}
+      runningWorkers: ${size(workerPods.items.filter(w, w.status.phase == 'Running'))}
+      allWorkersReady: ${workerPods.all(w, w.status.conditions.exists(c, c.type == 'Ready' && c.status == 'True'))}
+      workerNames: ${workerPods.items.map(w, w.metadata.name).join(',')}
```
ellistarn: Have we put any thought into shadowing? i.e., if you put `items.map(item, ...)`. Item is being shadowed. I suppose we have validations in place for this? Another option is to standardize `w` with a keyword.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+        kind: Pod
+        # ...
+```
+
+Status fields are continuously updated as the collection changes, providing
+real-time observability into the state of all resources managed by `forEach`.
+
+### resource references in CEL
+
+Colelction resourecs can be referenced by other resources using their "ID". For
```
ellistarn: lel

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,604 @@
...
+        kind: Ingress
+        # ...
+        spec:
+          rules:
+            - host: ${spec.domain}
+              http:
+                # collection resource can be referenced by its ID
+                paths:
+                  | # kro will do it's magic making sure to replace this with the proper type.
+                  ${microservices.map(svc, {
```
ellistarn: Will this cel expression re-evaluate on any change to any service? Maybe somewhat challenging. Would that lead users to optimize performance with something like `forEach: ${spec.services.map(s, s.status.loadbalancer.ingress.hostname[0])`. This would resolve the services in isolation on line 325 before doing the more expensive evaluation on line 341.

docs/design/proposals/collections.md
cheeseandcereal: +1, was reading and going to say the same thing. It would seem more logical to have this be separate, or maybe introduce another special variable like `collection`, etc, if you want to provide properties of the collection itself.

docs/design/proposals/collections.md
a-hilaly: That's correct - dependencies work at the collection level, not individual items. A collection is treated as a single node in the DAG, so for the node to be ready, all "resources" (expanded resources) must satisfty the `readyWhen` condition.

This is by design to maintain the predictable DAG behaviour. If we allow dependencies on individual collection items the DAG structure would become dynamic - edges could appear/disappear as collection items are added/removed during reconciliation. This would break kro's static analysis and make dependency resolution unpredictable.

docs/design/proposals/collections.md
a-hilaly: `forEach` would work with both `template` and `externalRef`. The mutual exclusion is between template and externalRef themselves (a resource can either create something or reference something, not both) but `forEach` enhances both

docs/design/proposals/collections.md
a-hilaly: I was referring more to situation where items inside a collection are dependent resources that need conditional branching. For example create item of index 5 after index 3 reach a specific state and if item 0 exists.

docs/design/proposals/collections.md
jakobmoellerdev: Agreed. However that means we should probably write a disclaimer on this in the design. I think this is a good tradeoff.

docs/design/proposals/collections.md
a-hilaly: Interresting, the more i think about it the more I do like the idea of using `self` or `this` in general (I think we should add them to https://github.com/kubernetes-sigs/kro/blob/main/pkg/graph/validation.go#L38-L65 ). I chose `each` because it clearly signals iteration context and to be very frank it's mainly inspired from [terraform's `for_each`](https://developer.hashicorp.com/terraform/language/meta-arguments#for_each). I do really want to use `self `with `readyWhen` conditions, i'm not entirely sure for collections.

a-hilaly: My concern with `self` is the the potential confusion if kro adds other self-referential features in the future, `self` might mean the current resource being created, not the iteration (as in `self.metadata.name` vs `self.item`

docs/design/proposals/collections.md
a-hilaly: You're absolutely both right in here, i'll eliminate `each.length`. @cheeseandcereal +1 for the another special variable to expose such properties - i purposefully shrank the proposal to not talk about them, i think there is a bigger picture where users can access resources via `resources[id]` or `resources[index]`. Also related to what @jakobmoellerdev proposed above https://github.com/kubernetes-sigs/kro/pull/679#discussion_r2356411293

a-hilaly:
@n3wscott BTW I called this one KREP-002 because technically you did write KREP-001 :)

docs/design/proposals/collections.md
a-hilaly: Yes, maps are support - i missed highlighting them clearly in the overview section.. The implementation in https://github.com/kubernetes-sigs/kro/pull/632 handles map iteration.

By struct, do you mean iterating over object fields as key-value pairs? or iterating over an array of objects?

docs/design/proposals/collections.md
a-hilaly: Nice catch! Update the section to include all the supported functions/macros by default.

docs/design/proposals/collections.md
a-hilaly: custom aggregations would be any combination of CEL functions, for example:
- `workers.filter(w, w.status.ready).size() >= 10`
- `workers.map(w, w.status.replicas).sum() >= 10`
- `workers.filter(w, w.metadata.labels.env == "productino").all(w, w.status.ready)`

docs/design/proposals/collections.md
a-hilaly: Yes, kro tracks "local" variables properly regardless of the variable name. We have an AST inspector that tries to handle variable scoping and shadowing correctly: https://github.com/kubernetes-sigs/kro/blob/main/pkg/cel/ast/inspector.go#L311-L333

docs/design/proposals/collections.md
a-hilaly: >Will this cel expression re-evaluate on any change to any service? 

Correct to change to any service or every "resyncPeriod" duration.

>This would resolve the services in isolation on line 325 before doing the more expensive evaluation on line 341.

Are you suggesting users could reference "static" values directly from the instance spec rather than from the dynamic service resources?

docs/design/proposals/collections.md
a-hilaly: I just updated the "readyWhen with Collections" section, L225

docs/design/proposals/collections.md
michaelhtm: what if we instead allowed users to define a variable?
eg:
```
forEach: ${worker: spec.workers}
```

or maybe something similar to how we handle existing functions (`.all(w, condition)`)
we can try to enforce a `var: collection` syntax only for `forEach`.
The same functionalities as `each.item` would turn into `worker.item`..
WDYT?

docs/design/proposals/collections.md
ellistarn: Discussed offline -- I think this is essentially the equivalent of a compiler optimization. We can essentially look at subtrees of the DAG, detect changes at specific nodes, and only re-evaluate children within the DAG. For large DAGs, this could result in significant performance improvements.

docs/design/proposals/collections.md
ellistarn: > By struct, do you mean iterating over object fields as key-value pairs

Yes this. Kind of wild -- I'd probably hold off until we knew the use case better. I could imagine wanting to walk the YAML though.

docs/design/proposals/collections.md
jakobmoellerdev: I think we can consider using evaluation result storage as a later iteration. Kubernetes famously does not recompute cel results everytime, and we could think of something similar. What do you think?

docs/design/proposals/collections.md
jakobmoellerdev: for me it is acceptable if self is used for more than the iterator values because it is effectively an inflection API. the only thing we would be careful on is which fields / keywords are available or blocked, but I think thats doable.

ellistarn: Would you ever have nested forEach that could cause collisions for self?

a-hilaly: @michaelhtm That's an interesting suggestion! The challenge with parsing variable name directly from the forEach expression (like `${worker: spec.workers}`) is that it conflicts with CEL's ternary operator syntax. In CEL `:` is a special token dedicated to expressions like `condition ? x : y`.
My worry is that trying to reliably distinguish between the two syntaxes would require a fragile prefix parsing that could break with comcplex expressions, we'd basically be extending CEL syntax in a non-standard way.

However your idea sparked a thought - we could use a CEL function apparoch:

```
forEach: ${iterate("worker", spec.workers)}
```

This would be a CEL compliant and allow custom variable names. We could leverage our ast inspector to detect those and only allow constant strings in the first argument of the function - and the function would return a special iterator context that kro recognizes. Thought this is a little unambiguis and might feed weird-ich.

We could also use a simpler syntax like:
```yaml
forEach:
  iterateOver: ${spec.workers}
  var: "worker"
```

it's slightly more expicit and avoid parsing complexity and leaves room for future extensions. Thoughts on these ideas?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+KRO (Kubernetes Resource Orchestrator) enables users to define custom Kubernetes
+APIs that create and manage multiple resources as a single unit. Users write
+ResourceGraphDefinitions (RGDs) that specify a directed acyclic graph (DAG) of
+resources and their relationships. KRO's engine uses CEL (Common Expression
+Language) for dynamic field computation while maintaining strict security
+properties: bounded execution time, runtime sandboxing, and predictable resource
+generation.
+
+Currently, KRO has a fundamental limitation: each resource block in an RGD
+creates exactly one resource. The number of resources must be known at design
```
brandonjbjelland: Nit: can't a CEL expression conditionally create resources today? If so, the following is technically more correct:


```suggestion
Currently, KRO has a fundamental limitation: each resource block in an RGD
creates zero or one of each declared resource. The number of resources must be known at design
```

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+- **Maps**: key-value pairs where iteration occurs over entries -
+  `${spec.labels}` or `${{"env": "prod", "tier": "web"}}`
+
+The expression can include CEL transformations or macros:
+`${spec.workers.filter(w, w.enabled)}` or `${spec.zones.map(z, z + "-a")}`.
+
+As the collection changes (items added/removed), KRO automatically reconciles
+the resources - creating new ones for added items and deleting resources for
+removed items. Each resource maintains a stable identity based on its position
+or key in the collection.
```
brandonjbjelland: Question / Potential Risk:

For array-based `forEach` resources, what is the stateful key that tracks each resource? My read is that position or numeric index is used. If so, that makes resources fragile: any change in element order or array size could cause churn. This is the same problem Terraform had before introducing `for_each`.

Options I see:

1. Set-based keys – Require arrays to track state via a guaranteed unique element (not index).
2. Maps only – Drop array-based iteration entirely; force arrays to be converted to maps.
3. Composite keys – Derive stable keys via a hash function or serialization utility (though duplicates like ["foo","foo"] still break).



Tradeoffs:

- Option 1 burdens users with ensuring key uniqueness and detecting collisions.
- Option 2 feels heavy-handed, but simplifies state stability and aligns with Terraform’s shift away from count.
- Option 3 may rely on api-machinery helpers, but isn’t a silver bullet for duplicates and runtime uniqueness violations (eg when the array is based on an externalRef) would make for a bad user experience.

Terraform analogy: `count` is fragile and considered an anti-pattern except for conditional resource creation. Iteration should use `for_each` with a `toset()` when executed on a list to guarantee uniqueness. My concern: `forEach` on arrays risks re-creating that “count trap.”

Am I over-thinking this? Or is this already solved in kro?

brandonjbjelland (commented):
This is exceptional! Thanks for the hard work here.

docs/design/proposals/collections.md
a-hilaly: @ellistarn regarding nested collections - i didn't plan on adding native support for nested collections with a single resource block. However nested for loops is already achievable by composing multiple RGDs: one RGDx creates a collection, and a parent RGDy can `forEach` over that new type (generated by RGDx). For example:

<table>
<tr>
<td>

```yaml
# RGDx: creates pods per zone #######
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
spec:
  schema:
    kind: ZonePods
    spec: { zones: []string }
  resources:
    - id: pods
      forEach: ${spec.zones}
      template:
        kind: Pod
        # ...
```

</td>
<td>

```yaml
# RGDy: creates ZonePods per region  
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
spec:
  schema:
    kind: RegionalApp
    spec: { regions: []string }
  resources:
    - id: regional
      forEach: ${spec.regions}
      template:
        kind: ZonePods
        # ...
``` 

</td>
</tr>
</table>

That said your question make me think about natively supporting nested iterations - we could potentially support multiple iterators:

```yaml
  forEach:
    - iterator: ${spec.regions}
      var: "var1"
    - iterator: ${spec.zones}
      var: "var2"
  template:
    kind: Pod
    metadata:
      name: ${'pod-' +var1.item + '-' +  var2.item}
```

This would create a cartesian product (us-east x zone-a = pod-us-east-zone-a etc...). Thought the semantics get tricky, what happens when one iterator is empty? I'm realizing this could be problematic we could endup evaluating `name: ${iter1.item}-${iter2.item}-${iter3.item}` evaluating to something like `a--` which isn't a valid kubernetes name foe example. Though this feels like it shouldn't be considered a design issue but rather an implementation issue that users needs to handle.

@jakobmoellerdev ^ if we ever go this route i think the above is a good reason to keep the runtime emulation system and protect users from generating awkward/invalid kuberentes names/namespaces.

vast-raymond-page: @a-hilaly `iterator` read redundant in the foreach context. Perhaps `itemCollection` or just `collection` for the list of items to iterator over. Instead of `var` perhaps just `name`. Also instead of `item`, `element` or `entry` are good alternatives if you need ideas.

```yaml
  forEach:
    - collection: ${spec.regions}
      name: "region"
    - collection: ${spec.zones}
      name: "zone"
  template:
    kind: Pod
    metadata:
      name: ${'pod-' +region.entry+ '-' +  zone.entry}
```

michaelhtm: @vast-raymond-page I like this suggestion as well. It's more `kro` like.

ellistarn: You could use a map to collapse this even further
```
forEach:
  region: ${spec.regions}
  zone: ${region.zones} # <-- depends on the first for loop.
```

You might require that customers order these correctly in case your loops depend on each other. 🤔 , maybe each loop is really a DAG node and you can reuse the dependency of the graph?

---

This approach does drop the `collection` information, so if you were ever iterating over a `map` instead of a `list`, you might need different syntax, like:
```
forEach:
  resource: ${ spec.resources } # <-- refer to this via resource.key and resource.value
```

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+      name: ${'pod-in-' + each.item.metadata.name}
+      namespace: ${each.item.metadata.name}
+```
+
+## Scope
+
+This proposal focuses on enabling dynamic resource creation through the
+`forEach` field and essential supporting features.
+
+### In Scope
```
ellistarn, line 351: What do we think about introducing a concept of "safe and progressive change" as part of our initial collections design? You currently have "rollout strategies" as out of scope for this design.

Collections introduce significant orchestration power in a cluster, but with great power comes great responsibility. I think it's fair to say that it's completely unsafe to execute a collection with full parallelism. i.e., if a collection is managing 500 objects, is it safe to attempt to create all 500 of those objects in parallel? Similarly, it's likely not viable to wait on each of those 500 objects to become ready serially. 

If we are interested in cutting scope, I could see launching with the latter, but definitely not the former. We should be thoughtful about breaking changes around changing the default in the future

A few initial questions in this space:

1. Should we treat creates and updates identically? What about deletes?
2. What should the right default be?
3. What configuration options make sense. i.e.Exponential with ceiling, `min(2^n, .1 * n)`
4. Should rollouts always be ordered in the same way, or randomized for each change?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+Language) for dynamic field computation while maintaining strict security
+properties: bounded execution time, runtime sandboxing, and predictable resource
+generation.
+
+Currently, KRO has a fundamental limitation: each resource block in an RGD
+creates exactly one resource. The number of resources must be known at design
+time and explicitly defined. This creates a gap between KRO's declarative model
+and real-world infrastructure needs, where the number of resources often depends
+on dynamic factors discovered at runtime - how many availability zones are
+healthy, how many tenants exist, how many namespaces match certain criteria.
```
tjamet:
```suggestion
healthy, how many tenants exist, how many namespaces match certain criteria,
how many instances or replicas are required to serve the current load.

```

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+
+Kubernetes environments are inherently dynamic, and the number of resources
+needed often depends on runtime configuration rather than design-time decisions.
+KRO's current model requires users to statically define each resource block in
+their ResourceGraphDefinition, creating a fundamental mismatch with real-world
+needs.
+
+The core issue is simple: each resource block in an RGD creates exactly one
+Kubernetes resource. If you need 5 Pods, you write 5 resource blocks. If you
+later need 10, you must modify the RGD to add 5 more blocks.
```
tjamet: It looks like we are talking here about the needs of a scaling factor, similar to what Deployments, ReplicaSet or StatefulSets are achieving.

For those use-cases it would be great to rely and implement Kubernetes's API through the scale subresource, so you can later connect it with the rest of the ecosystem, like Keda.

Scaling also introduces the idea of replicas and hence rollout.
When you have 5 replicas of your pod that you want to update, you don't want to update them all at once.
Instead, you want to update (delete then create, create then delete) 1 replica, observe everything is right and then proceed with the rest. For other resources (say you want to scale out the number of Kubernetes clusters or Prometheus shards), you probably can't afford the delete+create technique you must follow with pods.

```suggestion
#### Scaling sub-resources

The core issue is simple: each resource block in an RGD creates exactly one
Kubernetes resource. If you need 5 Pods, you write 5 resource blocks. If you
later need 10, you must modify the RGD to add 5 more blocks.

The number of pods you need is directly related to dynamic factors such as the
number of requests items or average CPU or memory usage. In the Kubernetes API,
a specific section is dedicated to this dynamic scaling purpose, the scale sub-resource.
The scale sub-resource acts an interface to support dynamic scales and benefits from
an active eco-system with command-lines like `kubectl scale` or automations like Keda.
With the current RGD you can't satisfy those constraints and automations. Scaling need
to be handled outside of KRO's guarantees.
```

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+
+Consider a batch processing application where users specify how many worker Pods
+to run through their custom resource: `workers: 5`. Each worker needs its own
+Pod with specific environment variables and configurations based on its index.
+With KRO today, you cannot express "create N Pods where N equals spec.workers".
+The same limitation appears in data processing pipelines where each stage might
+need a different number of resources:
+`stages: [{name: "extract", replicas: 3}, {name: "transform", replicas: 5}, {name: "load", replicas: 2}]`.
```
tjamet: Introducing this, we need to discuss unit of deployment.

```suggestion
### Rollout and unit of deployments

Consider a batch processing application where users specify how many worker Pods
to run through their custom resource: `workers: 5`. Each worker needs its own
Pod with specific environment variables and configurations based on its index.
With KRO today, you cannot express "create N Pods where N equals spec.workers".
The same limitation appears in data processing pipelines where each stage might
need a different number of resources:
`stages: [{name: "extract", replicas: 3}, {name: "transform", replicas: 5}, {name: "load", replicas: 2}]`.

Those stages may require more than just a pod, but also a ConfigMap or a PersistentVolumeClaim.
When applying a new version, the rollout must be controlled so that deployment stops if the pods do not come up healthy.

Similarly, consider a platform distributing its workloads across multiple Kubernetes clusters.
The cluster deployment of an EKS cluster using cluster-api is performed via
`Cluster`, `EKSConfigTemplate`, `MachineDeployment`, `AWSMachineTemplate`, `AWSManagedCluster`, `AWSManagedControlPlane` objects.
All those objects belong to the same logical entity, an EKS cluster. They should all be deployed concurrently.
When declaring multiple replicas of a cluster, cluster updates must be orchestrated in a
controlled manner to avoid propagating errors and control the blast radius.
```

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+
```
tjamet, line 21:
```suggestion

### Update strategies

There are several ways to update a specific workload, each matching particular needs. 
Consider you are running data processing pods, consuming workloads from a queue like Kafka. In this case, the pod identity is important to avoid "consumer rebalancing" and you need to delete the pod before creating the new one.
If the pods you are serving are handling live HTTP traffic, users prefer to temporarily increase capacity to avoid impacting the HTTP SLOs. You want hence to create the new pod then delete the old one.
In both of those cases, the update must go through a delete phase, as it is impossible to do an in-place update of pod specs.

If the collection you are deploying involves persistent volumes or clusters, you should consider an in-place upgrade to prevent service disruption.
```

tjamet (commented):
Great to see the idea of collections coming to KRO.
I do foresee 2 main use-cases for collections:
1. the "templating loop" (i.e. defining multiple variants of a single item like different ingresses for a given app for instance)
2. scaling purposes (i.e. generating replicas of a composed element)

I read a mix of both in the problem section, it's probably worth discussing further whether we want to cover both cases or focus on a single one

For the second part, a bit like what is mentioned [here](https://github.com/kubernetes-sigs/kro/pull/679#discussion_r2429794370), I think there is a discussion to have about how objects can be safely rolled-out, in particular when we start dealing with pods and more sensitive elements like PVCs

I made some suggestions to better cover the second case in the problem statement

docs/design/proposals/collections.md
matthchr: It would be cool to have kro support /scale, but it's not clear to me that it's really actually related to this feature. You could imagine scale scaling just deployment replicas (which doesn't require any collections support), _or_ scaling a collection in the RGD, or both.

It seems like they're orthogonal really? Scale subresource is about scaling the RGD CRD instance (replicas is an input to the RGD graph engine). That input may be used in multiple places when rendering the graph, possibly in collections or in just plain fields such as a Deployment `replicas` as well.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+expressions within `forEach` are subject to normal CEL cost limits, and the
+total number of resources created is visible and predictable before execution
+begins.
+
+**DAG Expansion**
+
+A resource block with the `forEach` field expands into multiple nodes in KRO's
+execution DAG at runtime. Each expanded node is treated as a distinct resource
+with its own reconciliation cycle, status tracking, and error handling.
+Dependencies between resources are preserved - if resource B depends on resource
```
matthchr: Would there be a way to depend on a subset of them? Or it's always all or nothing?

docs/design/proposals/collections.md
matthchr: > it's slightly more expicit and avoid parsing complexity and leaves room for future extensions. Thoughts on these ideas?

In terms of "user defining the variable name" I like:
```
forEach:
  iterateOver: ${spec.workers}
  var: "worker"
```
or 
```
forEach:
  worker: ${spec.workers}
```

more than the proposed CEL-function approach or the "almost-ternary operator" approach. Though not sold we really care about letting the user define the variable.

**On the topic of nested loops** as discussed previously it's probably possible to project nested collections into a single collection w/ something like a CEL `zip` function - though it might be really nasty if you have to dance between maps, lists, etc. It might be nice to support this more natively although agree that defining the semantics is important. As you say @a-hilaly if it's really a cartesian product, then the answer to "what happens if one set is empty" should probably be the "loop" runs 0 times - I'm not sure anything else makes sense.

It does feel like we should take a stance in this document on what the user experience for nested loops should be. Either we support it, or we explain how to do it by doing the cartesian product yourself (do we need to offer some CEL-helpers for this, if we take this route? I dont think `zip` exists)

docs/design/proposals/collections.md
matthchr: If they want the length, given that the expression itself is valid CEL, they can just repeat it and call size() on it?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+removed items. Each resource maintains a stable identity based on its position
+or key in the collection.
+
+### Iteration Context Variables
+
+When a resource uses forEach, special variables become available within CEL
+expressions in the template:
+
+- `each.item` - The current element being iterated over
+- `each.index` - The index (for arrays)
```
matthchr: Just confirming, this is not available for maps because we don't have stable iteration order?

Or do we need to have maps that iterate in a consistent order?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+
+While `readyWhen` is an existing field in KRO, it gains new capabilities with
+collections. For resources using forEach, readyWhen can now define conditions
+across all instances in the collection:
+
+```yaml
+- id: workers
+  forEach: ${spec.workers}
+  # 'workers' node is ready when ALL worker pods are running
+  readyWhen: ${workers.all(w, w.status.phase == 'Running')}
```
matthchr: are all references to `workers` (for example in subsequent graph entries) as a collection?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+kind: ResourceGraphDefinition
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: Application
+    spec:
+      workers: []string
+    status:
+      # Reference collections directly in status
+      totalWorkers: ${size(workerPods.items)}
```
matthchr: Inconsistent between `workerPods.items` and just `workerPods`?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+        spec:
+          schedule: "0 2 * * *"
+          jobTemplate:
+            # ...
+```
+
+### External references with collections
+
+Today, KRO supports `externalRef` to discover resources outside the RGD.
+external references does point to one and unique resource, discovered by name
```
matthchr: minor: Phrasing here seems wonky/wrong

docs/design/proposals/collections.md
matthchr: As discussed, maybe we need to talk about what exactly referencing the collection-resource "is". Is it a list? Can I do `workerPods[0]` (seems risky!)? is it a struct w/ `.items` (and possibly other fields?)

Maybe deserves a section of the document discussing it in more detail.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+- Resource references in CEL to enable collection chaining
+- Status field computations using collections
+- Integration with existing KRO fields (`readyWhen`, `includeWhen`)
+- Updates to resources when collection items change
+- Enhanced externalRef to watch collections of resources (via slectors)
+- Combining `forEach` with externalRef for dynamic external resource discovery
+
+### Out of Scope
+
+- Nested forEach loops or recursive expansion
```
matthchr, line 381: As mentioned, I think if we declare this out of scope we should at least:
- Have a plan for how we could (in a nonbreaking way) expand to support it in the future.
- Show how you can do cartesian product w/ maps/lists in CEL natively, meaning users can easily work around this lack.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+        kind: Ingress
+        # ...
+        spec:
+          rules:
+            - host: ${spec.domain}
+              http:
+                # collection resource can be referenced by its ID
+                paths:
+                  | # kro will do it's magic making sure to replace this with the proper type.
+                  ${microservices.map(svc, {
```
matthchr: What happens when `microservices` is len 0? In general for DAG ordering previously there was somewhat of a guarantee that a referenced resource had been rendered - now with collections that's not the case as length can be 0. When this happens, what happens w/ resources that depend on the collection?

I assume that they are rendered but whatever bits depend on the collection are empty - so in this case for example paths is empty.

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+resources are processed.
+
+KRO tracks resources created by collections using its existing ownership system,
+extended with collection specific labels. Each resource receives the standard
+KRO ownership labels plus `kro.run/collection-key` which identifies which item
+in the collection this resource represents. This labeling scheme, orthogonal to
+the ApplySet work, enables precise tracking of collection resources across
+reconciliation cycles.
+
+> NOTE(a-hilaly): Depending on how the ApplySet work shapes up, we might be able
```
matthchr: probably can update this given ApplySet has merged at this point.

docs/design/proposals/collections.md
jakobmoellerdev: Probably the only sane way we can introduce this and keep scope in check on the implementation now is by introducing a concurrency upper bound (you called it ceiling). I think we can add more behavior such as dynamic scaling limits under that ceiling quite easily. WDYT?

docs/design/proposals/collections.md
jakobmoellerdev: for me scale needs to supported by the instance API (RGD CRD) and simpleschema and we need to be able to react to scale dynamically. IMHO non-goal of the proposal.

docs/design/proposals/collections.md
jakobmoellerdev: TBH I think if at all, we should mention change behavior during update a specific non-goal of the design. This is and was always part of the instance API and the underlying resources, not of KROs / collections behavior. We dont need to promote best practices here (proposal is large enough as is imo)

docs/design/proposals/collections.md
ellistarn: Discussed with @a-hilaly offline. One path is to launch this with a simple approach (i.e. linear, one-at-a-time) and fast follow with a rollout KREP.

edit: saw your post after I submitted this comment. Yes agree -- suggest we start small (i.e. 1)

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+### `readyWhen` with Collections
+
+While `readyWhen` is an existing field in KRO, it gains new capabilities with
+collections. For resources using forEach, readyWhen can now define conditions
+across all instances in the collection:
+
+```yaml
+- id: workers
+  forEach: ${spec.workers}
+  # 'workers' node is ready when ALL worker pods are running
```
ellistarn: is it better to express readyWhen on a per resource basis, rather than the collection itself? e.g 
`readyWhen: item.status.phase == 'Running'`

What if other nodes in the graph take dependencies on individual items of this collection, rather than a dependency on the collection itself? I'm imagining another resource stanza that does a forEach on workers and makes a new object for each worker.

```
- id: deployments
  forEach: ${range(0, spec.workers)}
  template: ...
- id: hpas
  forEach: ${ deployments }
  template: ...
```

Under this mental model, the `readyWhen` would apply to each item in the collection, and the collection itself would have an implicit `readyWhen: items.all(item.Ready)`

ellistarn: I was going down a similar line of thinking when trying to reason about propagation control (i.e. `propagateWhen`).

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+
+### includeWhen with Collections
+
+The resource-level `includeWhen` determines whether any instances should be
+created at all. If `includeWhen` evaluates to false, the entire forEach is
+skipped and no resources are created.
+
+```yaml
+- id: backupJobs
+  includeWhen: ${size(spec.databases) > 0 && spec.backupEnabled}
```
ellistarn: Similar question to readyWhen (https://github.com/kubernetes-sigs/kro/pull/679/files#r2452982275). If we make includeWhen on a per resource basis, then the collection becomes an empty list if none of the resources pass the predicate.

docs/design/proposals/collections.md
tjamet: I can agree that it is out of the scope of this proposal and feature.
But we also need to be ready for users using it as such.

A way forward can also be to find examples in the KREP (and later documentation) that are not about horizontal scaling

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+## Proposal
+
+### Overview
+
+This KREP introduces the `forEach` field for resource definitions, enabling a
+single resource block to create multiple resources based on collections. The
+forEach field can iterate over:
+
+- Arrays from the custom resource spec
+- Ranges of numbers, characters, or other types
```
tjamet, line 38: let's take the idea of having `forEach: ["some.domain.tld", "other.domain.tld"]`
Do we want to support in-object iterations for example for DNSEndpoints or http routes

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: example-httproute
spec:
  parentRefs:
  - name: example-gateway
  hostnames: ${each.values}
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /login
    backendRefs:
    - name: example-svc
      port: 8080
---
apiVersion: externaldns.k8s.io/v1alpha1
kind: DNSEndpoint
metadata:
  name: cluster-operator
spec:
  endpoints:
  - dnsName: ${each.key}
    recordType: CNAME
    targets:
    - some.cloud.loadbalancerr
```

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,633 @@
...
+Example in an RGD:
+
+```yaml
+kind: ResourceGraphDefinition
+# ...
+spec:
+  resources:
+    # workers node in the DAG
+    - id: workers
+      forEach: ${range(1, 4)} # Creates 3 Pods
```
justinsb: I do think a common use case will be building (say) 4 replicas when `spec.replicas: 4`.  Can we come up with a good approach here for when users specify `forEach: ${spec.replicas}`.  Maybe we could recommend a function like `forEach: ${generateKeys(spec.replicas)}`

justinsb:
LGTM though I suggest we include examples of the user-facing syntax for key scenarios (to make it easier to understand what the user-experience would be).  Example scenarios:

* spec.replicas: 4
* forEach over a a list of map (maybe something like ports in a pod, or a map of zones where we also need the region)

jakobmoellerdev:
@a-hilaly we really should start pushing collections to completions. Do you need / want support to get the comments addressed? if not, can we start closing the various discussion threads and merging in the next few weeks? I think it would be a nice christmas gift for the project :christmas_tree:

docs/design/proposals/collections.md
a-hilaly: Excluded externalRefs with forEach for now. https://github.com/a-hilaly/kro/blob/krep-2/collections/docs/design/proposals/collections.md

docs/design/proposals/collections.md
a-hilaly: Compiled all the ideas in here an landed with something like:
```
- id: deployments
  forEach: # This is an array to keep cartesian products deterministic.
    - region: ${schema.spec.regions}  # ["us-east", "us-west"]
    - tier: ${schema.spec.tiers}      # ["web", "api"]
  template:
    kind: Deployment
    metadata:
      # Creates 4 deployments: us-east-web, us-east-api, us-west-web, us-west-api
      name: ${region + '-' + tier}
```

docs/design/proposals/collections.md
a-hilaly: Eliminated in new design, now things are simpler:
```yaml
- id: deployments
  forEach:
    - region: ${schema.spec.regions}  # ["us-east", "us-west"]
    - tier: ${schema.spec.tiers}      # ["web", "api"]
  template:
    kind: Deployment
    metadata:
      # Creates 4 deployments: us-east-web, us-east-api, us-west-web, us-west-api
      name: ${region + '-' + tier}
```

docs/design/proposals/collections.md
a-hilaly: Eliminated iteration over maps. Iterating over structs is not supported.

docs/design/proposals/collections.md
a-hilaly: Changed to use `each` e.g `${each.status.phase == 'Running'}`

docs/design/proposals/collections.md
a-hilaly: Thanks for bringing this @brandonjbjelland ! you're not over thinking it, but KRO sidesteps the worst of the Terraform count trap because k8s resource identity is gvk-namespace-name based, not index-based. When you use value based naming like `name: ${'worker-' + name}` where name comes from the array element, reordering ["alice", "bob"] to ["bob", "alice"] doesn't cause churn - `worker-alice` and worker-bob are the same K8s resources regardless of array order. KRO uses ApplySet which track resources by GVK/namespace/name, not by position. The collection-index label is just metadata for observability. That said, the trap still exists if users write index-based names (worker-0, worker-1), and duplicates like `["foo", "foo"]` will fail at runtime with a name collision. the proposal pushes naming responsibility to users rather than enforcing uniqueness - so it's not fully "solved", but more of a shared responsability model.

docs/design/proposals/collections.md
a-hilaly: Excluded in favor of progressive rollout KREP

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,607 @@
...
+- id: workers
+  forEach:
+    - name: ${schema.spec.workers}  # ["alice", "bob", "charlie"]
+  template:
+    kind: Pod
+    metadata:
+      name: ${'worker-' + name}  # worker-alice, worker-bob, worker-charlie
+```
+
+To access the index, iterate over a range instead:
```
spattk, line 188: Can we access both the index and the variable name inside the template?
For example, let's say, we might have a use-case something like `worker-0-alice`, `worker-1-bob`, `worker-2-charlie` where folks might need this?

docs/design/proposals/collections.md
```diff
@@ -0,0 +1,607 @@
...
+        # ...
+
+    - id: gateway
+      template:
+        kind: Ingress
+        spec:
+          rules:
+            - host: ${schema.spec.domain}
+              http:
+                # collection resource can be referenced by its ID
```
spattk, line 309: We are saying, that the whole collection will be referenced as one. I am just thinking out loud, will we ever have a situation where we would need to refer to the individual items under a collection?

docs/design/proposals/collections.md
a-hilaly: so currently, multiple iterators produce cartesian products - all combinations of independent dimensions:
```yaml
forEach:
  - region: ${schema.spec.regions}  # [us-east, us-west]
  - tier: ${schema.spec.tiers}      # [web, api]
# produces 4 resources: us-east-web, us-east-api, etc...
```

For dependent/recursive iteration over nested data, we can allow expressions to reference prior iterator variables:

```yaml
forEach:
  - region: ${schema.spec.regions}
  - zone: ${region.zones}
```

Not supported in the current proposal but easily addable since the syntax is already an ordered array

docs/design/proposals/collections.md
a-hilaly: Yes, individual item access is possible via indexing (`workers[0]`, `workers[i]`), but we can't statically validate bounds - users can write arbitrary expressions like `workers[i + 1 - 1]` on more complex ones. Out of bounds would be a runtime error, same as any programming language. That's expected behavior.

docs/design/proposals/collections.md
a-hilaly: Mainly for simplicity, iteration is on items only - no implicit index variable. Users can work around this with a zip pattern to pair items with their indices, for example we can build enumerate-like macros or functions that are available in the CEL environement:
```yaml
forEach:
  - w: ${enumerate(schema.spec.workers)} # [{index: 0, value: "alice"}, ...]
template:
  metadata:
    name: ${'worker-' + string(w.index) + '-' + w.name}
```

docs/design/proposals/collections.md
a-hilaly: `lists.range(n)` already does this - lists.range(schema.spec.replicas) generates `[0, 1, 2, 3]` when replicas is 4.

docs/design/proposals/collections.md
a-hilaly: Incorporated this feedback - the design now uses each for per-item `readyWhen`, with implicit `all()` semantics at collection level. When we add propagation control, we'll have finer mechanics for partial updates, gradual rollouts, etc..

docs/design/proposals/collections.md
a-hilaly: for now `includeWhen` stays collection level (should this resource block run at all?), per item filtering belongs in the `forEach` expression via `.filter()`. Keeps semantics clear - `includeWhen` gates the whole block, `forEach` controls which items to iterate. If filter results in empty array, zero resources are created and downstream dependencies see an empty collection.

docs/design/proposals/collections.md
a-hilaly: Correct - empty collection means downstream resources still render, but expressions referencing the collection evaluate to empty. In the gateway example, paths would be []. The collection is considered "ready" (by design, all zero items pass `readyWhen`) so the DAG proceeds normally.

docs/design/proposals/collections.md
a-hilaly: It's a list of rendered resources - you can use CEL list operations (map, filter, size, all). Direct indexing (workerPods[0]) works but isn't recommended - ordering depends on iteration order and you risk out-of-bounds at runtime.

docs/design/proposals/collections.md
a-hilaly: `forEach` creates multiple resources, not in object field expansion. For simple cases, CEL expressions work (`hostnames: ${schema.spec.domains}`). True in object iteration with each would require intermediate variable nodes that hold state but don't create resources - potential future work but not planned currently

docs/design/proposals/collections.md
a-hilaly: yes

docs/design/proposals/collections.md
a-hilaly: excluded maps from the scope. Users with map data should convert to sorted array with something like: `forEach: [{k: ${schema.spec.labels.map(k, k).sort()}}]`

docs/design/proposals/collections.md
a-hilaly: The design has been clarified - it's a single DAG node that creates multiple resources, not multiple DAG nodes. So dependencies are all-or-nothing at collection level. For now.

docs/design/proposals/collections.md
a-hilaly: Agreed - update/change behavior is out of scope for this KREP, that's instance API / propagation control territory (KREP-006).

docs/design/proposals/collections.md
a-hilaly: Agreed with @jakobmoellerdev and @matthchr - scale subresource is orthogonal to collections. Scale is about the instance API (RGD CRD / SimpleSchema)

Issue #680, opened 2025-09-17
Onboard onto prow / k8s community infra


justinsb:
Now that we are a kubernetes-sigs project 🎉  , we should onboard to prow and the k8s community infra (staging repos etc).

I am happy to get the ball rolling here:

- [ ] Get set up in kubernetes/k8s.io
- [ ] Get prow jobs set up and running in dashboard

(If there are other tasks I can add to the checklist, or we can add subissues)

jakobmoellerdev:
This should definetly first and foremost cover getting releasable artifacts out the door so we unblock releasing via K8s infra

justinsb:
Next snafu: https://github.com/kubernetes/test-infra/pull/35731

a-hilaly:
Thank you @justinsb and @jakobmoellerdev !

PR #681, opened 2025-09-17
Chore: refactor logic out of GHA

justinsb:
This supports moving to prow.

Putting scripts into a well-known location (dev/ci/presubmits)
should allow future generation of the prow config.

Issue #680


dev/ci/presubmits/unit-tests
a-hilaly, line 1: @justinsb can we move this to /hack? we already have many top level folders

barney-s: may be under scripts/dev instead ?

jakobmoellerdev: +1 to move to `scripts` top level

a-hilaly: i really loved hack!! snif snif

.github/workflows/unit-tests.yaml
```diff
@@ -6,12 +6,14 @@ on:
...
       - go.mod
       - go.sum
+      - dev/ci/**
   pull_request:
     branches: [ main ]
     paths:
       - "**.go"
       - go.mod
       - go.sum
+      - dev/ci/**
```
a-hilaly:
```suggestion
      - scripts/ci/**
  pull_request:
    branches: [ main ]
    paths:
      - "**.go"
      - go.mod
      - go.sum
      - scripts/ci/**
```

PR #684, opened 2025-09-17
Removing after signing CLA

bridgetkromhout:
Removing another contributor from the NOTICE file when they signed the CLA.


a-hilaly (approved):
/lgtm

Issue #683, opened 2025-09-17
Unable to create vault groupAlias using crossplane provider


fabualnaja:
### Description

# `identity.vault.upbound.io/v1alpha1` `GroupAlias` fails to create via RGD (instance ends in ERROR)

## **Summary**:

When reconciling an RGD that creates Vault `Group` and Vault `GroupAlias` (both under `identity.vault.upbound.io/v1alpha1`), the Group is created successfully but the `GroupAlias` never gets created. The `AppAccess` instance ends in `State=ERROR `with:

```pgsql
failed to create resource: the server could not find the requested resource
````

This happens regardless of namespacing (both resources are cluster-scoped, and the failure persists even when `metadata.namespace` is omitted from the template).

Other resources in the graph (ArgoCD roles/bindings, Keycloak roles, Vault policy/group) reconcile fine.

**Environment**:

- kro version: v0.4.1

- Kubernetes:
  - Client: v1.32.3
  - Server: v1.28.15

- Relevant providers/CRDs installed:
  - [Crossplane version 1.20.0](https://artifacthub.io/packages/helm/stable-crossplane/crossplane/1.20.0) (also I have tried v1.18.0 , v1.19.0, and 2.0.0) 
  - [Upbound Vault provider](https://marketplace.upbound.io/providers/upbound/provider-vault/v2.1.1) version 2.1.1 (`identity.vault.upbound.io`, `vault.vault.upbound.io`)
  - Keycloak Crossplane provider (`role.keycloak.crossplane.io`)
  - External Secrets Operator (external-secrets.io)

# RGD 
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: appaccess
spec:
  schema:
    apiVersion: v1alpha1
    kind: AppAccess
    spec:
      appName: string
      argoNamespace: string | default="argocd-system"
      enableKeycloak: boolean | default=true
      enableVault: boolean | default=true
      enableExec: boolean | default=true
      enableView: boolean | default=true

      # Keycloak (Argo roles)
      keycloakProviderConfig: string | default="keycloak-provider-config"
      argoClientRefName: string | default="argo-cd-client"
      realm: string | default="master-realm"

      # Keycloak (Vault client) - enable flag + client name
      enableKeycloakVaultRoles: boolean | default=true
      vaultClientRefName: string | default="vault-client"

      # Vault / ESO
      vaultProviderConfig: string | default="vault-provider-config"
[...]

PR #685, opened 2025-09-18
feat: add group mapping for special cases

TheCodingSheikh:
This is a very basic implementation, tested and working.
it allows to map special cases where the crd plural format is not correct using [github.com/gobuffalo/flect](github.com/gobuffalo/flect)

this implementation make it possible to set mapping using the environment variable `KRO_GROUP_RESOURCE_MAP`
to specify these custom conditions
for example
`KRO_GROUP_RESOURCE_MAP=groupalias=groupaliases,person=people`

this can be implemented in a better way, this is just a proof of concept, it's your choice of merging or not.

### DIFF
--- a/pkg/metadata/groupversion.go
+++ b/pkg/metadata/groupversion.go
@@ -16,6 +16,7 @@ package metadata
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/gobuffalo/flect"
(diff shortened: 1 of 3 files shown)


pkg/metadata/groupversion.go
```diff
@@ -87,8 +88,34 @@ func GVRtoGVK(gvr schema.GroupVersionResource) schema.GroupVersionKind {
...
+		if len(kv) == 2 {
+			cases[strings.ToLower(kv[0])] = strings.ToLower(kv[1])
+		}
+	}
+	return cases
+}
+
+var specialCases = loadSpecialCases()
+
 func GVKtoGVR(gvk schema.GroupVersionKind) schema.GroupVersionResource {
```
matthchr, line 112: I see what you're trying to do here, but wonder if it would make more sense for this to be a field on the RGD itself.

You could imagine something right [here](https://github.com/kubernetes-sigs/kro/blob/main/api/v1alpha1/resourcegraphdefinition_types.go#L58C2-L58C37), a new optional field like `Resource` or `PluralKind` or something. I'm not sure quite what the right name is.

matthchr: That way it doesn't require a weird env-based override on the pod

matthchr: Then that field could (if set) be used in place of flect entirely.

Though note that we also could just fix flect upstream to pluralize your resource correctly.

matthchr:
Also can you share what the actual resource name was you were trying to use that flect doesn't pluralize correctly?

TheCodingSheikh:
@matthchr it was GroupAlias
It pluralized it to groupalias while it should be groupaliases

your suggestion is great, i think it is a better approach

pkg/metadata/groupversion.go
a-hilaly: @matthchr @TheCodingSheikh how about leveraging simple schema syntax and doing something like:
```yaml
  kind: Peanut | plural=Peanuts
```

tjamet:
it looks like this is not required any longer: https://github.com/kubernetes-sigs/kro/issues/683#issuecomment-3527654356

a-hilaly:
Thank you for your PR @TheCodingSheikh - it looks like we won't need to support this now do we use proper RestMappers.
/close

PR #690, opened 2025-09-21
feat: Add TTL based caching to schema resolver

a-hilaly:
Performance testing with 100 RGDs and 100 concurrent reconcile workers revealed
the schema resolver was a major bottleneck. Memory profiling showed `ResolveSchema`
alone was consuming 78MB due to repeatedly fetching the same schemas (pods) from the
API server. The constant API calls were also triggering client side throttling,
with requests waiting over a second just due to rate limits.

This patch adds a caching layer that stores resolved schemas in memory with a 3 hour
TTL. To handle the concurrent access, using `x/singleflight` to ensure only one
worker fetches a schema while others wait for the result. The cache is capped
at 1000 entries to prevent unbounded growth.

After this change, ResolveSchema memory usage dropped to ~3MB and the
throttling warnings disappeared. Prometheus metrics were added to monitor cache
performance in production.

Tested with the same 100 RGDs and 100 concurrent workers - runs smoothly now
without the memory spikes or API server overloads.

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -37,6 +37,7 @@ import (
 	"github.com/kro-run/kro/pkg/graph/emulator"
 	"github.com/kro-run/kro/pkg/graph/parser"
 	"github.com/kro-run/kro/pkg/graph/schema"
+	schemaresolver "github.com/kro-run/kro/pkg/graph/schema/resolver"
 	"github.com/kro-run/kro/pkg/graph/variable"
 	"github.com/kro-run/kro/pkg/metadata"
 	"github.com/kro-run/kro/pkg/simpleschema"
@@ -46,7 +47,7 @@ import (
 func NewBuilder(
 	clientConfig *rest.Config,
 ) (*Builder, error) {
-	schemaResolver, dc, err := schema.NewCombinedResolver(clientConfig)
+	schemaResolver, dc, err := schemaresolver.NewCombinedResolver(clientConfig)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create schema resolver: %w", err)
 	}
--- a/pkg/graph/schema/resolver/metrics.go
+++ b/pkg/graph/schema/resolver/metrics.go
@@ -1,0 +1,69 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resolver
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	// Cache hit/miss metrics
+	cacheHitsTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "kro_schema_resolver_cache_hits_total",
+		Help: "Total number of schema resolver cache hits",
+	})
+	cacheMissesTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "kro_schema_resolver_cache_misses_total",
+		Help: "Total number of schema resolver cache misses",
+	})
+
+	// Cache size metric
+	cacheSize = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kro_schema_resolver_cache_size",
+		Help: "Current number of entries in the schema resolver cache",
+	})
+
+	// Cache evictions
+	cacheEvictionsTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "kro_schema_resolver_cache_evictions_total",
+		Help: "Total number of entries evicted from the schema resolver cache",
+	})
+
+	// API call duration
+	apiCallDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name:    "kro_schema_resolver_api_call_duration_seconds",
+		Help:    "Duration of API calls to fetch schemas",
+		Buckets: prometheus.ExponentialBuckets(0.01, 2, 10), // 10ms to ~10s
+	})
+
+	// Singleflight deduplicated requests
+	singleflightDeduplicatedTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "kro_schema_resolver_singleflight_deduplicated_total",
+		Help: "Total number of requests that were deduplicated by singleflight",
+	})
+)
+
+func init() {
+	// Register metrics with the controller-runtime metrics registry
+	metrics.Registry.MustRegister(
+		cacheHitsTotal,
+		cacheMissesTotal,
+		cacheSize,
+		cacheEvictionsTotal,
+		apiCallDuration,
+		singleflightDeduplicatedTotal,
+	)
+}
--- a/pkg/graph/schema/resolver/resolver.go
+++ b/pkg/graph/schema/resolver/resolver.go
@@ -12,29 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package schema
+package resolver
 
 import (
+	"time"
+
 	"k8s.io/apiextensions-apiserver/pkg/generated/openapi"
 	"k8s.io/apiserver/pkg/cel/openapi/resolver"
 	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 )
 
 // NewCombinedResolver creates a new schema resolver that can resolve both core and client types.
-func NewCombinedResolver(clientConfig *rest.Config) (resolver.SchemaResolver, *discovery.DiscoveryClient, error) {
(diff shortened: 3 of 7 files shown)


pkg/graph/schema/resolver/ttl_cached.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"golang.org/x/sync/singleflight"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/cel/openapi/resolver"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	"k8s.io/utils/clock"
+)
+
+// TTLCachedSchemaResolver caches schemas with time-based expiration and deduplicates concurrent requests
+type TTLCachedSchemaResolver struct {
```
jakobmoellerdev, line 28: I think we could just use https://pkg.go.dev/k8s.io/utils/lru and have full LRU capabilities without writing too much custom code. WDYT?

pkg/graph/schema/resolver/ttl_cached.go
```diff
@@ -0,0 +1,168 @@
...
+		evicted := 0
+		for k, v := range c.cache {
+			if c.clock.Since(v.timestamp) >= c.ttl {
+				delete(c.cache, k)
+				evicted++
+			}
+		}
+
+		// If still over limit, remove oldest
+		if len(c.cache) >= c.maxSize && c.maxSize > 0 {
```
jakobmoellerdev: Current eviction loop still checks >= c.maxSize which is zero. That means if len(c.cache) >= 0 (always true) (maxSize == 0), eviction will always fire after TTL sweep. Probably you want a guard: only enforce size limit if c.maxSize > 0. Otherwise no limit. WDYT?

a-hilaly:
@jakobmoellerdev i changed the implementation to use `hashicorp/golang-lru` - PTAL

pkg/graph/builder.go
```diff
@@ -46,7 +47,7 @@ import (
 func NewBuilder(
 	clientConfig *rest.Config,
 ) (*Builder, error) {
-	schemaResolver, dc, err := schema.NewCombinedResolver(clientConfig)
+	schemaResolver, dc, err := schemaresolver.NewCombinedResolver(clientConfig)
```
jakobmoellerdev, line 50: probably wanna move resolver to internal

pkg/graph/schema/resolver/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resolver
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
```
jakobmoellerdev: If you wanna offer this as package the registration should offer a `MustRegister` method instead of hardcoding against the controller runtime metrics registry

pkg/graph/schema/resolver/resolver.go
```diff
@@ -12,29 +12,37 @@
...
+func NewCombinedResolver(clientConfig *rest.Config) (resolver.SchemaResolver, discovery.DiscoveryInterface, error) {
+	// Create a regular discovery client first
 	discoveryClient, err := discovery.NewDiscoveryClientForConfig(clientConfig)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// Wrap it with memory-only caching
+	// This caches discovery information in memory with a TTL
+	memCachedDiscoveryClient := memory.NewMemCacheClient(discoveryClient)
```
jakobmoellerdev: this cached discovery client can result in behavior change because it needs a discovery reset. For example if a GVK was unregistered after it was discovered we need a way to reset it. Without this ability this cache is unsafe. Check https://pkg.go.dev/k8s.io/client-go@v0.34.1/discovery#CachedDiscoveryInterface for more details

pkg/graph/schema/resolver/resolver.go
```diff
@@ -47,5 +55,11 @@ func NewCombinedResolver(clientConfig *rest.Config) (resolver.SchemaResolver, *d
 	// Combine the two resolvers to create a single resolver that can resolve
 	// both core and client types.
 	combinedResolver := coreResolver.Combine(clientResolver)
-	return combinedResolver, discoveryClient, nil
+	// Wrap with TTL+LRU cache
+	cachedResolver := NewTTLCachedSchemaResolver(
+		combinedResolver,
+		500,           // maxSize: enough for 200 CRDs × 2-3 versions
+		5*time.Minute, // TTL: balance between freshness and performance
```
jakobmoellerdev: this definetly needs to be configurable and have these defaults configured via flag IMO.

pkg/graph/schema/resolver/ttl_cached.go
```diff
@@ -0,0 +1,104 @@
...
+		delegate: delegate,
+		cache:    cache,
+	}
+}
+
+// ResolveSchema resolves the schema for the given GroupVersionKind, using the cache if possible.
+// If multiple concurrent requests for the same GVK are made, they will be deduplicated.
+func (c *TTLCachedSchemaResolver) ResolveSchema(gvk schema.GroupVersionKind) (*spec.Schema, error) {
+	// Check cache first
+	if schema, ok := c.cache.Get(gvk); ok {
```
jakobmoellerdev, line 64: why dont we do this outside single flight when we do it here too? I  dont really see a lot of performance benefit (neglible) for this early check. Did you actually see behavior improvements?

pkg/graph/schema/resolver/ttl_cached.go
```diff
@@ -0,0 +1,104 @@
...
+			return schema, nil
+		}
+
+		// Actually fetch from delegate
+		start := time.Now()
+		schema, err := c.delegate.ResolveSchema(gvk)
+		apiCallDuration.Observe(time.Since(start).Seconds())
+
+		if err != nil {
+			return nil, err
```
jakobmoellerdev, line 85: if you add metrics I believe we should add failed resolutions as well to track error spikes

pkg/graph/schema/resolver/ttl_cached_test.go
```diff
@@ -0,0 +1,200 @@
...
+	_, err := cachedResolver.ResolveSchema(gvk)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if calls := mock.getCallCount(); calls != 1 {
+		t.Errorf("expected 1 initial call, got %d", calls)
+	}
+
+	time.Sleep(ttl + 50*time.Millisecond)
```
jakobmoellerdev, line 117: nit: maybe take a look at the new synctest utilities and see if you can use the test bubble

pkg/graph/schema/resolver/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// Cache evictions
+	cacheEvictionsTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "kro_schema_resolver_cache_evictions_total",
+		Help: "Total number of entries evicted from the schema resolver cache",
+	})
+
+	// API call duration
+	apiCallDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name:    "kro_schema_resolver_api_call_duration_seconds",
```
jakobmoellerdev: Im not sure if you can reliably track this because the resolver is not always directly pointing at the API. In your configuration it can use memory caching as well. (arguing that either the metric itself or the name of the metric is moot)

pkg/graph/schema/resolver/resolver.go
jakobmoellerdev: also, why do we need caching of GVK mappings here if we already cache one level above in the schema resolver? Feels like we got double layered caching and not sure if the tradeoff is good here.

a-hilaly: Good catch, left over from previous hack

pkg/graph/schema/resolver/resolver.go
a-hilaly: absolutely, will leave that for a different PR

pkg/graph/schema/resolver/ttl_cached.go
a-hilaly: mainly to avoid synchronization overhead for cache hits. Singleflight only deduplicates concurrent requests; after that, every request still goes through its locking without the outer check. Happy to remove if you prefer simpler code - both are correct

a-hilaly: >Did you actually see behavior improvements?

I didn't not compare both implementations, i mainly threw the implementation following the reasoning:
1/ if schema is cached, get it with an Rlock
2/ if not, only concurrent requests will enter the singleflight
3/ all subsequent Gets will get information from cache.

pkg/graph/schema/resolver/metrics.go
a-hilaly: I've refactored to wrap only the clientResolver (which actually makes API calls) with the TTL cache before combining it with the coreResolver

pkg/graph/schema/resolver/metrics.go
a-hilaly: Add MustRegister, but i'll actually do the rest of the changes (get rid of the init here) in a seperate api package. I would like to address all the metrics/package reusability stuff in a seperate PR

pkg/graph/schema/resolver/ttl_cached_test.go
a-hilaly: it looks like the LRU implementation spins a goroutine making it unusable with synctest

a-hilaly: also it looks like it's behind a GOEXPERIMENTAL flag, do we want to use these?

PR #691, opened 2025-09-21
Add `self` and `this` to reserved keywords

a-hilaly:
After [this conversation](https://github.com/kubernetes-sigs/kro/pull/679#discussion_r2361911400) i'd like to book these two keywords just in case they will be helpful for collections


jakobmoellerdev (approved):
/lgtm

I promise Im allowed to do this 😢

PR #692, opened 2025-09-21
perf: add caching layer to resource emulator to prevent redundant gen…

a-hilaly:
The emulator was generating identical sample resources multiple times when
processing multiple RGDs with the same resource types (e.g 100 RGDs all
referencing Pods). This caused excessive memory allocations and CPU usage

This patch adds a CachedEmulator wrapper that:
- Caches generated samples by `GroupVersionKind`
- uses `singleflight` pattern to prevent concurrent duplicate generation
- returns cached samples for subsequent requests of the same GVK

Performance impacted was neat ~66% reduction (tested with 100RGDs) reduce
from 33.5MiB to 11MiB

Note to reviewers: we could've used a similar pattern to resolver optimizations
but this package is used mainly for dry runs and not expecting much edge cases
with schema changes. Mid/Long term we could switch to an LRU implementation for
both package.


a-hilaly:
Closing a we need to get rid of the emulator instead.
/close

PR #693, opened 2025-09-21
feat: implement non-blocking CRD reconciliation with requeue support

a-hilaly:
Implement non blocking CRD reconciliation to improve controller throughput
when managing large amount of RGDs. The main idea is to free the workers to do
other tasks when we're waiting for the establishement.

Previously, the controller would block waiting for CRDs to become established,
preventing other ResourceGraphDefinitions from being processed. This change
introduces a requeue based approach where:

- CRD creation happens immediately without blocking
- controller checks CRD establishment status on each reconciliation
- if CRD exists but isn't established, reconciler requeues after 500ms
- other RGDs can be processed while waiting for CRD establishment

this patch also gets rid of the CRDWraper abstraction layer as it's no longer
needed.


pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -151,10 +168,44 @@ func buildResourceInfo(name string, deps []string) v1alpha1.ResourceInformation
...
+	}
+
+	// CRD exists, check if it's ready (Established condition)
+	established := false
+	for _, cond := range existing.Status.Conditions {
+		if cond.Type == v1.Established && cond.Status == v1.ConditionTrue {
+			established = true
+			break
+		}
+	}
```
a-hilaly: note to self: could've reused `isCRDEstablished`

test/integration/suites/core/crd_test.go
a-hilaly, line 1: I know this is not enough, and probably near impossible to do proper testing for this feature at the integration level. Planning on adding unit tests to the controller.go where we properly test these things.

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -158,6 +156,9 @@ func (r *ResourceGraphDefinitionReconciler) findRGDsForCRD(ctx context.Context,
 func (r *ResourceGraphDefinitionReconciler) Reconcile(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (ctrl.Result, error) {
 	if !o.DeletionTimestamp.IsZero() {
 		if err := r.cleanupResourceGraphDefinition(ctx, o); err != nil {
+			if needRequeue, duration := requeue.Check(err); needRequeue {
```
tjamet, line 159: What about extracting this logic to a distinct function
That way we keep our kro internal semantics and reduce risk of "forgetting error check for RequeueAfter" for future changes

```

// keep Reconcile(ctxm req) (result, obj) as this is the interface for controller-runtime
// TODO: check opportunities to converge interfaces with controllerRuntime for them
// to move to our requeue logic
func (r *ResourceGraphDefinitionReconciler) Reconcile(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (error) {
    err := r.reconcileRGD(ctx, o)
    if needRequeue, duration := requeue.Check(err); needRequeue {
        return ctrl.Result{RequeueAfter: duration}, nil
    }
   return ctrl.Result{}, nil
}

func (r *ResourceGraphDefinitionReconciler) reconcileRGD(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (ctrl.Result, error) {
// the current Reconcile
[...]

```

pkg/controller/resourcegraphdefinition/controller_cleanup.go
```diff
@@ -46,10 +56,16 @@ func (r *ResourceGraphDefinitionReconciler) cleanupResourceGraphDefinition(ctx c
...
-	if err := r.cleanupResourceGraphDefinitionCRD(ctx, crdName); err != nil {
+	completed, err := r.cleanupResourceGraphDefinitionCRD(ctx, crdName)
+	if err != nil {
 		return fmt.Errorf("failed to cleanup CRD %s: %w", crdName, err)
 	}
 
+	if !completed {
+		log.V(1).Info("CRD deletion in progress, requeuing", "crd", crdName)
+		return requeue.NeededAfter(nil, crdDeletionRequeueDuration)
+	}
```
tjamet, line 67: I sense this has some overlap with KREP-004 (#763 )
with KREP-004 we wouldn't need to manange requeues on our own. RGD reconciliation would be re-triggered as the CRD got deleted

pkg/controller/resourcegraphdefinition/controller_cleanup.go
```diff
@@ -63,17 +79,50 @@ func (r *ResourceGraphDefinitionReconciler) shutdownResourceGraphDefinitionMicro
...
+	// If CRD has no deletion timestamp, initiate deletion
+	if crd.DeletionTimestamp.IsZero() {
+		log.V(1).Info("initiating CRD deletion", "crd", name)
+		err = r.clientSet.CRD().Delete(ctx, name, metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, err
+		}
+		// Deletion initiated, but not complete yet
+		return false, nil
+	}
```
tjamet, line 121: Why is the `get` and `DeletionTimestamp` check required?

IIRC `err = r.clientSet.CRD().Delete(ctx, name, metav1.DeleteOptions{})` triggers the deletion of the CRD async already. Shall the CRD already have a `DeletionTimestamp` the API should also return immediately. Am I missing anything?

if so, we could eventually avoid some API calls and simplify the code with just this (we may lose some granularity on the log side though)

```suggestion
	log.V(1).Info("ensure CRD is marked for deletion", "crd", name)
	err = r.clientSet.CRD().Delete(ctx, name, metav1.DeleteOptions{})
	if err != nil && !apierrors.IsNotFound(err) {
		return false, err
	}
	if apierrors.IsNotFound(err) {
		log.V(1).Info("CRD deletion completed", "crd", name)
	}
	// Deletion initiated, but not complete yet
	return false, nil
```

tjamet (commented):
I wonder whether it makes sense to tackle it together with KREP-004 it would possibly simplify how we deal with CRD management and allow simplify the reconciliation steps (dropping needs for requeues)

a-hilaly:
Thank you for the feedback @tjamet - i'll address a part of the comments in my next CR. On the ownership reference for the CRD i believe that needs to be reviewed and done seperately.

PR #694, opened 2025-09-21
fix!: use kro-controller as leaderElectionID 

a-hilaly:
- Use descriptive lease name "kro-controller" instead of "6f0f64a5..."

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -152,22 +152,12 @@ func main() {
 		Metrics: metricsserver.Options{
 			BindAddress: metricsAddr,
 		},
-		HealthProbeBindAddress:  probeAddr,
-		LeaderElection:          enableLeaderElection,
-		LeaderElectionID:        "6f0f64a5.kro.run",
-		LeaderElectionNamespace: leaderElectionNamespace,
-		// LeaderElectionReleaseOnCancel defines if the leader should step down voluntarily
-		// when the Manager ends. This requires the binary to immediately end when the
-		// Manager is stopped, otherwise, this setting is unsafe. Setting this significantly
-		// speeds up voluntary leader transitions as the new leader don't have to wait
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -152,22 +152,12 @@ func main() {
 		Metrics: metricsserver.Options{
 			BindAddress: metricsAddr,
 		},
-		HealthProbeBindAddress:  probeAddr,
-		LeaderElection:          enableLeaderElection,
-		LeaderElectionID:        "6f0f64a5.kro.run",
```
jakobmoellerdev, line 157: this is a breaking change

cmd/controller/main.go
```diff
@@ -152,22 +152,12 @@ func main() {
...
-		// the manager stops, so would be fine to enable this option. However,
-		// if you are doing or is intended to do any operation such as perform cleanups
-		// after the manager stops then its usage might be unsafe.
-		// LeaderElectionReleaseOnCancel: true,
-		Logger: rootLogger,
+		HealthProbeBindAddress:        probeAddr,
+		LeaderElection:                enableLeaderElection,
+		LeaderElectionID:              "kro-controller",
+		LeaderElectionNamespace:       leaderElectionNamespace,
+		LeaderElectionReleaseOnCancel: true,
```
jakobmoellerdev: Maybe Im wrong but I think we cannot do this because of our dynamic controller. It shuts down after cancel, so this would be a conflict. You would have to merge https://github.com/kubernetes-sigs/kro/pull/616 first I think

matthchr: Yes I think @jakobmoellerdev is right here

cmd/controller/main.go
barney-s: ack. release noting it.

PR #699, opened 2025-09-28
refactor: use domain-based naming for leader election ID

a-hilaly:
Change `LeaderElectionID` from `kro-controller` to `controller.kro.run` to follow k8s convention of using domain notation


jakobmoellerdev (approved):
/lgtm

PR #700, opened 2025-09-29
chore(tests): properly use SpecContext from Ginkgo

jakobmoellerdev:
Standardize test cases to leverage interruptible nodes as per https://onsi.github.io/ginkgo/#spec-timeouts-and-interruptible-nodes

This is important because we used context.Background before which can cause stuck test suites on cancellation. By using the ginkgo spec context this is automatically managed for us by ginkgo

created as result from discussion: https://github.com/kubernetes-sigs/kro/pull/264#discussion_r2386193396


test/integration/suites/ackekscluster/suite_test.go
```diff
@@ -79,7 +77,7 @@ var _ = Describe("EKSCluster", func() {
 
 		// Verify ResourceGraphDefinition is created and becomes ready
 		createdRGD := &krov1alpha1.ResourceGraphDefinition{}
-		Eventually(func(g Gomega) {
+		Eventually(func(g Gomega, ctx SpecContext) {
```
fabianburth, line 80: You likely already know this, still wanted to mention that there is also the less verbose option to pass in the `SpecContext` as the the first argument to `Eventually`.

So `Eventually(ctx, func(g Gomega) {...})` is equivalent to `Eventually(ctx, func(g Gomega) {...}).WithContext(ctx)`

(not saying we should change this here)

jakobmoellerdev:
/hold

until I fix flakiness in the nested test

test/integration/suites/ackekscluster/suite_test.go
jakobmoellerdev: I think you meant to say `Eventually(ctx, func(g Gomega) {...})` is the same as `Eventually(func(g Gomega) {...}).WithContext(ctx)` right?

fabianburth: oops, yep

jakobmoellerdev:
@a-hilaly I fixed https://github.com/kubernetes-sigs/kro/pull/264 and reintroduced it here. The test had an unfortunate situation where the RGD managed another RGD that was shared between two test cases. This caused a race. I removed the duplicate case so the test runs fine now.

pkg/applyset/applyset.go
a-hilaly, line 1: out of scope for this PR?

pkg/controller/instance/controller_reconcile.go
a-hilaly, line 1: ditto

pkg/dynamiccontroller/dynamic_controller.go
a-hilaly, line 1: ditto

test/integration/suites/core/garbage_collection_test.go
a-hilaly, line 1: w00t!

test/integration/suites/core/include_when_test.go
a-hilaly, line 1: just realized we have a camelcase named file let's s/includeWhen_test.go/include_when_test.go` ?

test/integration/suites/core/nested_test.go
a-hilaly, line 1: Thanks a lot for doing this!

pkg/applyset/applyset.go
a-hilaly: let's move the logging enhancements to a seperate PR?

test/integration/suites/core/include_when_test.go
jakobmoellerdev: agreed, renamed

test/integration/suites/ackekscluster/suite_test.go
jakobmoellerdev: youre right, but i think this way its much more explicit what is used

PR #701, opened 2025-09-29
Revert "test(core): RGDs all the way down"

a-hilaly:
Reverts kubernetes-sigs/kro#264 - due to flaky tests


jakobmoellerdev:
/lgtm 

revert should go through since its clearly causing issues. we will readd the test once we are sure whats the core issue. 

Its interesting because our machines seem to not be able to replicate, so issue happens only on slower CI based runners...

PR #702, opened 2025-09-29
chore: ensure we patch RGD status for any Reconcile errors

michaelhtm:
Description:
During RGD reconcile, if there are any issues when setting RGD CR as
managed or unmanaged, any issues that may occur are not reflected on the
RGD CR status.

With these changes, we will ensure any issues are patched to the RGD
status


pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -154,7 +154,14 @@ func (r *ResourceGraphDefinitionReconciler) findRGDsForCRD(ctx context.Context,
 	}
 }
 
-func (r *ResourceGraphDefinitionReconciler) Reconcile(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (ctrl.Result, error) {
+func (r *ResourceGraphDefinitionReconciler) Reconcile(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (res ctrl.Result, reconcileErr error) {
+	var topologicalOrder []string
+	var resourcesInformation []v1alpha1.ResourceInformation
+	defer func() {
+		if err := r.updateStatus(ctx, o, topologicalOrder, resourcesInformation); err != nil {
+			reconcileErr = errors.Join(reconcileErr, err)
```
barney-s: Did  your `deferlinter` catch this @jakobmoellerdev  ?

barney-s: A comment on why this pattern works with defer would be good. This looks like one of those things that may trip people up if they are not paying attention.

Comments along the lines of named return vars can be updated in a defer call.

michaelhtm: Added comments

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -154,7 +155,23 @@ func (r *ResourceGraphDefinitionReconciler) findRGDsForCRD(ctx context.Context,
...
+	// patched to the resource
+	defer func() {
+		// if the updateStatus error is a k8s NotFound error we want to skip it.
+		// If not, we will merge it to the returned error
+		if err := r.updateStatus(ctx, o, topologicalOrder, resourcesInformation); err != nil {
+			// we don't want to reconcile if resource is deleted
+			if !apierrors.IsNotFound(err) {
+				reconcileErr = errors.Join(reconcileErr, err)
+			} else {
+				reconcileErr = nil
```
jakobmoellerdev: why would you set reconcileErr to nil explicitly here? because youre using err there is no conflict anyhow right?

pkg/controller/resourcegraphdefinition/controller.go
jakobmoellerdev: yeah, https://github.com/jakobmoellerdev/deferrlint spotted this:

```
deferrlint ./...
.../kro/pkg/controller/instance/controller_reconcile.go:120:6: deferred function assigns to error-typed variable `err`, but it is not a named return value
.../kro/pkg/controller/resourcegraphdefinition/controller.go:166:6: deferred function assigns to error-typed variable `err`, but it is not a named return value
```

jakobmoellerdev: I will say this is a strange one and the linter doesnt detect this correctly though, I believe reconcileErr should be fine here because it is reassigned

pkg/controller/resourcegraphdefinition/controller.go
jakobmoellerdev: also I hate to be a nit picker here, but would it be possible to write a test to make sure the RGD status is patched on error case?

michaelhtm: I wanted to add tests for `patch` behavior, but currently there's no package that mocks `finalize` failures.

>why would you set reconcileErr to nil explicitly here?

The main reason was to avoid triggering another reconciliation if the resource no longer exists..

api/v1alpha1/conditions.go
```diff
@@ -24,22 +24,6 @@ type ConditionType string
...
-	// CustomResourceDefinition (CRD) that kro uses to manage the resources in a
-	// ResourceGraphDefinition.
-	ResourceGraphDefinitionConditionTypeCustomResourceDefinitionSynced ConditionType = "CustomResourceDefinitionSynced"
-	// ResourceGraphDefinitionConditionTypeReconcilerReady indicates the state of the reconciler.
-	// Whenever an ResourceGraphDefinition resource is created, kro will spin up a
-	// reconciler for that resource. This condition indicates the state of the
-	// reconciler.
-	ResourceGraphDefinitionConditionTypeReconcilerReady ConditionType = "ReconcilerReady"
-)
-
```
michaelhtm, line 42: these are not used anywhere in the code base

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -154,26 +155,46 @@ func (r *ResourceGraphDefinitionReconciler) findRGDsForCRD(ctx context.Context,
...
+	// defer function will ensure any errors that occur during reconcile will eventually be
+	// patched to the resource
+	defer func() {
+		// if the updateStatus error is a k8s NotFound error we want to skip it.
+		// If not, we will merge it to the returned error
+		if patchErr := r.updateStatus(ctx, o, topologicalOrder, resourcesInformation); patchErr != nil {
+			if apierrors.IsNotFound(patchErr) {
+				// we don't want to return an error and requeue if 
+				// rgd is not found (deleted)
+				err = nil
```
michaelhtm, line 170: ignore the error if resource no longer exists

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -154,26 +155,46 @@ func (r *ResourceGraphDefinitionReconciler) findRGDsForCRD(ctx context.Context,
...
-func (r *ResourceGraphDefinitionReconciler) Reconcile(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (ctrl.Result, error) {
+func (r *ResourceGraphDefinitionReconciler) Reconcile(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (res ctrl.Result, err error) {
+	var topologicalOrder []string
+	var resourcesInformation []v1alpha1.ResourceInformation
+	// defer function will ensure any errors that occur during reconcile will eventually be
+	// patched to the resource
+	defer func() {
+		// if the updateStatus error is a k8s NotFound error we want to skip it.
+		// If not, we will merge it to the returned error
+		if patchErr := r.updateStatus(ctx, o, topologicalOrder, resourcesInformation); patchErr != nil {
```
cheeseandcereal, line 166: If this `updateStatus` is happening in a deferred func, now we _ALWAYS_ will run this function. This does not match the behavior that exists below. For example, if `!o.DeletionTimestamp.IsZero()`, we run `r.cleanupResourceGraphDefinition` instead and then return.

With this change, we would run `r.cleanupResourceGraphDefinition`, try to return, and then try to run `r.updateStatus` afterwards which definitely isn't correct.

michaelhtm: This change is mainly addressing issues when either cleanup or finalizer add/remove fails. In those cases, we want kro to report this information in the RGD status. This would allow users to see any issues in the resource status instead of looking for them in the logs..

cheeseandcereal: But doesn't this now run even if the cleanup succeeds where it wasn't before?

michaelhtm: It does. The information of the successful cleanup may be useful to users. Especially if the RGD has other finalizers that are preventing it from being removed.

cheeseandcereal: Ok, so it's explicit/intentional that we are changing from only updating this sometime to _always_ updating it. If that's the case then this makes sense

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -154,26 +155,46 @@ func (r *ResourceGraphDefinitionReconciler) findRGDsForCRD(ctx context.Context,
...
 		}
+		mark.ResourceGraphFinalized("removed")
 		return ctrl.Result{}, nil
 	}
 
 	if err := r.setManaged(ctx, o); err != nil {
+		mark.ResourceGraphUnfinalized(err.Error())
 		return ctrl.Result{}, err
 	}
+	mark.ResourceGraphFinalized("added")
```
fabianburth, line 195: IMHO, `ResourceGraphFinalized` feels a little off here, semantically. Still rather a nit though, because the message describes it reasonably well.

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -154,26 +155,46 @@ func (r *ResourceGraphDefinitionReconciler) findRGDsForCRD(ctx context.Context,
...
+func (r *ResourceGraphDefinitionReconciler) Reconcile(ctx context.Context, o *v1alpha1.ResourceGraphDefinition) (res ctrl.Result, err error) {
+	var topologicalOrder []string
+	var resourcesInformation []v1alpha1.ResourceInformation
+	// defer function will ensure any errors that occur during reconcile will eventually be
+	// patched to the resource
+	defer func() {
+		// if the updateStatus error is a k8s NotFound error we want to skip it.
+		// If not, we will merge it to the returned error
+		if patchErr := r.updateStatus(ctx, o, topologicalOrder, resourcesInformation); patchErr != nil {
+			if apierrors.IsNotFound(patchErr) {
```
fabianburth, line 167: Do we need an additional check for `err == nil` here, to prevent suppressing potential other errors? Or maybe at least log `err` if `err != nil`? 
```suggestion
			if apierrors.IsNotFound(patchErr) && err == nil {
```

michaelhtm: great catch! 
what if we set `Ready` condition to false if there's an error?

pkg/controller/resourcegraphdefinition/controller.go
michaelhtm: I'm not too opinionated here..i'm willing to remove it.
should we only set it on failures? and ignore it when it succeeds?

a-hilaly:
Hey @michaelhtm - can you please rebase this PR?

jakobmoellerdev:
closing this one until its further revisted / rebased. feel free to open again once youre ready!

/close

PR #704, opened 2025-10-02
chore!: remove service account spec and impersonation

jakobmoellerdev:
In the last community call we decided it would be smarter to get rid of the existing service account specification and implement a proper design on it in favor of the existing impersonation client that is currently underdeveloped


a-hilaly (approved):
👍

barney-s:
/approve
/lgtm 

Thanks @jakobmoellerdev

PR #705, opened 2025-10-02
!feat: Reserve item and items as keywords for collections

ellistarn:
Following discussion of the new collections KREP, it would be prudent to reserve these potential keywords. Discussed with @a-hilaly offline.


a-hilaly (approved):
/lgtm

PR #708, opened 2025-10-03
fix: Use controllerruntime's namespaced name correctly

ellistarn:
Simplified how we're using NamespacedNames to more naturally integrate for DynamicController consumers.

Tested manually by creating an RGD, RG, and deleting both.


a-hilaly (approved):
nice catch, thanks!
/lgtm

PR #710, opened 2025-10-03
chore: Hydrate goproxy after each push

ellistarn:
Modules don't land in goproxy until they are depended on by another package and pulled through the proxy. Some corporate firewalls (...) synchronize from go proxy, and until packages are hydrated there, they cannot be reflected into internal build systems. This change automates that pull and makes sure the proxy always has the latest version.

See: https://github.com/awslabs/operatorpkg/blob/main/.github/workflows/postsubmit.yaml

### DIFF
--- a/.github/workflows/postsubmit.yaml
+++ b/.github/workflows/postsubmit.yaml
@@ -1,0 +1,16 @@
+name: postsubmit
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+jobs:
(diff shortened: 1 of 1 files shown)


jakobmoellerdev (commented):
generally not opposed to this but we did want to migrate to kubernetes job infrastructure so any github action we add here we probably want to migrate off. its not a big deal here but I still think we should get another opinion

PR #711, opened 2025-10-06
fix: Broken mod name in postsubmit.yaml

ellistarn:
Fixes: https://github.com/kubernetes-sigs/kro/actions/runs/18279713855/job/52039714643


ellistarn:
Oh @a-hilaly already fixed this.

PR #712, opened 2025-10-08
feat: add concurrency support in applyset operations

jakobmoellerdev:
Introduce configurable concurrency limit for applyset operations using `sync/errgroup`. This improves performance by enabling concurrent processing of apply api calls. Defaults to the number of objects in the applyset if unspecified.

This makes it so by default we apply all objects in one apply set used concurrently. we can later on decide to make this configurable but for most folks with multiple resources in an instance this should significantly speed up reconciliation

### DIFF
--- a/pkg/applyset/applyset.go
+++ b/pkg/applyset/applyset.go
@@ -29,8 +29,10 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/go-logr/logr"
+	"golang.org/x/sync/errgroup"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


a-hilaly (approved):
Dankeshoen her Jakob!

barney-s:
This is good. We had a bunch of back and forth on the original applyset.

Added parallelization and removed it for pruning based on : https://github.com/kubernetes-sigs/kro/pull/561#issuecomment-3250128948 and followup.s

If we are warming up to parallelization, lets do it for pruning too.
Please take a look at `PruneGVKParallelizationLimit`.

jakobmoellerdev:
@barney-s thanks for the pointer! added pruning concurrency limit as well.
I believe concurrency is key to efficient apply set, otherwise we let a lot on the table

barney-s:
/approve
/lgtm 

We need a follow up to expose this via command line params ?

barney-s:
@matthchr @a-hilaly - How do we squash and merge this PR using automation ?

PR #713, opened 2025-10-08
feat: introduce ApplyMode to ResourceGraphDefinitions (switch between Server Side Apply and Client Side Delta)

jakobmoellerdev:
- Added support for `ApplyMode` to enable configurable reconciliation modes (CSA via Delta - our old implementation and SSA via ApplySets) for instances.
- Refactored instance reconciliation logic to accommodate new apply modes.
- Updated integration tests to validate `ApplyMode` functionalities so we test both

_I know that we have CRD conflicts in some tests when running both modes next to each other because the CRD names overlap and can cause issues. to fix this we need to make sure RGDs are only fully removed once the CRD is fully unregistered_

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -18,7 +18,7 @@ jobs:
   integration:
     strategy:
       matrix:
-        labels: [ "ClientSideDelta", "ApplySet" ]
+        mode: [ "ClientSideDelta", "ApplySet" ]
     name: Integration Tests
     runs-on: ubuntu-latest
     
@@ -40,4 +40,4 @@ jobs:
         echo "KUBEBUILDER_ASSETS=${KUBEBUILDER_ASSETS}" >> $GITHUB_ENV
 
     - name: Run integration tests
-      run: make test WHAT=integration INTEGRATION_LABEL_FILTER=${{ matrix.labels }}
+      run: make test WHAT=integration KRO_DEFAULT_RGD_RECONCILE_MODE=${{ matrix.mode }}
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -50,9 +50,10 @@ type ResourceGraphDefinitionReconcileSpec struct {
 // When this is set, all resources will be managed with an ApplySet based ServerSideApply implementation.
 // More info: https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/3659-kubectl-apply-prune
 //
+// If not set, default reconciliation behavior is defined by the controller.
+//
 // +kubebuilder:validation:Enum=ApplySet;ClientSideDelta
-// +kubebuilder:default=ClientSideDelta
-// +kubebuilder:validation:Required
+// +kubebuilder:validation:Optional
 // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="mode is immutable"
 type ResourceGraphDefinitionReconcileMode string
 
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -16,9 +16,11 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"os"
 	"time"
 
+	"github.com/kubernetes-sigs/kro/api/v1alpha1"
 	"go.uber.org/zap/zapcore"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -29,12 +31,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
-	xv1alpha1 "github.com/kubernetes-sigs/kro/api/v1alpha1"
 	kroclient "github.com/kubernetes-sigs/kro/pkg/client"
 	resourcegraphdefinitionctrl "github.com/kubernetes-sigs/kro/pkg/controller/resourcegraphdefinition"
 	"github.com/kubernetes-sigs/kro/pkg/dynamiccontroller"
 	"github.com/kubernetes-sigs/kro/pkg/graph"
-	//+kubebuilder:scaffold:imports
+	// +kubebuilder:scaffold:imports
 )
 
 var (
@@ -45,9 +46,9 @@ var (
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
-	utilruntime.Must(xv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(v1alpha1.AddToScheme(scheme))
 	utilruntime.Must(extv1.AddToScheme(scheme))
-	//+kubebuilder:scaffold:scheme
+	// +kubebuilder:scaffold:scheme
 }
 
 type customLevelEnabler struct {
@@ -66,6 +67,7 @@ func main() {
 		probeAddr                                   string
 		allowCRDDeletion                            bool
 		resourceGraphDefinitionConcurrentReconciles int
+		resourceGraphDefinitionDefaultReconcileMode string
 		dynamicControllerConcurrentReconciles       int
 		// dynamic controller rate limiter parameters
 		minRetryDelay time.Duration
@@ -97,6 +99,14 @@ func main() {
 	flag.IntVar(&resourceGraphDefinitionConcurrentReconciles,
 		"resource-graph-definition-concurrent-reconciles", 1,
 		"The number of resource graph definition reconciles to run in parallel",
+	)
+	flag.StringVar(&resourceGraphDefinitionDefaultReconcileMode,
+		"resource-graph-definition-default-reconcile-mode",
+		string(v1alpha1.ResourceGraphDefinitionReconcileModeClientSideDelta),
+		fmt.Sprintf("default reconcile mode for resource graph definition instance resources. one of %s or %s.",
+			v1alpha1.ResourceGraphDefinitionReconcileModeClientSideDelta,
+			v1alpha1.ResourceGraphDefinitionReconcileModeApplySet,
+		),
 	)
 	flag.IntVar(&dynamicControllerConcurrentReconciles,
 		"dynamic-controller-concurrent-reconciles", 1,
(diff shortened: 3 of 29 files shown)


barney-s:
> * Added support for `ApplyMode` to enable configurable reconciliation modes (CSA via Delta - our old implementation and SSA via ApplySets) for instances.

Is there a reason we want CSA ?

jakobmoellerdev:
Mostly CSA/SSA is a tradeoff on accuracy and performance. I would like to make SSA opt-in for the first few releases so that people can decide to push on SSA. Later on I have no problem with only allowing SSA. I think in general its just a much safer way to roll this out.

barney-s:
> Mostly CSA/SSA is a tradeoff on accuracy and performance. I would like to make SSA opt-in for the first few releases so that people can decide to push on SSA. Later on I have no problem with only allowing SSA. I think in general its just a much safer way to roll this out.

I am all in for rolling out the feature behind an optin. 
Just wanting to be clear it is not CSA vs SSA. Rather applyset vs existing get-diff-create/update method.

We need clear guidelines for the community.  Docs that clearly explains why we have 2 options. Make it clear to the community we are providing this as a stop gap option for users to migrate to SSA and use that CSA would be deprecated.

jakobmoellerdev:
@barney-s Ive now versioned off docs together with @a-hilaly so I could add docs on this:

Also renamed CSA to `ClientSideDelta` and SSA to `ApplySet` so one could have something like

```yaml
spec:
  reconcile:
    mode: ClientSideDelta # or ApplySet
```

hope that fits better! Also PTAL at the docs

.github/workflows/integration-tests.yaml
```diff
@@ -16,6 +16,9 @@ on:
 
 jobs:
   integration:
+    strategy:
+      matrix:
+        labels: [ "ClientSideDelta", "ApplySet" ]
```
barney-s: does this run it twice once for each label value ?

pkg/controller/instance/controller_reconcile.go
```diff
@@ -519,3 +348,61 @@ func (igr *instanceGraphReconciler) getResourceNamespace(resourceID string) stri
...
+	if err != nil {
+		return err
+	}
+	if patched != nil {
+		instance.Object = patched.Object
+	}
+	return nil
+}
+
+// setManaged ensures the instance has the necessary finalizer and labels.
```
barney-s: If this function has not changed would it be possible to move it back to reduce the diff ?

pkg/controller/instance/client_side_delta.go
```diff
@@ -0,0 +1,210 @@
...
+	"context"
+	"fmt"
+
+	"github.com/kubernetes-sigs/kro/pkg/controller/instance/delta"
+	"github.com/kubernetes-sigs/kro/pkg/runtime"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
+)
```
barney-s, line 27: Is this the same as prev implementation (pre-applyset merge). Or has there been improvements made.

pkg/controller/instance/controller_reconcile.go
```diff
@@ -102,7 +92,14 @@ func (igr *instanceGraphReconciler) reconcile(ctx context.Context) error {
 		return igr.handleReconciliation(ctx, igr.handleInstanceDeletion)
 	}
 
-	return igr.handleReconciliation(ctx, igr.reconcileInstance)
+	switch igr.reconcileConfig.Mode {
+	case v1alpha1.ResourceGraphDefinitionReconcileModeApplySet:
+		return igr.handleReconciliation(ctx, igr.reconcileInstanceApplySet)
+	case v1alpha1.ResourceGraphDefinitionReconcileModeClientSideDelta, "":
+		return igr.handleReconciliation(ctx, igr.reconcileInstanceCSA)
```
barney-s: rename to CSD ?

pkg/controller/instance/client_side_delta.go
```diff
@@ -0,0 +1,210 @@
+// Copyright 2025 The Kube Resource Orchestrator Authors
+//
```
barney-s, line 2: maybe rename the file to csd

.github/workflows/integration-tests.yaml
jakobmoellerdev: this should create one job for each value!

pkg/controller/instance/client_side_delta.go
barney-s: this was for the old file which was named just `delta`. the new file name is good.

pkg/controller/instance/client_side_delta.go
jakobmoellerdev: its exactly the same as before afaik. just renamed functions to be equivalent to the other mode.

pkg/controller/instance/controller_reconcile.go
jakobmoellerdev: good callout, moved back. Thanks!

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -33,6 +35,40 @@ type ResourceGraphDefinitionSpec struct {
 	Resources []*Resource `json:"resources,omitempty"`
 }
 
+type ResourceGraphDefinitionReconcileSpec struct {
```
barney-s, line 38: more of a meta Q. Do we need this per resource? 

I see value. Wondering if we should start with a global setting first.

cmd/controller/main.go
```diff
@@ -98,6 +99,11 @@ func main() {
 		"resource-graph-definition-concurrent-reconciles", 1,
 		"The number of resource graph definition reconciles to run in parallel",
 	)
+	flag.StringVar(&resourceGraphDefinitionDefaultReconcileMode,
+		"resource-graph-definition-default-reconcile-mode",
+		string(xv1alpha1.ResourceGraphDefinitionReconcileModeClientSideDelta),
+		"The default reconcile mode for resource graph definition instance resources if not explicitly set.",
```
barney-s: may be print allowed values ?

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: talked a few times with @a-hilaly on this and both dynamic watches (watchsets) and this were things we talked about many times to be configurable by spec field. Looking for other thoughts here on what we should do.

cmd/controller/main.go
jakobmoellerdev: good point, added.

cmd/controller/main.go
```diff
@@ -98,6 +100,14 @@ func main() {
 		"resource-graph-definition-concurrent-reconciles", 1,
 		"The number of resource graph definition reconciles to run in parallel",
 	)
+	flag.StringVar(&resourceGraphDefinitionDefaultReconcileMode,
```
barney-s, line 103: There is a way to use flag.Var with enums if we expose a Set() and String() on the type.

```
package main

import (
	"flag"
	"fmt"
)

// MyEnum represents an example enumeration.
type MyEnum string

const (
	ValueA MyEnum = "A"
	ValueB MyEnum = "B"
	ValueC MyEnum = "C"
)

// String implements the fmt.Stringer interface for MyEnum.
func (e MyEnum) String() string {
	return string(e)
}

// Set implements the flag.Value interface for MyEnum.
func (e *MyEnum) Set(s string) error {
	switch s {
	case string(ValueA):
		*e = ValueA
	case string(ValueB):
		*e = ValueB
	case string(ValueC):
		*e = ValueC
	default:
		return fmt.Errorf("invalid enum value: %s", s)
	}
	return nil
}

func main() {
	var myFlag MyEnum // Declare a variable of your custom enum type

	// Bind the enum variable to a command-line flag
	flag.Var(&myFlag, "myenum", "A flag that accepts specific enum values (A, B, C)")

	flag.Parse() // Parse command-line arguments

	fmt.Printf("MyEnum flag value: %s\n", myFlag)
}

```

jakobmoellerdev: sure we can do that. one caveats: the help will show "value" as type. apart from that can adjust

.github/workflows/integration-tests.yaml
```diff
@@ -16,6 +16,9 @@ on:
 
 jobs:
   integration:
+    strategy:
+      matrix:
+        mode: [ "ClientSideDelta", "ApplySet" ]
```
barney-s, line 21: we shoud do something similar for e2e ?  Can do in a follow up PR. we need to make sure e2e tests are not dropped for applyset.

barney-s:
@jakobmoellerdev - with the latest changes we don't have to set the `.spec.reconcile.mode` in the RGDs right  ?
If nothing is set it would take controller default (which can be switched via a flag)

cmd/controller/main.go
```diff
@@ -98,6 +100,14 @@ func main() {
...
+		"resource-graph-definition-default-reconcile-mode",
+		string(v1alpha1.ResourceGraphDefinitionReconcileModeClientSideDelta),
```
matthchr, line 105: Feel like this isn't the right default - we're alpha, and at least in theory this change is not breaking. We should default to the new way. If we're going to allow users to configure, allowing them to fix by moving back to old way is maybe fine, but IMO default should be the way we want to eventually go.

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -60,6 +60,30 @@ spec:
...
+                      More info: https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/3659-kubectl-apply-prune
+
+                      If not set, default reconciliation behavior is defined by the controller.
+                    enum:
+                    - ApplySet
+                    - ClientSideDelta
+                    type: string
+                    x-kubernetes-validations:
+                    - message: mode is immutable
+                      rule: self == oldSelf
```
matthchr, line 85: Does this immutability actually deliver on what we "want"? They have to delete + recreate the RGD anyway so it's not that easy for them to move forward (or move back, if we change the default to be ApplySets which I think we should)

api/v1alpha1/resourcegraphdefinition_types.go
matthchr: The thing I don't love about this is that at least for applysets, we don't really have a path to making ClientSide work correctly anyway, do we?

It's always going to fail on resources that are mutated by mutating webhooks and/or resources like Secrets/ConfigMaps that don't look the same as they're created (fields move around).

jakobmoellerdev:
/close

Spoken to a few maintainers and we have received overwhelming feedback to forward with SSA only

PR #714, opened 2025-10-08
 feat: Support custom type recursive reference

AlessioDiama:
With this PR we add support for recursive references in custom type definitions, i.e. a custom type can reference another custom type. Cyclic references are detected and not allowed.
Tests are extended to cover the new functionality and an example is provided


examples/kubernetes/recursive-custom-types/instance.yaml
```diff
@@ -0,0 +1,51 @@
+---
```
jakobmoellerdev: I believe this warrants an E2E / Chainsaw test

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+	}
+
+	// Build dependencies and construct the schema
+	for k, v := range obj {
+		dependencies := extractDependenciesFromMap(v)
+
+		// Add dependencies to the DAG and check for cycles
+		err := dagInstance.AddDependencies(k, dependencies)
+		if err != nil {
+			if cycleErr, isCycle := err.(*dag.CycleError[string]); isCycle {
```
jakobmoellerdev: couldnt we change this to errors.Is

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+			return err
+		}
+	}
+
+	// Perform a topological sort of the DAG to get the order of the types
+	// to be processed
+	orderedVertexes, err := dagInstance.TopologicalSort()
 	if err != nil {
-		return fmt.Errorf("failed to build pre-defined types schema: %w", err)
+		return fmt.Errorf("failed to sort DAG: %w", err)
```
jakobmoellerdev: it will be unaparrent to users which DAG youre talking about here

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
 
-	for k, properties := range jsonSchemaProps.Properties {
-		required := false
-		if slices.Contains(jsonSchemaProps.Required, k) {
-			required = true
+	// Build the pre-defined types from the sorted DAG
+	for _, vertex := range orderedVertexes {
+		objValueAtKey, ok := obj[vertex]
+		if !ok {
+			return fmt.Errorf("failed to assert type for vertex %s", vertex)
```
jakobmoellerdev: this is not the correct message and also it can never happen. you never look for a key that you didnt add, so this is a noop.

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+		}
+		objMap := map[string]interface{}{
+			vertex: objValueAtKey,
+		}
+		if schemaProps, err := t.buildOpenAPISchema(objMap); err == nil {
```
jakobmoellerdev: why dont you simply do a normal early return? no need for an if / else check because on error you return from the function. I think this was the behavior befor as well

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+	}
+
+	if rootMap, ok := obj.(map[string]interface{}); ok {
+		parseMap(rootMap)
+	}
+
+	// Remove duplicates using a set
+	dependencySet := make(map[string]struct{})
+	for _, dep := range dependencies {
+		dependencySet[dep] = struct{}{}
```
jakobmoellerdev: why did you do this after going through parseMap? wouldnt it be perfectly fine to collect that from the start instead of starting with an array?

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+
+func extractDependenciesFromMap(obj interface{}) []string {
+	var dependencies []string
+
+	// Use a recursive helper function to traverse the map and extract dependencies
+	var parseMap func(map[string]interface{})
+	parseMap = func(m map[string]interface{}) {
+		for _, value := range m {
+			switch v := value.(type) {
+			case string:
```
jakobmoellerdev: Because you now have semantic meaning attached to types, I believe the types that are allowed need much stricter naming restrictions to avoid issues in the future. Things I am thinking about:

- Disallow Special Characters (for example "|" would be a problem, or starting with "[" and "]")
- Optionally enforce UpperCamelCase
- Maybe some other things that im not aware of that could actually break dependency extraction What happens with "[][]Type" for example

In general, maybe think what can break the parser you built here and try to make it a bit more robust. I think those restrictions should be added onto the CRD itself as part of input validation, and not here in the parser.

barney-s:
I think its time we split off the simple schema into its own top level package with tests. In KRO we should just include new versions of the simple schema. 

We should be able to be faster evolving simple-schema if we decouple KRO usage from the schema. 

WDYT @jakobmoellerdev @a-hilaly @matthchr

jakobmoellerdev:
I understand that thought but would like separate that out into its own issue. Dont think we have to discuss it here on this PR, as this should be tackled separately.

cmd/controller/__debug_bin4030145734
fabianburth, line 1: Accidental check in, I assume?

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+// Types are loaded one by one so that each type can reference one of the
+// other custom types.
+// Cyclic dependencies are detected and reported as errors.
 // As of today, kro doesn't support custom types in the schema - do
 // not use this function.
 func (t *transformer) loadPreDefinedTypes(obj map[string]interface{}) error {
+
+	//Constructs a dag of the dependencies between the types
+	//If there is a cycle in the graph, then there is a cyclic dependency between the types
+	//and we cannot load the types
```
fabianburth, line 72: nit: 
```suggestion
	// Constructs a dag of the dependencies between the types
	// If there is a cycle in the graph, then there is a cyclic dependency between the types
	// and we cannot load the types
```

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+	"float":   {},
+	"object":  {},
+}
+
+func extractDependenciesFromMap(obj interface{}) []string {
+	var dependencies []string
+
+	// Use a recursive helper function to traverse the map and extract dependencies
+	var parseMap func(map[string]interface{})
+	parseMap = func(m map[string]interface{}) {
```
fabianburth: This anonymous function calling itself recursively, with the dependencies closure, is kind of wild! :) 
I think we should make this a named unexported function.

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
+	dependencySet := make(map[string]struct{})
+	for _, dep := range dependencies {
+		dependencySet[dep] = struct{}{}
+	}
+
+	// Convert the set back to a slice
+	result := make([]string, 0, len(dependencySet))
+	for dep := range dependencySet {
+		result = append(result, dep)
+	}
```
fabianburth: nit:

```suggestion
  result := slices.Collect(maps.Keys(dependencySet))
```
Could also use standard lib functionality here.

pkg/simpleschema/transform.go
a-hilaly: We could reuse some of the functions already defined in https://github.com/kubernetes-sigs/kro/blob/main/pkg/simpleschema/atomic.go#L49-L138

pkg/simpleschema/transform.go
```diff
@@ -57,27 +58,130 @@ func newTransformer() *transformer {
...
 
+// Define the set of basic types as
+// defined in https://kro.run/docs/concepts/simple-schema#basic-types
+var excludedTypes = map[string]struct{}{
+	"string":  {},
+	"integer": {},
+	"bool":    {},
+	"float":   {},
+	"object":  {},
+}
```
a-hilaly: I like this idea, let's address this in a seperate PR?

examples/kubernetes/recursive-custom-types/rg.yaml
```diff
@@ -0,0 +1,42 @@
...
+    types:
+      Person:
+        name: string
+        role: "Role"
+      Division:
+        name: string
+        employees: "[]Person"
+      Role:
+        name: string
+        seniority: string
```
tjamet, line 22: Will `types` get inside the `CompanyConfig` CRD? or is it a construct that stays private within the RGD?

From birds view, it would be interesting to keep it as a construct and exclude it from the `CompanyConfig`  CRD. I would hence advocate for moving it under spec

```suggestion
  types:
    Person:
      name: string
      role: "Role"
    Division:
      name: string
      employees: "[]Person"
    Role:
      name: string
      seniority: string
  schema:
```

pkg/simpleschema/transform.go
tjamet: As we already have a dependency on k8s apimachinery it could be interesting to re-use the existing `Set[comparable]` that could help code readability

```
depedencies := sets.Set[string]{}

[... in the loop]
     dependencies.Insert(depName)
[...]

return dependencies.UnsortedList()
```

a-hilaly:
ping @AlessioDiama - are still iterating on this?

AlessioDiama:
hi @a-hilaly, i am a little bit busy, will try to continue as soon as possible.

pkg/simpleschema/transform.go
AlessioDiama: not sure what output as a user-understandable error here. Something like:

`return fmt.Errorf("failed to get topological order for the types: %w", err)`

WDYT?

pkg/simpleschema/transform.go
AlessioDiama: thanks, removed.

pkg/simpleschema/transform.go
AlessioDiama: thanks for the comment. Implemented as suggested

pkg/simpleschema/transform.go
AlessioDiama: thanks for the hint! Changed to use sets.Set[string]{} as suggested

pkg/simpleschema/transform.go
AlessioDiama: i've pushed a new implementation for the `extractDependenciesFromMap` func. Now i am using the functions from https://github.com/kubernetes-sigs/kro/blob/main/pkg/simpleschema/atomic.go#L49-L138 to make sure to type parsing is more effective.

I believe it is still not 100% safe regarding to all the cases described by @jakobmoellerdev. I would need as well some indications on how to enforce restrictions "onto the CRD itself as part of input validation" as suggested above

a-hilaly:
@AlessioDiama can you please sign the CLA?

AlessioDiama:
> @AlessioDiama can you please sign the CLA?

it is now done.

jakobmoellerdev (commented):
For me this needs 2 more things before we can lgtm: I think it would be best if you can include a reference documentation on usage in the website.
I also still think an integration and/or e2e test is useful.
Once we have this, I think this is good for merge!

AlessioDiama:
> For me this needs 2 more things before we can lgtm: I think it would be best if you can include a reference documentation on usage in the website. I also still think an integration and/or e2e test is useful. Once we have this, I think this is good for merge!

How can i propose how to extend the documentation? I couldn't find any folder in the repo with the doc.
For the tests, i will add asap some integration/e2e.

jakobmoellerdev:
> > For me this needs 2 more things before we can lgtm: I think it would be best if you can include a reference documentation on usage in the website. I also still think an integration and/or e2e test is useful. Once we have this, I think this is good for merge!
> 
> How can i propose how to extend the documentation? I couldn't find any folder in the repo with the doc. For the tests, i will add asap some integration/e2e.

https://github.com/kubernetes-sigs/kro/tree/main/website/docs/docs

jakobmoellerdev:
@AlessioDiama gentle ping

jakobmoellerdev:
/close

in favor of https://github.com/kubernetes-sigs/kro/pull/950

Issue #720, opened 2025-10-10
Separate out simple schema into its own package


barney-s:
### Feature Description

**Problem Statement**:
Separate out simple schema into its own package.
Allows evolving simple schema into its own thing with tooling and cli support.
KRO would choose the simple schema version it is pinned to (default latest).

**Proposed Solution**:

Create a top level folder for simple-schema and make it a standalone go package.

**Alternatives Considered**:

Separate repo for simple-schema. 
We may still do this in future.

PR #721, opened 2025-10-11
feat: Move simpleschema to its own module

barney-s:
This PR moves the simple-schema package to its own Go module at the top-level of the repository. This is done to make the package more reusable and independent of the main kro module.

The following changes were made:
- Moved the simple-schema package to the top-level of the repository.
- Created a new go.mod file for the simple-schema module.
- Updated the import paths in the main module to use the new module.
- Updated the main go.mod file to use a replace directive for the new module.
- Moved the crd utils into simple-schema
- Add a new cli for generating CRD from simple schema file.

Fixes: #720


jakobmoellerdev:
I understand that you want to separate out development of simpleschema from kro mainline, but my question is if this really solves the problem of independance? if we think that development on KRO is too slow, we should fix that. 

I think if we introduce a separate module that wont change our review practices so im afraid it wont achieve the desired goal. Also happy to hear other thoughts though and maybe I am missing things.

On a different note: We will have to start running separate releases for this and document it as well, bit unsure if we want to maintain overhead

On a final note, we should not merge this without a two step approach because of the `replace` statement. Currently the statement would prohibit people from calling "go get" and one wouldnt be able to use the "pkg" packages as imports anymore.

simpleschema/cmd/main.go
a-hilaly, line 1: the CLI addition was unexpected and I'm not clear on why we need one now.

a-hilaly (commented):
Thank barni - i do have few thoughts:

1/ I think `simpleschema` is fundamentally about OpenAPI specs rather than CRDs. That's why I do not agree moving graph/crd into the simpleschema package. 

2/ the new `simpleschema/go.mod` pulls in heavy Kubernetes dependencies (`k8s.io/apiextensions-apiserver` and `k8s.io/apimachinery`). If the goal is to make it "more reusable and independent" adding these dependencies works against that.

3/ If we truly want `simpleschema` to be independnt, it probably belongs in its own repository (like `github.com/opensimpleschema`) but its probably too early to commit to that kind of distributed development workflow.

simpleschema/cmd/main.go
tjamet: > the CLI addition was unexpected and I'm not clear on why we need one now.

I also wonder the benefit of going through this change. It also introduces brings complexities when it comes to bumping the `simpleschema` version (development could go through workspaces but then we need either automation or discipline to make PRs as go needs at least a sha to resolve a dependency)

jakobmoellerdev:
/close

I will close this one out until we have either a KREP detailing the next steps for simpleschema. I dont think we should progress here until we are a bit more clear on where simpleschema should go.

PR #722, opened 2025-10-12
fix website type example typo

fabianburth:
Fixes `mytype` typo in the website example for simple schema.

### DIFF
--- a/website/docs/docs/concepts/10-simple-schema.md
+++ b/website/docs/docs/concepts/10-simple-schema.md
@@ -36,7 +36,7 @@ spec:
       ports: "[]integer"
 
       # Map type
-      env: "map[string]mytype"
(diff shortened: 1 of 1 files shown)


fabianburth:
While reading this example, I also noticed that the status fields might be misleading, as the whole `resources` property with `deployment` and `service` that are referenced by those fields are missing. While this is convenient for brevity, we might still want to think about adding something like:

```
# resources references by the status fields
resources:
  - id: deployment
    template:
      ...
  - id: service
    template:
      ...
```

WDYT?

a-hilaly:
@fabianburth i general tend to prefer brievity, but i see why we could use something like you suggested. Feel free to ship a seperate PR for that.

/lgtm

PR #723, opened 2025-10-13
Remove out-dated governance page from docs site

jlbutler:
We had a leftover governance page from pre-migration to kubernetes-sigs, this PR removes it. Also, cleaning up an old copy of the overview markdown.


jlbutler:
> could you please also remove it from previous website versions (versioned_docs), since all versions were donated to the SIG? thanks!

done, good call thanks!

jlbutler:
I've also removed the in-repo governance doc as it's no longer needed.

PR #730, opened 2025-10-15
chore: create scripts to build and push images

justinsb:
On every commit, we will run the cloudbuild job to build and push the
images to our staging image registry.

We use a script to avoid locking behaviour in cloudbuild.

Also remove some obsolete Makefile targets.

### DIFF
--- a/dev/tasks/push-images
+++ b/dev/tasks/push-images
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "$REPO_ROOT"
+
+if [[ -z ${IMAGE_PREFIX:-} ]]; then
+  echo "IMAGE_PREFIX is not set"
+  exit 1
+fi
+
+if [[ -z ${IMAGE_TAG:-} ]]; then
+  # Use a tag if the current commit is a tag, otherwise use a date+git-hash tag
+  if git describe --exact-match --tags HEAD >/dev/null 2>&1; then
+    IMAGE_TAG=$(git describe --exact-match --tags HEAD)
+  else
+    IMAGE_TAG="$(date +v%Y%m%d)-$(git rev-parse --short HEAD)"
+  fi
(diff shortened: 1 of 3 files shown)


a-hilaly:
@justinsb @barney-s were you able to test this and validate that it's working?

justinsb:
> this is missing the build /push of the chart I believe, whats your plan here?

I was going to do that next

justinsb:
Helm chart is in #731

justinsb:
>  @justinsb @barney-s were you able to test this and validate that it's working?

I can run it locally and it works.  It's par for the course that it won't work first time when hooked up to prow, but it's only pushing to a staging repo anyway

a-hilaly (approved):
/approve

delegating lgtm to @jakobmoellerdev for a final look

dev/tasks/push-images
```diff
@@ -0,0 +1,27 @@
...
+  # Use a tag if the current commit is a tag, otherwise use a date+git-hash tag
+  if git describe --exact-match --tags HEAD >/dev/null 2>&1; then
+    IMAGE_TAG=$(git describe --exact-match --tags HEAD)
+  else
+    IMAGE_TAG="$(date +v%Y%m%d)-$(git rev-parse --short HEAD)"
+  fi
+fi
+echo "Using IMAGE_TAG=${IMAGE_TAG}"
+
+# build first, technically this is not needed as publish does a build too
```
jakobmoellerdev, line 23: if we publish and build, why do you do it anyways? :D mostly confused by the comment

justinsb: `make publish-image` does everything that `make build-image` does, and then pushes the image.  Because it's `ko` all the work that `make build-image` does should be cached, so the duplication is not expensive.  And in return, we're less likely to end up "half pushing" something.  We can also tweak the script in future to e.g. upload the image to kind / skip the upload.

justinsb:
Looks like I forgot to configure the base dir: https://github.com/kubernetes/test-infra/pull/35731

PR #731, opened 2025-10-16
chore: push helm chart alongside image

justinsb:
- **chore: create scripts to build and push images**
  On every commit, we will run the cloudbuild job to build and push the
  images to our staging image registry.
  
  We use a script to avoid locking behaviour in cloudbuild.
  
  Also remove some obsolete Makefile targets.
  

- **chore: push helm chart alongside image**
  If we push the image to <prefix>/kro,
  we will push the helm chart to <prefix>/charts/kro
  
  This will likely mean that our primary locations will be:
  
  * image: registry.k8s.io/kro/kro
  * chart: registry.k8s.io/kro/charts/kro

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,9 +214,9 @@ publish-image: ko ## Publish the kro controller images to ghcr.io
 .PHONY: package-helm
 package-helm: ## Package Helm chart
 	cp ./config/crd/bases/* helm/crds/
-	@sed -i '' 's/tag: .*/tag: "$(RELEASE_VERSION)"/' helm/values.yaml
-	@sed -i '' 's/version: .*/version: $(RELEASE_VERSION)/' helm/Chart.yaml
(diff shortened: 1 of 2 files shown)


justinsb:
First commit is #730, we should probably merge that first

a-hilaly:
@justinsb @jakobmoellerdev how about making it:
```
    image: registry.k8s.io/kro/container/kro
    chart: registry.k8s.io/kro/kro
```

jakobmoellerdev:
> @justinsb @jakobmoellerdev how about making it:
> 
> ```
>     image: registry.k8s.io/kro/container/kro
>     chart: registry.k8s.io/kro/kro
> ```

not sure about this. a container is an instance of an image so it is not really accurate. I think kro/kro and kro/charts/kro is probably the least problematic one and works for me

Issue #729, opened 2025-10-15
Install using kubectl and raw manifests


barney-s:
### Documentation Issue

**Location**:
https://kro.run/docs/getting-started/Installation

**Current State**:
Install using helm 

**Suggested Changes**:
Release scripts to push manifests somewhere 
And docs point to usage. 

**Additional Context**:

chamodshehanka:
I like to work on this

tjamet:
@chamodshehanka happy to pair
did you have the chance to take a look at #732 ?

chamodshehanka:
> [@chamodshehanka](https://github.com/chamodshehanka) happy to pair did you have the chance to take a look at [#732](https://github.com/kubernetes-sigs/kro/pull/732) ?

sure I'll check that out

PR #732, opened 2025-10-16
Create ready-to-install kustomizations

tjamet:
As described in #729, there is a desire to install kro with a plain kubectl apply (or kustomize)

This commit is the first step to achieve this goal.

It enables installing kro with and without prometheus metrics directly with kubectl.

The simplest way to achieve the whole experience is to upload the generated manifests to the github release.

Alternatively, there can be an automation to update the manifests as part of the release process.

Manifests are generated with

```
kubectl kustomize  ./manifests/core-install/
kubectl kustomize  ./manifests/core-install-with-metrics
```

Next step is to publish the manifests automatically on release creation

### DIFF
--- a/website/docs/docs/getting-started/01-Installation.md
+++ b/website/docs/docs/getting-started/01-Installation.md
@@ -9,13 +9,6 @@ cluster using Helm.
 
 ## Prerequisites
 
-Before you begin, ensure you have the following:
-
-1. `Helm` 3.x installed
-2. `kubectl` installed and configured to interact with your Kubernetes cluster
-
-## Installation Steps
-
 :::info[**Alpha Stage**]
 
 kro is currently in alpha stage. While the images are publicly available, please
@@ -23,21 +16,38 @@ note that the software is still under active development and APIs may change.
 
 :::
 
-### Install kro using Helm
+Before you begin, ensure you have the following:
+
+1. `Helm` 3.x installed (for helm based installation)
+2. `kubectl` installed and configured to interact with your Kubernetes cluster
+3. chose the version you want to install or fetch the latest release from GitHub
+   Fetch the latest release version from GitHub
+   ```sh
+   export KRO_VERSION=$(curl -sL \
+       https://api.github.com/repos/kubernetes-sigs/kro/releases/latest | \
+       jq -r '.tag_name | ltrimstr("v")'
+     )
+   ```
+   Validate `KRO_VERSION` populated with a version
+   ```
+   echo $KRO_VERSION
+   ```
+4. For kubectl installation, select the variant you want to install
+   | variant                                    | Prometheus scrape config | Kro permissions |
+   | ------------------------------------------ | ------------------------ | --------------- |
+   | kro-core-install-manifests                 | Not installed            | least privilege |
+   | kro-core-install-manifests-with-prometheus | Installed                | least privilege |
+   | kro-unrestricted-manifests                 | Not installed            | cluster admin   |
+   | kro-unrestricted-manifests-with-prometheus | Installed                | cluster admin   |
+
+## Installation with helm
+
+### Installation Steps
+
+#### Install kro using Helm
 
 Once authenticated, install kro using the Helm chart:
 
-Fetch the latest release version from GitHub
-```sh
-export KRO_VERSION=$(curl -sL \
-    https://api.github.com/repos/kubernetes-sigs/kro/releases/latest | \
-    jq -r '.tag_name | ltrimstr("v")'
-  )
-```
-Validate `KRO_VERSION` populated with a version
-```
-echo $KRO_VERSION
-```
 Install kro using Helm
 ```
 helm install kro oci://registry.k8s.io/kro/charts/kro \
@@ -46,7 +56,7 @@ helm install kro oci://registry.k8s.io/kro/charts/kro \
   --version=${KRO_VERSION}
 ```
 
-## Verifying the Installation
+### Verifying the Installation
 
 After running the installation command, verify that Kro has been installed
 correctly:
@@ -73,7 +83,7 @@ correctly:
     kro-7d98bc6f46-jvjl5        1/1     Running            0           1s 
    ```
 
-## Upgrading kro
+### Upgrading kro
 
 To upgrade to a newer version of kro, use the Helm upgrade command:
 
@@ -97,7 +107,7 @@ documentation.
 
 :::
 
-## Uninstalling kro
+### Uninstalling kro
 
 To uninstall kro, use the following command:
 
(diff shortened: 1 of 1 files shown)


tjamet:
Added a sanity check that deployments are equivalent.
This should ensure that users can seamlessly migrate from helm to kustomize and back without generating conflicts that requires re-creation (in terms of namespaces, or selectors)

jakobmoellerdev:
@tjamet would you mind rebasing so tests trigger correctly? thanks!

tjamet:
just managed to test in a private fork:
image
image

tjamet:
Documentation update:

image


image
image

Makefile
```diff
@@ -89,7 +89,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=\"kro:controller:static\" crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases output:rbac:artifacts:config=config/rbac/aggregation
```
a-hilaly, line 92: Is this a breaking change?

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -108,3 +110,108 @@ helm uninstall kro -n kro
...
+1. `kubectl` installed and configured to interact with your Kubernetes cluster
+
+### Installation Steps
+
+:::info[**Alpha Stage**]
+
+kro is currently in alpha stage. While the images are publicly available, please
+note that the software is still under active development and APIs may change.
+
+:::
```
a-hilaly: I think this is relevant in general, let's move it up? also i feel like we're getting out of this mode soon, now that scale tests are showing good results :)

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -108,3 +110,108 @@ helm uninstall kro -n kro
...
+
+#### Install kro using Kubectl
+
+Once authenticated, install kro:
```
a-hilaly: nit: authenticated to what?

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -108,3 +110,108 @@ helm uninstall kro -n kro
...
+
+Fetch the latest release version from GitHub
+```sh
+export KRO_VERSION=$(curl -sL \
+    https://api.github.com/repos/kubernetes-sigs/kro/releases/latest | \
+    jq -r '.tag_name | ltrimstr("v")'
+  )
+```
```
a-hilaly: nit: feels like a repetition from previous steps.

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -108,3 +110,108 @@ helm uninstall kro -n kro
...
+# The basic installation with a metrics service and ServiceMonitor to be integrated with Prometheus
+# export KRO_VARIANT=kro-core-install-manifests-with-prometheus
+
+# The basic installation where KRO has full cluster administrator privileges
+# export KRO_VARIANT=kro-unrestricted-manifests
+
+# The basic installation where KRO has full cluster administrator privileges
+# with a metrics service and ServiceMonitor to be integrated with Prometheus
+# export KRO_VARIANT=kro-unrestricted-manifests-with-prometheus
+```
```
a-hilaly: nit: i really would like to stay out of the business of supporting multiple installations modes, but i'm happy to do it if there are folks willing to help us support this.

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -108,3 +110,108 @@ helm uninstall kro -n kro
...
+object that was removed in the new version of Kro.
+
+:::
+
+### Uninstalling kro
+
+To uninstall kro, use the following command:
+
+```bash
+kubectl apply https://github.com/kubernetes-sigs/kro/releases/download/$KRO_VERSION/$KRO_VARIANT.yaml
```
a-hilaly: apply?

Makefile
tjamet: no, this currently generates file `config/rbac/role.yaml` that unless I am mistaking is unused (the whole `config` directory contains somehow outdated files)

website/docs/docs/getting-started/01-Installation.md
tjamet: copy-paste left-over, thanks!

website/docs/docs/getting-started/01-Installation.md
tjamet: it is, I was thinking on actually having a tab view like [cluster-api](https://cluster-api.sigs.k8s.io/user/quick-start#install-clusterctl), replacing Linux, macOS, ... by `Helm`, `kubectl` but sounded like yet another big thing

image

website/docs/docs/getting-started/01-Installation.md
tjamet: I can relate with this. It's true that installing with kubectl is convenient, but brings challenges.

The goal of integration tests is to reduce the cognitive load of doing so, as well reducing the amount of undesired mismatches.
But it definitely comes with a higher maintenance cost.

website/docs/docs/getting-started/01-Installation.md
a-hilaly: super cool!

website/docs/docs/getting-started/01-Installation.md
tjamet: went for apply to support updates indeed, not sure we have a better option with kubectl to handle updates

a-hilaly: Mmmm i'm not a big kustomize user, but shouldn't the verb be `delete` instead?

tjamet: oh true! it should be delete!
I thought we were in the install section :D

website/docs/docs/getting-started/01-Installation.md
tjamet: Moved above all

website/docs/docs/getting-started/01-Installation.md
tjamet: Let's keep it for a future iteration and avoid PR scope creep

bschaatsbergen: This would be really nice to have, but I agree let’s follow up in a separate PR.

jakobmoellerdev:
/close

in favor of https://github.com/kubernetes-sigs/kro/pull/820

PR #733, opened 2025-10-16
Remove reference for S3 bucket, not used in this example

petrokashlikov:
Fixing example, to remove reference for not used S3 bucket


a-hilaly (approved):
/lgtm

PR #736, opened 2025-10-20
chore: run helm using go run

justinsb:
This lets us pin the helm version without requiring installation.

We don't use `go.mod` because we want to keep the dependency tree separated,
and we don't (yet) have `tools.mod`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ OCI_REPO ?= ghcr.io/kro-run/kro
 HELM_IMAGE ?= ${OCI_REPO}
 KO_DOCKER_REPO ?= ${OCI_REPO}/kro
 
+HELM ?= go run helm.sh/helm/v3/cmd/helm@v3.19.0
+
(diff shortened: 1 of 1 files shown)


a-hilaly:
Chatted offline, we can do something like if that works
``` bash
.PHONY: ko
ko: $(KO) ## Download ko locally if necessary. If wrong version is installed, it will be removed before downloading.
$(KO): $(LOCALBIN)
	@if test -x $(LOCALBIN)/ko && ! $(LOCALBIN)/ko version | grep -q $(KO_VERSION); then \
		echo "$(LOCALBIN)/ko version is not expected $(KO_VERSION). Removing it before installing."; \
		rm -rf $(LOCALBIN)/ko; \
	fi
	test -s $(LOCALBIN)/ko || GOBIN=$(LOCALBIN) GO111MODULE=on go install github.com/google/ko@$(KO_VERSION)
```

PR #738, opened 2025-10-21
KREP-003: Decorators (Collection Watching)

ellistarn:
KREP-003 introduces Decorators (a.k.a Collection Watching), an extension of [Collections](https://github.com/kubernetes-sigs/kro/pull/679) and [ExternalRefs](https://kro.run/docs/concepts/resource-group-definitions#using-externalref-to-reference-objects-outside-the-resourcegraphdefinition). We extend `externalRef` to support watching a collection of objects, rather than just a single object. 

See the full proposal here: https://github.com/ellistarn/kro/blob/krep/docs/design/proposals/decorators.md

### DIFF
--- a/docs/design/proposals/decorators.md
+++ b/docs/design/proposals/decorators.md
@@ -27,27 +27,39 @@ leverage existing and heterogeneous tooling. Further, all policies must be bundl
 single monolithic RGD. This requires cluster administrators to centralize all application guardrails
 into a single monolithic configuration. Decorators provide a solution to these problems.
 
-## Proposed API and Behavior Changes
+## API and Behavior Changes
 
-1. Introduce `Selector` and `NamespaceSelector` to `ExternalRef`
-2. Make `Metadata` optional and mutually exclusive with `Selector` and `NamespaceSelector`
-3. `Kind` must refer to a `ListKind` when `Metadata` is not defined (e.g. `DeploymentList`)
-4. If `Metadata` is defined, `ExternalRef` refers to a single resource, otherwise it refers to a
-   collection of resources
-5. Both `Selector` and `NamespaceSelector` are optional, and if omitted, all resources are included
+1. Introduce `Selector` to `ExternalRefMetadata`, mutually exclusive with `Name`
+2. If `Name` is set, `ExternalRef` refers to a single resource (scalar)
+3. If `Selector` is set, `ExternalRef` refers to a collection of matching resources
+4. `Kind` uses the standard resource kind (e.g. `Deployment`, not `DeploymentList`)
+5. `NamespaceSelector` is out of scope for this KREP — ordering items across multiple namespaces
+   into a single collection is undefined
 
 ```
+type ExternalRefMetadata struct {
+    // Name is the name of the external resource to reference.
+    // Mutually exclusive with Selector.
+    // +kubebuilder:validation:Optional
+    Name string `json:"name,omitempty"`
+    // Namespace is the namespace of the external resource.
+    // If empty, the instance's namespace will be used.
+    // +kubebuilder:validation:Optional
+    Namespace string `json:"namespace,omitempty"`
+    // Selector is a label selector for collection external references.
+    // When set, all resources matching the selector are included.
+    // Mutually exclusive with Name.
+    // +kubebuilder:validation:Optional
+    Selector *metav1.LabelSelector `json:"selector,omitempty"`
+}
+
 type ExternalRef struct {
     // +kubebuilder:validation:Required
     APIVersion string `json:"apiVersion"`
     // +kubebuilder:validation:Required
     Kind string `json:"kind"`
-    // +kubebuilder:validation:Optional # <---- Mutually exclusive with Selector, NamespaceSelector
+    // +kubebuilder:validation:Required
     Metadata ExternalRefMetadata `json:"metadata"`
-    // +kubebuilder:validation:Optional # <---- Mutually exclusive with Selector, NamespaceSelector
-    NamespaceSelector metav1.LabelSelector
-    // +kubebuilder:validation:Optional # <---- Mutually exclusive with Metadata
-    Selector metav1.LabelSelector
 }
 ```
 
@@ -59,7 +71,7 @@ The power of the decorator pattern is best understood through concrete examples.
 
 As a cluster administrator, I want to configure VPA in recommender mode for existing deployments in
 my cluster to see whether or not widespread rollout of VPA would provide significant cost savings.
-I’m going to start with an opt-in approach at the namespace level, though I plan to reduce scoping
+I'm going to start with an opt-in approach by labeling deployments, though I plan to reduce scoping
 as I gain confidence. Eventually, I plan to flip VPA into auto mode.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/design/proposals/decorators.md
```diff
@@ -0,0 +1,265 @@
...
+6. Rename `ExternalRef` to `Watch` in a future API Version (see discussion below)
+
+```
+type ExternalRef struct {
+    // +kubebuilder:validation:Required
+    APIVersion string `json:"apiVersion"`
+    // +kubebuilder:validation:Required
+    Kind string `json:"kind"`
+    // +kubebuilder:validation:Optional # <---- Mutually exclusive with Selector, NamespaceSelector
+    Metadata ExternalRefMetadata `json:"metadata"`
```
n3wscott, line 62: What if selector had an optional "Namespace" field to continue filtering labeled resources.

n3wscott (commented):
Lgtm, I think you can flatten the API a little.

docs/design/proposals/decorators.md
ellistarn: I wanted the full semantics to do things like opt namespaces in by label. Can you be a hair more concrete about what you're proposing? Structs?

docs/design/proposals/decorators.md
```diff
@@ -0,0 +1,265 @@
...
+
+Directionally, we posit that `singletons` could take the following form:
+
+1. The `schema` field is no longer required in RGDs
+2. An RGD without a `schema` would self-instantiate and create all of its resources as a singleton
+3. An RGD without a `schema` becomes the root node of the DAG, rather than the RGD's instance
+
+## The ExternalRef Naming Problem
+
+The name `ExternalRef` suggests a singular reference, which becomes misleading when referring to
```
jakobmoellerdev: I dont think this postulation is correct. A reference is as the naming suggests, referencing "something". based on the settings of the reference, this may be either an object OR a list of objects.

docs/design/proposals/decorators.md
```diff
@@ -0,0 +1,265 @@
...
+collections of resources. Additionally, the term `external` creates ambiguity about scope - it's
+unclear whether this refers to objects outside the RGD itself or objects outside the cluster
+entirely (as in multi-cluster scenarios). This awkwardness led me to explore alternative names.
+
+The name `watch` is a natural fit to both single resources and collections of resources. Watch is
```
jakobmoellerdev: For me the term watch is not a "natural fit" - because we dont even watch them yet: https://github.com/kubernetes-sigs/kro/pull/611 would introduce this (we have a dedicated issue list for watching child resources, its not just about that PR), but not just for the current externalRefs, but also for other resources. So, watch is a misnomer because we actually would watch them all. Also currently the proposal is to have periodic mode instead of watches so maybe that would clash.

TLDR: to have "watch" be better than "externalRefs" we need to actually start watching them and make that a core principle. 

Currently externalRefs are mostly present to be included into the graph as references, thus the name.

docs/design/proposals/decorators.md
```diff
@@ -0,0 +1,265 @@
...
+While KRO enables cluster administrators to create RGDs that define new APIs to enforce application
+guardrails, it requires application developers to “buy-in” and begin using them. This can be
+challenging in large multi-tenant environments that contain large numbers of application teams that
+leverage existing and heterogeneous tooling. Further, all policies must be bundled together into a
+single monolithic RGD. This requires cluster administrators to centralize all application guardrails
+into a single monolithic configuration. Decorators provide a solution to these problems.
+
+## Proposed API and Behavior Changes
+
+1. Introduce `Selector` and `NamespaceSelector` to `ExternalRef`
```
jakobmoellerdev: when using Selector and NamespaceSelector, there is even more reason to keep it as Reference because even Kubernetes calls it reference

docs/design/proposals/decorators.md
ellistarn: Fair!

docs/design/proposals/decorators.md
```diff
@@ -0,0 +1,225 @@
...
+2. Make `Metadata` optional and mutually exclusive with `Selector` and `NamespaceSelector`
+3. `Kind` must refer to a `ListKind` when `Metadata` is not defined (e.g. `DeploymentList`)
+4. If `Metadata` is defined, `ExternalRef` refers to a single resource, otherwise it refers to a
+   collection of resources
+5. Both `Selector` and `NamespaceSelector` are optional, and if omitted, all resources are included
```
tjamet: I'm unsure I get the difference between `Selector` and `NamespaceSelector`

What would be in `Metadata` that makes it impossible to craft a good `Selector`?

I smell that we would gain simplicity by having a single `Selector` to filter object of `Kind` whether it's a namespace or any other object.

There is another question, whether it is allowed, to watch resources from other namespaces (or cluster-wide for namespaced RGDs).

Maybe we can inspire from [NetworkPolicies](https://pkg.go.dev/k8s.io/api/networking/v1#NetworkPolicyPeer) and define:

`NamespaceSelector` a selector to match namespace labels in which the resource should be
`Selector` a selector to match the resource labels

ellistarn: The example from NetworkPolicies is exactly what I was thinking. The `NamespaceSelector` constrains what Namespaces we look at and the regular `Selector` constrains the individual objects by label.

```
namespaceSelector: # Only include namespaces with the opt-in label
  matchLabels:
    "opt-in-namespace": true
selector:
  matchLabels
    "opt-in-object": true
```

In this example, you would only include objects that are opted in in both their own metadata.labels, as well as the metadata.labels of the namespace. e.g.

```
kind: Namespace
metadata:
  labels:
    "opt-in-namespace": true
  ...
---
kind: Object
metadata:
  labels: 
    "opt-in-object": true

Comment (tjamet) [reply docs/design/proposals/decorators.md]: Why having metadata? for backward compatibility?

File: docs/design/proposals/decorators.md [hunk 6]
```diff
@@ -0,0 +1,225 @@
...
+```
+
+## The Blast Radius Problem
+
+Decorators have broad blast radius. A small change to a decorator has the potential to impact every
+object in the cluster of a given kind. However, this problem is not new to decorators is is
+identical to the blast radius problem inherent to collections. We consider this to be an opportunity
+for a future KREP.
+
+## The Singleton Problem
```
JamesGAWS, line 215: Any thoughts on adding ExternalRefs to the spec.schema field? That keeps the root object used to hydrate the RGD installation in the spec.schema field and avoids "empty" instance objects.

docs/design/proposals/decorators.md
ellistarn: Every kubernetes object has metadata.

docs/design/proposals/decorators.md
ellistarn: I am not following, but very curious. Can you provide an example?

jakobmoellerdev:
@ellistarn would we mind updating your KREP towards the latest actual state that was implemented / accepted? then we can merge this in I think

ellistarn:
Updated the KREP to reflect the implementation in #1079. The key changes:

1. `Selector` is now a field inside `ExternalRefMetadata` (mutually exclusive with `Name`), rather than a top-level field on `ExternalRef`
2. `NamespaceSelector` is dropped and noted as out of scope — ordering items across multiple namespaces into a single collection is undefined
3. `Kind` uses the standard resource kind (e.g. `Deployment`, not `DeploymentList`) since users naturally expect a selector on a kind to yield a collection
4. Updated all examples to use the implemented API shape (`metadata.selector` instead of top-level selectors/ListKinds)

PR #748, opened 2025-10-23
fix(applyset): make sure labels and annotations are applied

jakobmoellerdev:
we previously accidentally used the previous annotation/label map pointer to determine change detection of parent labels so we skipped setting the annotations after first apply. we now use a map copy which makes us use different pointers that will cause the labels to get applied. to not hardcode against reflect I also switched to semantic equality from k8s. I have exposed the ID func from apply set to make it testable

### DIFF
--- a/pkg/applyset/applyset.go
+++ b/pkg/applyset/applyset.go
@@ -26,13 +26,13 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
-	"reflect"
 	"sort"
 	"strings"
 	"sync"
 
 	"github.com/go-logr/logr"
 	"golang.org/x/sync/errgroup"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/applyset/applyset.go
```diff
@@ -374,13 +374,21 @@ func (a *applySet) Add(ctx context.Context, obj ApplyableObject) (*unstructured.
...
 
 // ID is the label value that we are using to identify this applyset.
-// Format: base64(sha256(<name>.<namespace>.<kind>.<apiVersion>)), using the URL safe encoding of RFC4648.
 func (a *applySet) ID() string {
+	return ID(a.parent)
+}
+
+// ID computes an apply set identifier for a given parent object.
+// Format: base64(sha256(<name>.<namespace>.<kind>.<apiVersion>)), using the URL safe encoding of RFC4648.
+func ID(parent interface {
```
a-hilaly, line 383: let's make this one private?

test/integration/suites/core/annotation_label_test.go
```diff
@@ -0,0 +1,161 @@
...
+				},
+			}, nil, nil),
+		)
+
+		Expect(env.Client.Create(ctx, rgd)).To(Succeed())
+		DeferCleanup(func(ctx SpecContext) {
+			Expect(env.Client.Delete(ctx, rgd)).To(Succeed())
+		})
+
+		Eventually(func(g Gomega) {
```
a-hilaly: Looking at other tests looks like all of them use `ctx SpecContext`

a-hilaly: same comment for the rest of the file

pkg/applyset/applyset_test.go
```diff
@@ -587,7 +654,7 @@ func TestApplySet_PruneOldGVKs(t *testing.T) {
 		assert.Contains(t, parentPatch.GetAnnotations(), ApplySetAdditionalNamespacesAnnotation)
 		assert.Contains(t, parentPatch.GetAnnotations(), ApplySetGKsAnnotation)
 		assert.Equal(t, parentPatch.GetAnnotations()[ApplySetAdditionalNamespacesAnnotation], "ns1,ns2")
-		//assert.Equal(t, parentPatch.GetAnnotations()[ApplySetGKsAnnotation], "ConfigMap,Foo.foo.bar")
+		// assert.Equal(t, parentPatch.GetAnnotations()[ApplySetGKsAnnotation], "ConfigMap,Foo.foo.bar")
```
a-hilaly, line 657: i wonder why those were commented out

a-hilaly (approved):
Thank you @jakobmoellerdev  - Implementation looks good to me. Left non blocking comments.

pkg/applyset/applyset.go
jakobmoellerdev: I have a cross usage between `pkg/applyset` and `test/integration/suites/core` to validate correct labels/annotations. is it okay to leave this public so its reproducible externally to have parent annotations?

pkg/applyset/applyset_test.go
jakobmoellerdev: @barney-s do you happen to remember?

test/integration/suites/core/annotation_label_test.go
jakobmoellerdev: good catch, thanks!

a-hilaly (approved):
Dankeshoen!
/lgtm

Issue #745, opened 2025-10-22
ResourceGraphDefinition -> CRD ownership


dudo:
### Feature Description

**Problem Statement**:

CRDs don't show up in ArgoCD

**Proposed Solution**:
Could we add ownerReferences and appropriate annotations to the generated CRDs that link to the ResourceGraphDefinition that created them?

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

sk31337:
Looking forward to this functionality :) (*[slack thread](https://kubernetes.slack.com/archives/C081TMY9D6Y/p1762868693341289)*)

matthchr:
Have you see the [FAQ entry](https://kro.run/docs/faq) on how to use kro with argo?

dudo:
Hello, @matthchr! I have, and it does work for the Kubernetes resources themselves. The CRDs are what I'm referring to - those aren't associated with anything.

a-hilaly:
@dudo we do have a design proposal for it in https://github.com/kubernetes-sigs/kro/pull/763 - your feedback would be valuable :)

PR #749, opened 2025-10-23
feat: update RGD controller to set CRD ownership and optimize watches

jakobmoellerdev:
Updated the ResourceGraphDefinition controller to set correct ownership references on CRDs and switched to metadata-only watches for improved performance and reduced memory consumption. Added integration tests to validate CRD ownership.

This would resolve both cleanup of the CRD with background deletion mode and resolution of owner refs in general but has the side effect that RGDs must stay non-namespaced (I think this is static by now and we can move to ownerrefs safely)

resolves https://github.com/kubernetes-sigs/kro/issues/745

### DIFF
--- a/helm/templates/validating-admission-policy.yaml
+++ b/helm/templates/validating-admission-policy.yaml
@@ -1,0 +1,33 @@
+{{- if not .Values.config.allowCRDDeletion }}
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: {{ include "kro.fullname" . }}-crd-protection-policy
+  labels:
+    {{- include "kro.labels" . | nindent 4 }}
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups: ["kro.run"]
+        apiVersions: ["v1alpha1"]
+        operations: ["DELETE"]
+        resources: ["resourcegraphdefinitions"]
+  validations:
+    - expression: |
+        has(oldObject.metadata.annotations) &&
+        oldObject.metadata.annotations.exists(w, w == '{{ .Values.validation.admission.policy.rgd.protectionKey }}') &&
+        oldObject.metadata.annotations['{{ .Values.validation.admission.policy.rgd.annotation.key }}'] == 'true'
+      reason: Invalid
+      message: |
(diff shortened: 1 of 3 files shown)


a-hilaly:
i'm not very sure we need ownerReferences on CRDs. This is conflicting with the fact that we have a --allow-crd-deletion flag https://github.com/kubernetes-sigs/kro/blob/main/helm/values.yaml#L97-L98.

jakobmoellerdev:
I think we have very static / classic controller ownership relations between RGD<->CRD. Thus I think also deletion behavior can be controlled by an ownerref and allow crd deletion can be deprecated in favor of ownerreferences. just my opinion of course, happy to hear other thoughts.

dudo:
Thanks for picking up my issue, @jakobmoellerdev! I'm obviously biased, but the owner relationship feels logical to me. Why would we want to keep CRDs around after an RGD is deleted?

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -109,50 +106,18 @@ func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) e
...
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return true
 			},
-		},
-	}
+			CreateFunc: func(e event.CreateEvent) bool {
+				return false
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return false
```
tjamet, line 117: I would say we do want to reconcile RGDs when the CRD they manage gets deleted so we re-create them

jakobmoellerdev: I would agree, will confirm the general impl in todays comm call

jakobmoellerdev:
Comm call agreement was that we need to ensure CRD protection still so setup a validation policy. Gonna create a KREP to support this PR by request

tjamet:
> Comm call agreement was that we need to ensure CRD protection still so setup a validation policy. Gonna create a KREP to support this PR by request

I would have managed CRDs to be deleted when the source RGD is deleted, by default.
But, indeed, ownerreferences should not be set (and removed) if `--allow-crd-deletion=false`

Issue #743, opened 2025-10-22
Regression: RGD instance is marked as ACTIVE before the managed resource reaches a ready state


devdanylo:
### Description

**Observed Behavior**:
The RGD instance is set to `ACTIVE` before the underlying resource meets its `readyWhen` condition.

**Expected Behavior**:
The RGD instance should stay in the `IN_PROGRESS` state until the underlying resource satisfies its `readyWhen` condition.

**Reproduction Steps**:

RGD:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: test-jobs
spec:
  schema:
    apiVersion: v1
    kind: TestJobs
  resources:
    - id: job1
      readyWhen:
        - ${job1.status.completionTime != null}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: job-1
        spec:
          template:
            spec:
              containers:
                - name: sleeper
                  image: busybox
                  command: ["sh", "-c", "echo 'Sleeping for 30 seconds' && sleep 30 && echo 'Hello from the Kubernetes Job #1'"]
              restartPolicy: Never
```

RGD instance:
```yaml
apiVersion: kro.run/v1
kind: TestJobs
metadata:
  name: test-jobs
  namespace: kro-tests
```

**Versions**:
- kro version: 0.5.0
- Kubernetes Version (`kubectl version`): 
```
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.34.0
```

**Logs**:
```
2025-10-22T16:45:30.580Z    DEBUG    dynamic-controller    Enqueueing object    {"objectIdentifiers": {"NamespacedKey":"kro-tests/test-jobs","GVR":{"Group":"kro.run","Version":"v1","Resource":"testjobs"}}, "eventType": "add"}
2025-10-22T16:45:30.580Z    DEBUG    dynamic-controller    Syncing object    {"gvr": "kro.run/v1/testjobs", "namespacedKey": "kro-tests/test-jobs"}
2025-10-22T16:45:30.581Z    DEBUG    testjobs-controller    Setting managed state    {"controller": "testjobs", "controllerGroup": "kro.run", "controllerKind": "TestJobs", "namespace": "kro-tests", "name": "test-jobs", "name": "test-jobs", "namespace": "kro-tests"}
[...]

PR #750, opened 2025-10-23
fix: replace error state check with error var check

jakobmoellerdev:
makes sure that for readiness errors that have state not on ready and error != nil we still fail out on the instance and stay on IN_PROGRESS

resolves https://github.com/kubernetes-sigs/kro/issues/743

still needs tests

/cc @a-hilaly

### DIFF
--- a/test/integration/suites/core/readiness_test.go
+++ b/test/integration/suites/core/readiness_test.go
@@ -21,6 +21,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


a-hilaly:
nice catch!

a-hilaly:
I wonder how didn't we catch the regression with https://github.com/kubernetes-sigs/kro/blob/main/test/integration/suites/core/readiness_test.go cc @michaelhtm

pkg/controller/instance/instance_state.go
matthchr, line 20: Can we add some comments to these about what each state actually means?

matthchr: (talking with @barney-s and @a-hilaly and we're wondering if we even want `InstanceStateError` at all or if just get rid of `InstanceStateError`?)

matthchr: let's maybe do the comments but we can do it in a different PR since this is fixing a pretty important bug.

a-hilaly:
@jakobmoellerdev I just pushed some tests, also opened https://github.com/kubernetes-sigs/kro/pull/754 to see them fail

PR #754, opened 2025-10-25
test: verify instance readiness waits for tail node completion

a-hilaly:
Add integration test to ensure instance state remains IN_PROGRESS
while waiting for the last node in the DAG to satisfy its readiness
gates, then transitions to ACTIVE once the readiness condition is met.


a-hilaly:
this is expected to fail
/hold

PR #756, opened 2025-10-27
Overhaul RGD static validation to eliminate emulation

a-hilaly:
**Note to reviewers:** This PR was crafted commit by commit to make the
refactoring easier to follow. Review each commit individually to
understand how the implementation evolved from emulation to native CEL
type checking.

When we built kro, we needed to validate that RGDs are correct and catch
as many errors as possible before users saw them when submitting
instances. Things like ensuring a string field doesnt get assigned an
integer expression, or that field paths actually exist on the resources
being referenced. We implemented an emulation based approach: for every
`RGD` at build time, generate fake data for every resource matching their
`OpenAPI` schemas, then evaluate every `CEL` expression against this fake
data. We needed this for three purposes: 1/ validate that resources
reference each other correctly and those links make sense. 2/ ensure
expressions resolved to something and validate that the result matched the
expected type. 3/ generate the instance status schema fields from the
expression output types. This was essentially a dry run of expression
evaluation at build time.

The emulation approach had fundamental limitations. Many schema
constraints were too complex to handle. Things like optional types, regex
patterns, and complex validation rules were either ignored or simplified.
The emulator couldn't generate valid data for all cases. More
fundamentally, we were executing expressions to observe their output types
rather than analyzing the expressions statically. Type coercion rules,
null handling, optional type behavior, and field access semantics became
approximations. Edge cases like `optional_type(T)` assignability to `T`,
dynamic type handling, and wrapper type conversions required custom logic
that diverged too far. Generating fake instances for every resource
consumed substantial memory, particularly for deeply nested schemas,
and all this data was discarded immediately after validation.

This series of commits replaces emulation with CELs built in type
checker. We leverage CEL's type checking and inference capabilities that
provide best-effort static analysis without needing to emulate data.
OpenAPI schemas are converted to CEL types and injected into the type
environment. CEL's compiler validates expressions during AST analysis,
checking field access and type compatibility against the actual schemas.
The type checker understands resource structures, validates field paths
exist, and ensures expressions return the expected types, all statically
without executing anything.

Type compatibility is checked using CEL's assignability rules, which
handle the nuances we previously approximated: optional types, null
handling, type coercion, and wrapper types all follow CEL's actual
semantics. Status schema generation now uses CEL's type inference directly
from checked ASTs. The compiler tells us what type an expression returns,
and we convert that CEL type back to an OpenAPI schema by walking through
the type structure. The same approach validates `readyWhen` and `includeWhen`
expressions, ensuring they return boolean types without executing anything.
No fake data generation, no expression execution, no emulator. We get
better correctness by using CEL's actual semantics and better performance
by eliminating all the overhead.


test/integration/suites/core/crd_test.go
```diff
@@ -75,7 +75,7 @@ var _ = Describe("CRD", func() {
 					},
 					"data": map[string]interface{}{
 						"key":  "value",
-						"key2": "${schema.spec.field2}",
+						"key2": "${string(schema.spec.field2)}",
```
ellistarn, line 78: Cool to see the type system actually catching type errors. I assume this was a bug in the emulator that allowed this test to succeed?

pkg/graph/schema/conversion_cel_type_test.go
```diff
@@ -0,0 +1,290 @@
...
+		},
+		{
+			name:    "convert uint",
+			celType: cel.UintType,
+			want:    &extv1.JSONSchemaProps{Type: "integer"},
+		},
+		{
+			name:    "convert double",
+			celType: cel.DoubleType,
+			want:    &extv1.JSONSchemaProps{Type: "number"},
```
ellistarn, line 54: Curious if you've run into trouble with floating point precision causing flip flops in graph evaluation.

pkg/graph/parser/parser.go
```diff
@@ -68,6 +70,24 @@ func parseResource(resource interface{}, schema *spec.Schema, path string) ([]va
 	}
 }
 
+// getCelType converts an OpenAPI schema to a CEL type using the Kubernetes OpenAPI library.
+// This replaces manual type conversion with the library's schema-to-CEL type conversion.
+func getCelType(schema *spec.Schema) *cel.Type {
+	if schema == nil {
+		return cel.DynType
+	}
```
ellistarn: is this just defense programming, or can schema actually be nil?

pkg/graph/schema/conversion_schema.go
```diff
@@ -162,3 +165,32 @@ func ConvertJSONSchemaPropsToSpecSchema(props *extv1.JSONSchemaProps) (*spec.Sch
 
 	return schema, nil
 }
+
+// GetObjectMetaSchema extracts the ObjectMeta schema from Kubernetes OpenAPI definitions.
+// This returns the fully resolved ObjectMeta schema including all nested types like
+// OwnerReference, ManagedFieldsEntry, Time, etc.
+func GetObjectMetaSchema() (*spec.Schema, error) {
```
ellistarn: Can you do this once at startup?

pkg/graph/builder.go
```diff
@@ -714,155 +674,192 @@ func extractDependencies(env *cel.Env, expression string, resourceNames []string
...
-	// include all resources, and remove individual ones
-	// during the validation
-	// this is done to avoid having to create a new context for each resource
-	for resourceName, contextResource := range resources {
-		expressionContext[resourceName] = contextResource
+// validateNode validates all CEL expressions for a single resource node:
+// - Template expressions (resource field values)
+// - includeWhen expressions (conditional resource creation)
+// - readyWhen expressions (resource readiness conditions)
+func validateNode(resource *Resource, templatesEnv, schemaEnv *cel.Env, resourceSchema *spec.Schema) error {
```
ellistarn, line 681: Completely optional but something I noticed. 

Each of these validations are derivations of the data `resource` and `schema`. In the function outside `validateNode`, we scope down `templatesEnv` and `schemaEnv` and then inside this function, we scope down the `resourceEnv` for the readyWhen validation. 

It made me wonder two things:
1. If this pattern/separation was intentional (i.e., we create a schemaEnv outside, even if includeWhenExpressions == 0, but we check readyWhenExpressions exist before constructing the limited env)
2. If we ever might want to expand the language to support cross-resource readyWhens or status-based includeWhens. The latter is maybe more likely?

test/integration/suites/core/crd_test.go
a-hilaly: correct :)

pkg/graph/schema/conversion_cel_type_test.go
a-hilaly: Haven't seen it in practice yet, but yeah, could definitely cause issues. Any float comparisons expressions in resource templates, readyWhen and includeWhen could cause issues... worth documenting this to steering users towards using more of string/bool/int expression rather than floats, but it's inevitable i guess.

a-hilaly: but yeah i can imagine cases where nodes are oscillating between ready/non ready state or included/excluded caused floating points precision issues.

pkg/graph/schema/conversion_schema.go
a-hilaly: Agreed, fixed in [d111e9d](https://github.com/kubernetes-sigs/kro/pull/756/commits/d111e9d34e5954137356b1608cb64b89ae55250a)

pkg/graph/builder.go
a-hilaly: 1. Yeah, intentional. wanted to start restrictive and relax later once we've solved the core problems. cross referencing resources in `includeWhen`/`readyWhen` adds complexity - theoritically they introduce edges to the DAG, but when do we want to consider them circular dependencies? I'm thinking of example where resource B depends on A's status, but A `readyWhen` checks if B reached a specific state - that looks circular but might still be a valid DAG in practice. In short there are cases where dependencies appear circular but are actually resolvable at runtime.

2. Yep, definitely something we'll want eventually (short term). Status based includeWhens especially feels like a straightforward one. the new type system makes raesoning about these features easier though, not just for this but for collections too.

cmd/kro/go.mod
```diff
@@ -0,0 +1,79 @@
+module github.com/kro-run/kro/cmd/kro
+
+go 1.24.0
+
+require (
+	github.com/go-echarts/go-echarts/v2 v2.6.5
+	github.com/kubernetes-sigs/kro v0.5.0
```
tjamet, line 7: How do we plan to handle version bumps?
in other words, how do we make sure that the CLI uses the same version, in particular for development and `latest` builds.

In my experience, you need to have the version pushed to be able to do so, which is a bit annoying when developing (for example updating the CRD definition or the emulator itself)

a-hilaly: good callout. honnestly I haven't thought about that yet, but we'll definitely need some automation there. For local development we can use go workspaces to keep things clean.

It's an open problem for now - we'll need to solve it before we start deistributing the kro binary official. Open to ideas if you have suggestions on how other projects handle this.

pkg/cel/environment.go
```diff
@@ -66,9 +94,64 @@ func DefaultEnvironment(options ...EnvOption) (*cel.Env, error) {
...
+			if declType != nil {
+				declType = declType.MaybeAssignTypeName(name)
+				declTypes = append(declTypes, declType)
+			}
+		}
+
+		if len(declTypes) > 0 {
+			provider := apiservercel.NewDeclTypeProvider(declTypes...)
+
+			registry, err := types.NewRegistry()
```
tjamet: I'm unsure I get how the registry works and what capability it provides.

It's clearly a CEL library implementation detail but it could be worth explaining why it's required and what it brings

pkg/graph/builder.go
```diff
@@ -563,17 +564,74 @@ func buildStatusSchema(
...
+				}
+
+				outputType := checkedAST.OutputType()
+				// multiple expressions per field - all must be strings for now.
+				if err := validateExpressionType(outputType, []string{"string"}, expression, "status", fieldDescriptor.Path); err != nil {
+					return nil, nil, err
+				}
+			}
+			// All expressions are strings - result type is string
+			statusTypeMap[fieldDescriptor.Path] = cel.StringType
```
tjamet, line 633: This is when we have a status like 

```yaml
status:
    someField: ${ expressionA }${ expressionB }
```

correct?

Could we find/test any case where `type(expressionA)type(expressionB)` is not a string and returns something valid at the CEL level? 

2 side notes/nit (just thinking out loud)

I wonder if we wouldn't gain readability flexibility building directly the `extva.JSONSchemaProps` . Eventually, we could also declare `statusTypeMap := make(map[string][]*cel.Type)` so we can handle options like ``

pkg/graph/schema/conversion_cel_type_test.go
ellistarn: Thoughts on just blocking floats?

cmd/kro/commands/generate/instance.go
```diff
@@ -57,8 +59,21 @@ func generateInstance(rgd *v1alpha1.ResourceGraphDefinition) error {
 	if err != nil {
 		return fmt.Errorf("failed to create resource graph definition: %w", err)
 	}
+	em := emulator.NewEmulator()
+
+	crd := rgdGraph.Instance.GetCRD()
+	gvk := rgdGraph.Instance.GetCRD().GroupVersionKind()
+
+	s, err := schema.ConvertJSONSchemaPropsToSpecSchema(crd.Spec.Versions[0].Schema.OpenAPIV3Schema)
```
jakobmoellerdev, line 67: I think this schema conversion has to happen exactly once for any given crd version so maybe its worth caching it. worth a todo?

pkg/cel/environment.go
```diff
@@ -66,9 +94,86 @@ func DefaultEnvironment(options ...EnvOption) (*cel.Env, error) {
...
+// CEL expressions against OpenAPI schemas.
+func TypedEnvironment(schemas map[string]*spec.Schema) (*cel.Env, error) {
+	return DefaultEnvironment(WithTypedResources(schemas))
+}
+
+// UntypedEnvironment creates a CEL environment without type declarations.
+//
+// This is theoretically cheaper to use as there are no Schema conversions
+// required. NOTE(a-hilaly): maybe use this for runtime? undecided.
+func UntypedEnvironment(resourceIDs []string) (*cel.Env, error) {
```
jakobmoellerdev, line 153: I think we should reuse an existing ENV and then add type declarations on top of it. Schema conversions are not very expensive if they happen within CEL I think.

pkg/graph/builder.go
```diff
@@ -203,39 +192,70 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
...
-	// resources in the resource graph definition a.k.a the dependency graph.
+	// include the instance spec schema in the context as "schema". This will let us
+	// validate expressions such as ${schema.spec.someField}.
 	//
-	// The dependency graph is a directed acyclic graph that represents the
-	// relationships between the resources in the resource graph definition. The graph is
-	// used to determine the order in which the resources should be created in the
-	// cluster.
+	// not that we only include the spec and metadata fields, instance status references
+	// are not allowed in RGDs (yet)
```
jakobmoellerdev, line 209: nit: probably worth a TODO?

pkg/graph/builder.go
jakobmoellerdev: +1 here, the CEL type of status.someField could be checked against the typed schema type for its assigneability. I think there are cases when expression results could be integers or nested objects. otherwise its a missing limitation doc I think

pkg/graph/builder.go
```diff
@@ -714,155 +674,187 @@ func extractDependencies(env *cel.Env, expression string, resourceNames []string
...
-func ensureResourceExpressions(env *cel.Env, context map[string]*Resource, resource *Resource) error {
-	// We need to validate the CEL expressions in the resource.
-	for _, resourceVariable := range resource.variables {
-		for _, expression := range resourceVariable.Expressions {
-			_, err := ensureExpression(env, expression, []string{resource.id}, context)
+// validateTemplateExpressions validates CEL template expressions for a single resource.
+// It type-checks that expressions reference valid fields and return the expected types
+// based on the OpenAPI schemas.
+func validateTemplateExpressions(env *cel.Env, resource *Resource) error {
+	for _, templateVariable := range resource.variables {
```
jakobmoellerdev, line 714: I think this could be concurrent for every single expression

pkg/graph/builder.go
```diff
@@ -714,155 +674,187 @@ func extractDependencies(env *cel.Env, expression string, resourceNames []string
...
-			// ignore apiVersion and kind from readyWhenExpression context
-			delete(resourceEmulatedCopy.Object, "apiVersion")
-			delete(resourceEmulatedCopy.Object, "kind")
-		}
-		context := map[string]*Resource{}
-		context[resource.id] = &Resource{
-			emulatedObject: resourceEmulatedCopy,
-		}
+	// "dyn" is a special case. output type matches anything (x-kubernetes-int-or-string, etc)
+	if outputType.String() == "dyn" {
```
jakobmoellerdev: nit: this probably has a constant defined for the Type that you can check on

jakobmoellerdev: Im not sure about output type matching anything, shouldnt this match based on the annotation?

pkg/graph/builder.go
```diff
@@ -714,155 +674,187 @@ func extractDependencies(env *cel.Env, expression string, resourceNames []string
...
+		"type mismatch in resource %q at path %q: expression %q returns type %q but expected %q",
+		resourceID, path, expression, outputType.String(), expectedType.String(),
+	)
+}
+
+// parseAndCheckCELExpression parses and type-checks a CEL expression.
+// Returns the checked AST on success, or the raw CEL error on failure.
+// Callers should wrap the error with appropriate context.
+func parseAndCheckCELExpression(env *cel.Env, expression string) (*cel.Ast, error) {
+	parsedAST, issues := env.Parse(expression)
```
jakobmoellerdev, line 778: this AST check could be cached later, maybe worth a todo

pkg/graph/builder.go
```diff
@@ -714,155 +674,187 @@ func extractDependencies(env *cel.Env, expression string, resourceNames []string
...
-// ensureExpression validates the CEL expression in the context of the resources
-func ensureExpression(env *cel.Env, expression string, resources []string, context map[string]*Resource) (ref.Val, error) {
-	err := validateCELExpressionContext(env, expression, resources)
-	if err != nil {
-		return nil, fmt.Errorf("failed to validate expression %s: %w", expression, err)
+// getSchemaWithoutStatus returns a schema from the CRD with the status field removed.
+func getSchemaWithoutStatus(crd *extv1.CustomResourceDefinition) (*spec.Schema, error) {
+	crdCopy := crd.DeepCopy()
+
+	if len(crdCopy.Spec.Versions) != 1 || crdCopy.Spec.Versions[0].Schema == nil {
```
jakobmoellerdev, line 839: this should probably get a TODO when we start managing upgrades to CRDs. Whats our stance here?

pkg/graph/parser/parser.go
```diff
@@ -275,36 +297,56 @@ func parseString(field string, schema *spec.Schema, path string, expectedTypes [
...
+func getSchemaTypeName(v interface{}) string {
+	switch v.(type) {
+	case bool:
+		return "boolean"
+	case int, int8, int16, int32, int64:
+		return "integer"
 	default:
-		return nil, fmt.Errorf("unexpected type for path %s: %T, expectedTypes: [%s]", path, field, strings.Join(expectedTypes, ","))
+		// For other types (including float), use the Go type name
+		return fmt.Sprintf("%T", v)
```
jakobmoellerdev, line 348: why is this case in here? I was expecting an error and more explicit mappings

pkg/graph/schema/conversion_cel_type.go
```diff
@@ -0,0 +1,253 @@
...
+		return &extv1.JSONSchemaProps{Type: "boolean"}, true
+	case "int", "uint":
+		return &extv1.JSONSchemaProps{Type: "integer"}, true
+	case "double":
+		return &extv1.JSONSchemaProps{Type: "number"}, true
+	case "string":
+		return &extv1.JSONSchemaProps{Type: "string"}, true
+	case "bytes":
+		return &extv1.JSONSchemaProps{Type: "string", Format: "byte"}, true
+	case "null_type":
```
jakobmoellerdev, line 58: where does this null_type come from? my assumption is that null_type in json schema is something different. Also I believe that XPreserveUnknownFields can only apply to objects, whereas null type would also allow primitives

pkg/graph/schema/schema.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"fmt"
+
+	"k8s.io/apiextensions-apiserver/pkg/generated/openapi"
+	"k8s.io/apiserver/pkg/cel/openapi/resolver"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+// ObjectMeta holds the k8s ObjectMeta schema, populated once at startup.
+var ObjectMetaSchema spec.Schema
```
jakobmoellerdev, line 26: could this be internal? my assumption is that this is only used internally inside the graph pkg

cmd/kro/commands/generate/instance.go
a-hilaly: Agreed there are caching opportunities here. I think we should aim to address caching (schema conversions, CEL envs etc..) in a separate PR to keep this one focused on the emulation -> type checking migration. There are also runtime optimizations we can tackle separately.

pkg/cel/environment.go
a-hilaly: Agreed, same response as above, i'll open a github issue for it.

a-hilaly: https://github.com/kubernetes-sigs/kro/issues/761

pkg/graph/builder.go
a-hilaly: Created https://github.com/kubernetes-sigs/kro/issues/762

pkg/graph/builder.go
a-hilaly: Agreed. I purposefully avoided these kind of optimizations in this PR. i want to get to logic right first, optimize later on in a seperate PR.

Side thought: curious on how much we'll win by doing concurrency here, maybe useful for very large rgds.

pkg/graph/builder.go
a-hilaly: When `outputType` is `dyn` it means the expression can produce any type at runtime, so there's no static type verification we can do. We have to allow it through.

The reason this happens: there's no 1:1 mapping between CEL types and OpenAPI schemas. We do best efforrt type conversion, but in K8s OpenAPISpec features like `x-kubernetes-int-or-string` and `x-kubernetes-preserve-unknown-fields` both get mapped to CEL's `dyn` type because CEL's type system can't represent them (can't do int | string) 

lmk if missed your point here.

pkg/graph/parser/parser.go
a-hilaly: I think the git diff isn't properly mapping things. The errors do return an explicit mapping

a-hilaly: https://github.com/kubernetes-sigs/kro/blob/edbce61559116d39526d7f014097db2d826c2ae6/pkg/graph/parser/parser.go#L335-L336

pkg/graph/schema/conversion_cel_type.go
a-hilaly: Yes, `null_type` is a real CEL type - it's the type of the `null` literal value. https://github.com/google/cel-go/blob/a36d461be5219280506c29e9eaabf24fee2033cc/cel/io.go#L255

I'm not sure this case actually gets hit in practice. The mapping to `XPreserveUnknownFields` could be - I made it permissive to be safe rather than risk the wrong type. 

It should likely map to `nullable: true` in the OpenAPI schema, but I haven't given it much thought honestly. Worth revisiting if we actually encounter this case.

pkg/graph/schema/schema.go
a-hilaly: Agreed. I think maybe things in this repo should be internal. Maybe worth addressing seperately moving some packages to internal later on?

pkg/graph/builder.go
jakobmoellerdev: I was saying that DynType is a type definition that you should have access to and that should probably have a constant or even type that you can check on.

pkg/graph/parser/parser.go
jakobmoellerdev: thanks for the explanation, feel free to resolve

pkg/graph/builder.go
a-hilaly: This is intentional. In status field schema generation, for "single expression fields" any type allowed:
```yaml
status:
  replicas: ${deployment.status.replicas}  # int
  metadata: ${deployment.metadata}  # object
```

However with "multi expression fields" we do have restrictions:
```yaml
status:
  allowed: "Count: ${string(a.spec.replicas)}, Ready: ${string(a.status.ready)}" #allowed
  notAllowed: "${deployment.metadata}-hellol-world-${deployment.status.replicas}"
```

Concatenating arbitrary types (objects, bools, ints) is complex and error prone. I'm just starting restrictive, we can relax later if there are good use cases and it's technically feasible. For now, users need explicit `string()` calls for multi exprssion fields.

a-hilaly: Some historical context: Initially kro didn't support multi expression fields at all. We added it because users wanted simple string templating instead of complex CEL string manipulation and concatenation expressions. For example IAM policies:
```yaml
status:
  policy: |
    {
      "Effect": "Allow",
      "Resource": "arn:aws:s3:::${bucket.name}/*",
      "Principal": "${serviceAccount.metadata.name}"
    }
```
vs:
```yaml
status:
  policy: | ${json.marshal({ # json.Marshall doesn't exist yet AFAIK
      "Effect": "Allow",
      "Resource": "arn:aws:s3:::" + bucket.name + "/*",
      "Principal": serviceAccount.metadata.name
    })}
```

pkg/graph/builder.go
a-hilaly: ohh yeah i'm addressing that part, my reply was more on your second comment :)

pkg/graph/builder.go
jakobmoellerdev: I have to admit I completely missed that we are in the else branch which only deals with concatenated expressions.

1. A single expression, strictly typed
2. Multiple expressions concatenated, always strings

The argument right now is that status fields only accept strings, whereas in theory they can be object types or integers or others when working with the first case. for the second case, they must stay strings (as you outlined). I subscribe to that 100% and I think we should leave this as is

pkg/graph/builder.go
a-hilaly: https://github.com/kubernetes-sigs/kro/pull/756/files#diff-3393f5070e9756c4a6e99e84aa8e5b7c2061cdac97c143468f073d36c6d536f2R753-R756

pkg/cel/environment.go
a-hilaly: The newest version does a better job than this implementation i think, checkout https://github.com/kubernetes-sigs/kro/blob/eddbdea33e5143a4dd94c318fa93704e1a7dd8bb/pkg/cel/environment.go#L96-L136 - happy to add more comments if it's still obscure.

pkg/graph/schema/conversion_cel_type_test.go
a-hilaly: We can block floats in simple schema, nothing we can do now for CRDs that already have a float fields. Maybe we can warn where users are manipulating float fields?

cmd/kro/go.mod
tjamet: what I did in another project is to have a script in actions that automatically updates the `importer module` on each push
Definitely not the thing I'm the proudest of (it generates other problems like prohibiting rebase/squash merges and remote branch drift from local one) but it kind of works

pkg/graph/builder.go
tjamet: The more I think about it, the less I see an instance of `${ expressionA }${ expressionB }` where `expressionA` or `expressionB` are not strings and the expression makes sense.

edit:

Actually, yes, there is a case, when crafting a json as you shown for example

```
status:
   someField: |
      {
          "generation": ${resource.metadata.generation}
      }
```

Then, generation is an int and gets into a string.
How hard would it be to force a `.String)()` call?

jakobmoellerdev:
/unhold

to ~Infinity~ CEL and beyond

PR #763, opened 2025-10-31
KREP-004: OwnerReferences and deletion policy for ResourceGraphDefinitions

jakobmoellerdev:
Add `OwnerReferences` between ResourceGraphDefinitions (RGDs) and generated CRDs to align with Kubernetes-native lifecycle management. Introduce a `ValidatingAdmissionPolicy` to prevent accidental RGD deletion and refine the `allowCRDDeletion` flag for clear controller-side cleanup behavior. Update documentation and tests accordingly.

Backing KREP for https://github.com/kubernetes-sigs/kro/pull/749 and https://github.com/kubernetes-sigs/kro/issues/745


docs/design/proposals/rgd-ownership-and-deletion.md
```diff
@@ -0,0 +1,203 @@
...
+Introduce proper **Kubernetes ownership semantics** between RGDs and their generated CRDs.
+Combine this with a **safe-deletion mechanism** for user-initiated RGD removals and **clearer internal deletion control** for KRO.
+
+### Overview
+
+This proposal introduces three coordinated changes:
+
+1. **Ownership introduction** — CRDs now include an `OwnerReference` pointing to the creating RGD.
+2. **Deletion protection** — A `ValidatingAdmissionPolicy` prevents accidental user deletion of RGDs.
+3. **Controller deletion control** — The Helm flag `allowCRDDeletion` now cleanly governs whether KRO is permitted to delete CRDs as part of reconciliation.
```
tjamet, line 25: Does this mean the current implementation does not fully honor the flag?

What about making the default to `true`?

There's also a discussion to have about finalizers that plays a different role IMO (the best place is probably  below in the discarded section)

```suggestion
1. **Ownership introduction** — CRDs now include an `OwnerReference` pointing to the creating RGD.
2. **Deletion protection** — A `ValidatingAdmissionPolicy` prevents accidental user deletion of RGDs (blocks deletion of RGDs with instances)
3. **Controller deletion control** — The Helm flag `allowCRDDeletion` now cleanly governs whether KRO is permitted to delete CRDs as part of reconciliation.
```

docs/design/proposals/rgd-ownership-and-deletion.md
```diff
@@ -0,0 +1,203 @@
...
+      uid: 12345-abcde
+      controller: true
+      blockOwnerDeletion: true
+```
+
+**Resulting behavior:**
+
+* When an RGD is deleted, its owned CRD is automatically garbage-collected. This is proper behavior
+  as we can still delete RGDs with policy "orphan" to preserve CRDs.
+* Manual CRD deletion (while RGD still exists) will trigger re-reconciliation, as ownership implies controller responsibility.
```
tjamet, line 62: 💯

docs/design/proposals/rgd-ownership-and-deletion.md
```diff
@@ -0,0 +1,203 @@
...
+Now, with `OwnerReferences` managing lifecycle automatically, the flag’s meaning is refined:
+
+* When `allowCRDDeletion: false`:
+  KRO still sets owner references, but **does not itself delete CRDs**.
+  Deletion is handled by Kubernetes GC only when the RGD is removed.
+  The ValidatingAdmissionPolicy remains active to prevent accidental RGD deletion.
+
+* When `allowCRDDeletion: true`:
+  KRO may directly delete CRDs it manages (e.g., during reconciliation or cleanup).
+  The Helm chart skips installation of the ValidatingAdmissionPolicy, allowing full manual control.
```
tjamet, line 118: I'm unsure I make sense of the semantics.
It feels weird to me to have `allowCRDDeletion: false` and that CRD would be deleted anyhow (by k8s).

There's also a case that could lead to inconsistent behaviour, when the RGD `Kind`, `Group` or `APIVersion` changes. In this case, when `allowCRDDeletion: false` we wouldn't delete the dangling CRD but the CRD would be deleted in case of RGD deletion.

The proposed behaviour and lifecycle sounds more like `confirmCRDDeletion` or `requireManualJudgementForCRDDeletion`
This semantics would also bring consistency when `Kind`, `Group` or `APIVersion` changes.
Maybe the validation webhook needs to be on the CRD itself then?

docs/design/proposals/rgd-ownership-and-deletion.md
```diff
@@ -0,0 +1,203 @@
...
+* Ensure disabling the policy (via Helm) removes admission resources
+
+---
+
+## Other solutions considered
+
+| Option                               | Reason Rejected                                                                         |
+| ------------------------------------ |-----------------------------------------------------------------------------------------|
+| Continue using label linkage only    | No lifecycle tracking or automatic cleanup, against k8s API semantics                   |
+| Finalizer-based blocking             | Overcomplicates reconciliation; redundant with OwnerReferences and foreground finalizer |
```
tjamet, line 159: Can you develop on what makes reconciliation more complex and redundant with ownerreferences?

From kubernetes' [documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/), finalizer seem to pursue this objective: hold a resource until all clean-up is done.

> The target object remains in a terminating state while the control plane, or other components, take the actions defined by the finalizers

The only difference with the current approach is that the RGD would be marked for deletion and have to be entirely deleted before being re-applied, while the proposed approach would block deletion

tjamet (commented):
💯 on aligning with standard k8s behaviours.

I would be in favour of moving forward with this proposal with 2 changes:
- change the parameter from `allowCRDDeletion` to `protectCRDDeletion` or similar
- [not mandatory] default behaviour is to delete "dangling" CRDs

docs/design/proposals/rgd-ownership-and-deletion.md
jakobmoellerdev: That is correct. However, kubernetes sets its own foregroundDeletion as part of client based lifecycle handling: https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion . The proposed approach effectively makes client deletion protected by a label and then allows users to control deletion behavior based on propagation policy.

docs/design/proposals/rgd-ownership-and-deletion.md
```diff
@@ -0,0 +1,203 @@
...
+
+Introduce proper **Kubernetes ownership semantics** between RGDs and their generated CRDs.
+Combine this with a **safe-deletion mechanism** for user-initiated RGD removals and **clearer internal deletion control** for KRO.
+
+### Overview
+
+This proposal introduces three coordinated changes:
+
+1. **Ownership introduction** — CRDs now include an `OwnerReference` pointing to the creating RGD.
+2. **Deletion protection** — A `ValidatingAdmissionPolicy` prevents accidental user deletion of RGDs.
```
ellistarn, line 24: I can understand protecting against the deletion of the Kinds managed by RGDs, but what's the thinking behind stopping a user that says `k delete rgd foo` from deleting the rgd foo. Delete means delete, right?

It strikes me that this could be something that organizations choose to opt in more broadly (deletion protection for any k8s resource), but odd that KRO would take the opinion for users when other projects don't.

docs/design/proposals/rgd-ownership-and-deletion.md
```diff
@@ -0,0 +1,203 @@
...
+metadata:
+  annotations:
+    kro.run/allow-delete: "true"
+```
+
+This ensures intentional deletions only.
+
+#### 3. Controller-side deletion control (`allowCRDDeletion`)
+
+Previously, the flag `allowCRDDeletion` determined whether KRO would delete CRDs it had created if the RGD was removed or changed.
```
ellistarn, line 108: I'm curious to explore the use cases for setting this flag one way or another. If we're certain that all of the instances managed by the RGD are gone, is it really dangerous to remove the Kind? I'm not seeing the upsides of leaking it if we're confident we're not losing data. I can see downsides of not removing the kind, such as attempting to do a fresh installation of a new version where the Kind has changed substantially.

docs/design/proposals/rgd-ownership-and-deletion.md
jakobmoellerdev: the main reason is that by default kubernetes deletes with background propagation. if we delete a RGD, that would automatically start deletion of the CRD as well during the delete. I only added this after there was an opinion against simply allowing a delete. I personally favor this too but adjusted after community call feedback

docs/design/proposals/rgd-ownership-and-deletion.md
tjamet: I'm unsure how a label protects deletion.
It only applies to the KRO controller's own reconciliations, right?
I mean, it does not prevent anybody else from doing `kubectl delete RGD` or `kubectl delete RGDInstance` correct?

docs/design/proposals/rgd-ownership-and-deletion.md
tjamet: Indeed, k8s standard leans towards no protection and delete everything, cascade.
Moreover, once marked for deletion, there is no way back, you must complete deletion before being able to re-use it again (`kubectl delete` starts by setting the deletionTimestamp that can't be undone)

As there is a clear 1-1 ownership link RGD <-> CRD I would indeed favour the standard behaviour

docs/design/proposals/rgd-ownership-and-deletion.md
jakobmoellerdev: No thats not correct. If we use a policy this is on API server level so the deletion protection is enforced globally on the API, not just for the controller

docs/design/proposals/rgd-ownership-and-deletion.md
ellistarn: >  I personally favor this too but adjusted after community call feedback

I think this is the kind of thing where you can easily find voices in both directions. You'll see this debate surface through many AWS APIs, too. Just try deleting an S3 bucket with objects in it.

Given the controversy and the fact that it's broader than just our API, I'd just stick with the existing standard and let a separate solution handle deletion protection holistically.

chrisdoherty4: Given the potential blast radius of RGD deletion it might be beneficial to consider modeling deletions as its own API with approval mechanisms similar to certificate signing requests; administrators can opt-in by configuring RBAC to restrict deletions and force the use of the deletion request API.

As written, the proposal feels a little more like an operational hurdle than protection mechanism.

Issue #744, opened 2025-10-22
Regression: readyWhen condition is ignored when orchestrating dependencies


devdanylo:
### Description

**Observed Behavior**:
Job 2 is created immediately after job 1, without waiting for job 1 to satisfy its `readyWhen` condition.

**Expected Behavior**:
Job 2 is created after job 1 becomes ready according to its `readyWhen` condition.

**Reproduction Steps** :

RGD:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: test-jobs
spec:
  schema:
    apiVersion: v1
    kind: TestJobs
  resources:
    - id: job1
      readyWhen:
        - ${job1.status.completionTime != null}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: job-1
          namespace: kro-tests
        spec:
          template:
            spec:
              containers:
                - name: sleeper
                  image: busybox
                  command: ["sh", "-c", "echo 'Sleeping for 30 seconds' && sleep 30 && echo 'Hello from the Kubernetes Job #1'"]
              restartPolicy: Never
    - id: job2
      readyWhen:
        - ${job2.status.completionTime != null}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: job-2
          namespace: ${job1.metadata.namespace}
        spec:
          template:
            spec:
              containers:
                - name: sleeper
                  image: busybox
                  command: [ "sh", "-c", "echo 'Sleeping for 30 seconds' && sleep 30 && echo 'Hello from the Kubernetes Job #1'" ]
              restartPolicy: Never
```

RGD instance:
```yaml
apiVersion: kro.run/v1
kind: TestJobs
metadata:
  name: test-jobs
  namespace: kro-tests
```

**Versions**:
- kro version: 0.5.0
- Kubernetes Version (`kubectl version`): 
```
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.34.0
```

**Logs**:
```
2025-10-22T17:02:44.096Z    DEBUG    dynamic-controller    Enqueueing object    {"objectIdentifiers": {"NamespacedKey":"kro-tests/test-jobs","GVR":{"Group":"kro.run","Version":"v1","Resource":"testjobs"}}, "eventType": "add"}
[...]

PR #764, opened 2025-11-02
fix: add dependency readiness check in instance reconciliation

govindup63:
Implement a new method to verify if all dependencies of a resource are ready before proceeding with reconciliation. This enhances the instance reconciliation process by ensuring that resources are only applied when their dependencies are resolved and meet readiness conditions.

fix : #744

### DIFF
--- a/test/integration/suites/core/dependency_readiness_test.go
+++ b/test/integration/suites/core/dependency_readiness_test.go
@@ -90,9 +90,9 @@ var _ = Describe("Dependency Readiness", func() {
 				},
 				"data": map[string]interface{}{
 					"value": "${schema.spec.configA.data}",
-					"ready": "${schema.spec.configA.ready}",
+					"ready": "${string(schema.spec.configA.ready)}",
 				},
-			}, []string{"${configmapA.data.ready == \"true\"}"}, nil),
+			}, []string{"${configmapA.data.?ready.orValue(\"false\") == \"true\"}"}, nil),
 			// ConfigMap B - no dependencies, has readyWhen
 			generator.WithResource("configmapB", map[string]interface{}{
 				"apiVersion": "v1",
@@ -102,9 +102,9 @@ var _ = Describe("Dependency Readiness", func() {
 				},
 				"data": map[string]interface{}{
 					"value": "${schema.spec.configB.data}",
-					"ready": "${schema.spec.configB.ready}",
+					"ready": "${string(schema.spec.configB.ready)}",
 				},
-			}, []string{"${configmapB.data.ready == \"true\"}"}, nil),
+			}, []string{"${configmapB.data.?ready.orValue(\"false\") == \"true\"}"}, nil),
 			// Deployment - depends on both configmaps
 			generator.WithResource("deployment", map[string]interface{}{
 				"apiVersion": "apps/v1",
@@ -133,11 +133,11 @@ var _ = Describe("Dependency Readiness", func() {
 									"env": []interface{}{
 										map[string]interface{}{
 											"name":  "CONFIG_A",
-											"value": "${configmapA.data.value}",
+											"value": "${configmapA.data.?value.orValue(\"\")}",
 										},
 										map[string]interface{}{
 											"name":  "CONFIG_B",
-											"value": "${configmapB.data.value}",
+											"value": "${configmapB.data.?value.orValue(\"\")}",
 										},
 									},
 								},
@@ -243,11 +243,13 @@ var _ = Describe("Dependency Readiness", func() {
 		}, 20*time.Second, time.Second).WithContext(ctx).Should(Succeed())
 
 		// Verify Deployment is NOT created yet (dependencies not ready)
-		err := env.Client.Get(ctx, types.NamespacedName{
-			Name:      instanceName,
-			Namespace: namespace,
-		}, &appsv1.Deployment{})
-		Expect(err).To(MatchError(errors.IsNotFound, "deployment should not be created yet"))
+		Consistently(func(g Gomega, ctx SpecContext) {
+			err := env.Client.Get(ctx, types.NamespacedName{
+				Name:      instanceName,
+				Namespace: namespace,
+			}, &appsv1.Deployment{})
+			g.Expect(err).To(MatchError(errors.IsNotFound, "deployment should not be created yet"))
+		}, 10*time.Second, time.Second).WithContext(ctx).Should(Succeed())
 
 		// Verify instance state is IN_PROGRESS
 		Eventually(func(g Gomega, ctx SpecContext) {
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (commented):
thanks for the contrib!

I think this needs a test where this triggers, too important to not be covered. Im also wondering if this shouldnt have already been present in one of our other checks. I.e. ReadyToProcessResource. Maybe thats worth a look. WDYT?

govindup63:
Thanks for the review!
I’ll check if `ReadyToProcessResource` already covers this or if a dedicated test is needed, and follow up accordingly.

Also, could you clarify who usually runs the tests before merging? I’d be willing to take up adding the automated test if it’s not already assigned.

jakobmoellerdev:
The tests run automatically as part of our CI. please check the github checks (E2E and integration tests). Usually we want the fix to be accompanied to the test as well, with rare exceptions.

devdanylo:
Hi @govindup63,
> igr.runtime.ResourceDescriptor(resourceID).GetDependencies()

Do you know how these dependencies are calculated? Are they based on the topological order?

test/integration/suites/core/dependency_readiness_test.go
```diff
@@ -0,0 +1,344 @@
...
+			g.Expect(configMapB.Data["value"]).To(Equal("valueB"))
+			g.Expect(configMapB.Data["ready"]).To(Equal("false"))
+		}, 20*time.Second, time.Second).WithContext(ctx).Should(Succeed())
+
+		// Verify Deployment is NOT created yet (dependencies not ready)
+		err := env.Client.Get(ctx, types.NamespacedName{
+			Name:      instanceName,
+			Namespace: namespace,
+		}, &appsv1.Deployment{})
+		Expect(err).To(MatchError(errors.IsNotFound, "deployment should not be created yet"))
```
a-hilaly: We can wrap this with `Consistently`

test/integration/suites/core/dependency_readiness_test.go
```diff
@@ -0,0 +1,344 @@
...
+		configMapA.Data["ready"] = "true"
+		Expect(env.Client.Update(ctx, configMapA)).To(Succeed())
+
+		// Wait a bit and verify deployment is still NOT created (ConfigMap B not ready yet)
+		time.Sleep(5 * time.Second)
+		err = env.Client.Get(ctx, types.NamespacedName{
+			Name:      instanceName,
+			Namespace: namespace,
+		}, &appsv1.Deployment{})
+		Expect(err).To(MatchError(errors.IsNotFound, "deployment should still not be created"))
```
a-hilaly: ditto

test/integration/suites/core/dependency_readiness_test.go
```diff
@@ -0,0 +1,344 @@
...
+				"apiVersion": "v1",
+				"kind":       "ConfigMap",
+				"metadata": map[string]interface{}{
+					"name": "${schema.spec.name}-config-b",
+				},
+				"data": map[string]interface{}{
+					"value": "${schema.spec.configB.data}",
+					"ready": "${schema.spec.configB.ready}",
+				},
+			}, []string{"${configmapB.data.ready == \"true\"}"}, nil),
```
a-hilaly:
```suggestion
			}, []string{"${configmapB.data.?ready.orValue("false") == \"true\"}"}, nil),
```

test/integration/suites/core/dependency_readiness_test.go
```diff
@@ -0,0 +1,344 @@
...
+				"apiVersion": "v1",
+				"kind":       "ConfigMap",
+				"metadata": map[string]interface{}{
+					"name": "${schema.spec.name}-config-a",
+				},
+				"data": map[string]interface{}{
+					"value": "${schema.spec.configA.data}",
+					"ready": "${schema.spec.configA.ready}",
+				},
+			}, []string{"${configmapA.data.ready == \"true\"}"}, nil),
```
a-hilaly:
```suggestion
			}, []string{"${configmapA.data.?ready.orValue("false") == \"true\"}"}, nil),
```

a-hilaly: This should fix the RGD validaiton

a-hilaly (commented):
Thank @govindup63 !! leaving a few comments below

a-hilaly:
@govindup63 Since this is critical and your PR is very close, i just pushed a commit to your branch to get the test fixed.

jakobmoellerdev (approved):
/lgtm

a-hilaly:
Thank you @govindup63 for helping with this!!

devdanylo:
Thanks!
@a-hilaly will you release a hotfix?

a-hilaly:
@devdanylo we're making 0.6 this week, ideally tomorrow or after tomorrow. In the mean time you prow does publish images every time we push to main, you can use the tag `v20251104-d2267c1`. 

You can see the list of published images using `crane ls us-central1-docker.pkg.dev/k8s-staging-images/kro/kro`

devdanylo:
@a-hilaly @govindup63 I've tested it with `v20251104-d2267c1` and the issue is still there. Test RGD (no dependencies, only topological order):
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: test-jobs
spec:
  schema:
    apiVersion: v1
    kind: TestJobs
  resources:
    - id: job1
      readyWhen:
        - ${job1.status.completionTime != null}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: job-1
        spec:
          template:
            spec:
              containers:
                - name: sleeper
                  image: busybox
                  command: ["sh", "-c", "echo 'Sleeping for 30 seconds' && sleep 30 && echo 'Hello from the Kubernetes Job #1'"]
              restartPolicy: Never
    - id: job2
      readyWhen:
        - ${job2.status.completionTime != null}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: job-2
        spec:
          template:
            spec:
              containers:
                - name: sleeper
                  image: busybox
                  command: [ "sh", "-c", "echo 'Sleeping for 30 seconds' && sleep 30 && echo 'Hello from the Kubernetes Job #2'" ]
              restartPolicy: Never
```

a-hilaly:
@devdanylo thank you for flagging that, i'll add a integration test for this exact case and see what happens

a-hilaly:
@devdanylo i think i might've shared the wrong image can you please test `v20251104-9996de3` (i assumed they are in chronological order but that's not entirely true) (thanks @jakobmoellerdev !)

devdanylo:
@a-hilaly the same with `v20251104-9996de3`

a-hilaly:
looking ... @devdanylo are you using `us-central1-docker.pkg.dev/k8s-staging-images/kro/kro` as registry?

devdanylo:
> are you using us-central1-docker.pkg.dev/k8s-staging-images/kro/kro as registry?

@a-hilaly that's the only option I have. I assume the staging images are not pushed to other registries.

a-hilaly:
@devdanylo can you describe the behaviour you're observing vs expecting?

devdanylo:
@a-hilaly 
Expected: job2 is created after job1 is ready
Actual: both jobs are created at the same time

jakobmoellerdev:
The reason that your test is not working as expected is because the test case you used does not have a dependency like in the issue you linked. Pay attention to `job2`. In the first issue example you send we had this graph here:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: test-jobs
spec:
  schema:
    apiVersion: v1
    kind: TestJobs
  resources:
    - id: job1
      readyWhen:
        - ${job1.status.completionTime != null}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: job-1
          namespace: kro-tests
        spec:
          template:
            spec:
              containers:
                - name: sleeper
                  image: busybox
                  command: ["sh", "-c", "echo 'Sleeping for 30 seconds' && sleep 30 && echo 'Hello from the Kubernetes Job #1'"]
              restartPolicy: Never
    - id: job2
      readyWhen:
        - ${job2.status.completionTime != null}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: job-2
          namespace: ${job1.metadata.namespace}
        spec:
          template:
            spec:
              containers:
                - name: sleeper
                  image: busybox
                  command: [ "sh", "-c", "echo 'Sleeping for 30 seconds' && sleep 30 && echo 'Hello from the Kubernetes Job #1'" ]
              restartPolicy: Never
```

because of `namespace: ${job1.metadata.namespace}` the jobs get created after each other.

With your example, you did not include a namespace reference, so now they get created concurrently.

devdanylo:
@jakobmoellerdev correct, but then as I understood from @a-hilaly's comment [here](https://github.com/kubernetes-sigs/kro/issues/744#issuecomment-3457323508) it's irrelevant - to have resources created sequentially it should be enough 1) to define them in the right topological order and 2) to provide "readWhen" conditions for the resources:

> The thing is that kro will only move to the next resources in topological order when previous resources all reach ready state.

This is how it was working in `v0.4.1`.

P.S. If I define the depends-on relationship between the jobs, as in the original issues description, it works as expected.

jakobmoellerdev:
I believe you misinterpreted the comment:

> In your situation you don't need includeWhen, job2 will be created after job1 is completed. The thing is that kro will only move to the next resources in topological order when previous resources all reach ready state.

this means that you need a dependency on job1, so that then the readyWhen condition is waited for. @a-hilaly just meant that in this case we dont need includeWhen, but your dependency statement is still required.

devdanylo:
@jakobmoellerdev sorry for the confusion. Due to lack of details in the documentation I was not entirely sure whether the explicit depends-on relationship is a requirement for sequential readiness-based resource creation.

Nonetheless, I've retested `v0.4.1`, and the resources are created sequentially without the explicit depends-on relationship. I would expect KRO to work the same way in `v0.5.x`. That's the reason why I provided a modified RGD [here](https://github.com/kubernetes-sigs/kro/issues/744#issuecomment-3457437772).

jakobmoellerdev:
No worries! I just think one of the main switches we did was move to applysets which is why non-dependant resources now get applied concurrently. I believe this is expected behavior as I mentioned before.

devdanylo:
Okay, got it. In such case we should be good, the fix works as expected.

devdanylo:
@jakobmoellerdev @a-hilaly would you guys consider introducing a new syntax, similar to `readyWhen`, for dependencies:
```yaml
- id: job2
  readyWhen:
    - ${job2.status.completionTime != null}
  dependsOn:
    - job1
```
Because the current approach with the namespace from the example feels rather artificial.

PR #767, opened 2025-11-04
test: add jobs test case for dependency readiness

a-hilaly:
Adds test case with two jobs where `job2` depends on `job1` via CEL reference.
Validates that `job2` waits for `job1` to satisfy its `readyWhen` condition
before being created.


jakobmoellerdev (commented):
/lgtm

a-hilaly:
fyi @devdanylo

PR #768, opened 2025-11-04
chore(applyset): add missing resource status markers for reconciliation process

jakobmoellerdev:
previously if the resources were not ready in the apply, we did not set the resource state correctly because the marker updates were missing.


a-hilaly (approved):
👍

a-hilaly:
@jakobmoellerdev please unhold if you think this is ready to go

jakobmoellerdev:
im struggling to provision a good testing flow here because instance controller has not the best test setup outside of integration tests. I vote to merge this in and then I will take a look at providing a stable test base for instance controller. WDYT?

a-hilaly:
sgtm
/unhold

PR #776, opened 2025-11-07
docs: enhance RGD, SimpleSchema, and Instance documentation

a-hilaly:
Update documentation to reflect recent architectural improvements in kro v0.6.0.

### DIFF
--- a/website/docs/docs/concepts/00-resource-group-definitions.md
+++ b/website/docs/docs/concepts/00-resource-group-definitions.md
@@ -137,7 +137,10 @@ correctness and set up the necessary components:
    - Checks that referenced values exist and are of the correct type
    - Confirms resource dependencies form a valid Directed Acyclic Graph(DAG)
      without cycles
-   - Validates all CEL expressions in status fields and conditions
+   - Validates all CEL expressions in status fields and conditions using CEL's native type system
+     - Validates field references exist in the actual resource schemas
+     - Ensures expressions return types compatible with their target fields
+     - Checks type compatibility statically without executing expressions
 
 2. **API Generation**: kro generates and registers a new CRD in your cluster
    based on your schema. For example, if your **ResourceGraphDefinition** defines a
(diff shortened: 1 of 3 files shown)


website/docs/docs/concepts/00-resource-group-definitions.md
```diff
@@ -137,7 +137,10 @@ correctness and set up the necessary components:
    - Checks that referenced values exist and are of the correct type
    - Confirms resource dependencies form a valid Directed Acyclic Graph(DAG)
      without cycles
-   - Validates all CEL expressions in status fields and conditions
+   - Validates all CEL expressions in status fields and conditions using CEL's native type system
+     - Validates field references exist in the actual resource schemas
+     - Ensures expressions return types compatible with their target fields
```
matthchr, line 142: Not sure if it's worth calling out here, but presumably this includes validating that the functions you're calling exist so it's not only that expressions return compatible types, it validates the expression "correctness" (not sure how to call it) as well, right?

website/docs/docs/concepts/10-simple-schema.md
```diff
@@ -268,22 +268,70 @@ username: string | minLength=3 maxLength=20 pattern="^[a-zA-Z0-9_]+$"
...
+  metadata: ${deployment.metadata}  # object
+  name: ${deployment.metadata.name}  # string
+  ready: ${deployment.status.conditions.exists(c, c.type == 'Available')}  # boolean
 ```
 
+**Multi-Expression Fields** (string templating) must contain only string expressions:
+```yaml
+status:
+  # ✓ Valid - all expressions return strings
+  endpoint: "https://${service.metadata.name}.${service.metadata.namespace}.svc.cluster.local"
```
matthchr, line 310: This is nice although CEL also has a way to do this AFAIK, `"%s:%d".format([self.spec.location, 7])` for example - maybe reads slightly less clearly though.

I assume we added special handling for this to make this work?

website/docs/docs/concepts/15-instances.md
```diff
@@ -88,17 +125,64 @@ Every instance includes:
    - `DELETING`: Indicates that the instance is in the process of being deleted.
    - `ERROR`: Indicates that an error occurred during instance processing.
 
-2. **Conditions**: Detailed status information
+2. **Conditions**: Detailed status information structured hierarchically
+
+   kro provides a top-level `Ready` condition that reflects the overall instance health. This condition is supported by three sub-conditions that track different phases of the reconciliation process:
```
matthchr, line 130: Maybe worth highlighting that users should use this condition as the condition to wait on and that we reserve the right to change/add additional sub-conditions?

a-hilaly: Good callout will add that.

website/docs/docs/concepts/00-resource-group-definitions.md
a-hilaly: Good callout will add that.

website/docs/docs/concepts/10-simple-schema.md
a-hilaly: It works :) i'll add the example. nice call out

PR #782, opened 2025-11-09
feat: enable CEL reserved keyword recognition for field access

a-hilaly:
The Kubernetes apiserver CEL library automatically translates field
lookups for reserved keywords (namespace, return, package, etc...) when
`SetRecognizeKeywordAsFieldName` is enabled. This provides a better UX
by hiding implementation details from users.

this patch enables automatic recognition of CEL reserved keywords as field
names in `DeclTypeProvider`. This allows users to access fields like
'namespace' naturally in CEL expressions without manual escaping. e.g
`schema.spec.namespace` instead of `schema.spec.__namespace__`

Context:
- https://github.com/kubernetes/apiserver/blob/master/pkg/cel/escaping.go#L90-L100


jakobmoellerdev (approved):
/lgtm

thanks for the fix

PR #787, opened 2025-11-10
Allow RGD resources to reference instance status fields

### LINKED ISSUE #762: Support referencing instance status fields in RGD resources
### Feature Description

Right now resources can reference `schema.spec.*` and `schema.metadata.*`, but not `schema.status.*`.

This would be useful for resources that need to get information directly from instance status.

example: 

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: app-with-monitoring
spec:
  schema:
    apiVersion: v1alpha1
    kind: Application
    spec:
      replicas: integer
    status:
      endpoint: string  # computed from service
      
  resources:
    - id: service
      template:
        apiVersion: v1
        kind: Service
        # ... service definition
        
    - id: monitoring-alert
      template:
        apiVersion: monitoring.example.com/v1
        kind: Alert
        spec:
          # Want to reference the computed status field
          target: ${schema.status.endpoint}  # not allowed today
```

bschaatsbergen:
Closes #762

Currently we explicitly exclude `schema.status` fields from being referenced in CEL expressions on RGD resources. The [getSchemaWithoutStatus](https://github.com/kubernetes-sigs/kro/blob/548e4db0531c3bfddbda3922e66620b224c4ce4a/pkg/graph/builder.go#L835) function deletes the status property before validation, preventing `schema.status.*` references, as previously intended.

This does however block the ability to reference computed status values (like `status.endpoint: ${service.status.loadBalancer.ingress[0].hostname}`) in dependent resources, forcing users to duplicate expressions instead of implementing a cleaner dependency pattern.

This removes that restriction by renaming the function to `getInstanceSchema` and including `schema.status` alongside `schema.spec` and `schema.metadata`.

Quick question for the maintainers; what was the reasoning behind initially excluding the status fields? Understanding the background here would help me add more appropriate safeguards around including `schema.status`.


bschaatsbergen:
> Hey there! thanks for the contrib. Mainly the limitations we must consider are:
> 
> 1. Status Objects are always optional and may or may not be present. Referencing should only be possible with optional value reference closures IMHO.
> 2. Status Objects were previously unstructured so to get a proper setup, you would also need to replace usage here https://github.com/kubernetes-sigs/kro/blob/548e4db0531c3bfddbda3922e66620b224c4ce4a/pkg/graph/builder.go#L572
> 
> I think if you manage to get rid of dynamic status value computation on top of your change, this will be more "complete". Otherwise your RGD creation will have the schema but your cluster based instance will still not use this

Thanks for the quick review, @jakobmoellerdev! Just to confirm my understanding; we should exclude status from required fields before type-checking the instance schema, so that CEL expressions in dependent resources can reference the instance’s status without validation errors when it’s still empty? Basically, clean up/unmark any status entry from the `schema.Required` list? 

In b4128a9, I've ensured status fields remain optional before type-checking the instance schema.

jakobmoellerdev:
Before I push you further into any direction and you potentially have a waste of time, I want to have a conceptual guidance of @a-hilaly here as he wrote the original status schemaless logic. Expect a bit of turnaround due to kubecon please.

My understanding is that rgd resources should be able to FILL instance status fields, but not the other way around. an instance should aggregate status and we didnt have good cases yet in which the other way would have made sense.

tjamet:
First, I agree that, conceptually, we must be able to bubble up the resources status to the RGD instance the same way we cascade parameters down.

It is true that this brings a complexity when the status does not exist yet, but I think it is manageable and worth the effort

pkg/graph/builder_test.go
```diff
@@ -1766,6 +1766,163 @@ func TestGraphBuilder_CELTypeChecking(t *testing.T) {
...
+		{
+			name: "resource references instance status field (string)",
+			resourceGraphDefinitionOpts: []generator.ResourceGraphDefinitionOption{
+				generator.WithSchema(
+					"Application", "v1alpha1",
+					map[string]interface{}{
+						"region": "string",
+					},
+					map[string]interface{}{
+						"vpcID": "${vpc.status.vpcID}",
```
tjamet, line 1806: We also need tests to ensure that:
1. when the status field can't be resolved (i.e. `vpc.status == nil`) the instance can be reconciled without error
2. when the status field exists (i.e. `vpc.status.vpcID = "vpc-xzy"), after reconcile, the instance has status `status.vpcID: vpc-xyz`

pkg/graph/builder.go
```diff
@@ -831,8 +829,28 @@ func validateReadyWhenExpressions(env *cel.Env, resource *Resource) error {
...
+	}
+
+	// Remove "status" from required fields if present
+	required := make([]string, 0, len(s.Required))
+	for _, field := range s.Required {
+		if field != "status" {
+			required = append(required, field)
+		}
+	}
+	s.Required = required
```
tjamet, line 960: Would we gain readability on the intent using `slices.DeleteFunc`

```suggestion
	s.Required = slices.DeleteFunc(s.Required, func(e string){
		return e == "status"
	})
```

a-hilaly:
We've decided to not go with this - referencing instance status won't be allowed for now. Thank you @bschaatsbergen

PR #790, opened 2025-11-10
api: add AdditionalPrinterColumnPolicy and merge logic for additional…

yashpal2104:
… printer columns

### API change

### What the policy does 

### Backwards compatibility


yashpal2104:
The prow bot keeps removing the label `ok-to-test`

jakobmoellerdev:
I think it was something different. we needed https://github.com/kubernetes/test-infra/pull/35781 to land and I didnt lgtm that yet. lets wait a bit and see if the jobs pop up

yashpal2104:
A little help @jakobmoellerdev here I don't seem to understand why are the tests failing, I locally tested it and it did run fine

jakobmoellerdev:
were missing the hooked files in main: https://github.com/kubernetes-sigs/kro/tree/main/scripts/ci/presubmits

yashpal2104:
> were missing the hooked files in main: https://github.com/kubernetes-sigs/kro/tree/main/scripts/ci/presubmits

Anything I can do to help/fix out

jakobmoellerdev:
no worries, I created https://github.com/kubernetes-sigs/kro/pull/792 but will need an approver to ack this. Currently its kubecon time so give us a bit of turnaround

jakobmoellerdev:
please rebase, after that we should be good

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -16,7 +16,28 @@ package v1alpha1
...
+	runtime "k8s.io/apimachinery/pkg/runtime"
+)
+
+// AdditionalPrinterColumnPolicy defines how additional printer columns are applied to the generated CustomResourceDefinition (CRD). This policy controls whether user-provided columns replace or merge with the default columns.
+type AdditionalPrinterColumnPolicy string
+
+const (
+	// DefaultServiceAccountKey is the key to use for the default service account
+	// in the serviceAccounts map.
+	DefaultServiceAccountKey = "*"
```
jakobmoellerdev: I think this is a leftover from an accidental rebase from an older state where we still had service account selection. we removed this atm though, so lets not add the field here

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -116,13 +116,33 @@ spec:
...
+                  additionalPrinterColumnPolicy:
+                    default: Replace
+                    description: |-
+                      AdditionalPrinterColumnPolicy defines additional printer columns
+                      that will be passed down to the created CRD. The default policy is "Replace".
+
+                      AdditionalPrinterColumnPolicy controls how provided additional printer
+                      columns via AdditionalPrinterColumns are applied to the generated CRD.
+
+                      - "Replace": uses the provided columns as is; if no or empty AdditionalPrinterColumns are provided, fallback to the default columns.
```
jakobmoellerdev, line 128: this looks very strange, please double check the entire description field and make sure it generates correctly

pkg/graph/crd/crd.go
```diff
@@ -103,10 +113,34 @@ func newCRDSchema(spec, status extv1.JSONSchemaProps, statusFieldsOverride bool)
...
+func newCRDAdditionalPrinterColumns(
+	policy v1alpha1.AdditionalPrinterColumnPolicy,
+	userCols []extv1.CustomResourceColumnDefinition,
+) []extv1.CustomResourceColumnDefinition {
+	if len(userCols) == 0 {
 		return defaultAdditionalPrinterColumns
 	}
 
-	return additionalPrinterColumns
+	// For backwards compatibility treat an empty policy string as Replace.
```
jakobmoellerdev: this needs more explanation i believe

pkg/graph/crd/crd.go
```diff
@@ -103,10 +113,34 @@ func newCRDSchema(spec, status extv1.JSONSchemaProps, statusFieldsOverride bool)
 	}
 }
 
-func newCRDAdditionalPrinterColumns(additionalPrinterColumns []extv1.CustomResourceColumnDefinition) []extv1.CustomResourceColumnDefinition {
-	if len(additionalPrinterColumns) == 0 {
+func newCRDAdditionalPrinterColumns(
```
jakobmoellerdev, line 129: nit: the function deserves some docs

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
yashpal2104: Right I may have missed this since it was a merge conflict while I was commiting

api/v1alpha1/resourcegraphdefinition_types.go
yashpal2104: Yup I cherry picked the commit and it might have added it since I rebased it there

jakobmoellerdev (commented):
/lgtm
/hold

this is looking good to me now. just holding in favor of upcoming 0.6.1. Thanks for the work!

bschaatsbergen:
FYSA https://github.com/kubernetes-sigs/kro/pull/817 fixes the failing verify-lint job by marking the invoked script as executable (chmod).

yashpal2104:
> FYSA #817 fixes the failing verify-lint job by marking the invoked script as executable (chmod).

Is there something needed to do by me? Or will this issue be addressed in other PR? I cant seem to understand what seems to be failing since all the tests locally pass

jakobmoellerdev:
please rebase

a-hilaly:
looks like this PR needs a rebase. Can you PTAL @yashpal2104 ?

yashpal2104:
Not able to see the option to rebase here, Do I need to rebase it from VSCode/locally

yashpal2104:
Hey @a-hilaly tried manually resolving conflicts and it got way worse than it should have, should I create a new PR and cherry-pick the commits?

Help needed @jakobmoellerdev

jakobmoellerdev:
No worries. I recommend you to 

1. Checkout latest remote main of kubernetes-sigs/kro into a new branch
2. Then take the entire diff of this branch (your changes) and apply them in one commit to that new branch you just created
3. Then force push the contents of this branch into your `add-printercolumn-policy-crd` remote branch.

config/crd/bases/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -0,0 +1,329 @@
+---
```
tjamet, line 1: This file was deleted in a previous commit.
It should not reappear, maybe a rebase issue or this is a bug in the generation?

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -38,6 +38,10 @@ import (
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
 )
 
+// +kubebuilder:rbac:groups=kro.run,resources=resourcegraphdefinitions,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=kro.run,resources=resourcegraphdefinitions/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=kro.run,resources=resourcegraphdefinitions/finalizers,verbs=update
+
```
tjamet, line 44: This also got removed in #819 we should not re-introduce it unless we re-introduce a working rbac generation

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -16,7 +16,24 @@ package v1alpha1
...
+
+// AdditionalPrinterColumnPolicy defines how additional printer columns are applied to the generated CustomResourceDefinition (CRD). This policy controls whether user-provided columns replace or merge with the default columns.
+type AdditionalPrinterColumnPolicy string
+
+const (
+	// AdditionalPrinterColumnPolicyReplace specifies that the provided additional
+	// printer columns should completely replace the default columns. If no
+	// AdditionalPrinterColumns are provided or the list is empty, the system
+	// will fall back to using the default columns.
+	AdditionalPrinterColumnPolicyReplace AdditionalPrinterColumnPolicy = "Replace"
```
tjamet, line 30: I don't think we should expose this. IMO there should be only supported way that behaves similarly to CRDs:
- If a user defines an `additionalPrinterColumns`, it gets added to the columns
- If a user overrides a column, there should be an additional column with the same title (like how CRD works with `Name`)
- There should probably be a policy forbidding to override KRO's defined columns (STATE, READY, AGE). That's an addition to CRDs, but it doesn't change the behaviour.

I see no reason where the KRO's added printed columns should be removed.
Kro objects should have consistent column data.

A side note that the proposed `Add` is confusing. It is currently some kind of `Merge` where the definition of a column content can be overridden rather than `Add`

jakobmoellerdev:
/close

closing as being superceded by https://github.com/kubernetes-sigs/kro/pull/825

Issue #785, opened 2025-11-10
LogLevel does not seem to work as advertised


anthonyh-quilt:
### Description

The bug #780 generates a _lot_ of noise in the logs. When I tried adjusting the log level (using 0-5 values)- I could not get the debug/warning logs to stop. I had noticed in the logs that it was showing as Level(-2)...

**Observed Behavior**:
Setting level to -3, -2 and -1 actually felt like I was switching between Info, Warning, Error

I wonder if it has to do with this particular line?
https://github.com/kubernetes-sigs/kro/blob/main/cmd/controller/main.go#L59
```
return -int(lvl) <= c.level
       ^
```

**Expected Behavior**:
Setting log level lower (towards 0) reduces log verbosity
Setting log level higher (towards 5) increases log verbosity

**Versions**:
- kro version: 0.5.1 / 0.6.0
- Kubernetes Version (kubectl version):
```
Client Version: v1.30.2
Server Version: v1.32.9-eks-113cf36
```

**Error Logs** (if applicable)**:
```
2025-11-10T06:23:35.084Z	LEVEL(-2)	statelessservices-controller	applied object	{"controller": "statelessservices", "controllerGroup": "kro.run", "controllerKind": "StatelessService", "namespace": "dev", "name": "webapp", "object": "[traefik.io/v1alpha1, Kind=IngressRoute:dev/webapp]", "applied-revision": "118872308", "error": null}
```

---

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

a-hilaly:
@anthonyh-quilt Would like to take a stab at this?

bschaatsbergen:
I also noticed the default is set to 10, which seems unusual given that Zap’s highest log level is 5 ([Fatal](https://github.com/uber-go/zap/blob/master/zapcore/level.go#L52)￼). Since Zap log levels are hierarchical, setting a level shows logs from that level and all higher-severity levels, a default of 10 effectively shows all logs, including debug, info, warn, and error. Was this intentional to ensure verbose logging by default?

[...]

PR #791, opened 2025-11-10
refactor: adopt controller-runtime standard logging flags

bschaatsbergen:
Relates to #785

This PR adopts controller-runtime's standard logging flags, replacing the custom `--log-level` flag with the widely-used
  `--zap-log-level` flag and related options. See https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options.BindFlags.

**Warning**: This is a breaking change for existing deployments as this will replace the `--log-level` flag as well as changes the default verbosity.


jakobmoellerdev:
> Currently, the default log level is [set to 10](https://github.com/kubernetes-sigs/kro/blob/main/cmd/controller/main.go#L123), which is unusual because Zap’s highest defined level is 5 ([Fatal](https://github.com/uber-go/zap/blob/master/zapcore/level.go#L52)). Because Zap log levels are hierarchical, meaning a configured level shows logs at that level and all higher-severity levels, a default of 10 effectively enables all logs, including debug, info, warn, and error (if that is intended, discard this PR however.)

AFAIU this was intended, but I think its time to change that :D We had too verbose logging from development in kro and defaulting to a less verbose level like info seems sensible to me. The breaking change here is acked from my side but I believe we need to do it anyhow.

cmd/controller/main.go
```diff
@@ -120,7 +124,7 @@ func main() {
 	flag.IntVar(&queueMaxRetries, "dynamic-controller-default-queue-max-retries", 20,
 		"maximum number of retries for an item in the queue will be retried before being dropped")
 	// log level flags
-	flag.IntVar(&logLevel, "log-level", 10, "The log level verbosity. 0 is the least verbose, 5 is the most verbose.")
+	flag.IntVar(&logLevel, "log-level", 0, "The log level verbosity. Lower values reduce verbosity (e.g., -1=debug, 0=info, 1=warn, 2=error).")
```
jakobmoellerdev: I think we wouldnt exactly need a custom level enabler. shouldnt controller runtime already bring a log enabler zap logger that follows similar (not same) behavior? PTAL at https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap#Options.BindFlags and try to adopt the controller runtime logger where possible.

From flag bindings

> zap-log-level: Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', 'panic' or any integer value > 0 which corresponds to custom debug levels of increasing verbosity").

for me this is a good behavior we can also use. I would be willing to accept breaking change on the flags even to normalize across other controllers (using controller runtime flags is very common nowadays thanks to kubebuilder)

bschaatsbergen: Good callout, if we’re introducing a breaking change, we might as well standardize on the controller-runtime default logging flags. [a769115](https://github.com/kubernetes-sigs/kro/pull/791/commits/a769115300c9aaaf0e77b92b2da1829c60be72c1) refactors that. It is however good to note that ACK controllers do use `--log-level`.

bschaatsbergen: Regarding the Helm chart, I’ll avoid introducing a breaking change by **not** renaming anything there, to maintain compatibility with existing configurations. What do you think @jakobmoellerdev?

cmd/controller/main.go
```diff
@@ -171,7 +158,6 @@ func main() {
 		// if you are doing or is intended to do any operation such as perform cleanups
 		// after the manager stops then its usage might be unsafe.
 		LeaderElectionReleaseOnCancel: false,
-		Logger:                        rootLogger,
```
bschaatsbergen, line 174: Removing this, as per the package docs on the [Options](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.22.4/pkg/manager#Options): 

> // Logger is the logger that should be used by this manager.
	// If none is set, it defaults to log.Log global logger.
	Logger [logr](https://pkg.go.dev/github.com/go-logr/logr).[Logger](https://pkg.go.dev/github.com/go-logr/logr#Logger)

Since we're setting it in `cmd/controller/main.go`: `ctrl.SetLogger(rootLogger)`

jakobmoellerdev:
/hold

we may need to create 0.6.1 which would discourage now introducing a breaking change. Lets save this one for 0.7.0

helm/values.yaml
```diff
@@ -120,8 +120,8 @@ config:
   dynamicControllerDefaultResyncPeriod: 36000
   # The maximum number of retries for an item in the queue will be retried before being dropped
   dynamicControllerDefaultQueueMaxRetries: 20
-  # The log level verbosity. 0 is the least verbose, 5 is the most verbose
-  logLevel: 3
+  # Log level verbosity: 'debug', 'info', 'error', 'panic', or integer > 0
+  logLevel: "info"
```
tjamet, line 124: We need to inform users that the level changed in release notes

a-hilaly: i'll take care of that

bschaatsbergen:
https://github.com/kubernetes-sigs/kro/pull/817 fixes the failing verify-lint job by marking the invoked script as executable (chmod).

cmd/controller/main.go
```diff
@@ -119,22 +109,19 @@ func main() {
...
-	flag.IntVar(&logLevel, "log-level", 10, "The log level verbosity. 0 is the least verbose, 5 is the most verbose.")
 	// qps and burst
 	flag.Float64Var(&qps, "client-qps", 100, "The number of queries per second to allow")
 	flag.IntVar(&burst, "client-burst", 150,
 		"The number of requests that can be stored for processing before the server starts enforcing the QPS limit")
 
-	flag.Parse()
-
 	opts := zap.Options{
 		Development: true,
```
tjamet, line 118: Just to follow-up on #828, even if we have `Development: true` `-zap-encoder=json` will enable json logs

Image

jakobmoellerdev (commented):
/unhold

a-hilaly:
Thank you @bschaatsbergen !
/approve
/lgtm

PR #798, opened 2025-11-11
refactor: migrate to golangci-lint v2

### LINKED ISSUE #797: update to golangci-lint v2
We should update golangci-lint from v1 to v2 to take advantage of the revamped configuration, simplified linters management, and new commands like `golangci-lint fmt`. The migration should be possible using `golangci-lint migrate`, but we need to update the golangci-lint GHA workflow.

bschaatsbergen:
Closes #797

This migrates our project to golangci-lint v2, announced earlier this year. The configuration was updated using `golangci-lint migrate`, and the GitHub Actions workflow has been bumped accordingly.

**Note for reviewers**: The additional code changes were applied using `golangci-lint run --fix`. They mainly involve fields that can now be accessed directly through embedding and some unnecessary typing that was removed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,6 +33,10 @@ linters:
       - third_party$
       - builtin$
       - examples$
+  settings:
+    staticcheck:
+      checks:
+        - -QF1008
 formatters:
   enable:
     - gofmt


pkg/graph/parser/parser.go
```diff
@@ -233,7 +233,7 @@ func parseObject(field map[string]interface{}, schema *spec.Schema, path string,
 
 func parseArray(field []interface{}, schema *spec.Schema, path string, expectedTypes []string) ([]variable.FieldDescriptor, error) {
 	// Look for vendor schema extensions first
-	if len(schema.VendorExtensible.Extensions) > 0 {
```
jakobmoellerdev, line 236: nit: i think we dont need to enforce a linter that makes you NOT use embedded struct fields. WDYT?

bschaatsbergen: Valid nit, let me look into this

bschaatsbergen: Disabled [QF1008](https://staticcheck.dev/docs/checks#QF1008) in 104930e.

PR #799, opened 2025-11-11
chore: bump and pin GitHub Actions by SHA instead of tag

bschaatsbergen:
This updates our GitHub Actions workflows to use (including bumped versions) commit SHAs for actions instead of floating tags. Pinning by SHA ensures that the workflow always uses the exact version of the action, preventing unexpected behavior caused by upstream changes in action tags.

There are no functional changes in this pull request, apart from removing a matrix which had a single entry, see `.github/workflows/unit-tests.yaml`.

See https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions


.github/workflows/unit-tests.yaml
```diff
@@ -18,16 +18,13 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
```
bschaatsbergen, line 21: Apart from the commit SHA pinning and bumping some actions from `vX` to `vY`, this is the only _logical_ change. Happy to revert this if our plan was to run tests also against other versions.

bschaatsbergen:
> Thanks @bschaatsbergen! looks like we just migrated a few tests to prow, should we remove them completely? thinking unit/e2e/integrations/build-image

Good point, given https://github.com/kubernetes/test-infra/pull/35781, the Prow jobs are stable, I think we could remove them 👍🏼

PR #800, opened 2025-11-12
chore: bump to go 1.25.0 and bump controller-gen to v0.19.0

bschaatsbergen:
Closes #802

When running `make build`, `make deploy-kind` or `make test` I kept running into the same issue:
 ```console
$ make build
test -s /Users/bschaatsbergen/github/bschaatsbergen/kro/bin/controller-gen && /Users/bschaatsbergen/github/bschaatsbergen/kro/bin/controller-gen --version | grep -q v0.16.2 || \
        GOBIN=/Users/bschaatsbergen/github/bschaatsbergen/kro/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.2
# golang.org/x/tools/internal/tokeninternal
../../../go/pkg/mod/golang.org/x/tools@v0.24.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
make: *** [/Users/bschaatsbergen/github/bschaatsbergen/kro/bin/controller-gen] Error 1
```

After some digging, I found the cause was related to the `controller-gen` version we pin in the `Makefile`. The build error was caused by an incompatibility between Go 1.25.4 and the old controller-gen version (v0.16.2), which depended on an outdated `golang.org/x/tools@v0.24.0` that doesn't compile with newer Go versions (1.25.x).

[The previous controller-gen version (v0.16.2) was built for Go 1.22](https://github.com/kubernetes-sigs/controller-tools/blob/v0.16.2/go.mod#L3), which is now past its LTS support window. 
The newer version of controller-tools (v0.19.0) uses `golang.org/x/tools@v0.36.0`, which is [pinned to Go 1.24](https://github.com/kubernetes-sigs/controller-tools/blob/v0.19.0/go.mod#L3) (our current project requirement) and Go 1.25+.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,17 @@ KO_PUSH ?= true
 KO_LOCAL ?= true
 export KIND_CLUSTER_NAME ?= kro
 
+# Adapted from etcd-operator and k/k:
+# Set the GOTOOLCHAIN to force the toolchain defined in go.mod
+GOTOOLCHAIN ?= auto
+ifeq (auto,$(GOTOOLCHAIN)) # User didn't specify the GOTOOLCHAIN, or is set to auto.
+ifeq (,$(FORCE_HOST_GO)) # User didn't provide FORCE_HOST_GO, use go.mod's toolchain
+    export GOTOOLCHAIN=$(shell grep '^toolchain go' go.mod | cut -d' ' -f2)
+else # User provided FORCE_HOST_GO, use the local version
+    export GOTOOLCHAIN=local
+endif
+endif
+
 GIT_TAG ?= dirty-tag
 GIT_VERSION ?= $(shell git describe --tags --always --dirty)
 GIT_HASH ?= $(shell git rev-parse HEAD)


jakobmoellerdev:
hmm strange unit failure, maybe related to the cov tool its looking for. could be that were missing runner sh wrap like i introduced in the e2e kind test, but just a guess for now..

bschaatsbergen:
> hmm strange unit failure, maybe related to the cov tool its looking for. could be that were missing runner sh wrap like i introduced in the e2e kind test, but just a guess for now..

The build breaks due to a Go 1.25 toolchain switching [bug](https://github.com/golang/go/issues/75031) where auto-downloaded toolchains can't locate the `covdata` tool. If you look closely at the Prow logs you notice:
> go: no such tool "covdata" 

This is an upstream Go bug. We're able to work around it by forcing `GOTOOLCHAIN` to match the version in `go.mod`.

I suspect that when Go automatically downloads toolchains to resolve version mismatches between local environments and the `go.mod` `toolchain` directive, it fetches a minimal runtime that lacks certain tools, either due to a bug or by design (not sure which). In our case, the auto-downloaded Go 1.25.4 toolchain was missing the `covdata` tool, causing the unit test Prow job to fail with no such tool `"no such tool 'covdata'"` errors. By explicitly setting `GOTOOLCHAIN` to match the version specified in `go.mod`, we ensure that Go uses a complete, locally installed toolchain that includes all required tools.

The etcd-operator team, among others, encountered the same issue: https://github.com/etcd-io/etcd-operator/pull/196#issuecomment-3250392465

a-hilaly:
Nice sleuthing @bschaatsbergen ! Thank you so much!
/lgtm

PR #810, opened 2025-11-13
Delete unused metadata functions

tjamet:
There are several functions in `pkg/metadata/groupversion.go` that were left unused over time.

With this commit, we aim at giving better clarity about what is actually used.

In addition, there is a simplification opportunity with `GVKtoGVR` and `metadata.GetResourceGraphDefinitionInstanceGVk`

The only place where we used it was in the builder to then provide it to `GVKtoGVR`

In other terms, we were doing `gvr := metadata.GVKtoGVR(metadata.GetResourceGraphDefinitionInstanceGVK(group, apiVersion, kind))` which is strictly equivalent to `gvr := metadata.GetResourceGraphDefinitionInstanceGVR(group, apiVersion, kind)`

TO further clean unused `pkg/metadata` methods, use directly `gvr := metadata.GetResourceGraphDefinitionInstanceGVR(group, apiVersion, kind)` This helps identify that the `Plurilize` function is only used for RGD resources where users only exposes a `kind` that KRO declines into `singular` and `plural` to match Kubernetes' CRD

```
  names:
    plural: crontabs
    singular: crontab
    kind: CronTab
```

This helps solving issues like #683

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -465,7 +465,7 @@ func (b *Builder) buildInstanceResource(
 	// CRD declarations.
 
 	// The instance resource is a Kubernetes resource, so it has a GroupVersionKind.
-	gvk := metadata.GetResourceGraphDefinitionInstanceGVK(group, apiVersion, kind)
+	gvr := metadata.GetResourceGraphDefinitionInstanceGVR(group, apiVersion, kind)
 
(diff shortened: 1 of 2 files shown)


tjamet:
As shown below, there is no use of the deleted functions:
image

PR #813, opened 2025-11-13
fix: Add structural type compatibility checking for CEL validation

a-hilaly:
This patch implements structural type compatibility checking to resolve
false positive type errors in CEL expression validation. The core issue
is that CEL's `IsAssignable` performs nominal type checking, which
rejects structurally compatible types that have different names. This
becomes problematic in kro where we generate unique type names for each
resource's schema (e.g `__type_pod.spec.containers` vs
`__type_deployment.spec.template.speccontainers`), causing validation to
fail even when the underlying structure is identical.

In a nutshell, we introduces `AreTypesStructurallyCompatible`, a function
that performs deep structural comparison of CEL types using proper type
checking semantics. Wehn nominal type checking fails, we fall back to
structural checking to determine if types are truly incompatible. This
approach checks that the output type is a valid subset of the expected
type, meaning it can have fewer fields but cannot have extra fields or
mismatched types

cases handled:

- Primitive types: simple kind equality checking (string, int, bool,
  double...)
- List types: recursive element type compatibility checking with
  proper handling of nested arrays
- Map types: both key and value type compatibility with recursive
  validation
- Struct types: field by field compatibility using `DeclTypeProvider`
  for introspection, supporting subset semantics where output can omit
  fields but cannot add unexpected ones
- Nested structures: deep traversal through complex type hierarchies
  including arrays of structs and maps of structs
- Cross resource references: enables copying entire structs between
  resources (e.g `${pod.spec.containers}` into another pod's spec)

The parser now leaves `ExpectedType` nil for field descriptors, deferring
type resolution to the builder. The builder implements
`setExpectedTypeOnDescriptor` which walks through schema paths,
constructing proper type names that include the full path (e.g
`__type_schema.spec.ports.[at]idx`). This ensures that when we perform
structural checking, the `DeclTypeProvider` can correctly resolve and
introspect the type definitions.

Future work:

The current implementation works at the CEL type level, which requires
converting OpenAPI schemas to CEL types and then performing structural
comparison on the CEL representation. A more efficient approach would be
to implement schema-level structural comparison directly on the OpenAPI
`spec.Schema` types. This would avoid the conversion overhead and provide
more direct access to schema metadata like required fields, default
values, and validation constraintz. However, the current CEL based
approach is sufficient for our needs and integrates cleanly with the
existing validation infrastructure.

### DIFF
--- a/pkg/cel/compatibility.go
+++ b/pkg/cel/compatibility.go
@@ -1,0 +1,270 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cel
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/cel-go/cel"
+	apiservercel "k8s.io/apiserver/pkg/cel"
+)
+
+const (
+	// TypeNamePrefix is the prefix used for CEL type names when converting OpenAPI schemas.
+	// Used to namespace custom types and avoid conflicts with built-in CEL types.
+	// Example: "__type_schema.spec.ports" for a ports field in the schema resource.
+	TypeNamePrefix = "__type_"
+)
+
+// AreTypesStructurallyCompatible checks if two CEL types are structurally compatible,
+// ignoring type names and using duck-typing semantics.
+//
+// This performs deep structural comparison:
+// - For primitives: checks kind equality
+// - For lists: recursively checks element type compatibility
+// - For maps: recursively checks key and value type compatibility
+// - For structs: uses DeclTypeProvider to introspect fields and check all required fields exist with compatible types
+//
+// The provider is required for introspecting struct field information.
+// Returns true if types are compatible, false otherwise. If false, the error describes why.
+func AreTypesStructurallyCompatible(output, expected *cel.Type, provider *DeclTypeProvider) (bool, error) {
+	if output == nil || expected == nil {
+		return false, fmt.Errorf("nil type(s): output=%v, expected=%v", output, expected)
+	}
+
+	// Dynamic type is compatible with anything
+	if expected.Kind() == cel.DynKind || output.Kind() == cel.DynKind {
+		return true, nil
+	}
+
+	// Kinds must match
+	if output.Kind() != expected.Kind() {
+		return false, fmt.Errorf("type kind mismatch: got %q, expected %q", output.String(), expected.String())
+	}
+
+	switch expected.Kind() {
+	case cel.ListKind:
+		return areListTypesCompatible(output, expected, provider)
+	case cel.MapKind:
+		return areMapTypesCompatible(output, expected, provider)
+	case cel.StructKind:
+		return areStructTypesCompatible(output, expected, provider)
+	default:
+		// For primitives (int, string, bool, etc.), kind equality is enough
+		return true, nil
+	}
+}
+
+// areListTypesCompatible checks if list element types are structurally compatible.
+func areListTypesCompatible(output, expected *cel.Type, provider *DeclTypeProvider) (bool, error) {
+	outputParams := output.Parameters()
+	expectedParams := expected.Parameters()
+
+	// Both must have element type parameters
+	if len(outputParams) == 0 || len(expectedParams) == 0 {
+		if len(outputParams) != len(expectedParams) {
+			return false, fmt.Errorf("list parameter count mismatch: got %d, expected %d", len(outputParams), len(expectedParams))
+		}
+		return true, nil
+	}
+
+	// Recursively check element type compatibility
+	compatible, err := AreTypesStructurallyCompatible(outputParams[0], expectedParams[0], provider)
+	if !compatible {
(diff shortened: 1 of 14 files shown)


pkg/cel/compatibility.go
```diff
@@ -0,0 +1,236 @@
...
+// ignoring type names and using duck-typing semantics.
+//
+// This performs deep structural comparison:
+// - For primitives: checks kind equality
+// - For lists: recursively checks element type compatibility
+// - For maps: recursively checks key and value type compatibility
+// - For structs: uses DeclTypeProvider to introspect fields and check all required fields exist with compatible types
+//
+// The provider is required for introspecting struct field information.
+func AreTypesStructurallyCompatible(output, expected *cel.Type, provider *apiservercel.DeclTypeProvider) bool {
```
jakobmoellerdev: if you do struct based comparison im wondering if we should use our go native compat system instead. since we dont actually do ref comparison on CEL but the underlying type it might make sense to convert cel values and decltypes back to their go equivalent and compare on that.

pkg/cel/compatibility.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// Split the path into segments
+	segments := strings.Split(typePath, ".")
+	if len(segments) == 0 {
+		return nil
+	}
+
+	// Get the root name, stripping __type_ prefix if present
+	rootName := segments[0]
+	if strings.HasPrefix(rootName, "__type_") {
```
jakobmoellerdev: nit: this seems like a hack thats based on your previous type prefixing, maybe this should become a const or things will break

pkg/graph/builder.go
```diff
@@ -674,13 +687,102 @@ func extractDependencies(env *cel.Env, expression string, resourceNames []string
...
 
+// setExpectedTypeOnDescriptor sets the ExpectedType field on a FieldDescriptor by:
+// 1. Taking the path (e.g., "spec.containers[0].name")
+// 2. Stripping array indices (e.g., "spec.containers.name")
+// 3. Constructing type name with __type_ prefix
+// 4. Looking up schema for that path
+// 5. Converting schema to CEL type
+func setExpectedTypeOnDescriptor(descriptor *variable.FieldDescriptor, rootSchema *spec.Schema, resourceID string) {
+	// String templates always expect StringType - already set by parser
+	if !descriptor.StandaloneExpression {
```
jakobmoellerdev, line 726: feels weird to have 2 places where expectation types are set. is it worth unifying this to this place here?

jakobmoellerdev (commented):
acked as discussed, but needs a follow issue in mid-future to convert this back to json schema based comparison instead of CEL type comparison which is inherently not as accurate.

pkg/cel/compatibility.go
a-hilaly: Yep, i'm not super happy with this. I'll re-iterate.

bschaatsbergen:
https://github.com/kubernetes-sigs/kro/pull/817 fixes the failing verify-lint job by marking the invoked script as executable (chmod).

pkg/cel/compatibility.go
a-hilaly: I hesitated about using Go native comparison (I know we have a CEL to Go converstion function) but I think the current approach is equivalent. Converting to Go types and comparing there doesn't really buy us anything - we'd still be doing structural field comparison, just in a different layer. Please correct me if i'm missing something.

a-hilaly: What I'm thinking is maybe the best path forward is OpenAPI spec comparison instead of either CEL type comparison. the main benefit would be properly checking required fields - right now we're only checking that fields exist and types match

pkg/cel/compatibility.go
```diff
@@ -0,0 +1,269 @@
...
+	if !found {
+		return nil
+	}
+
+	// Walk through remaining path segments
+	for i := 1; i < len(segments); i++ {
+		segment := segments[i]
+
+		// Handle list element type (@idx) and map value type (@elem)
+		if segment == "@idx" || segment == "@elem" {
```
jakobmoellerdev, line 180: I believe this opinionates (rightfully so) on the decl type provider, so I believe it needs a bit of comment as this doesnt come from cel

pkg/graph/builder.go
```diff
@@ -672,13 +684,125 @@ func extractDependencies(env *cel.Env, expression string, resourceNames []string
...
+	descriptor.ExpectedType = getCelTypeFromSchema(schema, typeName)
+}
+
+// getCelTypeFromSchema converts an OpenAPI schema to a CEL type with the given type name
+func getCelTypeFromSchema(schema *spec.Schema, typeName string) *cel.Type {
+	if schema == nil {
+		return cel.DynType
+	}
+
+	declType := openapi.SchemaDeclType(schema, false)
```
jakobmoellerdev: I think this is an inconsistent schema decl type since we now have our own

pkg/graph/builder.go
```diff
@@ -753,6 +877,21 @@ func validateExpressionType(outputType, expectedType *cel.Type, expression, reso
 		return nil
 	}
 
+	// Try structural compatibility checking (duck typing)
+	if typeProvider != nil {
```
jakobmoellerdev: Do  we have a case where the type provider is nil?

pkg/graph/fieldpath/builder.go
```diff
@@ -28,18 +28,19 @@ func Build(segments []Segment) string {
...
-		}
-
 		if segment.Index != -1 {
 			b.WriteString(fmt.Sprintf("[%d]", segment.Index))
 			continue
 		}
 
-		if strings.Contains(segment.Name, ".") {
+		// Use bracket notation for field names with dots or empty names
+		if strings.Contains(segment.Name, ".") || segment.Name == "" {
```
jakobmoellerdev, line 37: good find on the index...

pkg/graph/builder.go
a-hilaly: Good catch!

PR #819, opened 2025-11-14
Delete unused kustomize configuration

tjamet:
The Kustomize configuration was automatically generated by kube-builder, but never maintained, as revealed in #732.

This may invite users try to use it but fail.

This commit aims removes possible interpretations that kro supports plain kustomizations.

It removes unnecessary kube-builder role generation (re-written in helm) and generated CRDs directly in the helm CRDs folder.


jakobmoellerdev (commented):
/lgtm

PR #820, opened 2025-11-14
Publish static kro manifests on release

tjamet:
This is an alternate implementation to #732
that addresses #729.

#732 was designed following standard industry practices, producing static manifests with `kustomize`. However, as raised during the review, there is a concern about the additional cognitive load and maintenance burden. We can notice it in argoCD where they maintain the [manifests](https://github.com/argoproj/argo-cd/blob/master/manifests/core-install.yaml) as well as the helm [chart](https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd) (even in a different repository).

This commit aims at providing static bundles to benefit from static manifests without the additional maintenance burden.
It keeps Helm as the source of truth for deployment description and picks a list of well-known values combination to generate static manifests out of it.

fixes: #729

### DIFF
--- a/scripts/ci/presubmits/e2e-raw-manifests-tests
+++ b/scripts/ci/presubmits/e2e-raw-manifests-tests
@@ -14,5 +14,5 @@ mkdir -p ./bin
 curl -Lo ./bin/kind https://kind.sigs.k8s.io/dl/v0.30.0/kind-linux-amd64
 chmod +x ./bin/kind
 
-# Helm based end to end tests, with helm defaults
-make test-e2e-kind KIND=./bin/kind
+# Raw manifests based end to end tests, with minimal installation and "aggregated" mode
+make test-e2e-kind-core-install-manifests KIND=./bin/kind


a-hilaly:
@tjamet looks like prow bot isn't happy with the commit message.

tjamet:
> @tjamet looks like prow bot isn't happy with the commit message.

I've rebased it should help

a-hilaly:
@tjamet Prow bot is still unhappy https://github.com/kubernetes-sigs/kro/pull/820#issuecomment-3533677973

tjamet:
> @tjamet Prow bot is still unhappy [#820 (comment)](https://github.com/kubernetes-sigs/kro/pull/820#issuecomment-3533677973)

removed `#` `@` from the commit message

a-hilaly (commented):
LGTM. i'll leave it up to @jakobmoellerdev to share his opinion on the new helm chart values.

tjamet:
> LGTM. i'll leave it up to @jakobmoellerdev to share his opinion on the new helm chart values.

I agree, it's not ideal but a trade-off I found to keep it simple while not introducing wrong information (i.e. managed-by: helm when we do `kubectl apply`)

I tried it with

```
# path: manifests/kustomization.yaml
resources:
- ./bundled/kro-core-install-manifests.yaml
```

```
make bundle-static-manifests RELEASE_VERSION=v0.6.0
kustomize build ./manifests
```

generated a set of manifests, including the CRD

a-hilaly:
@tjamet fully understand the tradeoff, maybe doing a sed/replace or some `yq`-fu could help keeping helm chart unaware of the kustomize usage?

tjamet:
I indeed thought about yq, but then, it's likely that it generates a non-human friendly yaml (the one I know generated json as a good wrapper around jq), even `grep -v` but then we risk breaking the yaml layout

And we probably want those manifests to be human readable.

Another thing I thought about is to write a simple go program (here we're pretty used to write anything that's more than 10 lines in go)  and then `go run ./cmd/generate-manifests` but it doesn't seem it's aligned with the practices of the repo :D 

trade-offs all the way

.gitignore
```diff
@@ -44,3 +44,4 @@ TODO.*
 node_modules/
 
 tools/lsp/client/out/
+manifests/bundled
```
jakobmoellerdev: I think you could just push the manifests under `rendered`. WDYT?

Makefile
```diff
@@ -44,6 +44,14 @@ LDFLAGS="-buildid= -X sigs.k8s.io/release-utils/version.gitVersion=$(GIT_VERSION
 
 WITH_GOFLAGS = GOFLAGS="$(GOFLAGS)"
 
+HELM_STATIC_MANIFESTS_FLAGS ?= --set metadata.includeHelmChart=false --set metadata.includeManagedBy=false --include-crds
+
+ifeq ($(shell uname -s),Darwin)
```
jakobmoellerdev, line 50: It was a matter of time when platform args come into the makefile... :'(

helm/values.yaml
```diff
@@ -7,6 +7,10 @@ fullnameOverride: ""
 additionalLabels: {}
   # app: kro
 
+metadata:
+  includeHelmChart: true
```
jakobmoellerdev, line 11: nit: probably could be named a bit better

```
.Values.metadata.labels.includeHelmChart
```

helm/values.yaml
```diff
@@ -7,6 +7,10 @@ fullnameOverride: ""
...
+  includeManagedBy: true
```
jakobmoellerdev, line 12: nit: ditto, nothing too major though

Makefile
tjamet: indeed, not the best but otherwise `make helm-package` and others fails on mac
Happy to drop if that's too noisy

helm/values.yaml
tjamet: I thought the same but then I thought it would be misleading like.
Users may interprete as 
> add label `includeHelmChart : true` to objects

tjamet:
> Im in general not too much in favor of generating static manifests without having them tested in E2E at least on a smoke test.
> 
> Apart from that, I dont really buy the whole "industry standard of generating manifests". Many public tools only offer helm charts. That said, I get the argument. If there is general agreement this can help the project, lets put it in.
> 
> HELM-wise all good. not too many concerns to be honest.

Happy to add some tests if needed

a-hilaly:
agreed on having some smoke tests for these manifests. (in ideal world we want smoke tests for our helm chart too)

Makefile
jakobmoellerdev: I think i wanna migrate to task midterm, so fine for me :+1:

tjamet: task? mind sharing more about it?

a-hilaly (approved):
👍 
/cc @jakobmoellerdev

Makefile
jakobmoellerdev: https://taskfile.dev/

Makefile
```diff
@@ -282,6 +319,20 @@ cli:
...
+deploy-kind: ko deploy-kind-helm ## Deploy kro to a kind cluster
+	@true
+
+# Default end to end tests uses helm deployments
 .PHONY: test-e2e-kind
-test-e2e-kind: deploy-kind
-	make test-e2e
\ No newline at end of file
+test-e2e-kind: deploy-kind-helm
+	@true
```
jakobmoellerdev: why do we need a true here and above? i dont think this should be required with makefiles since aliases dont need their own run

Makefile
```diff
@@ -252,19 +277,31 @@ install: manifests ## Install CRDs into the K8s cluster specified in ~/.kube/con
...
+deploy-kind-helm: ko start-kind
 	make install
 	# This generates deployment with ko://... used in image.
 	# ko then intercepts it builds image, pushes to kind node, replaces the image in deployment and applies it
 	${HELM} template kro ./helm --namespace kro-system --set image.pullPolicy=Never --set image.ko=true --set config.allowCRDDeletion=true | $(KO) apply -f -
 	kubectl wait --for=condition=ready --timeout=1m pod -n kro-system -l app.kubernetes.io/component=controller
 	$(KUBECTL) --context kind-${KIND_CLUSTER_NAME} get pods -A
 
+.PHONY: deploy-kind-%
+deploy-kind-%: export KO_DOCKER_REPO=kind.local
```
jakobmoellerdev, line 301: why do we have a parameterized deploy-kind for manifests, but helm is not one of these parameters? I can see that this would be nice but it should maybe be unified. WDYT?

manifests/variants.yaml
```diff
@@ -0,0 +1,29 @@
+variants:
+- name: kro-core-install-manifests
+  values:
+    rbac:
+      mode: aggregation
+- name: kro-unrestricted-manifests
+  values:
+    rbac:
+      mode: unrestricted
+      legacy: true
```
jakobmoellerdev: i feel like we should not even ship unrestricted versions. WDYT @a-hilaly

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -108,3 +118,69 @@ helm uninstall kro -n kro
...
 except for the ResourceGraphDefinition CRD and any other custom resources you may have
 created.
+
+## Installation with kubectl
+
+### Installation Steps
+
+#### Install kro using Kubectl
+
+Select the variant you want (see available variants in [prerequisites](#prerequisites))
```
jakobmoellerdev: do we have an overview of these variants somewhere? if not, maybe we should include them as part of the release (where you started including them) or listing them here

jakobmoellerdev (commented):
Well done, thanks so much for your effort. My suggestions are mostly focused on trying to keep the makefile and the variants maintainable. Gonna wait for your response on most of these but this should definetly land the next few days. terrific work :medal_military:

manifests/variants.yaml
a-hilaly: Our tutorial quick starter do use unrestricted mode but i think we'll only keep it helm for now. Okkey with removing unrestricted from kustomize manifests.

tjamet:
rendered docs:
image

image

image

manifests/variants.yaml
tjamet: unrestricted is gone

Makefile
tjamet: Do you mean being able to run `make deploy-kind-helm-with-prometheus` for example?

My take was rather that given manifests are the output of `helm template`, we should test all supported helm combinations. Happy to read your thoughts

website/docs/docs/getting-started/01-Installation.md
tjamet: I have included the table in the docs (under raw install) let me know if we need to futrher edit
Screenshot 2025-12-02 at 20 31 24

Makefile
tjamet: otherwise the target does not exist.
There needed to be a command.
Trying again it seems to work without `@true`. Removing it then

Makefile
jakobmoellerdev: hmm I was thinking that there should be a list of 

```
make deploy-kind-helm
make deploy-kind-manifests
```

and then there would be n permutations for all the various manifest bundles we support. for helm we already today opinionate on the values we set ever. WDYT?

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -23,7 +25,7 @@ Install the latest version:
 
 ```bash
 helm install kro oci://registry.k8s.io/kro/charts/kro \
-  --namespace kro \
+  --namespace kro-system \
```
jakobmoellerdev, line 28: why do we want kro-system? i feel like this opinionation should be a separate PR, feel to big of a change here.

test/e2e/chainsaw/check-arbitary-objects/chainsaw-test.yaml
```diff
@@ -9,6 +9,9 @@ spec:
   - name: install-rgd
     try:
     #description: Install the RGD that creates an Instance CRD
+    - apply:
```
jakobmoellerdev, line 12: if i understand correctly this should only be run when we are in the manifests run so my feeling is that this is not the correct place for this. It should probably be part of the end to end setup. WDYT?

jakobmoellerdev: nvm this actually has individual corrected permissions that also tests aggregation, this is good!

website/docs/docs/getting-started/01-Installation.md
a-hilaly: let not make "drastic" changes that could be considered breaking for existing users. Let's make this PR about publishing static manifests and leave changing the default namespace discussion to a seperate issue/slack thread.
/hold

jakobmoellerdev: While I concur, this will not surface as a drastic change for most because its a manual flag that will be controlled by deployment tooling anyhow and will not break folks.

a-hilaly: i'm thinking from the perspective of leases namespace changing too, which could cause two controllers to be running in parallel. Maybe not that big of a deal.

Looks like everyone wants kro-system, let's do it :)

/unhold

a-hilaly:
Awesome work, merci beaucoup @tjamet !!
/approve

Makefile
a-hilaly: I believe this can be addressed in a follow up PR. /cc @tjamet

a-hilaly:
leaving unhold to @jakobmoellerdev 
/hold

Makefile
```diff
@@ -256,19 +281,31 @@ install: manifests ## Install CRDs into the K8s cluster specified in ~/.kube/con
...
+
+.PHONY: deploy-kind-helm
+deploy-kind-helm: export KO_DOCKER_REPO=kind.local
+deploy-kind-helm: ko start-kind
 	make install
 	# This generates deployment with ko://... used in image.
 	# ko then intercepts it builds image, pushes to kind node, replaces the image in deployment and applies it
 	${HELM} template kro ./helm --namespace kro-system --set image.pullPolicy=Never --set image.ko=true --set config.allowCRDDeletion=true | $(KO) apply -f -
 	kubectl wait --for=condition=ready --timeout=1m pod -n kro-system -l app.kubernetes.io/component=controller
 	$(KUBECTL) --context kind-${KIND_CLUSTER_NAME} get pods -A
```
a-hilaly, line 298: @tjamet did you mean to change the `rbac.mode` to `aggregation` here?

Makefile
tjamet: Happy to add simple target to list available commands, it's in the end easy to get through yq:

```
cat manifests/variants.yaml | yq '.variants[]|.name'
```

Happy to create a follow-up PR

Makefile
tjamet: No, I didn't want to change the default helm install to avoid breaking development flows
Moving to dev using "aggregation mode" seems to be a change that could be discussed individually, WDYT?

tjamet:
> I really dont want this to be part of an existing e2e-test, it should be its own separate CI imho. This is simply due to the turnaround time of the CI. it would be quite long otherwise.

Fine for me, let's remove it (at least with the current proposal, we know test passes in CI)

Makefile
a-hilaly: I don't mind changing that for dev environment, as long as same test continue to pass. Happy to do that seperately

website/docs/docs/getting-started/01-Installation.md
```diff
@@ -11,7 +11,9 @@ Install kro on your Kubernetes cluster using Helm.
 
 ## Prerequisites
 
-- Helm 3.x installed
+Kro can be installed via Helm or raw manifests. Pick the solution that best fits you.
```
a-hilaly, line 14:
```suggestion
kro can be installed via Helm or raw manifests. Pick the solution that best fits you.
```

Issue #821, opened 2025-11-14
`x-kubernetes-preserve-unknown-fields` causes CEL Type checking failures on nested objects


jakobmoellerdev:
### Description

**Observed Behavior**:

`x-kubernetes-preserve-unknown-fields` causes CEL Type checking failures on nested objects

**Expected Behavior**:

Im expecting `x-kubernetes-preserve-unknown-fields` to allow optional `?` / unsafe object access and not break the RGD

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: examples.example.com
spec:
  group: example.com
  versions:
    - name: v1alpha1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                config:
                  type: object
                  x-kubernetes-preserve-unknown-fields: true
      additionalPrinterColumns: []
  scope: Namespaced
  names:
    plural: examples
    singular: example
    kind: Example
    shortNames:
      - ex
```

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: example-rgd
spec:
  schema:
    apiVersion: v1alpha1
    kind: Demo
    spec:
      name: string
  resources:
    - id: example
      template:
        apiVersion: example.com/v1alpha1
        kind: Example
        metadata:
          name: ${schema.spec.name}
        spec:
          config:
            foo: bar
    - id: configmap
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${example.metadata.name}
        data:
          myvalue: ${example.spec.config.?foo}
```

**Versions**:
- kro version: 0.6.0+
- Kubernetes Version (`kubectl version`): v1.34.0

**Involved Controllers**:
- Controller URLs and Versions (if applicable):

**Error Logs** (if applicable)**:

```
failed to validate node "configmap": failed to type-check template expression "example.spec.config.foo" at path "data.myvalue": ERROR: <input>:1:20: undefined field 'foo'     │
│        | example.spec.config.?foo                                                                                                                                                     │
│        | ...................^
```

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
[...]

PR #823, opened 2025-11-14
fix: loosen up static type checker unknown field handling

jakobmoellerdev:
Mainly makes sure we dont break folks who dont define CRDs with allow unknown properties. This will make us not as strict as the K8s API Server which enforces field ignores. KRO wants to be a bit more lenient here (even though the DAG parsing still produces a warning)

I have now decided to migrate to our own, more lenient port of the K8s API Server CEL Type Provider and Declarator. This is mainly different in that it sets a DeclType Metadata Attribute for unknown fields, which KRO can use to identify potential object candidates for dyn types

Introduced tests to validate behavior of `x-kubernetes-preserve-unknown-fields` in ResourceGraphDefinition and CustomResourceDefinition management.

fix https://github.com/kubernetes-sigs/kro/issues/821

### DIFF
--- a/test/integration/suites/core/unknown_fields_test.go
+++ b/test/integration/suites/core/unknown_fields_test.go
@@ -1,0 +1,198 @@
+package core_test
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/ptr"
+
+	krov1alpha1 "github.com/kubernetes-sigs/kro/api/v1alpha1"
+	"github.com/kubernetes-sigs/kro/pkg/testutil/generator"
+)
+
+var _ = Describe("Unknown Fields", func() {
+	var namespace string
+
+	BeforeEach(func(ctx SpecContext) {
+		namespace = fmt.Sprintf("test-%s", rand.String(5))
+
+		By("creating namespace")
+		Expect(env.Client.Create(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Name: namespace},
+		})).To(Succeed())
+	})
+
+	AfterEach(func(ctx SpecContext) {
+		By("cleaning up namespace")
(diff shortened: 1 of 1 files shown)


jakobmoellerdev:
we have now one failing unit test and also a possible panic because were no longer prepared for a failing runtime. this still needs a bit of work :/

pkg/graph/builder.go
```diff
@@ -710,18 +718,32 @@ func validateNode(resource *Resource, templatesEnv, schemaEnv *cel.Env, resource
...
 			if err != nil {
-				return fmt.Errorf("failed to type-check template expression %q at path %q: %w", expression, templateVariable.Path, err)
+				// we are on a dynamic expression - there are cases in which
+				// the type-checker will fail to type-check the expression,
+				// but the resource can still be resolved by our ecosystem.
+				//
+				// For example, if the object is not available in the schema
+				// due to x-kubernetes-preserve-unknown-fields,
+				// or when we have arbitrary object schema fields
+				outputType = cel.DynType
```
tjamet: Can we make it more granular?
I mean, err could be because syntax error and then we delay this error at runtime.

But, indeed, as we have a regression, it's better to accept and then see how to bring errors earlier

test/e2e/chainsaw/check-unknown-fields/allow-unknown-crd.yaml
```diff
@@ -0,0 +1,28 @@
...
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                config:
+                  type: object
+                  x-kubernetes-preserve-unknown-fields: true
```
tjamet, line 20: could we manage to get this in a unit test so we have faster feedback loops?

pkg/graph/builder.go
jakobmoellerdev: I removed this in favor of a much safer take that doesnt kill our static type checker

test/e2e/chainsaw/check-unknown-fields/allow-unknown-crd.yaml
jakobmoellerdev: 100% will add one tomorrow.

jakobmoellerdev: added into envtest core suite. Im gonna keep the e2e test for now since I believe its good to visualize conformance.

cmd/controller/main.go
```diff
@@ -188,9 +188,7 @@ func main() {
 		BurstLimit:      burstLimit,
 	}, set.Metadata(), set.RESTMapper())
 
-	resourceGraphDefinitionGraphBuilder, err := graph.NewBuilder(
```
barney-s: can we remove this formatting change ?

test/integration/suites/core/unknown_fields_test.go
a-hilaly, line 1: 👍

PR #824, opened 2025-11-16
Add printer column policy CRD changes

yashpal2104:
### API change

### What the policy does 

### Backwards compatibility


jakobmoellerdev:
/close

please keep your changes to https://github.com/kubernetes-sigs/kro/pull/790. if youre still having trouble you need to give us a proper diff so we can support you. currently the change list is too big.

yashpal2104:
> /close
> 
> please keep your changes to https://github.com/kubernetes-sigs/kro/pull/790. if youre still having trouble you need to give us a proper diff so we can support you. currently the change list is too big.

Yes I am having difficulty applying diff to my new branch I applied diff from main to the #790

PR #825, opened 2025-11-16
Add printer column policy CRD changes

yashpal2104:
### API change

### What the policy does 

### Backwards compatibility


config/crd/bases/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -0,0 +1,329 @@
+---
```
jakobmoellerdev: this currently doesnt exist in the codebase. I think you messed up the merge/rebase from your last PR again :D

yashpal2104: I'm trying my best to understand I did got diff main and applied the changes through git diff still I am not able to do it

jakobmoellerdev:
if you want, you can add me as contributor and I will help you cleanup the patch here.

yashpal2104:
Yeah sure, I have been a little busy currently will hop on this back but I'll add you as a contributor. Thanks so much for your help

yashpal2104:
@jakobmoellerdev @a-hilaly PTAL and merge this ASAP else more rebase conflicts might arise

PR #826, opened 2025-11-17
fix: remove RGD owner ID check to allow RGDs to adopt their CRD

michaelhtm:
Currently, when a user deletes an RGD, and recreates it, the new RGD
is no longer able to adopt the CRD the first RGD left dangling.

This change removes the RGD ID check, ensuring, if an RGD with the same
name as the one that created the CRD is created, it can adopt it.


pkg/metadata/labels.go
```diff
@@ -58,12 +58,10 @@ func IsKROOwned(meta metav1.Object) bool {
 // resources before calling this function.
 func HasMatchingKROOwner(a, b metav1.ObjectMeta) bool {
 	aOwnerName := a.Labels[ResourceGraphDefinitionNameLabel]
-	aOwnerID := a.Labels[ResourceGraphDefinitionIDLabel]
 
 	bOwnerName := b.Labels[ResourceGraphDefinitionNameLabel]
-	bOwnerID := b.Labels[ResourceGraphDefinitionIDLabel]
```
jakobmoellerdev, line 64: I believe if we have an existing UID that is conflicting we should log a warning that mentions this "adoption"

michaelhtm: Fixed

pkg/client/crd.go
```diff
@@ -205,12 +205,12 @@ func (w *CRDWrapper) waitForReady(ctx context.Context, name string) error {
 // ResourceGraphDefinition that is attempting to update it. This prevents conflicts
 // where multiple ResourceGraphDefinitions try to manage the same CRD, or where a
 // CRD created outside of KRO is accidentally overwritten.
-func (w *CRDWrapper) verifyOwnership(existingCRD *v1.CustomResourceDefinition, newCRD v1.CustomResourceDefinition) error {
+func (w *CRDWrapper) verifyOwnership(ctx context.Context, existingCRD *v1.CustomResourceDefinition, newCRD v1.CustomResourceDefinition) error {
```
a-hilaly: Instead of passing a context, we can return two booleans, `rgdNameMatch`, `rgdIDMatching` and let the caller decide what to do with this information

michaelhtm: I can change `HasMatchingKROOwner` to have it return `rgdNameMatch` and `rgdIDMatching`.
Maybe verifyOwnership can still get the context and log the ownership change?

jakobmoellerdev: +1 to that, lets do that.

jakobmoellerdev: maybe you can rename to `EvaluateKROOwner`

bschaatsbergen: Reviewing the call stack, `AllowMutation` would by my suggestion. The below reflects this and updates the function arguments to be consistent with the existing/desired terminology used throughout this package:

```go
// AllowMutation checks if an existing CRD may be updated.
//
// It allows mutation only when both:
//  1. The CRD is KRO-managed (metadata.IsKROOwned).
//  2. The CRD's recorded RGD owner matches the updating RGD (metadata.HasMatchingKROOwner).
//
// Otherwise it returns a conflict to prevent takeover.
func (w *CRDWrapper) AllowMutation(existingCRD *v1.CustomResourceDefinition, desiredCRD v1.CustomResourceDefinition) error {
	if !metadata.IsKROOwned(&existingCRD.ObjectMeta) {
		return fmt.Errorf("conflict detected: CRD %s exists and is not KRO-managed", existingCRD.Name)
	}
	if !metadata.HasMatchingKROOwner(existingCRD.ObjectMeta, desiredCRD.ObjectMeta) {
		return fmt.Errorf("conflict detected: CRD %s owned by different ResourceGraphDefinition", existingCRD.Name)
	}
	return nil
}
```

bschaatsbergen (approved):
Small nit on naming — see https://github.com/kubernetes-sigs/kro/pull/826#discussion_r2535444833. Also, maybe UID validation could be opt-in instead (something we could address in a future PR).

a-hilaly:
@michaelhtm @bschaatsbergen @jakobmoellerdev got inspired from all of you and pushed to this branch, planned for 0.6.2

jakobmoellerdev (commented):
/lgtm

PR #827, opened 2025-11-17
level-based topological sorting and execution

### LINKED ISSUE #822: Implement parallel DAG walking
Resource reconciliation currently happens sequentially, which creates a bottleneck for complex graphs. Many resources are independent of each other, yet they are processed one by one because the existing walker enforces sequential execution. This limits throughput and increases processing times unnecessarily.

A simpler approach could be to rely on levelled topological execution and push all resources at the same level into an apply set. Each apply set would contain only the nodes that are ready to be processed in parallel, while still respecting all dependency guarantees. This avoids introducing a fully separate concurrent DAG walker, but still allows independent resources to be applied concurrently within each level.

Before moving forward, it would be useful to discuss how people typically write RGDs and wire up dependencies. This will help us understand patterns and ensure that applying resources in parallel preserves expected behavior.

bschaatsbergen:
Closes #822

This change is about how kro processes ResourceGraphDefinition instances, moving from sequential resource-by-resource processing to level-based topological sorting with parallelized execution within each level, while maintaining a single ApplySet for the entire instance lifecycle.

The reconciliation logic currently processes resources in discrete topological levels rather than a flat sequential order. The DAG (`pkg/graph/dag/dag.go`) now uses `TopologicalSortLevels()` based on [Kahn's algorithm](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm), which returns `[][]string` representing levels where all resources within a level have no dependencies on each other, which allows us to do parallel execution within levels.

Instead of creating multiple ApplySets or applying resources individually, we maintain a single ApplySet instance for the entire ResourceGraphDefinition instance. The ApplySet follows the [kubectl ApplySet specification](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/3659-kubectl-apply-prune/README.md#design-details-applyset-specification), using GKNN-based (Group/Kind/Namespace/Name) inventory tracking stored in parent resource annotations.

The reconciliation for instances (`controller_reconcile.go:reconcileInstance`):
  - Creates one `ApplySet` at the start
  - Iterates through topological levels sequentially
  - For each level:
    - Processes all resources in parallel, adding them to the `ApplySet`
    - Applies the ApplySet without pruning (`Apply(ctx, false)`)
    - Synchronizes runtime state so later levels can reference earlier resource outputs via CEL
    - Validates all resources in the level are ready before proceeding
  - After all levels complete successfully, performs a final apply with pruning enabled (`Apply(ctx, true)`)

Resource deletion now mirrors the creation flow in reverse order, processing topological levels from bottom to top (last level to first). Resources within each level are deleted in parallel, leveraging the same concurrency patterns used during creation to improve deletion performance. The controller validates that all resources in a level are fully deleted before moving to the next level.

Reviewers, please focus on the parallelization logic, whether the mutexes I introduced are actually protecting state properly, and the potential impact of multiple calls to a single applyset on cluster mutations beyond level 1.


pkg/controller/instance/controller_reconcile.go
bschaatsbergen, line 1: Reviewers, please focus on the parallelization logic, whether the mutexes I introduced are actually protecting state properly, and the potential impact of multiple calls to a single applyset on cluster mutations beyond level 1.

pkg/applyset/applyset.go
```diff
@@ -251,14 +251,24 @@ type applySet struct {
 	currentLabels      map[string]string
 	currentAnnotations map[string]string
 
+	// mu guards all maps and sets in applySet.
+	// These fields are accessed and mutated from multiple goroutines during
+	// reconciliation, so the lock must be held for every read or write to
+	// avoid race conditions and ensure consistent state.
+	mu sync.Mutex
```
jakobmoellerdev, line 251: nit: I think the mutex should be the first level argument and protect literally every map as you advertise. protected by mu is not on all of them

pkg/controller/instance/controller_reconcile.go
```diff
@@ -189,60 +203,171 @@ func (igr *instanceGraphReconciler) reconcileInstance(ctx context.Context) error
...
+			return fmt.Errorf("failed to synchronize after level %d: %w", levelNum, err)
+		}
+
+		igr.log.V(1).Info("Level processed successfully", "level", levelNum)
+	}
+
+	// Prune resources that are no longer part of the desired state
+	// The ApplySet tracks inventory via GKNN in parent annotations
+	// We do a final apply with prune=true to clean up orphaned resources
+	pruneResult, err := aset.Apply(ctx, true)
```
jakobmoellerdev, line 291: I have an issue with prune and levels. Technically we need to prune in a very specific levelled order.

Let me visualize.

Assuming we have a set of resources `A,B,C,D` and they visualize to this graph:

`A->B->C->D`

Then the removal order for pruning must be ordered as 

`D->C->B->A`. Currently pruning does not respect this order. I think this is a pretty heavy issue that is even present right now as pruning doesnt respect topology.

I think this is partially covered by deleteResourcesInOrder in the delete branch but is ignored here.

pkg/controller/instance/controller_reconcile.go
```diff
@@ -252,63 +377,57 @@ func (igr *instanceGraphReconciler) reconcileInstance(ctx context.Context) error
...
-			igr.updateResourceReadiness(applied.ID)
-		}
+		return nil
 	}
 
-	if err != nil {
-		mark.ResourcesNotReady("failed to reconcile the apply set: %v", err)
-		return igr.delayedRequeue(fmt.Errorf("failed to apply/prune resources: %w", err))
-	}
+	// Use errgroup to process all resources in parallel
```
jakobmoellerdev: I think errgroup is a nice interface but doesnt have the best fit here. Errgroup cancels on first error but you want to continue with the level even if one request fails.

pkg/controller/instance/controller_reconcile.go
```diff
@@ -252,63 +377,57 @@ func (igr *instanceGraphReconciler) reconcileInstance(ctx context.Context) error
...
+	// Resources in the same level have no dependencies on each other
+	g, gCtx := errgroup.WithContext(ctx)
+	g.SetLimit(stdruntime.NumCPU()) // Limit parallelism to number of CPUs
```
jakobmoellerdev: we have a flag for concurrency limit that I think we should respect

pkg/controller/instance/controller_reconcile.go
```diff
@@ -361,64 +487,126 @@ func (igr *instanceGraphReconciler) handleInstanceDeletion(ctx context.Context)
 
 // initializeDeletionState prepares resources for deletion by checking their
 // current state and marking them appropriately.
-func (igr *instanceGraphReconciler) initializeDeletionState() error {
-	for _, resourceID := range igr.runtime.TopologicalOrder() {
-		if _, err := igr.runtime.Synchronize(); err != nil {
-			return fmt.Errorf("failed to synchronize during deletion state initialization: %w", err)
-		}
+func (igr *instanceGraphReconciler) initializeDeletionState(levels [][]string) error {
```
jakobmoellerdev, line 489: Super high level but why do we even need initializeDeletionState still? I was assuming that this state should be covered by a proper pruned diff.

pkg/graph/dag/dag.go
```diff
@@ -136,6 +136,84 @@ func (d *DirectedAcyclicGraph[T]) AddDependencies(from T, dependencies []T) erro
...
+//
+//	A -> C
+//	B -> C
+//	C -> D
+//
+// The result would be: [[A, B], [C], [D]]
+// where A and B can be processed in parallel, then C, then D.
+func (d *DirectedAcyclicGraph[T]) TopologicalSortLevels() ([][]T, error) {
+	// Calculate in-degrees (number of dependencies) for each vertex
+	inDegree := make(map[T]int)
```
jakobmoellerdev, line 155: currently  for each vertex in the current level, you scan all vertices to find dependents. This balloons runtime to O(V²) for dense cases. I dont think its a major issue but you can probably optimize for

```
reverse := make(map[T][]T)
for id, v := range d.Vertices {
    for dep := range v.DependsOn {
        reverse[dep] = append(reverse[dep], id)
    }
}

for _, vertex := range currentLevel {
    for _, dependentID := range reverse[vertex.ID] {
        inDegree[dependentID]--
        if inDegree[dependentID] == 0 {
            nextLevel = append(nextLevel, d.Vertices[dependentID])
        }
    }
}
```

That produces O(V + E) as intended

pkg/graph/dag/dag.go
```diff
@@ -136,6 +136,84 @@ func (d *DirectedAcyclicGraph[T]) AddDependencies(from T, dependencies []T) erro
...
+				}
+			}
+		}
+
+		currentLevel = nextLevel
+	}
+
+	// Check if all vertices were processed (if not, there's a cycle)
+	if processed != len(d.Vertices) {
+		hasCycle, cycle := d.hasCycle()
```
jakobmoellerdev: nit: This is non optimal. Kahn has implicit cycle detection and this is a very expensive op

pkg/runtime/runtime.go
```diff
@@ -191,25 +191,24 @@ func (rt *ResourceGraphDefinitionRuntime) ResourceDescriptor(id string) Resource
...
+	defer rt.mu.RUnlock()
+
+	if r, ok := rt.resolvedResources[id]; ok {
 		return r, ResourceStateResolved
 	}
-
-	// If not, can we process the resource?
-	resolved := rt.canProcessResource(id)
-	if resolved {
+	if rt.canProcessResource(id) {
```
jakobmoellerdev: you have quite a bunch of microchanges in this pr that make it tough to review. in this case for example the lock addition is fine but the other refactor messes up the diff quite a bit. maybe try to make this pr really as atomic as it needs to be as the changeset is huge

jakobmoellerdev (commented):
First Review from me (need to do a checkout and another proper one): Thanks for the work on this, I remember from our slack discussions that you were very invested in this over a few days :rocket:  

Most importantly: I would love you to especially look at the deletion / pruning case again, I DONT think we are good on this. Apart from that mostly style and nits to get the PR in easier as the changeset is large.

Also missing tests for the levelled apply

Thanks also for adopting Kahns algorithm as I proposed. I believe this is the right way to do levelled apply. I do wish that you would share a bit more of implementation intent in the code as if you dont know that people wont know why youre doing what youre doing.

Last but not least, a lot of this change will have implicit changes to the reconcile paths in the controller which is why we need to adjust docs and prep release notes. Not something to fix immediately but we need to prep some information for our users.

pkg/applyset/applyset.go
bschaatsbergen: Addressed in 4e3550ae094ea163d004bc3c2e39b5c9203497ac

pkg/controller/instance/controller_reconcile.go
bschaatsbergen: Addressed in 16accff98ce7a51365c929eb2bce1f299734b1eb

pkg/graph/dag/dag.go
bschaatsbergen: Addressed in 585907c97c66c98d93acd639af09552e2ad69201

pkg/graph/dag/dag.go
bschaatsbergen: Addressed in 1c9c2bbe124998f072560cf50e2e9530d1ebda95

pkg/runtime/runtime.go
bschaatsbergen: This particular microchange I've reverted in 11289f3ec5a4ebdb2384e75d8fd92b4844fcd6a6

pkg/controller/instance/controller_reconcile.go
bschaatsbergen: Thanks for catching this. You're right, pruning needs to respect reverse topological order to safely remove dependents before their dependencies. We'll need to get this _sorted out_ (pun intended).

I explored adding `PruneByLevels` to mirror the existing `deleteResourcesInOrder` logic, but the implementation is proving more complex than I want it to be, given I prefer to keep `applyset` changes minimal. Would appreciate another set of eyes on this. How would you approach integrating topological pruning without major refactoring?

jakobmoellerdev: TBH Im not sure if you can easily achieve this. I think fundamentally apply sets are not designed to be levelled, so the upstream package copy also isnt. anything we build with applysets would be on top of the already existing specification.

bschaatsbergen:
Thanks everyone for the discussion so far. After a productive conversation this morning with @jakobmoellerdev, we’ve decided to defer this change for now, hence I'll convert it to a draft and attempt to keep it up-to-date with the default branch for the time being.

At the moment, it’s challenging to integrate this into the controller due to current limitations in applyset and pruning. Before we can move forward, we need to address these areas, specifically ensuring that our apply and prune logic is designed with level-based topological sorting in mind.

To move this in the right direction, @jakobmoellerdev and I will be drafting a KREP over the coming weeks to outline the path forward.

For those who are waiting on this feature: thank you for your patience. We want to make sure we’re aligned on the _why_ before we commit to the _how_. This level of rigor is what keeps kro predictable, stable, and production-ready.

jakobmoellerdev:
I will take this know-how here and try to port it into our new runtime world. Thanks so much for all the effort here.

Since its such a large work I will co-attribute you in the upcoming changes but close this one out

/close

PR #829, opened 2025-11-18
perf: optimize isKROReservedWord function

kennygt51:
Replace linear search through reservedKeyWords.List() with direct set lookup using Has() method for better performance when validating reserved word.

ref: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Byte.Has


a-hilaly (approved):
Thank you @kennygt51 !
/ok-to-test

PR #833, opened 2025-11-18
fix: various fixes on static type analysis in our structural type checking

jakobmoellerdev:
this will mainly address these issues:

1. we started introducing a map any type in 0.6.2 that was not correct, it should have been dyn on additional properties. This is a regression imo
2. we allow `?` since ages but since 0.6.0 are forcing people to use `orValue` in all cases. This loosens this restriction because a lot of the times not having a default value is acceptable and the zero value can be derived from KRO. If it cannot, we will fail at runtime still.
3. we stopped allowing struct<>map assignments. this will make that possible again
4. we stopped assignments to structs that were defined by allow unknown, this loosens that up again

Long term we need to move away from struct comparison towards JSON schema based one but that would explode this fix even more as we need to introduce a custom decl type. I will do this in a separate PR.

Added significant amount of unit tests. PTAL if youre interested in new cases that are now (re)allowed

### DIFF
--- a/pkg/cel/compatibility.go
+++ b/pkg/cel/compatibility.go
@@ -37,6 +37,7 @@ const (
 // - For lists: recursively checks element type compatibility
 // - For maps: recursively checks key and value type compatibility
 // - For structs: uses DeclTypeProvider to introspect fields and check all required fields exist with compatible types
+// - For map → struct and struct → map compatibility if fields/keys are structurally compatible
 //
 // The provider is required for introspecting struct field information.
 // Returns true if types are compatible, false otherwise. If false, the error describes why.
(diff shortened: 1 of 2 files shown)


pkg/cel/compatibility.go
a-hilaly, line 1: Can we have some unit tests show two compatible types pass with optionals? thinking:
- optional(X) and X
- X and optional(X)
- optional(X) and optional(X)

pkg/cel/compatibility.go
```diff
@@ -219,6 +222,10 @@ func areStructFieldsCompatible(output, expected *apiservercel.DeclType, provider
 	if expected == nil {
 		return true, nil
 	}
+	// PreserveUnknownFields is set on the expected type, so everything we would pass from output would be okay
+	if expected.Metadata[XKubernetesPreserveUnknownFields] == "true" {
+		return true, nil
+	}
```
a-hilaly, line 234: Not sure i understand this. Do you have examples in mind?

pkg/cel/schemas.go
```diff
@@ -104,7 +105,7 @@ func SchemaDeclTypeWithMetadata(s common.Schema, isResourceRoot bool) *apiserver
 	case "object":
 		if s.AdditionalProperties() != nil && s.AdditionalProperties().Schema() != nil {
 			propsType := SchemaDeclTypeWithMetadata(s.AdditionalProperties().Schema(), s.AdditionalProperties().Schema().IsXEmbeddedResource())
-			mt := apiservercel.MapType
+			mt := apiservercel.NewMapType(apiservercel.StringType, apiservercel.DynType, math.MaxInt)
```
a-hilaly, line 108: what's the reasoning here

jakobmoellerdev: The apiserver cel type generates MapType based on `Any` by default but our entire runtime is based on Dyn. If I would continue generating based on Any assigneability defaults based on Dyn would not work on nested maps where additional properties are set to allow unknown

pkg/cel/compatibility.go
jakobmoellerdev: if expected is a DeclType of type struct, and has allow unknown properties set to true, then this metadata field is set, but we wont have any properties. Without this, we would try to assert fields into expected Fields but an unknown object has no field definitions so without this we would not allow an assign

pkg/cel/compatibility.go
```diff
@@ -63,6 +62,10 @@ func AreTypesStructurallyCompatible(output, expected *cel.Type, provider *DeclTy
 	case cel.StructKind:
 		return areStructTypesCompatible(output, expected, provider)
 	default:
+		// Kinds must match
+		if output.Kind() != expected.Kind() {
```
jakobmoellerdev, line 72: this change is incorrect. i did this originally to allow assign from maps<>structs and the other way around but I think this also accidentally allowed passing an output type that is not a map to areMapTypesCompatible. As such I think I might have to revert this but this will make the type checker a lot stricter. Probably json field comparison would be a lot better here

jakobmoellerdev: I have now included map <> struct conversions that are safely possible as part of the compatibility layer. one of these cases is also part of the integration test.

PR #846, opened 2025-11-21
docs: add examples for Secret use-case with base64 decoding

sk31337:


website/docs/examples/basic/optionals.md
```diff
@@ -57,4 +57,49 @@ spec:
...
+    status:
+      creationTimestamp: ${secret.metadata.creationTimestamp}
+  resources:
+    - id: test
+      externalRef:
+        apiVersion: v1
+        kind: Secret
+        metadata:
+          name: test
+          namespace: ""
```
jakobmoellerdev: why do you set an empty namespace?

website/docs/examples/basic/optionals.md
```diff
@@ -57,4 +57,49 @@ spec:
...
+  name: secret-transformation
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: test
+    spec:
+      prefix: string | default="test"
+      namespace: string | default="default"
+    status:
+      creationTimestamp: ${secret.metadata.creationTimestamp}
```
jakobmoellerdev: this is not needed for this example

website/docs/examples/basic/optionals.md
```diff
@@ -57,4 +57,49 @@ spec:
...
+apiVersion: kro.run/v1alpha1
+kind: ResourceGraphDefinition
+metadata:
+  name: secret-transformation
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: test
+    spec:
+      prefix: string | default="test"
```
jakobmoellerdev: you need none of these spec fields for the example

website/docs/examples/basic/optionals.md
```diff
@@ -57,4 +57,49 @@ spec:
...
+          namespace: ""
+    - id: secret
+      template:
+        apiVersion: v1
+        kind: Secret
+        metadata:
+          name: "${schema.spec.prefix}-transformed"
+          namespace: ${schema.spec.namespace}
+        stringData:
+          Token-url: "${ string(base64.decode(string(test.data.uri))) }/oauth/token"
```
jakobmoellerdev: lets keep it sane and name it token

jakobmoellerdev:
```suggestion
          token: "${ string(base64.decode(string(test.data.uri))) }/oauth/token"
```

website/docs/examples/basic/optionals.md
```diff
@@ -57,4 +57,49 @@ spec:
...
+```
+
+## Secret
+
+```yaml title="secret.yaml"
+apiVersion: v1
+kind: Secret
+metadata:
+  name: test
+  namespace: default
```
jakobmoellerdev: this namespace field is not needed because you anyhow use default

website/docs/examples/basic/optionals.md
```diff
@@ -4,7 +4,7 @@ sidebar_position: 104
 
 # Optional Values & External References
 
-
+## Config Map
```
jakobmoellerdev, line 7:
```suggestion
## Config Map

```

jakobmoellerdev:
please also add a quick description when people wanna know what you do here.

website/docs/examples/basic/optionals.md
sk31337: I need to set `namespace` otherwise there is an error.

Empty namespace -> https://kro.run/docs/concepts/resource-group-definitions#using-externalref-to-reference-objects-outside-the-resourcegraphdefinition

website/docs/examples/basic/optionals.md
sk31337: removed

website/docs/examples/basic/optionals.md
sk31337: removed

website/docs/examples/basic/optionals.md
sk31337: done

sk31337: done

website/docs/examples/basic/optionals.md
sk31337: removed

sk31337:
> please also add a quick description when people wanna know what you do here.

done

website/docs/examples/basic/optionals.md
jakobmoellerdev: It says very clearly that if empty the instance namespace is used. so what is the error? `# optional, if empty uses instance namespace`

are you forced to set the field to empty? if so thats a bug!

sk31337: I don't want to contribute an example which does not work at the moment:

```
Error from server (Invalid): error when creating "/Users/.../kro.yaml": ResourceGraphDefinition.kro.run "test" is invalid: [spec.resources[0].externalRef.metadata.namespace: Required value, <nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]
```

sk31337: fix -> https://github.com/kubernetes-sigs/kro/pull/847

Issue #842, opened 2025-11-20
Panic with CEL format func


t3mi:
### Description

Creation of the instance leads to controller panic. Using `${...}${...}` works fine.

**Observed Behavior**:
Controller panics

**Expected Behavior**:
Instance successfully created

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):
```yaml
# ResourceGraphDefinition
---
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: test-format
spec:
  schema:
    apiVersion: v1alpha1
    kind: TestFormat
  resources:
    - id: serviceAccount
      template:
        apiVersion: v1
        kind: ServiceAccount
        metadata:
          name: ${schema.metadata.name}

    - id: configMap
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.metadata.name}
        data:
          key: ${"%s:%s".format([schema.metadata.namespace, serviceAccount.metadata.name])}
```

```yaml
# Instance
---
apiVersion: kro.run/v1alpha1
kind: TestFormat
metadata:
  name: test-format
  namespace: default

```

**Versions**:
- kro version: 0.6.3
- Kubernetes Version (`kubectl version`): 1.33.2

**Error Logs** (if applicable)**:
```
2025-11-20T17:32:40.738Z	INFO	dynamic-controller	Starting dynamic controller
2025-11-20T17:32:40.895Z	INFO	rgd-controller	Starting workers	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition", "worker count": 1}
2025-11-20T17:32:40.899Z	DEBUG	rgd-controller	setting resourcegraphdefinition as managed	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition", "name": "test-format", "reconcileID": "c49b9d46-3c71-499e-bbd9-ef53a9534a35"}
2025-11-20T17:32:40.899Z	DEBUG	rgd-controller	reconciling resource graph definition graph	{"controller": "ResourceGraphDefinition", "controllerGroup": "kro.run", "controllerKind": "ResourceGraphDefinition", "name": "test-format", "reconcileID": "c49b9d46-3c71-499e-bbd9-ef53a9534a35"}
[...]

PR #849, opened 2025-11-21
fix: add list expression handling and format statement tests

jakobmoellerdev:
This commit enables the CEL AST inspector to properly parse and analyze list expressions (e.g., `[1, 2, 3]`), which it previously ignored. It adds a new handler for `Expr_ListExpr` that recursively inspects elements for dependencies and registers a synthetic `createList` function call for the list literal itself.

The commit also refactors the test suite to use stricter sorting for verifying dependencies and adds extensive test cases covering simple lists, nested lists, and lists containing function calls. Finally, it introduces a new integration test to verify that strings powered by these list expressions (like `"%s:%s".format([namespace, name])`) correctly resolve resource dependencies.

fix https://github.com/kubernetes-sigs/kro/issues/842

### DIFF
--- a/pkg/cel/ast/inspector_test.go
+++ b/pkg/cel/ast/inspector_test.go
@@ -16,7 +16,9 @@ package ast
 
 import (
 	"reflect"
+	"slices"
 	"sort"
+	"strings"
 	"testing"
 )
 
@@ -47,7 +49,6 @@ func TestInspector_InspectionResults(t *testing.T) {
 				{ID: "bucket", Path: "bucket.spec.name"},
 			},
 		},
-
 		{
 			name:       "bucket name with function",
 			resources:  []string{"bucket"},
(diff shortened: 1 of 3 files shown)


jakobmoellerdev:
Note that this only fixes the underlying parser. I will open a second fix to fixup the potential panics in case the runtime fails

pkg/cel/ast/inspector_test.go
```diff
@@ -57,7 +58,7 @@ func TestInspector_InspectionResults(t *testing.T) {
 				{ID: "bucket", Path: "bucket.name"},
 			},
 			wantFunctions: []FunctionCall{
-				{Name: "toLower"},
+				{Name: "toLower", Arguments: []string{"bucket.name"}},
```
a-hilaly, line 61: Thank you for catching those.

Issue #848, opened 2025-11-21
Generated OpenAPI Schema - `unknown CEL type: google.protobuf.Timestamp`


sk31337:
### Description

When using following simple schema, the generated OpenAPI schema includes `unknown CEL type: google.protobuf.Timestamp`
**Observed Behavior**:

```yaml
---
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: test
spec:
  schema:
    apiVersion: v1alpha1
    kind: Test
    spec:
      name: string | default="test"
    status:
    #  creationTimestamp: "${string(secret.metadata.creationTimestamp)}" # works
      creationTimestamp: "${secret.metadata.creationTimestamp}" # does not work
  resources:
    - id: secret
      template:
        apiVersion: v1
        kind: Secret
        metadata:
          name: "${schema.spec.name}"
        stringData:
          hello: "world"
---
apiVersion: kro.run/v1alpha1
kind: Test
metadata:
  name: test
  namespace:
    "default"

```

**Expected Behavior**:

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):

**Versions**:
- kro version: 0.6.3
- Kubernetes Version (`kubectl version`):v1.34.1

**Involved Controllers**:
- Controller URLs and Versions (if applicable):

**Error Logs** (if applicable)**:
Generated OpenAPI Schema on cluster
```yaml
schema:
      openAPIV3Schema:
        properties:
          apiVersion:
            type: string
          kind:
            type: string
          metadata:
            type: object
          spec:
            default: {}
            properties:
              name:
                default: test
                type: string
            type: object
          status:
            properties:
              conditions:
                items:
                  properties:
                    lastTransitionTime:
                      type: string
                    message:
                      type: string
                    observedGeneration:
                      type: integer
                    reason:
                      type: string
                    status:
                      type: string
                    type:
                      type: string
                  type: object
                type: array
              creationTimestamp:
                description: 'unknown CEL type: google.protobuf.Timestamp'
                x-kubernetes-preserve-unknown-fields: true
              state:
                type: string
            type: object
        type: object
```

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
[...]

PR #850, opened 2025-11-21
fix: allow properly referencing timestamp expressions

jakobmoellerdev:
Allows referencing timestamp values in simpleschema by adding a proper infer case. Tested and validated with unit and an integration test.

fix https://github.com/kubernetes-sigs/kro/issues/848


a-hilaly (approved):
👍 
/lgtm

PR #851, opened 2025-11-23
chore: add managedby label to kro owned resources

michaelhtm:
This change adds the kubernetes recommended managedby label to resources
kro owns. https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

This change overrides the managed by label if it already exists.

### DIFF
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -37,6 +37,9 @@ const (
 	OwnedLabel      = LabelKROPrefix + "owned"
 	KROVersionLabel = LabelKROPrefix + "kro-version"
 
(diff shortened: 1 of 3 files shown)


bschaatsbergen (approved):

pkg/metadata/labels.go
```diff
@@ -37,6 +37,9 @@ const (
 	OwnedLabel      = LabelKROPrefix + "owned"
 	KROVersionLabel = LabelKROPrefix + "kro-version"
 
+	ManagedByLabelKey = "app.kubernetes.io/managed-by"
```
jakobmoellerdev, line 40: I know im being very nitty here but would it be possible to mention this setting somewhere here? https://kro.run/docs/concepts/instances

Issue #629, opened 2025-08-14
allow new preserveDefaultPrinterColumns field to RGDs to augment existing additionalPrinterColumns field


GeraldLoeffler:
### Feature Description

**Problem Statement**:
Adding `additionalPrinterColumns` to an RGD hides all default printer columns. This is correct, if one knows that this field mirrors the field of the same name in standard CRDs, which behaves like that. However, to the uninitiated this behaviour is entirely unexpected given the name of the field, which implies **addition** and not **replacement**.

**Proposed Solution**:
Keep `additionalPrinterColumns` in the current form to preserve compatibility with standard CRDs. Add a separate field `preserveDefaultPrinterColumns` which, if set to `true` makes `additionalPrinterColumns` not **replace** but **add to** the printer columns.

**Alternatives Considered**:
One could add a nested field to `additionalPrinterColumns`, maybe called `preserveDefault`. But this would make  `additionalPrinterColumns` in RGDs and CRDs deviate.

**Additional Context**:
Currently, `additionalPrinterColumns` behaves like this:
When my RGD has no additionalPrinterColumns field, it prints like this:

```
NAME                                                 STATE      SYNCED   AGE
containerenvironment.kro.run/container-environment   DELETING   False    164m
```

But when I add this to the RGD:
```
    additionalPrinterColumns:
      - name: Static IP
        type: string
        jsonPath: .status.staticIp
```

then it prints like this:

```
NAME                                                 STATIC IP
containerenvironment.kro.run/container-environment   10.64.226.172
```

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

jakobmoellerdev:
I can  take a jab at tackling this if we gather a bit more interest here.

yashpal2104:
@jakobmoellerdev I can help you with this, just point me to the resources I'll try to help you as much as I can possible.

jakobmoellerdev:
Of course, happy to have you involved. You will need a few things here:

- Extend https://github.com/kro-run/kro/blob/main/api/v1alpha1/resourcegraphdefinition_types.go#L96 with the new field suggestion. I suggest you do NOT implement a bool as prepared in the issue but a Policy field: `additionalPrinterColumnPolicy` that can be set to `Replace`, `Add`, with `Replace` being the default, to not break existing behavior.
[...]

PR #860, opened 2025-11-24
feat: add AdditionalPrinterColumnPolicy to control CRD printer column…

yashpal2104:
… merging

Add AdditionalPrinterColumnPolicy field to Schema with two modes:
- Replace (default): User columns replace defaults; falls back to defaults if empty
- Add: Merges user columns with defaults; user columns override on Name/JSONPath match

This allows users to either fully control printer columns or extend the defaults, maintaining backward compatibility with existing RGDs.

Includes comprehensive unit and e2e tests for both policy modes.

### DIFF
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -16,7 +16,7 @@ package v1alpha1
 import (
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	runtime "k8s.io/apimachinery/pkg/runtime"
 )
 
 // ResourceGraphDefinitionSpec defines the desired state of ResourceGraphDefinition.
@@ -87,12 +87,9 @@ type Schema struct {
 	// Example: {"connectionName": "${database.status.connectionName}", "endpoint": "${service.status.loadBalancer.ingress[0].hostname}"}
 	Status runtime.RawExtension `json:"status,omitempty"`
 
-	// AdditionalPrinterColumns defines additional printer columns
-	// that will be passed down to the created CRD. If set, no
-	// default printer columns will be added to the created CRD,
-	// and if default printer columns need to be retained, they
-	// need to be added explicitly.
-	//
+	// AdditionalPrinterColumns is the list of user-specified printer columns
+	// to merge with the default columns (State, Ready, Age). User columns
+	// override defaults when they share the same Name.
 	// +kubebuilder:validation:Optional
 	AdditionalPrinterColumns []extv1.CustomResourceColumnDefinition `json:"additionalPrinterColumns,omitempty"`
 }
--- a/helm/crds/kro.run_resourcegraphdefinitions.yaml
+++ b/helm/crds/kro.run_resourcegraphdefinitions.yaml
@@ -162,11 +162,9 @@ spec:
                 properties:
                   additionalPrinterColumns:
                     description: |-
-                      AdditionalPrinterColumns defines additional printer columns
-                      that will be passed down to the created CRD. If set, no
-                      default printer columns will be added to the created CRD,
-                      and if default printer columns need to be retained, they
-                      need to be added explicitly.
+                      AdditionalPrinterColumns is the list of user-specified printer columns
+                      to merge with the default columns (State, Ready, Age). User columns
+                      override defaults when they share the same Name.
                     items:
                       description: CustomResourceColumnDefinition specifies a column
                         for server side printing.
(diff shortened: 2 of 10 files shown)


helm/crds/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -209,6 +223,18 @@ spec:
                       the SimpleSchema spec.
                     type: object
                     x-kubernetes-preserve-unknown-fields: true
+                  validation:
```
jakobmoellerdev: this is a leftover. you really need to rebase properly.

yashpal2104:
I am sorry but I don't know how can I add you as a collaborator. Didn't mean to create problems for you, but I thought creating fesh PR and cherry picking will solve the problem.
Please do let me know how do I add you as a collaborator.
Sorry for the trouble caused

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -16,7 +16,18 @@ package v1alpha1
 import (
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+)
+
+// AdditionalPrinterColumnPolicy defines how additional printer columns are applied to the generated CustomResourceDefinition (CRD).
+type AdditionalPrinterColumnPolicy string
```
jakobmoellerdev: leftover from spec change I believe

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,17 +97,30 @@ type Schema struct {
...
+	// Validation is a list of validation rules that are applied to the
+	// resourcegraphdefinition.
+	Validation []Validation `json:"validation,omitempty"`
+	// AdditionalPrinterColumnPolicy controls how custom printer columns are merged
+	// with defaults. Currently only "Add" is supported, which merges user columns
+	// with defaults (user columns override defaults by Name).
+	//
+	// +kubebuilder:validation:Enum=Add
+	// +kubebuilder:default=Add
+	AdditionalPrinterColumnPolicy AdditionalPrinterColumnPolicy `json:"additionalPrinterColumnPolicy,omitempty"`
```
jakobmoellerdev: should be removed as discussed in comm call

jakobmoellerdev (changes requested):

api/v1alpha1/resourcegraphdefinition_types.go
yashpal2104: is the policy not needed I thought as you said ( or ig amine said) this needs to be implemented as a policy can you pls confirm?

api/v1alpha1/resourcegraphdefinition_types.go
yashpal2104: sorry was not present in the call so the policy needs to be removed?

yashpal2104: sorry was not present in the call so the policy needs to be removed?

jakobmoellerdev: https://github.com/kubernetes-sigs/kro/issues/629#issuecomment-3666312810

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: https://github.com/kubernetes-sigs/kro/issues/629#issuecomment-3666312810

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: yes correct, if we only have one policy there is no need for a field anymore!

yashpal2104: so will we be using string instead of policy ("Add")? and also then we have to use string literals in tests instead of custom types

yashpal2104: Oh now I get it I think

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -86,17 +97,30 @@ type Schema struct {
 	// This allows you to surface important information from managed resources at the instance level.
 	// Example: {"connectionName": "${database.status.connectionName}", "endpoint": "${service.status.loadBalancer.ingress[0].hostname}"}
 	Status runtime.RawExtension `json:"status,omitempty"`
+	// Validation is a list of validation rules that are applied to the
+	// resourcegraphdefinition.
+	Validation []Validation `json:"validation,omitempty"`
```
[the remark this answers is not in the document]
jakobmoellerdev: +1 to that. please get rid of validation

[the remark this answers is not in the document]
yashpal2104: pls check the changes made

a-hilaly:
thanks for your contribution! we decided we no longer want to expose this parameter in the RGD. at least in v1alpha1.

PR #861, opened 2025-11-24
KREP-006: Propagation Control

ellistarn:
KREP-006 introduces `propagateWhen`, a per-resource mechanism to conditionally gate mutation as
changes propagate through the graph. Both `propagateWhen` and `readyWhen` are complementary and
bookend when mutation for a node in the graph can start and is considered complete.

### DIFF
--- a/docs/design/proposals/propagation-control.md
+++ b/docs/design/proposals/propagation-control.md
@@ -41,9 +41,9 @@ change.
 1. Add `propagateWhen` to `ResourceGraphDefinitionSpec` to control propagation to graph instances
 2. Add `propagateWhen` to `Resource` to control propagation to a resource in a graph instance
 3. Add `ready()` and `updated()` lifecycle methods available on all resources in CEL expressions
-4. Add built-in CEL functions `exponentiallyUpdated` and `linearlyUpdated` for propagation control
+4. Add built-in CEL functions `exponentiallyReady` and `linearlyReady` for propagation control
 5. The default `propagateWhen` for collections (including RGDs) is
-   `[exponentiallyUpdated(item, collection)]`
+   `[exponentiallyReady(item, collection)]`
 6. The default `propagateWhen` for individual resources is `[]`
 
 ### API Types
@@ -84,16 +84,16 @@ pods.filter(p, p.ready()).size() >= pods.size() * 0.8
 These functions are syntactic sugar over the lifecycle methods:
 
 ```cel
-// linearlyUpdated(item, collection, batchSize) -> bool
+// linearlyReady(item, collection, batchSize) -> bool
 // Item can proceed when its batch is reached
-linearlyUpdated(pod, pods, 3)
+linearlyReady(pod, pods, 3)
 
 // Equivalent to:
 indexOf(pod, pods) < (pods.filter(p, p.updated()).size() / 3 + 1) * 3
 
-// exponentiallyUpdated(item, collection) -> bool
+// exponentiallyReady(item, collection) -> bool
 // Item can proceed when exponential batch (1, 2, 4, 8...) is reached
-exponentiallyUpdated(pod, pods)
+exponentiallyReady(pod, pods)
 
 // Equivalent to (where u = pods.filter(p, p.updated()).size()):
 indexOf(pod, pods) < int(math.pow(2, math.ceil(math.log(double(u + 1)) / math.log(2.0))))
@@ -109,7 +109,7 @@ Variables are derived from the resource definition:
 - id: pods
   forEach:
     - pod: ${schema.spec.pods}
-  propagateWhen: ["exponentiallyUpdated(pod, pods)"]
+  propagateWhen: ["exponentiallyReady(pod, pods)"]
 ```
 
 **For RGD-level propagateWhen (instances of the RGD):**
@@ -121,7 +121,7 @@ from the schema's kind. If KRO were to support custom plurals, it would inherit 
 spec:
   schema:
     kind: Application # CRD generates singular: application, plural: applications
-  propagateWhen: ["exponentiallyUpdated(application, applications)"]
+  propagateWhen: ["exponentiallyReady(application, applications)"]
 ```
 
 ## Design Questions
@@ -189,14 +189,14 @@ propagations are in flight, the latest takes precedence and the graph is reevalu
 This is similar to Kubernetes Deployments, which pivot mid-rollout towards the desired state.
 
 Propagation control must reason about overlapping propagations when determining
-`exponentiallyUpdated` and `linearlyUpdated` built-in functions. When determining whether or not a
+`exponentiallyReady` and `linearlyReady` built-in functions. When determining whether or not a
 node in the graph can mutate, it's not enough to measure `readyWhen`, we must also consider whether
 or not a resource is outdated with respect to the latest desired state of the graph.
 
 Take the following double mutation:
 
 ```
-Graph: A → B, C, D (collection with linearlyUpdated)
+Graph: A → B, C, D (collection with linearlyReady)
 
 T1: Mutation   T2: Mutation Propagates   T3: Rollback Starts   T4: Rollback Propagates
       A'                A'                       A''                     A''
@@ -205,7 +205,7 @@ T1: Mutation   T2: Mutation Propagates   T3: Rollback Starts   T4: Rollback Prop
 
 ```
 
-Node D can propagate when node C is ready, according to the `linearlyUpdated` function. However,
+Node D can propagate when node C is ready, according to the `linearlyReady` function. However,
 node C is yet not updated to the latest version of the graph. Should `propagateWhen` allow us to
 update D to D' or D'' or should we wait for C' to reach C'' before propagating the change? This
 question fundamentally boils down to the number of concurrent propagations allowed in the graph.
@@ -277,8 +277,8 @@ spec:
     group: example.com
     kind: Application
   propagateWhen:
-    - exponentiallyUpdated(application, applications)
-    - linearlyUpdated(application, applications, 10)
+    - exponentiallyReady(application, applications)
+    - linearlyReady(application, applications, 10)
   resources:
     - id: deployment
       template: ...
@@ -343,7 +343,7 @@ spec:
       forEach:
         - stage: ${ schema.spec.stages }
       propagateWhen:
-        - "${ linearlyUpdated(stage, stages, 1) }" # Sequential rollout
+        - "${ linearlyReady(stage, stages, 1) }" # Sequential rollout
         - "${ !releaseBlockers.data[stage.value] }" # Per-stage blockers
       template:
         apiVersion: example.com/v1
@@ -384,7 +384,7 @@ spec:
         # Custom: wait for half to be ready before continuing
         - "${ pods.filter(p, p.ready()).size() >= pods.size() / 2 }"
         # And use exponential for the rollout
-        - "${ exponentiallyUpdated(pod, pods) }"
+        - "${ exponentiallyReady(pod, pods) }"
       template:
         apiVersion: v1
         kind: Pod
--- a/docs/design/proposals/propagation-control.md
+++ b/docs/design/proposals/propagation-control.md
@@ -1,0 +1,405 @@
+# KREP-006 Propagation Control
+
+## Summary
+
+KREP-006 introduces `propagateWhen`, a per-resource mechanism to conditionally gate mutation as
+changes propagate through the graph. Both `propagateWhen` and `readyWhen` are complementary and
+bookend when mutation for a node in the graph can start and is considered complete.
+
+## Motivation
+
+KREP-002 introduces Collections, a mechanism for managing resources based on data discovered at
+runtime. It's important to recognize that each ResourceGraphDefinition is itself implicitly a
+collection, where each instance of the ResourceGraphDefinition is a member of that collection. There
+are risks associated with any change to a ResourceGraphDefinition or the explicit collections
+defined within. For example, an organization that has used KRO to unify application deployment with
+an Application CRD risks cluster-wide impact from a bad change to the ResourceGraphDefinition. A
+ResourceGraphDefinition that loops over a collection of zones to deploy a set of zonal Deployments
+risks regional impact from a bad change in the deployment's configuration.
+
+Propagation risks are exacerbated by collections, but exist even for single instances of simple
+ResourceGraphDefinitions. Any change to production involves risk, which creates opportunities to
+manage that risk. This observation leads to us viewing propagation control for a single resource a
+base case for propagation control over a collection. As KRO takes an increasingly large role for
+deciding what is running in production, it must provide controls to decide when production can
+change.
+
+## Use Cases
+
+1. **Rate Controls**: As an administrator, I want to limit the rate of change to collections of
+   resources defined in my ResourceGraphDefinitions when the inputs to the graph change. As an
+   administrator, I want to limit the rate of change to the instances of a ResourceGraphDefinition
+   when the definition itself changes.
+2. **Time Controls**: As an administrator, I want to prevent changes from happening outside of
+   business hours, including evenings, weekends, and holidays.
+3. **Reactive Controls**: As an administrator, I want to prevent changes when an alarm fires.
+4. **Manual Controls**: As an administrator, I want to manually prevent changes while I root cause
+   an incident.
+
+## Proposed API and Behavior
+
+1. Add `propagateWhen` to `ResourceGraphDefinitionSpec` to control propagation to graph instances
+2. Add `propagateWhen` to `Resource` to control propagation to a resource in a graph instance
+3. Add `ready()` and `updated()` lifecycle methods available on all resources in CEL expressions
+4. Add built-in CEL functions `exponentiallyUpdated` and `linearlyUpdated` for propagation control
+5. The default `propagateWhen` for collections (including RGDs) is
+   `[exponentiallyUpdated(item, collection)]`
+6. The default `propagateWhen` for individual resources is `[]`
+
+### API Types
+
+```go
+type ResourceGraphDefinitionSpec struct {
+    // ... existing fields ...
+
+    // PropagateWhen defines CEL expressions that allow the object to be mutated when true
+    PropagateWhen []string `json:"propagateWhen,omitempty"`
+}
+
+type Resource struct {
+    // ... existing fields ...
+
+    // PropagateWhen defines CEL expressions that allow the object to be mutated when true
+    PropagateWhen []string `json:"propagateWhen,omitempty"`
+}
+```
+
+### Lifecycle Methods
+
+KRO exposes lifecycle state on resource objects via two methods, available in all CEL expressions:
+
+```cel
+pod.ready()    // true if readyWhen conditions are satisfied
+pod.updated()  // true if updated to the current graph generation
+```
+
+These methods enable custom propagation logic. For example, wait for 80% of pods to be ready:
+
+```cel
+pods.filter(p, p.ready()).size() >= pods.size() * 0.8
+```
+
+### Built-in CEL Functions
+
+These functions are syntactic sugar over the lifecycle methods:
+
+```cel
+// linearlyUpdated(item, collection, batchSize) -> bool
+// Item can proceed when its batch is reached
+linearlyUpdated(pod, pods, 3)
+
+// Equivalent to:
+indexOf(pod, pods) < (pods.filter(p, p.updated()).size() / 3 + 1) * 3
+
+// exponentiallyUpdated(item, collection) -> bool
+// Item can proceed when exponential batch (1, 2, 4, 8...) is reached
+exponentiallyUpdated(pod, pods)
+
+// Equivalent to (where u = pods.filter(p, p.updated()).size()):
+indexOf(pod, pods) < int(math.pow(2, math.ceil(math.log(double(u + 1)) / math.log(2.0))))
+```
+
+### Built-in CEL Variables
+
+**For resource-level propagateWhen (collections within a graph):**
+
+Variables are derived from the resource definition:
+
+```yaml
+- id: pods
+  forEach:
+    - pod: ${schema.spec.pods}
+  propagateWhen: ["exponentiallyUpdated(pod, pods)"]
+```
+
+**For RGD-level propagateWhen (instances of the RGD):**
+
+Variables are derived from the CRD's singular/plural names, which are generated automatically be KRO
+from the schema's kind. If KRO were to support custom plurals, it would inherit automatically.
+
+```yaml
+spec:
+  schema:
+    kind: Application # CRD generates singular: application, plural: applications
+  propagateWhen: ["exponentiallyUpdated(application, applications)"]
+```
+
+## Design Questions
+
+### Should propagateWhen be a separate concept from readyWhen?
+
+Yes.
+
+On each reconciliation, we compute a topological ordering of the DAG. As we move through the
+ordering, we check that the dependencies of the node are satisfied, ensure the desired state of the
+node, compute the `readyWhen` condition for the node, and when true, move to the next node in the
+ordering. When all nodes are ready, the graph evaluation is considered complete.
+
+This currently happens serially, such that only one node is propagating at a time. It's
+straightforward to extend this algorithm to operate in parallel, and necessary for propagation
+control to be meaningful. In essence, the existing serial approach is a simple method of propagation
+control, where only one resource is mutating at a time.
+
+We introduce a new concept, `propagateWhen`, which controls when a node's mutation can start. This
+is in contrast to `readyWhen` which controls when a node's mutation ends. It is essential for these
+concepts to be decoupled because they serve fundamentally different purposes in the graph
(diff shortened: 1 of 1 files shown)


docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+rollback example above, we would rely on the result of `propagateWhen` to prevent a bad change from
+continuing to progress through the graph. Multiple concurrent changes also results in increased
+mutation compared to single concurrent mutation, but we consider this an acceptable tradeoff. We
+could explore exposing this tradeoff to customers via a `propagationPolicy`, but we consider this
+out of scope for this KREP.
+
+Mechanically, supporting concurrent mutations will require new machinery in KRO. We defer the exact
+details of this discussion to the implementation phase, due to the magnitude of the change.
+
+Directionally, we could introduce a new `ResourceGraphRevision` CRD for each unique set of inputs to
```
jakobmoellerdev, line 233: resource graph revision will become absolutely critical for long term stability of kro and its graph reconciliation. 

BUT I believe the graph revision should be included in its own KREP in separation of the propagation policy. We need a stable implementation plan first and foremost. I love the ideas but considering kros state right now we need to start thinking how to get this in without causing significant breaks.

IMHO Both ApplySets and Static Type Eval have IMHO caused too many regressions because we didnt focus enough on test plans.

jamal:
I wanted to +1 this, it would be fantastic to have this! I'm currently iterating on how to solve a problem I have where propagation of RGD changes can cause some impact to developers. To explain my use case a bit, I'm using kro to manage ephemeral development environments (somewhat similar to what Tilt, DevSpace or Skaffold will let you do). This is targeting game developers/designers who don't have kubectl or work with infrastructure/backend at all, so I wanted to avoid requiring installing things like kubectl, or even giving them cluster access. The RGD deploys a set of services and their dependencies so that a developer can have an isolated environment to work on.

But, one of the issues I'm running into is that the developer may be actively working and have state / configuration on the service that gets lost when the pod gets replaced. I'm iterating on options from trying to persist that state (which would add a ton of complexity) or just controlling when the instance can be updated.

Anyhow, having a way to control propagation would make that a lot simpler to solve. I'm still debating on what the control would be but either time based (to try to do things outside normal working hours) or manually managed by the developer using the cli tool.

Thank you! Definitely looking forward to see how this evolves.

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+## Motivation
+
+KREP-002 introduces Collections, a mechanism for managing resources based on data discovered at
+runtime. It's important to recognize that each ResourceGraphDefinition is itself implicitly a
+collection, where each instance of the ResourceGraphDefinition is a member of that collection. There
+are risks associated with any change to a ResourceGraphDefinition or the explicit collections
+defined within. For example, an organization that has used KRO to unify application deployment with
+an Application CRD risks cluster-wide impact from a bad change to the ResourceGraphDefinition. A
+ResourceGraphDefinition that loops over a collection of zones to deploy a set of zonal Deployments
+risks regional impact from a bad change in the deployment's configuration.
```
barney-s, line 18: Just to clarify, is the controls proposed scoped to a single instance ? 
Or does it propose across instances of RGD ?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+5. The default `propagateWhen` for individual resources is `[]`
+
+**CEL Functions:**
+
+```cel
+// exponentiallyUpdated(collection, each) -> bool
+// Returns true when updated items grow exponentially: 1, 2, 4, 8, 16...
+// An item is considered updated when its generation annotation matches the graph revision generation
+exponentiallyUpdated(collection, each) =
+  size(collection.filter(i, i.metadata.annotations['kro.run/generation'] == string(schema.metadata.generation))) >=
```
barney-s: Performance consideration: `collection.filter(...)` iterates over the entire collection.
If `propagateWhen` is evaluated for *every* resource in the collection during a reconciliation loop, this logic becomes O(N^2).
For large collections (e.g. thousands of resources), this could be a performance bottleneck.

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+  ((each.index / batchSize) + 1) * batchSize
+```
+
+```go
+type ResourceGraphDefinitionSpec struct {
+    // ... existing fields ...
+
+    // PropagateWhen defines CEL expressions that allow the object to be mutated when true
+    PropagateWhen []string `json:"propagateWhen,omitempty"`
```
barney-s, line 56: Please clarify the semantics of the `[]string` slice.
Are these CEL expressions evaluated as a logical AND (all must be true) or logical OR (at least one must be true)?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+}
+
+type Resource struct {
+    // ... existing fields ...
+
+    // PropagateWhen defines CEL expressions that allow the object to be mutated when true
+    PropagateWhen []string `json:"propagateWhen,omitempty"`
```
barney-s, line 63: Consider adding a `FailurePolicy` field.
If a CEL expression fails to evaluate (e.g., division by zero, missing field, type error), should the propagation be allowed (FailOpen) or blocked (FailClosed)?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+deciding what is running in production, it must provide controls to decide when production can
+change.
+
+## Use Cases
+
+1. **Rate Controls**: As an administrator, I want to limit the rate of change to collections of
+   resources defined in my ResourceGraphDefinitions when the inputs to the graph change. As an
+   administrator, I want to limit the rate of change to the instances of a ResourceGraphDefinition
+   when the definition itself changes.
+2. **Time Controls**: As an administrator, I want to prevent changes from happening outside of
```
barney-s, line 33: For "Time Controls", will the CEL environment provide a safe and deterministic way to access the current timestamp (e.g., `time.now()`)?
This is often restricted in standard CEL environments to ensure determinism.

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+   business hours, including evenings, weekends, and holidays.
+3. **Reactive Controls**: As an administrator, I want to prevent changes when an alarm fires.
```
barney-s, line 35: For "Reactive Controls" (alarms), how is this data made available to the CEL context?
Does this rely on an external controller patching the RGD status, or is there a mechanism to query external metrics?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+4. **Manual Controls**: As an administrator, I want to manually prevent changes while I root cause
+   an incident.
+
+## Proposed API and Behavior
+
+1. Add `propagateWhen` to `ResourceGraphDefinitionSpec` to control propagation to graph instances
+2. Add `propagateWhen` to `Resource` to control propagation to a resource in a graph instance
+3. Add built-in CEL functions `exponentiallyUpdated` and `linearlyUpdated` for propagation control
+4. The default `propagateWhen` for collections (including RGDs) is `[exponentiallyUpdated]`
```
barney-s: what does including RGDs mean here ?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Motivation
+
+KREP-002 introduces Collections, a mechanism for managing resources based on data discovered at
+runtime. It's important to recognize that each ResourceGraphDefinition is itself implicitly a
+collection, where each instance of the ResourceGraphDefinition is a member of that collection. There
+are risks associated with any change to a ResourceGraphDefinition or the explicit collections
+defined within. For example, an organization that has used KRO to unify application deployment with
+an Application CRD risks cluster-wide impact from a bad change to the ResourceGraphDefinition. A
+ResourceGraphDefinition that loops over a collection of zones to deploy a set of zonal Deployments
```
barney-s, line 17: Are you suggesting an RGD instance would be used to define a multi regional/zonal resource changes ? 
Even if KRO were a mature product, this would still not be recommended.

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Design Questions
+
+### Should propagateWhen be a separate concept from readyWhen?
+
+Yes.
+
+On each reconciliation, we compute a topological ordering of the DAG. As we move through the
+ordering, we check that the dependencies of the node are satisfied, ensure the desired state of the
+node, compute the `readyWhen` condition for the node, and when true, move to the next node in the
```
barney-s, line 135: If iam not wrong, readyWhen is used to include a resource when computing a dependency. If the resource is not ready, it is not included and dependent resources will not be generated. Things still happen in parallel.

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+ordering. When all nodes are ready, the graph evaluation is considered complete.
+
+This currently happens serially, such that only one node is propagating at a time. It's
+straightforward to extend this algorithm to operate in parallel, and necessary for propagation
+control to be meaningful. In essence, the existing serial approach is a simple method of propagation
+control, where only one resource is mutating at a time.
+
+We introduce a new concept, `propagateWhen`, which controls when a node's mutation can start. This
+is in contrast to `readyWhen` which controls when a node's mutation ends. It is essential for these
```
barney-s, line 144: how does this control node mutation ending ? it just signals dependent resources that the values from this node is available for use.

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+ordering, we check that the dependencies of the node are satisfied, ensure the desired state of the
+node, compute the `readyWhen` condition for the node, and when true, move to the next node in the
+ordering. When all nodes are ready, the graph evaluation is considered complete.
+
+This currently happens serially, such that only one node is propagating at a time. It's
+straightforward to extend this algorithm to operate in parallel, and necessary for propagation
+control to be meaningful. In essence, the existing serial approach is a simple method of propagation
+control, where only one resource is mutating at a time.
+
+We introduce a new concept, `propagateWhen`, which controls when a node's mutation can start. This
```
barney-s, line 143: so if propagateWhen returns false, the current node is not created ?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+may choose to alert when ResourceGraphDefinitions remain in a propagating state for too long (see:
+[Status Conditions](#status-conditions)) and manually revert the desired spec to trigger the inverse
+propagation. We may explore automation to detect and revert changes in a higher order abstraction,
+but we defer this discussion to a future KREP.
+
+### Should we allow overlapping propagations?
+
+Probably.
+
+Changes can be made to the inputs of the graph while other changes are still propagating through.
```
barney-s, line 180: if we make a change to an RGD, existing propagations would see the latest RGD. How can we have multiple views of RGD in multiple propagations ?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+This is similar to Kubernetes deployments, which can be mutated mid-rollout. A common use case for
+this is Rollback, described above. A mutation is made to the graph, and then the inverse of the
+mutation is made before it completes. Overlapping propagations can be more complicated, with up to
```
barney-s, line 183: are propagations stateful. If so where is state stored ?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+`exponentiallyUpdated` and `linearlyUpdated` built-in functions. When determining whether or not a
+node in the graph can mutate, it's not enough to measure `readyWhen`, we must also consider whether
+or not a resource is outdated with respect to the latest desired state of the graph.
+
+Take the following double mutation:
+
+```
+Graph: A → B, C, D (collection with linearlyUpdated)
+
+T1: Mutation   T2: Mutation Propagates   T3: Rollback Starts   T4: Rollback Propagates
```
barney-s, line 201: effectively indication there is only one view or propagation at a given time ?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,325 @@
...
+node C is yet not updated to the latest version of the graph. Should `propagateWhen` allow us to
+update D to D' or D'' or should we wait for C' to reach C'' before propagating the change? This
+question fundamentally boils down to the number of concurrent propagations allowed in the graph.
+
+Kubernetes Deployments allow a single concurrent propagation. This is a reasonable position for KRO
+to take, but it is not without tradeoffs. With this approach, if propagations are made on an
+interval T and each propagation takes > T, then the tail of the topological ordering will become
+increasingly stale. One clear example of this is when using KRO to model software release pipelines.
+Given an RGD for `SoftwareReleaseEnvironment` and a pipeline that deploys this environment to many
+stages and regions, each of which are dependent on each other, it may take O(days) to propagate a
```
barney-s, line 218: again a reason why this may not be the right abstraction for such use cases. We may need to define what is scoped and what is not.

docs/design/proposals/propagation-control.md
barney-s: this is what my proposal talks about https://docs.google.com/document/d/1kmi9hXK7tF5JkBBT-FLNDnz5j37RtiXuRIQYYBiFbUE/edit?usp=sharing

barney-s (commented):
thanks for the proposal.

docs/design/proposals/propagation-control.md
ellistarn: Both.

chrisdoherty4:
I like the approach. 

KREP-0003 discusses decorators and how generally they should be considered singletons; however it doesn't enforce the constraint and acknowledges there could be unknown use cases where a schema can be provided and result in non-conflicting changes between decorator instances. Is it worth considering how propagation control works in that context or at least making it an explicit non-goal?

ellistarn:
> KREP-0003 discusses decorators and how generally they should be considered singletons; however it doesn't enforce the constraint and acknowledges there could be unknown use cases where a schema can be provided and result in non-conflicting changes between decorator instances. Is it worth considering how propagation control works in that context or at least making it an explicit non-goal?

I owe this community a doc on "singletons" or "Schemaless Graphs".

docs/design/proposals/propagation-control.md
ellistarn: This was just an example. Optimizing this in the impl.

docs/design/proposals/propagation-control.md
ellistarn: It's identical to the ReadyWhen semantics. All must pass (AND)

docs/design/proposals/propagation-control.md
ellistarn: Let's consider this in future scope. Fail closed is appropriate, I think.

docs/design/proposals/propagation-control.md
ellistarn: Agreed -- punting `time` to an external KREP. You can always drive it with a cronjob and externalRef.

docs/design/proposals/propagation-control.md
ellistarn: externalRefs

docs/design/proposals/propagation-control.md
ellistarn: I see KRO as inhabiting the space of terraform. Anything done in terraform is within KRO's scope.

docs/design/proposals/propagation-control.md
ellistarn: Not currently, but I believe we need to make them so. Without this, we can only support a single propagation at a time. I am deferring this work to a future KREP (ResourceGraphRevisions). I think this work stands on it's own, though.

docs/design/proposals/propagation-control.md
ellistarn: These are the same thing. The node is "ready" and no longer mutating, and dependents can continue.

docs/design/proposals/propagation-control.md
ellistarn: The current node exists in the graph, but the graph does not propagate.

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+manage that risk. This observation leads to us viewing propagation control for a single resource a
+base case for propagation control over a collection. As KRO takes an increasingly large role for
+deciding what is running in production, it must provide controls to decide when production can
+change.
+
+## Use Cases
+
+1. **Rate Controls**: As an administrator, I want to limit the rate of change to collections of
+   resources defined in my ResourceGraphDefinitions when the inputs to the graph change. As an
+   administrator, I want to limit the rate of change to the instances of a ResourceGraphDefinition
```
NicholasBlaskey, line 31: what is the method for overriding? Would it be updating the RGD to no longer have these propagate controls?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+}
+```
+
+### Lifecycle Methods
+
+KRO exposes lifecycle state on resource objects via two methods, available in all CEL expressions:
+
+```cel
+pod.ready()    // true if readyWhen conditions are satisfied
+pod.updated()  // true if updated to the current graph generation
```
ellistarn, line 73: I wonder if these should be collapsed. The both the linearlyReady and exponentiallyReday functions below use `updated`

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+// linearlyReady(item, collection, batchSize) -> bool
+// Item can proceed when its batch is reached
+linearlyReady(pod, pods, 3)
+
+// Equivalent to:
+indexOf(pod, pods) < (pods.filter(p, p.updated()).size() / 3 + 1) * 3
+
+// exponentiallyReady(item, collection) -> bool
+// Item can proceed when exponential batch (1, 2, 4, 8...) is reached
+exponentiallyReady(pod, pods)
```
cheeseandcereal, line 96: Should the exponential factor be exposed as an optional parameter instead of always being forced to `2.0`?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+2. **Time Controls**: As an administrator, I want to prevent changes from happening outside of
+   business hours, including evenings, weekends, and holidays.
+3. **Reactive Controls**: As an administrator, I want to prevent changes when an alarm fires.
+4. **Manual Controls**: As an administrator, I want to manually prevent changes while I root cause
+   an incident.
+
+## Proposed API and Behavior
+
+1. Add `propagateWhen` to `ResourceGraphDefinitionSpec` to control propagation to graph instances
+2. Add `propagateWhen` to `Resource` to control propagation to a resource in a graph instance
```
michaelhtm, line 42: can we define the CEL context `propagateWhen` will have for each scenario?

docs/design/proposals/propagation-control.md
ellistarn: i.e., what if ready returned false when the resource was itself ready, but the generation has moved on. Thoughts on modeling this @a-hilaly ?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+Per KREP-001, KRO's status conditions are as follows:
+
+```
+  Ready
+  ├─ GraphResolved - Runtime graph has been created and resources resolved
+  ├─ ResourcesReady - All resources in the graph are created and ready
+  └─ InstanceManaged - Instance finalizers and labels are properly set
+```
+
+Above, we assert that readiness and propagation are separate concepts, and thus we introduce a
```
NicholasBlaskey, line 255: this may tie into resource graph revisions, but what would be the way for a user to know if an instance had the RGD propagated to it? 

Is this check all or nothing or is there a way to have per instance understanding of what changes have been applied or not

docs/design/proposals/propagation-control.md
k1ranpk, line 42: Is it possible for the propagation to be stuck in a cycle between the different controls expressed across multiple resources and RGDs?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
+# KREP-006 Propagation Control
+
+## Summary
+
+KREP-006 introduces `propagateWhen`, a per-resource mechanism to conditionally gate mutation as
+changes propagate through the graph. Both `propagateWhen` and `readyWhen` are complementary and
```
JamesGAWS, line 6: How will propagateWhen interact with includeWhen? If an RGD update changes an includeWhen condition to remove or add a node from a graph, is the node removal controlled by propagateWhen?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+```cel
+// linearlyReady(item, collection, batchSize) -> bool
+// Item can proceed when its batch is reached
+linearlyReady(pod, pods, 3)
+
+// Equivalent to:
+indexOf(pod, pods) < (pods.filter(p, p.updated()).size() / 3 + 1) * 3
+
+// exponentiallyReady(item, collection) -> bool
+// Item can proceed when exponential batch (1, 2, 4, 8...) is reached
```
JamesGAWS, line 95: How are instance batched for update? Are they chosen randomly per RGD generation or is the ordering fixed between generations?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+  name: cautious-rollout
+spec:
+  schema:
+    apiVersion: v1
+    group: example.com
+    kind: CautiousApp
+  resources:
+    - id: pods
+      forEach:
+        - pod: ${ schema.spec.pods }
```
ellistarn, line 382: This needs a revisit. Cannot reference schema.spec.pods

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+[Status Conditions](#status-conditions)) and manually revert the desired spec to trigger the inverse
+propagation. We may explore automation to detect and revert changes in a higher order abstraction,
+but we defer this discussion to a future KREP.
+
+### Should we allow overlapping propagations?
+
+Probably.
+
+Changes can be made to the inputs of the graph while other changes are still propagating through.
+This is similar to Kubernetes deployments, which can be mutated mid-rollout. A common use case for
```
JamesGAWS, line 181: Deployments use an intermediate object (replica sets) between deployments and pods. Does KRO needs a similar construct to allow overlapping propagations?

docs/design/proposals/propagation-control.md
```diff
@@ -0,0 +1,405 @@
...
+   when the definition itself changes.
+2. **Time Controls**: As an administrator, I want to prevent changes from happening outside of
+   business hours, including evenings, weekends, and holidays.
+3. **Reactive Controls**: As an administrator, I want to prevent changes when an alarm fires.
+4. **Manual Controls**: As an administrator, I want to manually prevent changes while I root cause
+   an incident.
+
+## Proposed API and Behavior
+
+1. Add `propagateWhen` to `ResourceGraphDefinitionSpec` to control propagation to graph instances
```
michaelhtm, line 41: in what order would we process the instances?

docs/design/proposals/propagation-control.md
ellistarn: If you specify `propagateWhen: []`, then there are no gates.

docs/design/proposals/propagation-control.md
ellistarn: I was imagining an annotation on the resource the shows a graph revision. And then KRO is managing those revisions internally.

docs/design/proposals/propagation-control.md
ellistarn: My understanding is that includeWhen removes the node from the graph, so it would not be counted in the propagation math. If it resolves to true, it is included in the graph and is taken into account. Though I suppose order matters here -- if it poofs into existence, we probably want to start with it, depending on the order.

This may be a problem to solve with resource graph revisions

docs/design/proposals/propagation-control.md
ellistarn: Yes! ResourceGraphRevision

ellistarn:
Playing a little bit with CEL expressions, I think this might be a simpler way to model linear/exponential rollout:

```yaml
# Exponential rollout — budget doubles each wave
- id: deployments
  forEach:
    app: ${apps}
  propagateWhen:
    - >-
      ${deployments.filter(d, d.updated() && !d.ready()).size()
       < max(1, deployments.filter(d, d.updated() && d.ready()).size())}
  template: ...
```

```yaml
# Linear rollout — 2 at a time
- id: deployments
  forEach:
    app: ${apps}
  propagateWhen:
    - ${deployments.filter(d, d.updated() && !d.ready()).size() < 2}
  template: ...
```

cc: @chrisdoherty4

Issue #862, opened 2025-11-25
rgd-controller: Observed a panic - with nested RGD


grodzik:
### Description

Hello,
I'm trying out Kro and as a first step for more advanced setup I attempted to create some basic RGD and extend it with another RGD. This failed with Panic error in Kro controller. This seems similar to #852 however, I don't use if statement to conditionally include a resource, and panic seems to be in reconcile controller.

**Observed Behavior**:
Panic error in RGD controller and CRD not getting applied/deployed. ResourceGraphDefinition is created, however doesn't get to Active state

**Expected Behavior**:
ResourceGrapghDefinition being successfully deployed and CRD created

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):
1. Deploy BasicApp RGD:
```yaml
---
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: basic-app
spec:
  schema:
    apiVersion: v1alpha1
    kind: BasicApp
    spec:
      name: string
      image: string | default="nginx"
      replicas: integer | default=3
    status:
      deploymentConditions: ${deployment.status.conditions}
      availableReplicas: ${deployment.status.availableReplicas}
  resources:
    - id: deployment
      template:
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: ${schema.spec.name}
        spec:
          replicas: ${schema.spec.replicas}
          selector:
            matchLabels:
              app: ${schema.spec.name}
          template:
            metadata:
              labels:
                app: ${schema.spec.name}
            spec:
              containers:
                - name: ${schema.spec.name}
                  image: ${schema.spec.image}
                  ports:
                    - containerPort: 80
```

2. Deploy AdvancedApp RGD:**Versions**:
```yaml
---
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: advanced-app
spec:
  schema:
    apiVersion: v1alpha1
    kind: AdvancedApp
    spec:
      name: string
      image: string | default="nginx"
      replicas: integer | default=3
  resources:
    - id: rgddep
      template:
        apiVersion: kro.run/v1alpha1
        kind: BasicApp
        metadata:
          name: my-advanced-app-${schema.spec.name}
        spec:
          replicas: ${schema.spec.replicas}
          name: ${schema.spec.name}
          image: ${schema.spec.image}
```

**Kro and Kubernetes versions**:
- kro version: 0.6.3
[...]

PR #864, opened 2025-11-25
fix: handle nil schemas in additional properties and add RGD reference test

jakobmoellerdev:
Ensure `AdditionalProperties` handles nil schemas without panicking by adding an upstream bug workaround. Introduce integration test for resource graph definitions referencing other RGD statuses to prevent regressions.

fixed by Joe in https://github.com/kubernetes/apiserver/commit/363607a41032472134a2a902958f3f1fdce6047d
will remove once we have dep on k8s >0.34.2 with this fix

fixes https://github.com/kubernetes-sigs/kro/issues/862

### DIFF
--- a/pkg/cel/schemas.go
+++ b/pkg/cel/schemas.go
@@ -22,9 +22,9 @@ import (
 
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/common/types"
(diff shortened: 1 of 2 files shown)


jakobmoellerdev:
whoopsie, classic forgot to remove focus in ginkgo.

PR #866, opened 2025-11-26
KREP-007: Horizontal scaling in KRO (Scale subresource + RGD Set)

tjamet:
Integrate KRO with the Kubernetes autoscaling ecosystem by exposing the Scale subresource on RGDs, enabling `kubectl scale`, HPA, and external autoscalers like keda.

- Treat each RGD instance as a “unit of deployment”; rely on graph resolution for intra-unit ordering and propagation without progressive rollout inside a single instance.
- Introduce an optional Set abstraction to replicate RGD instances and control rollout across replicas (ordering, concurrency, readiness), reducing blast radius during updates.
- Simplify user workflows with familiar tooling and declarative controls while improving operator observability and separation of concerns.
- Clearly scopes initial functionality (replica-count scaling with randomized names) an outlines future paths without coupling rollout logic into the core RGD reconciler.

### DIFF
--- a/docs/design/proposals/horizontal-scaling.md
+++ b/docs/design/proposals/horizontal-scaling.md
@@ -1,0 +1,223 @@
+# Horizontal scaling in KRO
+
+This proposal complements [KREP-002] and defines:
+- An interface to control `ResourceGraphDefinition` (RGD) instance scale using the Kubernetes “scale subresource”.
+- An opt-in mechanism to replicate RGD instances and declaratively control rollout across replicas.
+
+It is inspired by Kubernetes Set controllers (for example, `ReplicaSet`) to achieve safe, observable, and controllable rollouts.
+
+## Goals
+
+- Integrate KRO with the Kubernetes horizontal scaling ecosystem (for example, `kubectl scale`, [HPA], and [keda]) via the [scale-subresources].
+- Keep a single RGD instance as a “unit of deployment”; KRO’s graph resolution ensures correct ordering and propagation within that unit.
+- Provide an optional “Set” abstraction to replicate RGD instances and control rollout across replicas.
+
+## Non-goals
+
+- Implement all possible scaling or rollout strategies.
+- Replace or reimplement HPA/VPA behavior.
+- Manage progressive rollout within a single RGD instance. The Set abstraction controls rollout across replicas only.
+
+## Background and problem statement
+
+In [KREP-002] we introduced looping/collections. Collections may be required for:
+- Different environments (for example, private vs. public ingresses).
+- Different customers (duplicated stacks, different hostnames).
+- Horizontal scale based on load (request rate, queue depth, number of replicas, and so on).
+
+Required replica counts can vary rapidly. Kubernetes already provides a rich ecosystem (`kubectl scale`, [HPA], [keda]). KRO should integrate with this ecosystem.
+
+Replicas can be:
+- A single resource (for example, a Pod or a Deployment), or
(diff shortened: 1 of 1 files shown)


docs/design/proposals/horizontal-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+This proposal proceeds in two phases:
+
+1) Scale subresource on RGD (integration with the Kubernetes scaling ecosystem).
+2) RGD Set controller (replicate RGD instances and control rollout across replicas).
+
+### Phase 1: Scale subresource on RGD
+
+Users add explicit schema annotations on RGD fields to expose the Scale subresource:
+- `| scale` on integer fields to represent desired/current replicas.
+- `| scaleSelector` on a status string field that holds a label selector for Pods (used by HPA/VPA).
```
ellistarn, line 47: Firstly, this KREP is really cool.

I wonder a bit about the proliferation of keywords. WDYT about building this concept to be a bit more aware of kubernetes subresources? e.g.



https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource

```
integer | subresource:autoscaling/v1.Scale:specReplicasPath
integer | subresource:autoscaling/v1.Scale:statusReplicasPath
map[string]string | subresource:autoscaling/v1.Scale:labelSelectorPath
```

docs/design/proposals/horizontal-scaling.md
```diff
@@ -0,0 +1,223 @@
...
+
+```yaml
+apiVersion: kro.run/v1alpha1
+kind: ResourceGraphDefinition
+schema:
+  spec:
+    desiredReplicas: "integer | scale"
+  status:
+    replicas: "integer | scale"
+    podSelector: "string | scaleSelector"
```
ellistarn, line 64: Noticing that a podSelector is not a string, it's a map[string]string. How would you model this?

tjamet: In case of the scale API, it seems to be the serialized version of a selector
At least this is what is pointed by the [doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource)

```
metadata:
  name: crontabs.stable.example.com
spec:
  group: stable.example.com
  versions:
    - name: v1
[...]
            status:
              type: object
              properties:
                replicas:
                  type: integer
                labelSelector:
                  type: string
      # subresources describes the subresources for custom resources.
      subresources:
        # status enables the status subresource.
        status: {}
        # scale enables the scale subresource.
        scale:
          # specReplicasPath defines the JSONPath inside of a custom resource that corresponds to Scale.Spec.Replicas.
          specReplicasPath: .spec.replicas
          # statusReplicasPath defines the JSONPath inside of a custom resource that corresponds to Scale.Status.Replicas.
          statusReplicasPath: .status.replicas
          # labelSelectorPath defines the JSONPath inside of a custom resource that corresponds to Scale.Status.Selector.
          labelSelectorPath: .status.labelSelector
[...]
```

docs/design/proposals/horizontal-scaling.md
tjamet: It sounds a good direction to take a look at indeed.
Happy to get your syntax proposal in this krep

jakobmoellerdev:
/close

at the moment it does not seem like there is traction on this proposal and we do not see the proposal developing into that direction, closing it out here unless you wanna take a new stab at or revisit it

Issue #786, opened 2025-11-10
Add examples of Multi-tenant SaaS application


kennygt51:
### Feature Description

**Problem Statement**:
The examples collection lacks a **Multi-tenant SaaS** example, which is a common and complex use case in modern cloud-native applications. I would say, a lot of developers need guidance on implementing secure tenant isolation, resource management, RBAC, and scalability patterns using kro ResourceGraphDefinitions.

**Proposed Solution**:
Add a new example at `examples/kubernetes/multi-tenant-saas/` with:

- **TenantEnvironment RGD**: Creates isolated namespaces with RBAC, ResourceQuotas (Basic/Premium plans), and NetworkPolicies
- **SaaSApplication RGD**: Deploys applications with auto-scaling, health checks, and ingress configuration
- **Sample instances**: Demonstrates different service plans and tenant configurations
- **Comprehensive documentation**: Quick start guide, architecture diagrams, and troubleshooting

Key features:
- Complete tenant isolation using Kubernetes namespaces
- Production-ready security (RBAC, NetworkPolicies)
- Plan-based resource management
- Auto-scaling integration

File structure:
```
examples/kubernetes/multi-tenant-saas/
├── README.md
├── tenant-environment.yaml
├── saas-application.yaml
├── tenant-instance.yaml
└── app-instance.yaml
```

_This is just an example and may change during implementation_

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

kennygt51:
If I get permission to implement it, I’ll create a PR myself

a-hilaly:
Please go for it!

/assign @kennygt51

PR #874, opened 2025-11-28
chore: Add K8s examples to doc

kennygt51:
The documentation examples have not been updated, so the examples added #786 cannot be referenced from the website.

Modify script to add Kubernetes examples, then execute it and include the results.

The following is the result of building the website in the local environment.

image

### DIFF
--- a/website/versioned_docs/version-0.7.0/examples/kubernetes/job-example.md
+++ b/website/versioned_docs/version-0.7.0/examples/kubernetes/job-example.md
@@ -1,0 +1,65 @@
+---
+sidebar_position: 708
+---
+
+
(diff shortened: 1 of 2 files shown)


kennygt51:
> do you mind also backporting this to 0.7.0 docs? thanks!

Thank you for your review.
I've made the corretcions you pointed out.

Below image is local development.

The following is the result of building the website in the local environment.

image

a-hilaly (approved):
Thanks @kennygt51 !
/ok-to-test
/approve

jakobmoellerdev (commented):
/lgtm

PR #875, opened 2025-11-28
Update FAQ5 in README to match the website FAQ content.

a-hilaly:
Replace "Can I use this in production?" with "What should I know about API changes?" to align with the official documentation. in https://kro.run/docs/faq

cc @jlbutler @bridgetkromhout @nicslatts @justinsb


jlbutler:
oh good catch, we shouldn't have mismatch between readme and website. 

/lgtm

PR #877, opened 2025-12-01
Allow includeWhen expressions to reference externalRef resources

tenminusthreeseven:
Previously, includeWhen CEL expressions were validated in an environment
containing only the `schema` variable, preventing users from using
externalRef resources within includeWhen.

This PR updates the CEL environment to include all resources (template + externalRef),
allowing externalRef references in includeWhen expressions while keeping validation secure.

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -244,14 +244,13 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
 		return nil, fmt.Errorf("failed to create typed CEL environment: %w", err)
 	}
 
-	// Create a CEL environment with only "schema" for includeWhen expressions
-	var schemaEnv *cel.Env
-	if schemas["schema"] != nil {
-		schemaEnv, err = krocel.TypedEnvironment(map[string]*spec.Schema{"schema": schemas["schema"]})
-		if err != nil {
-			return nil, fmt.Errorf("failed to create CEL environment for includeWhen validation: %w", err)
-		}
-	}
+	
+	// Create CEL environment for includeWhen expressions including all resources (templates + externalRef)
+includeEnv, err := krocel.TypedEnvironment(schemas)
+if err != nil {
+    return nil, fmt.Errorf("failed to create CEL environment for includeWhen validation: %w", err)
+}
+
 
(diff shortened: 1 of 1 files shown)


pkg/graph/builder.go
```diff
@@ -799,11 +798,12 @@ func validateNode(resource *Resource, templatesEnv, schemaEnv *cel.Env, resource
 	}
 
 	// Validate includeWhen expressions if present
-	if len(resource.includeWhenExpressions) > 0 {
-		if err := validateIncludeWhenExpressions(schemaEnv, resource); err != nil {
-			return err
-		}
-	}
+if len(resource.includeWhenExpressions) > 0 {
```
jakobmoellerdev, line 801: im very certain that you didnt test and lint this pr. there hasnt been go fmt run on it as well.

pkg/graph/builder.go
```diff
@@ -244,14 +244,13 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
...
-	var schemaEnv *cel.Env
-	if schemas["schema"] != nil {
-		schemaEnv, err = krocel.TypedEnvironment(map[string]*spec.Schema{"schema": schemas["schema"]})
-		if err != nil {
-			return nil, fmt.Errorf("failed to create CEL environment for includeWhen validation: %w", err)
-		}
-	}
+	
+	// Create CEL environment for includeWhen expressions including all resources (templates + externalRef)
+includeEnv, err := krocel.TypedEnvironment(schemas)
```
jakobmoellerdev, line 249: even though you include these schemas, that doesnt mean they actually get interpreted and referenced correctly. Are you sure this works as intended?

conqquest: yes, you can check

jakobmoellerdev: the environment you create here is equivalent to the templatesEnv. However unlike templatesEnv, includeEnv is never used. what you committed here is dead code. you actually need to update the schemaEnv.

conqquest: okay sir , thank you for your response , i"ll update it

jakobmoellerdev:
closing this PR for now until this has been significantly reworked. feel free to open once you have a bit more of a testable state for us! Cheers

/close

PR #887, opened 2025-12-07
feat(website): add custom syntax highlighting for kro code blocks (RGDs)

a-hilaly:
Add a custom CodeBlock component with syntax highlighting for kro, rgd, and
simpleschema languages. Uses regex patterns to identify and style CEL expressions,
schema definitions, and YAML structure. Updated docs to use the new language tags.

Updated Prism theme for bash/shell to be more subdued and terminal-like.
Added smooth theme transitions, image zoom plugin, and Jest tests for the
tokenizer.

Screenshots:

Screenshot 2025-12-07 at 12 50 15 AM

Screenshot 2025-12-07 at 12 50 53 AM

Screenshot 2025-12-07 at 12 52 12 AM

### DIFF
--- a/website/src/theme/CodeBlock/highlighter.ts
+++ b/website/src/theme/CodeBlock/highlighter.ts
@@ -1,0 +1,226 @@
+// TODO: These keywords are duplicated from the Go codebase. We need a centralized
+// place (e.g., a shared JSON/YAML file) to keep these in sync automatically between
+// the Go parser and this TypeScript highlighter.
+// Go locations:
+//   - MARKER_KEYWORDS: pkg/simpleschema/parser/markers.go
+//   - KRO_KEYWORDS: pkg/graph/parser/parser.go (resource fields like id, template, etc.)
+
+// Marker keywords for SimpleSchema validation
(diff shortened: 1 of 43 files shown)


website/src/theme/CodeBlock/highlighter.ts
```diff
@@ -0,0 +1,226 @@
...
+      tokens.push(...processYAML(code.slice(pos, m.index!)));
+    }
+    tokens.push({ type: 'cel', value: m[0] });
+    pos = m.index! + m[0].length;
+  }
+  if (pos < code.length) {
+    tokens.push(...processYAML(code.slice(pos)));
+  }
+
+  function processYAML(yaml: string): Token[] {
```
tjamet, line 65: how hard would it be to rather extend the yaml highlighter and handle kro specifics (only for string values IIUC):
- markers
- cel expression (it could also be interesting to have highlights on that)

a-hilaly: I tried, unfortunetly wasn't successful. I didn't want to spend more time on this that it deserves too :) - happy to re-iterate another time.

PR #888, opened 2025-12-07
Add docs-tests presubmit script

a-hilaly:
Runs npm test, typecheck, and build for the website.

k/test-infra PR: https://github.com/kubernetes/test-infra/pull/36041


scripts/ci/presubmits/docs-tests
```diff
@@ -0,0 +1,20 @@
...
+npm ci
+
+# Run tests
+npm test
+
+# Type check
+npm run typecheck
+
+# Build the site
+npm run build
```
tjamet, line 20: Would it be a good idea and possible to post the build website to a temporary URL while the PR is alive so reviewers can also review the rendered version?

PR #889, opened 2025-12-07
[do-not-merge][experiment]: asynchronous watch registration

a-hilaly:
Previously, when a ResourceGraphDefinition had issues (e.g., missing RBAC
permissions), the `Register()` method would block indefinitely on
`cache.WaitForCacheSync()` while holding a global lock, causing all other RGDs
to stall. This change makes registration non-blocking by moving the cache sync
wait into a goroutine and using channel-based notification when sync completes.

/hold

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,6 +269,18 @@ deploy-kind: ko ## Deploy kro to a kind cluster
 	kubectl wait --for=condition=ready --timeout=1m pod -n kro-system -l app.kubernetes.io/component=controller
 	$(KUBECTL) --context kind-${KIND_CLUSTER_NAME} get pods -A
 
+.PHONY: deploy-kind-aggregation
+deploy-kind-aggregation: export KO_DOCKER_REPO=kind.local
+deploy-kind-aggregation: ko ## Deploy kro to a kind cluster with aggregation RBAC mode
+	$(KIND) delete clusters ${KIND_CLUSTER_NAME} || true
+	$(KIND) create cluster --name ${KIND_CLUSTER_NAME}
+	$(KUBECTL) --context kind-$(KIND_CLUSTER_NAME) create namespace kro-system
+	make install
+	# Deploy with aggregation RBAC mode - requires ClusterRoles with rbac.kro.run/aggregate-to-controller label
+	${HELM} template kro ./helm --namespace kro-system --set image.pullPolicy=Never --set image.ko=true --set config.allowCRDDeletion=true --set rbac.mode=aggregation | $(KO) apply -f -
+	kubectl wait --for=condition=ready --timeout=1m pod -n kro-system -l app.kubernetes.io/component=controller
+	$(KUBECTL) --context kind-${KIND_CLUSTER_NAME} get pods -A
+
 .PHONY: ko-apply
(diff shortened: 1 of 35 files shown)


tjamet:
IIRC the problem is then `RGDController` tries to [register](https://github.com/kubernetes-sigs/kro/blob/28313f62599a844d4978f8b71fed3cc8ea4d913f/pkg/controller/resourcegraphdefinition/controller_reconcile.go#L184) an instance controller to the dynamicController, it blocks any other RGD reconciliation because the dynamic controller [holds a global lock](https://github.com/kubernetes-sigs/kro/blob/28313f62599a844d4978f8b71fed3cc8ea4d913f/pkg/dynamiccontroller/dynamic_controller.go#L376) that never gets released because [lazyInforner.AddHandler](https://github.com/kubernetes-sigs/kro/blob/28313f62599a844d4978f8b71fed3cc8ea4d913f/pkg/dynamiccontroller/dynamic_controller.go#L480) never returns [waiting for cache sync](https://github.com/kubernetes-sigs/kro/blob/28313f62599a844d4978f8b71fed3cc8ea4d913f/pkg/dynamiccontroller/internal/gvr_watch.go#L115) that never happens, correct?

In the current implementation,  there are 2 identified cases:
- failing to start any [parent informer](https://github.com/kubernetes-sigs/kro/blob/28313f62599a844d4978f8b71fed3cc8ea4d913f/pkg/dynamiccontroller/dynamic_controller.go#L480) (because permission error)
- failing to start any [child informer](https://github.com/kubernetes-sigs/kro/blob/28313f62599a844d4978f8b71fed3cc8ea4d913f/pkg/dynamiccontroller/dynamic_controller.go#L538) (because permission error)

We should recover when:
- KRO got granted permissions to list/watch the parent resource
- KRO got granted permissions to list/watch the child resource
- The RGD got updated and removes the need to watch resources without permissions

From quick read, I we detect permission changes that would recover a functional state.
It looks like we are already watching for [informer errors](https://github.com/kubernetes-sigs/kro/blob/28313f62599a844d4978f8b71fed3cc8ea4d913f/pkg/dynamiccontroller/internal/gvr_watch.go#L86).
Would updating this handler for something like
```
	_ = inf.SetWatchErrorHandler(func(_ *cache.Reflector, err error) {
		w.log.V(1).Error(err, "watch error for lazy informer", "gvr", w.gvr)
		if isNoPermissions(err) {
			w.startErr = err
			w.Shutdown()
		}
	})
```

and update the handler registration:
```
		if !cache.WaitForCacheSync(w.done, w.informer.HasSynced) {
			w.log.Error(fmt.Errorf("cache sync failed"), "lazy informer sync failure", "gvr", w.gvr)
			if w.cancel!= nil{
				w.cancel()
			}
			w.cancel = nil
			w.informer = nil
			w.handlers = make(map[string]cache.ResourceEventHandlerRegistration)
			if w.startErr != nil {
				return w.startErr
			}
			return fmt.Errorf("failed to sync informer for %s", w.gvr)
		}
```

We would also need to make sure this error is correctly propagated to the RGD reconciler and the RGD status is updated as expected

PR #894, opened 2025-12-11
Draft: Allow aaS mode: running kro outside of a cluster against a generic control plane

sttts:
This pull request adds flags and helm chart values to allow separating the kubeconfig for leader election (in-cluster) from where the RGDs live and optionally where the workload objects (CRDs, CRs, composed objects) live.

Goal: have kro running outside of the control plane, but allow the use of RGDs to define APIs and consume them. I.e. the control plane does not have to be a full kubernetes cluster, but any kind of generic control plane (compare https://github.com/kubernetes/enhancements/blob/22e1b5c6c40908f498e55ed7f33a62478ad23a96/keps/sig-api-machinery/4080-generic-controlplane/README.md).

Key changes include:

**Multi-cluster configuration and client management:**

* Added new flags and configuration options to specify separate kubeconfigs, API server URLs, and CA certificates for RGD and workload clusters in both the controller code (`main.go`) and Helm values (`values.yaml`). [[1]](diffhunk://#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554dR75-R81) [[2]](diffhunk://#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554dR125-R138) [[3]](diffhunk://#diff-81801117ec01136c8a49bfcd42af8e104f4efad1f2daccda9da9d960eaad5279R103-R139)
* Updated the client package to support building REST configs from provided kubeconfigs or server URLs, and to allow using in-cluster credentials with custom endpoints and CAs. [[1]](diffhunk://#diff-d12234b4bf9ad71044e02892733242def8fc7b85c3236ba8c91544107e2c97acR76-R126) [[2]](diffhunk://#diff-d12234b4bf9ad71044e02892733242def8fc7b85c3236ba8c91544107e2c97acL86-R147)
* Refactored the main controller setup to create and manage separate `cluster.Cluster` and client sets for RGD and workload clusters as needed, falling back to single-cluster mode if not configured. [[1]](diffhunk://#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554dR190-R252) [[2]](diffhunk://#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554dL175-R273)

**Controller refactoring for multi-cluster:**

* Refactored the `ResourceGraphDefinitionReconciler` to accept and use separate client sets and clusters for RGDs and workload resources, ensuring correct resource watching and operations in multi-cluster mode. [[1]](diffhunk://#diff-45937ce2d4611b9e57fba30c7a979aa3ef257d6e0af7444befa518721ed9cc9aL51-R57) [[2]](diffhunk://#diff-45937ce2d4611b9e57fba30c7a979aa3ef257d6e0af7444befa518721ed9cc9aR66-R86) [[3]](diffhunk://#diff-45937ce2d4611b9e57fba30c7a979aa3ef257d6e0af7444befa518721ed9cc9aR97-R99)

**Helm chart updates:**

* Enhanced the Helm deployment template to mount the appropriate kubeconfig and CA secrets as volumes and to pass the new multi-cluster flags to the controller container when enabled. [[1]](diffhunk://#diff-a63ea33c81187c63a5163e4042419e5cce470581f485a1c99914c655f5570435R99-R151) [[2]](diffhunk://#diff-a63ea33c81187c63a5163e4042419e5cce470581f485a1c99914c655f5570435R176-R201)

**Testing and client interface improvements:**

* Updated the fake client set to implement the new `Metadata()` method, aligning with the expanded client interface for multi-cluster scenarios. [[1]](diffhunk://#diff-78e3b52b2bbf97b82bce59f3793588f2be179d340989779630ca8d6cdd0db15eR27-R35) [[2]](diffhunk://#diff-78e3b52b2bbf97b82bce59f3793588f2be179d340989779630ca8d6cdd0db15eR66-R70) [[3]](diffhunk://#diff-d12234b4bf9ad71044e02892733242def8fc7b85c3236ba8c91544107e2c97acR43-R45)

These changes enable flexible deployment of the controller in environments where RGDs and workloads are managed in separate clusters, improving scalability and security.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -113,8 +113,14 @@ config:
       # Key in the secret containing the kubeconfig data
       secretKey: "kubeconfig"
     # API server URL for RGD resources (alternative to kubeconfig)
-    # If set, uses in-cluster credentials with this server (CA reset to system roots)
+    # If set, uses in-cluster credentials with this server
     rgdServer: ""
+    # CA certificate for RGD server (use with rgdServer)
+    rgdCA:
+      # Name of an existing Secret containing the CA certificate
+      secretName: ""
+      # Key in the secret containing the CA certificate
+      secretKey: "ca.crt"
     # Kubeconfig for CRDs and CR instances (workload cluster)
     # If not specified, uses in-cluster config
     workloadKubeconfig:
@@ -123,8 +129,14 @@ config:
       # Key in the secret containing the kubeconfig data
       secretKey: "kubeconfig"
     # API server URL for CRDs and CR instances (alternative to kubeconfig)
-    # If set, uses in-cluster credentials with this server (CA reset to system roots)
+    # If set, uses in-cluster credentials with this server
     workloadServer: ""
+    # CA certificate for workload server (use with workloadServer)
+    workloadCA:
+      # Name of an existing Secret containing the CA certificate
+      secretName: ""
+      # Key in the secret containing the CA certificate
+      secretKey: "ca.crt"
   # The maximum number of queries per second to allow
   clientQps: 100
   # The number of requests that can be stored for processing before the server starts enforcing the QPS limit
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -75,8 +75,10 @@ func main() {
 		// kubeconfig paths for multi-cluster support
 		rgdKubeconfig      string
 		rgdServer          string
+		rgdCA              string
 		workloadKubeconfig string
 		workloadServer     string
+		workloadCA         string
 	)
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8078", "The address the metric endpoint binds to.")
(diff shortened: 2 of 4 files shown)


dims:
cc @a-hilaly @jlbutler !

ellistarn:
Have you explored a proxy based approach that peels off traffic and routes it to a different API server? This could be a generic solution so as to avoid implementing this in every relevant controller. Alternatively, something native to controller-runtime also seems reasonable to me. (i.e. KUBECONFIG=... REMOTEKUBECONFIG=...)

tjamet:
Hi!
In addition to @ellistarn 's thoughts, I am unsure I understand the problem solved.
I get that the goal is to have the "leader-election configmap" in a certain API endpoint and the RGD, instances and managed objects in another one.
What I don't get is what problem this architecture solves, what benefits it brings
It would be great if you can walk us through it
Thanks

helm/values.yaml
```diff
@@ -100,6 +100,43 @@ deployment:
 config:
   # Allow kro to delete CRDs
   allowCRDDeletion: false
+  # Multi-cluster kubeconfig settings
+  # When set, kro will use separate kubeconfigs for RGD and workload clusters
+  multiCluster:
```
a-hilaly, line 105: I'd suggest to rename `multiCluster` to something like `remoteWorkload` or `splitPlane`/`remotePlane` (or anything better). As we probably want to keep `multiCluster` for 1:n archiecture if we ever head towards that direction.

a-hilaly:
>Have you explored a proxy based approach that peels off traffic and routes it to a different API server? This could be a generic solution so as to avoid implementing this in every relevant controller. Alternatively, something native to controller-runtime also seems reasonable to me. (i.e. KUBECONFIG=... REMOTEKUBECONFIG=...)

+1 on exploring this, it would keep the controller simple and solve this generically. One of the tricky edges i'm thinking of is turtle RGDs (where an RGD can create another RGD as one of its resoucres), simple routing rules like RGDs -> cluster A, everything -> ClusterB, would break since the child RGD is technically a workload from the parent's perspective. Unless kro sets some headers to hint the proxy where resources should be routed?

For controller-runtime native support - I can see `KUBECONFIG` + `REMOTEKUBECONFIG` being useful at least for leases, that could apply to any controller. Though maybe what we need is a `COORDINATION_LEASES_KUBECONFIG`? Thought i think kro and dynamically configured controllers might be a different category of controllers (i think crossplane falls here as well) that needs more than just 2 environment variables/kubeconfigs

jakobmoellerdev:
hey @sttts, are you interested in incorporating the feedback above? if not, please let us know so someone could iterate that approach instead. Otherwise I would close out for now

PR #897, opened 2025-12-14
feat: add new golang script generatig examples doc on website

kennygt51:
Add a Go script that automatically generates website documentation
from example directories. The script reads examples.config.yaml to
determine which examples to process and generates Docusaurus-compatible
markdown files with frontmatter and tabbed manifest file displays.

The following is the sample docs generated by script:

image

This PR, as an MVP, targets the Go script and two examples (`examples/s3-buckets` and `kubernetes/saas-multi-tenant`) for automatic generation. The script is executed locally before creating the PR.

Next steps:
- Update the script so that it can generate other examples as well.
- delete existing update-docs.sh
- Run on the CI (Add `go run go run scripts/generate-examples-doc.go` to `deploy-docs.yaml`)

### DIFF
--- a/website/docs/examples/aws/s3-bucket.md
+++ b/website/docs/examples/aws/s3-bucket.md
@@ -1,0 +1,101 @@
+---
+sidebar_position: 101
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# S3 Buckets
+
+This example demonstrates how to create and manage AWS S3 buckets using kro's ResourceGraphDefinition.
+
+## Overview
+
+The S3 bucket example shows how to define an S3 bucket resource with common configurations such as versioning, encryption, and access policies. This is useful for applications that need object storage with consistent configuration across environments.
+## Manifest files
+
+
+
+<Tabs>
+  <TabItem value="instance" label="instance.yaml">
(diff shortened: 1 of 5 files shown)


website/docs/examples/aws/s3-bucket.md
```diff
@@ -0,0 +1,101 @@
+---
+sidebar_position: 101
```
tjamet, line 2: Shall we plan to at least have a check that this file is updated when the examples are updated?

As we start to have more generated files (and even more true with documentation) we may face 2 problems:
1. we update the examples but we forget to run the update script. The documentation results outdated
2. we manually update the generated file (because for example we don't remember there is this script) and then all changes are lost in the next run.

I believe this adds cognitive load we should try to lower.
While we have a similar problem with CRDs, which is lowered by 2 factors:
1. It's standard in the k8s ecosystem to run controller-gen somehow
2. Usually, not running controller-gen breaks either compilation or the tests, making it easily visible

An alternative could be to include it with all the other code generation in a single and simple `go generate ./...` for everything (CRDs, docs, ...) 
WDYT?

jakobmoellerdev:
closing this as stale, happy to revisit if theres still interest
/close

Issue #898, opened 2025-12-14
Wrong apiVersion on Quick start -> Create your Application Instance


birapjr:
### Documentation Issue

**Documentation Issue**:

The apiVersion in the Application example, instance.yaml, is not correct.

**Location**:
https://kro.run/docs/getting-started/deploy-a-resource-graph-definition#create-your-application-instance

**Current State**:
apiVersion: v1alpha1

**Suggested Changes**:
apiVersion: kro.run/v1alpha1

**Additional Context**:
Tested with:
kind cluster - v1.34.0
k3s cluster - v1.33.5+k3s1

PR #899, opened 2025-12-14
docs: correct apiVersion for Application example

birapjr:
Correct the apiVersion in the Application example, at the Quick start -> Create your Application Instance.

resolves: #898


a-hilaly (approved):
/lgtm

PR #901, opened 2025-12-14
Replace panic in CRD Graph Builder with proper error handling

AnshulPatil2005:
Summary

Replaces a panic() in the CRD graph builder with proper error handling to prevent controller crashes.

Changes

Removed panic() on invalid CRD configurations

Return clear errors when:

CRD has multiple versions

CRD schema is missing


jakobmoellerdev (commented):
/ok-to-test
/lgtm

this should usually not happen, very rare to have CRDs without a single version submitted. anyhow okay fix for me 👍

PR #910, opened 2025-12-19
cleanup: use variadic append for enum values

PhantomInTheWire:
cleanup: use variadic append for enum values in pkg/cel/schemas.go


a-hilaly (approved):
/lgtm

Issue #911, opened 2025-12-19
Quickstart Document about instance.yaml should be fixed


ricky9408:
### Documentation Issue

**Location**:

https://kro.run/docs/getting-started/deploy-a-resource-graph-definition#create-your-application-instance

**Current State**:

It says the example "instance.yaml" would be this

```yaml
apiVersion: v1alpha1
kind: Application
metadata:
  name: my-app-instance
spec:
  name: my-app
  replicas: 1
  ingress:
    enabled: true
```

**Suggested Changes**:

However, the apiVersion of the sample above seems wrong.

It should be fixed to this

```yaml
apiVersion: kro.run/v1alpha1
kind: Application
metadata:
  name: my-app-instance
spec:
  name: my-app
  replicas: 1
  ingress:
    enabled: true
```

**Additional Context**:

This quickstart page would be the first document for new comers who uses kro like me.
I can create new PR to fix this, thanks.

ARYANPATEL-BIT:
Hi 👋
I’d like to work on this issue.
Please assign it to me if it’s available. Thanks!

ARYANPATEL-BIT:
Hi maintainers 👋
I’m interested in working on this issue.
Please assign it to me if no one is working on it. Thanks!

PhantomInTheWire:
@ARYANPATEL-BIT this already has a pr that has been lgtm'ed, try looking for issues with help needed / good first issue labels

ARYANPATEL-BIT:
Thanks for clarifying.

PR #913, opened 2025-12-19
fix: Fixed Quickstart instance.yaml sample on document

ricky9408:
Issue: https://github.com/kubernetes-sigs/kro/issues/911

**Location**:

https://kro.run/docs/getting-started/deploy-a-resource-graph-definition#create-your-application-instance

**Current State**:

It says the example "instance.yaml" would be this

```yaml
apiVersion: v1alpha1
kind: Application
metadata:
  name: my-app-instance
spec:
  name: my-app
  replicas: 1
  ingress:
    enabled: true
```

**Suggested Changes**:

However, the apiVersion of the sample above seems wrong.

It should be fixed to this

```yaml
apiVersion: kro.run/v1alpha1
kind: Application
metadata:
  name: my-app-instance
spec:
  name: my-app
  replicas: 1
  ingress:
    enabled: true
```

**Additional Context**:

This quickstart page would be the first document for new comers who uses kro like me.
Would be good to fix this document as soon as possible :)

### DIFF
--- a/website/versioned_docs/version-0.7.0/docs/getting-started/02-deploy-a-resource-graph-definition.md
+++ b/website/versioned_docs/version-0.7.0/docs/getting-started/02-deploy-a-resource-graph-definition.md
@@ -160,7 +160,7 @@ an `Ingress`. Let's use it!
    with the following content:
 
    ```yaml title="instance.yaml"
-   apiVersion: v1alpha1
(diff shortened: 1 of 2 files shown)


a-hilaly (approved):
Thank you @ricky9408 !
/lgtm

Issue #914, opened 2025-12-19
Add Labels and Annotations to generated CRDs


cnvergence:
### Feature Description

**Problem Statement**:

Currently, ResourceGraphDefinitions generate CRDs without user-defined labels or annotations. Adding these would be helpful for users in integrating with kro in their clusters.

**Proposed Solution**:

Add support for specifying labels and annotations on CRDs generated by ResourceGraphDefinitions. 

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

cnvergence:
I would gladly send the draft PR for this and work on it :)

RafPe:
Nice work @cnvergence

cnvergence:
thanks @RafPe for linking the older issue, I haven't found it before. 🙂 
My PR will add the labels to the CRD object created from RGD, did you also want to have them copied to the grouped resources?

RafPe:
> thanks [@RafPe](https://github.com/RafPe) for linking the older issue, I haven't found it before. 🙂 My PR will add the labels to the CRD object created from RGD, did you also want to have them copied to the grouped resources?

It was my mistake! I do like what your PR introduces but yes my idea is to have a spec in RGD that allows for setting up labels/annotations on all resources created. I will make a PR for this specific case or check if there are some active developments on related issues

cnvergence:
Got it, I might also help with this, but let's sort my PR first, then.

a-hilaly:
released in 0.8.X

a-hilaly:
Thank you folks!

PR #915, opened 2025-12-19
add support for labels and annotations in the generated CRDs

cnvergence:
Currently, ResourceGraphDefinitions generate CRDs without user-defined labels or annotations. Adding these would be helpful for users in integrating with kro in their clusters.

This PR will add support for specifying labels and annotations for generated CustomResourceDefinitions in ResourceGraphDefinitions.

Issue: 
- https://github.com/kubernetes-sigs/kro/issues/914


cnvergence:
sorry for the noise, organizational copilot started to do reviews without being requested..

Issue #909, opened 2025-12-19
0.7.1 - found unknown resources in CEL expression: [[{random random}]]


shwetambry:
### Description

**Observed Behavior**: Using CEL function `random` in RGD is failing in v0.7.1

**Expected Behavior**: random.seededString(xx) should work

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):
I used the examples given in KRO github repository.
RGD: https://github.com/kubernetes-sigs/kro/blob/main/examples/kubernetes/cel-functions/rg.yaml
Instance: https://github.com/kubernetes-sigs/kro/blob/main/examples/kubernetes/cel-functions/instance.yaml
rgd
```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: somethings.kro.run
spec:
  schema:
    apiVersion: v1alpha1
    kind: Something
    spec:
      name: string
      fullDNSName: string | default="a.b.c.d.example.com"
    status:
      deploymentConditions: ${deployment.status.conditions}
      availableReplicas: ${deployment.status.availableReplicas * 10}
  resources:
    - id: deployment
      template:
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: ${schema.spec.name}
        spec:
          replicas: 1
          selector:
            matchLabels:
              app: deployment
          template:
            metadata:
              labels:
                app: deployment
            spec:
              containers:
                - name: ${schema.spec.name}-deployment
                  image: nginx
                  ports:
                    - containerPort: 80
                  env:
                    - name: FULL_DNS_NAME
                      value: ${schema.spec.fullDNSName}
                    - name: DNS_FIRST_PART
                      value: ${schema.spec.fullDNSName.split('.')[0]}
                    - name: DNS_PARTIALS
                      value: ${schema.spec.fullDNSName.split('.').slice(0, 1).join('.')}
                    - name: RANDOM_SEEDED_STRING
                      value: ${random.seededString(10, schema.metadata.uid)}
    - id: service
      template:
        apiVersion: v1
        kind: Service
        metadata:
          name: ${schema.spec.name}
          labels:
            app: deployment
            dns: ${schema.spec.fullDNSName}
            partials: "${size(schema.spec.fullDNSName.split('.')) > 2 ? schema.spec.fullDNSName.split('.').slice(0, 3).join('.') : schema.spec.fullDNSName}"
[...]

PR #919, opened 2025-12-22
Prevent random.* from being classified as a resource

benzaidfoued:
Fix Bug : 0.7.1 - found unknown resources in CEL expression: [[{random random}]] #909

### DIFF
--- a/pkg/cel/ast/inspector.go
+++ b/pkg/cel/ast/inspector.go
@@ -266,7 +266,7 @@ func (a *Inspector) inspectCall(ast *celast.AST, call celast.CallExpr, path stri
 			if _, ok := a.functions[full]; !ok {
 				out.merge(a.inspectExpr(ast, t, path))
 			}
-			// Record the member function invocation as part of CEL dependency graph construction.
+			// Treat chained method calls as unknown unless they resolve to a known namespaced function
 			out.FunctionCalls = append(out.FunctionCalls, FunctionCall{
 				Name: full,
 			})


pkg/cel/ast/inspector.go
```diff
@@ -262,8 +262,19 @@ func (a *Inspector) inspectCall(ast *celast.AST, call celast.CallExpr, path stri
...
+					Name: full,
+				})
+				return out
+			}
+
+			// Otherwise, it will be a normal method call on a value
 			out.merge(a.inspectExpr(ast, t, path))
+			out.FunctionCalls = append(out.FunctionCalls, FunctionCall{
+				Name: full,
+			})
```
a-hilaly: Isn't this just a duplicate of L280-282?

a-hilaly (commented):
Thank you @benzaidfoued! i think this deserves an integration test to prevent from future regressions - can you please add some?

pkg/cel/ast/inspector.go
benzaidfoued: ur right

benzaidfoued:
> Thank you @benzaidfoued! i think this deserves an integration test to prevent from future regressions - can you please add some?

Sure .

a-hilaly (commented):
Thanks @benzaidfoued ~ could we also add a unit test that explicitly covers the case where a resourceID matches a function namespace prefix? i'm thinking something like declaring resources: `[]string{"random"}` and using both `random.status.field` and `random.seededString()` in the same expression. I'm curious to see the behavior in this case.

benzaidfoued:
> Thanks @benzaidfoued ~ could we also add a unit test that explicitly covers the case where a resourceID matches a function namespace prefix? i'm thinking something like declaring resources: `[]string{"random"}` and using both `random.status.field` and `random.seededString()` in the same expression. I'm curious to see the behavior in this case.

umm sounds interesting, let me try something

jakobmoellerdev (commented):
/lgtm

thanks for the contrib and happy holidays!

PR #925, opened 2025-12-24
Docs: manifests download URL changed to match actual URL

hatofmonkeys:
Hi,
Thanks for the awesome project!

The docs currently generate a URL that misses the "v" in the download URL. This PR is a quick fix.

I'm not sure where you'd like to add the "v", or whether you want to change the directory structure of the releases to not have a "v", but I thought I'd submit a simple pull request and let maintainers choose if it should be fixed elsewhere.

Thanks again!
Colin

### DIFF
--- a/website/docs/docs/getting-started/01-Installation.md
+++ b/website/docs/docs/getting-started/01-Installation.md
@@ -87,7 +87,7 @@ echo $KRO_VERSION
 
 ```bash
 kubectl create namespace kro-system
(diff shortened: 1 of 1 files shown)


website/docs/docs/getting-started/01-Installation.md
a-hilaly, line 82: Might be better to remove the `| ltrimstr("v")` here? cc @tjamet 

Can iterate later on

a-hilaly (approved):
Nice catch, thank you @hatofmonkeys !

/lgtm

PR #935, opened 2026-01-05
KREP-009: Versioning support in KRO

barney-s:
Proposal for Versioning for KRO.

related:
#861

### DIFF
--- a/docs/design/proposals/versioning.md
+++ b/docs/design/proposals/versioning.md
@@ -1,0 +1,268 @@
+# KRO: Versioning and Rollouts
+
+# Problem Statement
+
+As the KRO project matures, platform teams (RGD authors) need a structured way to evolve their \`ResourceGraphDefinition\` (RGD) resources over time. Currently, any change to an RGD in a cluster is immediately propagated to all existing instances across all namespaces upon their next reconciliation. This "rollout-on-reconcile" behavior is fast but lacks the control necessary for some production environments. The goal is to allow RGD authors to introduce changes while giving control over how the changes are safely rolled out in the cluster. This aligns with GitOps principles, where version pinning and controlled automation are crucial for stability.
+
+# Background
+
+To understand the scope of the problems to be addressed, we need to understand a few things that will be used as inputs in the proposal.
+
+## What can change in an RGD
+
+When we say an RGD author makes a change, it could mean one or more of the following:
+
+1. **Resource Adjustments:** Changes to `.spec.resources`:  
+   1. Resources are added.  
+   2. Resources are removed.  
+   3. Resources are modified (e.g., changing a container image, updating a label).  
+2. **Schema Evolution:** Changes to the `.spec.schema`:  
+   1. Backward-compatible changes:  
+      1. A new optional field is added.  
+      2. A new field with a default value is added.  
+      3. A default value is removed  
+   2. Backward-incompatible changes:  
+      1. A field is removed.  
+      2. Adding/Changing defaults.  
+      3. ~~A field is renamed. This is done by removing the old field and adding a new field.~~  
+3. **CEL Environment Dependency:** The version of KRO determines the available CEL environments and functions.  
+   1. An RGD author could change the CEL snippets used in RGD that require a different CEL environment.  
+   2. KRO authors could fix CEL functions that would cause behavioural changes in the resulting resources.
+
+## Challenges in current model
+
+Key challenges with the current model include:
+
+* **Lack of Controlled Rollouts:** There is no mechanism to gradually roll out RGD changes. All instances are updated at once, which is risky.  
+* **No Version Pinning**: Application teams (instance authors) cannot pin their instances to a specific, known-good version of an RGD. This makes them vulnerable to breaking changes introduced by the platform team and complicates bug remediation.  
+* **Handling Breaking Changes:** There is no formal process for managing backward-incompatible changes to an RGD's schema or resource composition.  
+* **GitOps Integration:** The current model complicates GitOps workflows. GitOps relies on declarative state, but the effective state of an instance can change implicitly due to an RGD update, without any change to the instance's own manifest in git.
+
+## Common questions
+
+1. How should modifications to the ResourceGraphDefinition itself be managed ?  
+2. Should RGDs explicitly reference a specific kro-cel environment version ?  
+3. **Rollout Authority:** Who retains control over the rollout process: the infrastructure team (RGD author) or the application teams (instance authors)?  
+4. **Rollout Methodologies:** What distinct rollout strategies should be supported (e.g., granular rollouts by namespace or label)?
+
+# Proposal
+
+We may have to pick one or more mechanisms based on the changes needed.
+
+* We propose using the [Replicaset pattern](https://docs.google.com/document/d/1-9lWOJEDDN2Qsl6KVfrEbEZHgLm6Twk06L6-OsVJuho/edit?resourcekey=0-lLY3WBJ-0vlHfi5l0nypxg&tab=t.0#heading=h.mva37kvbhn51) for non-breaking schema changes. This allows authors to make changes without worrying about breaking users. Also allows users to select which version of RGD to use.  
+* For breaking schema changes or behavioural changes, we recommend creating  a [new RGD with a different CRD](#new-rgd-for-the-change). We shall support migration tooling for such use cases. A Cli helper `kro migrate` can be added.
+
+# Design Options
+
+This section outlines various design options for implementing versioning. The discussion of these approaches aims to provide a better understanding of the available solutions, enabling a robust discussion that will inform the final design proposal.
+
+## Representing RGD changes
+
+In Kubernetes, RGD is an object. When any change is made to an RGD, we lose the previous configuration. There is no notion of the history of an object. Only the current view exists. 
+
+To deal with this we have these options:
+
+* **Single canonical RGD**: Only one instance of RGD exists, latest.  
+* **Different RGD for each change**: Each change requires a new RGD and Instance CRD.  
+* **Replicaset pattern**: Replicas automatically created and instances pinned to a replica  
+* **Sections with RGD for each iteration number:** Schema and Resources are defined per iteration. Instances are pinned to a section.
+
+## Replicaset Pattern
+
+Introduce a new `ResourceGraphDefinitionReplica` CRD that is a point-in-time snapshot of a RGD.  This is automatically created and managed by KRO reconciler. Instances reconcile against a specific version of the replica.  
+**Mechanism**
+
+* Crucially, every modification to the RGD specification automatically triggers the creation of a new `ResourceGraphDefinitionReplica` by the KRO.   
+* Instance reconciliation picks current `ResourceGraphDefinitionReplica` during first reconciliation and is pinned to that replica (via annotation).  
+* We can have a mode where pinning is not done and all instances are updated to the latest replica always.  
+* To mimic existing behavior we can set the history depth to 1 and enable no pinning mode where instances are always updated to the latest replica.
+
+**Migration**
+
+* **User-Controlled Opt-in:** Users have control over when they migrate their instance to a new version. Instance can be moved to a different version/replica by changing the annotation.  
+* **Automated Migration:** A mechanism can be built to automate the migration process that changes the annotation. No special handling of resource migration is required since we are dealing with the same instance object.
+
+**Pros**
+
+* **Easy rollback**: This automated process ensures that a historical record of all RGD spec versions is maintained. Each `ResourceGraphDefinitionReplica` essentially represents a snapshot of the RGD at a particular point in time, encapsulating all the details of that specific version. This automatic generation is fundamental to enabling versioning and allowing for the tracking and potential rollback to previous configurations.  
+* **Version Pinning:** During its initial reconciliation cycle, an instance will identify and select a `ResourceGraphDefinitionReplica` to which it will be "pinned." This pinning is achieved through an annotation applied to the instance. The primary purpose of this pinning is to ensure stability and predictability. Once an instance is pinned to a particular replica, it will continue to operate based on the definition provided by that specific `ResourceGraphDefinitionReplica`, even if newer replicas are subsequently created. This behavior is critical for preventing unexpected disruptions to running instances due to RGD spec changes. It allows for a controlled rollout of new RGD versions, where existing instances continue to use their original, validated configurations.  
+* **Velocity:** While the default behavior emphasizes pinning for stability, the system also offers a configurable mode where this pinning is deliberately *not* performed. In this alternative mode, all instances are designed to continuously update themselves to the latest available `ResourceGraphDefinitionReplica`. This dynamic updating ensures that instances are always running with the most current RGD specification. This mode would be particularly useful in environments where rapid adoption of the latest RGD definitions is paramount, or for stateless services that can tolerate frequent configuration updates without significant impact. The choice between pinning and continuous updates provides flexibility, allowing users to select the behavior that best suits their application's requirements and deployment strategy.  
+* **No Concept of immediate rollout**: Any change to a RGD results in a new replica. So automated rollout does not happen.
+
+**Cons**
+
+* **Breaking schema change handling:** If the schema has a breaking change across versions,  it would require conversion webhooks to be defined. There is no mechanism to do it in KRO.  
+* 
+
+## Single Canonical RGD
+
+This approach maintains the current model where only one version of the RGD exists, which is always the most recently applied one. This simplifies reasoning, mirroring how most Kubernetes resources function, as changes are immediately visible across all RGD instances. The RGD contains a single schema definition and a single resources section.
+
+**Mechanism**
+
+* **Resource Updates:** Changes (additions, removals, modifications) are directly updated within the RGD's resource section.  
+* **Schema Compatibility:** Only backward-compatible schema changes are allowed.  
+* **Automatic Rollout:** When the RGD is modified, these changes are automatically rolled out across all instances upon reconciliation.
+
+**Pros**
+
+* **Simplicity:** This is the most straightforward method for managing RGD changes, easy to understand and implement.  
+* **Single Source of Truth:** Maintains one definitive source for the RGD definition.  
+* **Automatic Propagation:** Changes are automatically propagated, which is beneficial for rapid updates.
+
+**Cons**
+
+* **Uncontrolled Rollouts:** All instances update simultaneously, which can be risky, especially for critical applications. Users cannot pin to a specific version to avoid a buggy rollout.  
+* **Irreversible Changes:** This model has a "unidirectional and irreversible" change model, which can be a drawback for rollbacks.  
+* **Breaking schema change handling:** Handling backward-incompatible schema changes is challenging and likely necessitates downtime or complex migrations.  
+* **No Version Pinning**: Since changes are immediately rolled out, there is no way to pin instances to a specific revision of RGD.
+
+## RGD with version sections
+
+This approach integrates versioning directly into a single RGD (Resource Group Definition) resource. A single RGD exists for multiple versions with sections within the RGD.
+
+**Mechanism**
+
+* The RGD `spec` will include a map or list of versions (e.g., `v1`, `v2`).  
+* Each version will have its own schema and set of resources, as shown in the example below:
+
+```
+spec:
+   - version: v1
+     schema:
+        # schema for v1...
+     resources:
+        # resources for v1...
+   - name: v2
+     schema:
+       # schema for v2...
+     resources:
+       # resources for v2...
+  servedVersion: v2
+```
+
+* When a new version is added to the RGD, the KRO controller will create a new version of the corresponding Custom Resource Definition (CRD) (e.g., `MyWebApp/v2`).  
+* Instances are not automatically migrated. Users must explicitly change their instance's `apiVersion` (e.g., from `my.api/v1` to `my.api/v2`) to opt into the new version.  
+* This option requires supporting a "migrate" lifecycle for instances and an "adoption" functionality for new instances.  
+* Changes within a version are immediately applied across instances using that version.
+
+**Migration**
+
+* **User-Controlled Opt-in:** Users have control over when they migrate their instance to a new version.  
+* **Automated Migration:** A mechanism can be built to automate the migration process. This would involve creating a new instance object at the new version, copying relevant fields, and deleting the old one. This requires careful handling of resource adoption and state.
+
+**Pros**
+
+* **Controlled Rollouts:** Users can adopt new versions at their own pace. Provides fine-grained control over when instances adopt new versions, crucial for critical systems or phased rollouts.  
+* **Version Pinning**: Instances are by default pinned to a specific RGD. Migration is explicit.  
+* **Clear Separation:** Versions are clearly defined and managed within a single RGD.  
+* **GitOps Friendly:** Users can pin their instance to a specific version in their Git repository.  
+* **No Impact on Existing Instances:** Introducing a new version has no effect on consumers of older versions.
+
+**Cons**
+
+* **Increased Complexity:** The RGD resource becomes more complex.  
+* **Breaking schema change handling:** If the schema has a breaking change across versions,  it would require conversion webhooks to be defined. There is no mechanism to do it in KRO.  
+* **CRD Management:** The controller needs to manage the lifecycle of multiple CRD versions.  
+* **Migration Tooling:** Requires building robust mechanisms for instance migration and adoption.  
+* Changes within a version are immediately applied across instances using that version.  
+* **KRM Object size limitation**: With multiple versions in the same RGD, it is more likely to hit the KRM object size limitations.   
+* **Immediate rollout within version**: Changes within a version are immediately rolled out with no control.  
+* **Instance Migration for Minor Changes:** Instances need to be migrated to the new CRD even for minor changes in the RGD.  
+* **Resource Abandonment and Adoption:** Resources managed by the old CRD must be abandoned and then adopted by the new CRD.
+
+## New RGD for the change {#new-rgd-for-the-change}
+
+This approach treats each version of a definition as a completely distinct ResourceGraphDefinition resource.
+
+**Mechanism**
+
+* To create a new version, the platform team generates a new RGD resource with a versioned name (e.g., `my-web-app-v1`, `my-web-app-v2`).  
+* Each RGD (e.g., `my-web-app-v2`) defines its own Custom Resource Definition (CRD) (e.g., `MyWebAppV2`), with the CRD name itself being versioned.  
+* Existing instances of the older RGD are entirely unaffected by the new version.  
+* Users must create new instances using the new CRD to adopt the updated version.  
+* This necessitates a well-defined migration strategy. This involves supporting a lifecycle for migrating an instance from one RGD/CRD to another, including the adoption of underlying resources.
+
+**Migration**
+
+* **User-Controlled Opt-in:** Users have control over when they migrate their instance to a new version.  
+* **Automated Migration:** A mechanism can be built to automate the migration process. This would involve creating a new instance object at the new version, copying relevant fields, and deleting the old one. This requires careful handling of resource adoption and state.
+
+**Pros**
+
+* **Clear Separation:** Each version exists as a distinct, immutable resource, offering a very clean and easily understandable structure.  
+* **No Impact on Existing Instances:** Introducing a new version has no effect on consumers of older versions.  
+* **Version Pinning**: Instances are by default pinned to a specific RGD. Migration is explicit.  
+* **Controlled Rollouts:** Users can adopt new versions at their own pace. Provides fine-grained control over when instances adopt new versions, crucial for critical systems or phased rollouts.  
+* **GitOps Friendly:** Users can pin their instance to a specific version in their Git repository.  
+* **Breaking schema change handling:** Since new CRD is created there is no concept of breaking changes in schema. It's a new CRD that we need to migrate to.
+
+**Cons:**
+
+* **Resource Proliferation:** This method can lead to a significant increase in the number of RGD and CRD resources within the cluster.  
+* **Discovery Challenges:** Users may find it more difficult to discover the latest or recommended version of a definition, potentially requiring a new grouping mechanism.  
+* **Migration Tooling:** Substantial effort may be required to develop tooling that supports instance migration and resource adoption.  
+* **Impractical for Velocity for small changes:** This approach essentially makes an RGD immutable. A new RGD is created for every change, using a different Instance CRD. This can be impractical for RGD authors or organizations that prioritize rapid development.  
+* **Instance Migration for Minor Changes:** Instances need to be migrated to the new CRD even for minor changes in the RGD. Making easy things harder.  
+* **Resource Abandonment and Adoption:** Resources managed by the old CRD must be abandoned and then adopted by the new CRD.  
+* **Immediate rollout within version**: Changes within a version are immediately rolled out with no control.
+
+## Rollouts
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+1. How should modifications to the ResourceGraphDefinition itself be managed ?  
+2. Should RGDs explicitly reference a specific kro-cel environment version ?  
+3. **Rollout Authority:** Who retains control over the rollout process: the infrastructure team (RGD author) or the application teams (instance authors)?  
+4. **Rollout Methodologies:** What distinct rollout strategies should be supported (e.g., granular rollouts by namespace or label)?
+
+# Proposal
+
+We may have to pick one or more mechanisms based on the changes needed.
+
+* We propose using the [Replicaset pattern](https://docs.google.com/document/d/1-9lWOJEDDN2Qsl6KVfrEbEZHgLm6Twk06L6-OsVJuho/edit?resourcekey=0-lLY3WBJ-0vlHfi5l0nypxg&tab=t.0#heading=h.mva37kvbhn51) for non-breaking schema changes. This allows authors to make changes without worrying about breaking users. Also allows users to select which version of RGD to use.
```
justinsb, line 52: The doc wasn't accessible to the account I am logged in to right now (but I'll try my work account).  Regardless, I think it would be good to have a one sentence summary because I'm not sure whether you're scoping this to the RGD or the created objects.  For example: `At a high level, we propose automtically "snapshotting" each significant change to ResourceGraphDefinition objects, creating a set of ResourceGraphDefinitionVersion objects that are immutable and have their own identity`

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+* **Velocity:** While the default behavior emphasizes pinning for stability, the system also offers a configurable mode where this pinning is deliberately *not* performed. In this alternative mode, all instances are designed to continuously update themselves to the latest available `ResourceGraphDefinitionReplica`. This dynamic updating ensures that instances are always running with the most current RGD specification. This mode would be particularly useful in environments where rapid adoption of the latest RGD definitions is paramount, or for stateless services that can tolerate frequent configuration updates without significant impact. The choice between pinning and continuous updates provides flexibility, allowing users to select the behavior that best suits their application's requirements and deployment strategy.
```
justinsb, line 89: Another pro: It always takes _some_ time to apply a new version, and with versions we can give the users a breadcrumb trail to understanding what change is being made.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+* **Velocity:** While the default behavior emphasizes pinning for stability, the system also offers a configurable mode where this pinning is deliberately *not* performed. In this alternative mode, all instances are designed to continuously update themselves to the latest available `ResourceGraphDefinitionReplica`. This dynamic updating ensures that instances are always running with the most current RGD specification. This mode would be particularly useful in environments where rapid adoption of the latest RGD definitions is paramount, or for stateless services that can tolerate frequent configuration updates without significant impact. The choice between pinning and continuous updates provides flexibility, allowing users to select the behavior that best suits their application's requirements and deployment strategy.  
+* **No Concept of immediate rollout**: Any change to a RGD results in a new replica. So automated rollout does not happen.
+
+**Cons**
+
+* **Breaking schema change handling:** If the schema has a breaking change across versions,  it would require conversion webhooks to be defined. There is no mechanism to do it in KRO.
```
justinsb, line 94: Honestly I'm not sure this is a con of this approach.  I'd say the most obvious con (IMO) is that there is another layer of objects and thus more complexity, though I believe it is very "kubernetes": making the state very explicit, but hiding that state in a kind that you can generally ignore (it is pretty rare that I `kubectl get replicaset -oyaml`)

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+* 
+
+## Single Canonical RGD
+
+This approach maintains the current model where only one version of the RGD exists, which is always the most recently applied one. This simplifies reasoning, mirroring how most Kubernetes resources function, as changes are immediately visible across all RGD instances. The RGD contains a single schema definition and a single resources section.
+
+**Mechanism**
+
+* **Resource Updates:** Changes (additions, removals, modifications) are directly updated within the RGD's resource section.  
+* **Schema Compatibility:** Only backward-compatible schema changes are allowed.
```
justinsb, line 104: We should certainly only support backward compatible changes, but I think we have choices about how we enforce this.  e.g. do we want a webhook right away, or is it good enough to simply have the RGD go into an error mode "IncompatibleSchemaChange"?  (I think this actually applies to all alternatives)

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+* **Velocity:** While the default behavior emphasizes pinning for stability, the system also offers a configurable mode where this pinning is deliberately *not* performed. In this alternative mode, all instances are designed to continuously update themselves to the latest available `ResourceGraphDefinitionReplica`. This dynamic updating ensures that instances are always running with the most current RGD specification. This mode would be particularly useful in environments where rapid adoption of the latest RGD definitions is paramount, or for stateless services that can tolerate frequent configuration updates without significant impact. The choice between pinning and continuous updates provides flexibility, allowing users to select the behavior that best suits their application's requirements and deployment strategy.  
+* **No Concept of immediate rollout**: Any change to a RGD results in a new replica. So automated rollout does not happen.
+
+**Cons**
+
+* **Breaking schema change handling:** If the schema has a breaking change across versions,  it would require conversion webhooks to be defined. There is no mechanism to do it in KRO.  
+* 
+
+## Single Canonical RGD
```
justinsb, line 97: Maybe we make it clearer that this is an "Alternative Considered (but rejected)" - perhaps by moving it to its own section?  We might also consider whether we want to include alternatives considered in our permanent documentation (or some way to hide it from humans/LLMs).

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+### Rollout Patterns
+
+To accommodate diverse deployment needs, the following rollout patterns can be supported:
+
+* **Rollout by Namespace:** This allows for rolling out changes to instances within specific Kubernetes namespaces, useful for phased rollouts across different environments or teams.  
+* **Rollout by Labels:** This enables targeted rollouts based on labels applied to instances, providing flexibility for canary deployments, A/B testing, or rolling out to a subset of instances based on specific criteria.
+
+### Other Considerations
+
+* **CEL Environment:** How should changes in the CEL environment, tied to the KRO controller version, be handled? Should a field be added to the RGD spec to specify a required `kro-cel-env` version, allowing the controller to report unmet dependencies?
```
justinsb, line 226: My 2c: Yes, but we can treat this as another version, i.e. this field should be one of the fields that triggers a snapshot (I _think_ it'll be exactly the set of .spec fields, but IDK)

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+Regardless of the chosen versioning strategy, it's critical to define who controls the rollout:
+
+* **Infrastructure Team (RGD Author):** This team is responsible for defining and maintaining the RGD. They would have the authority to initiate and manage rollouts of RGD changes. This is suitable for scenarios where changes are broad and impact many users.  
+* **User Teams (Instance Author):** These teams create and manage instances of the RGD. They would control when their specific instances adopt new versions, empowering them with autonomy over their deployments.
+
+### Rollout Patterns
+
+To accommodate diverse deployment needs, the following rollout patterns can be supported:
+
+* **Rollout by Namespace:** This allows for rolling out changes to instances within specific Kubernetes namespaces, useful for phased rollouts across different environments or teams.
```
justinsb, line 221: I think we want to support:

* the "yolo" rollout pattern we have today - users should be able to opt in to it.
* per-instance pinning - maybe in a well-known field (e.g. spec.version), maybe in an annotation
* automatic pinning of instances

Then I think the different rollout strategies are orthogonal and we can add more strategies as needed.  We can even support external rollout strategies.

My _guess_ is that we will have a field like `spec.rolloutStrategy` on the RGD that will define how the new version should be rolled out, which means the Platform Team persona determines this.  And we'll likely support "yolo" as the default initially (because this is what we have today).  And my guess is that we'll start with "external" rollout as our second strategy, and prototype strategies in our CLI before baking them in.

I think this also implies that we want to track the version that a instance is on, I'm thinking status.targetVersion and status.desiredVersion is probably a good approach there (rather than writing into a spec.version field and that field having multiple writers)

justinsb (commented):
The "ReplicaSet" design LGTM and I think if we have consensus there we should more clearly separate out the alternatives considered (but rejected) - I personally stumbled there.

I do think we need a few more details about the initial rollout strategies and the fields we will need to support that, but I don't think we need much more and I don't think there are also that many alternative paths once you have made the ReplicaSet pattern decision (a good thing!)

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+2. Should RGDs explicitly reference a specific kro-cel environment version ?  
+3. **Rollout Authority:** Who retains control over the rollout process: the infrastructure team (RGD author) or the application teams (instance authors)?  
+4. **Rollout Methodologies:** What distinct rollout strategies should be supported (e.g., granular rollouts by namespace or label)?
+
+# Proposal
+
+We may have to pick one or more mechanisms based on the changes needed.
+
+* We propose using the [Replicaset pattern](https://docs.google.com/document/d/1-9lWOJEDDN2Qsl6KVfrEbEZHgLm6Twk06L6-OsVJuho/edit?resourcekey=0-lLY3WBJ-0vlHfi5l0nypxg&tab=t.0#heading=h.mva37kvbhn51) for non-breaking schema changes. This allows authors to make changes without worrying about breaking users. Also allows users to select which version of RGD to use.  
+* For breaking schema changes or behavioural changes, we recommend creating  a [new RGD with a different CRD](#new-rgd-for-the-change). We shall support migration tooling for such use cases. A Cli helper `kro migrate` can be added.
```
jakobmoellerdev, line 53: IMHO this goes against kubernetes primitives around migration of CRDs and using exactly one CRD object that has multiple apiVersions.

Why cant we allow CRD upgrades and run storage version migration witWh kubernetes tooling. We could even use the k8s api for storage version migration after the last resource has been migrated? https://kubernetes.io/docs/tasks/manage-kubernetes-objects/storage-version-migration/#update-the-preferred-storage-schema-of-a-crd

I would like to avoid having to recreate RGDs. It will complicate GitOps workflows, and does not allow in place migrations. Using a CLI helper also is problematic because it forces me to use SRE practices of connecting directly to the cluster, instead of being able to migrate with GitOps alone.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+**Pros**
+
+* **Simplicity:** This is the most straightforward method for managing RGD changes, easy to understand and implement.  
+* **Single Source of Truth:** Maintains one definitive source for the RGD definition.  
+* **Automatic Propagation:** Changes are automatically propagated, which is beneficial for rapid updates.
+
+**Cons**
+
+* **Uncontrolled Rollouts:** All instances update simultaneously, which can be risky, especially for critical applications. Users cannot pin to a specific version to avoid a buggy rollout.  
+* **Irreversible Changes:** This model has a "unidirectional and irreversible" change model, which can be a drawback for rollbacks.
```
jakobmoellerdev, line 116: I would argue that because we are talking about API upgrades, the way to fix an issue is by "fixing forward". if there is a change coming up in v1alpha1 and you fail to fix it with v1alpha2, v1alpha3 would need to be rolled out to unify.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+
+* **Simplicity:** This is the most straightforward method for managing RGD changes, easy to understand and implement.  
+* **Single Source of Truth:** Maintains one definitive source for the RGD definition.  
+* **Automatic Propagation:** Changes are automatically propagated, which is beneficial for rapid updates.
+
+**Cons**
+
+* **Uncontrolled Rollouts:** All instances update simultaneously, which can be risky, especially for critical applications. Users cannot pin to a specific version to avoid a buggy rollout.  
+* **Irreversible Changes:** This model has a "unidirectional and irreversible" change model, which can be a drawback for rollbacks.  
+* **Breaking schema change handling:** Handling backward-incompatible schema changes is challenging and likely necessitates downtime or complex migrations.
```
jakobmoellerdev, line 117: I dont see why breaking schema change handling is our go to here? The answer is to migrate instances onto a new version before a storage version migration and to allow users of RGD based CRDs to migrate, just like in Kubernetes. Then, once migration is completed (which can also be phased), the old version can be completely disabled for use.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+* **Simplicity:** This is the most straightforward method for managing RGD changes, easy to understand and implement.  
+* **Single Source of Truth:** Maintains one definitive source for the RGD definition.  
+* **Automatic Propagation:** Changes are automatically propagated, which is beneficial for rapid updates.
+
+**Cons**
+
+* **Uncontrolled Rollouts:** All instances update simultaneously, which can be risky, especially for critical applications. Users cannot pin to a specific version to avoid a buggy rollout.  
+* **Irreversible Changes:** This model has a "unidirectional and irreversible" change model, which can be a drawback for rollbacks.  
+* **Breaking schema change handling:** Handling backward-incompatible schema changes is challenging and likely necessitates downtime or complex migrations.  
+* **No Version Pinning**: Since changes are immediately rolled out, there is no way to pin instances to a specific revision of RGD.
```
jakobmoellerdev, line 118: Again, this assumes that we do not do phased migration through a webhook and 2 concurrently active APIs. I dont see why it wouldnt be possible to maintain previous apiVersions through status / state tracking until all instances were migrated.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+To deal with this we have these options:
+
+* **Single canonical RGD**: Only one instance of RGD exists, latest.  
+* **Different RGD for each change**: Each change requires a new RGD and Instance CRD.  
+* **Replicaset pattern**: Replicas automatically created and instances pinned to a replica  
+* **Sections with RGD for each iteration number:** Schema and Resources are defined per iteration. Instances are pinned to a section.
+
+## Replicaset Pattern
+
+Introduce a new `ResourceGraphDefinitionReplica` CRD that is a point-in-time snapshot of a RGD.  This is automatically created and managed by KRO reconciler. Instances reconcile against a specific version of the replica.
```
jakobmoellerdev, line 72: You will need an upper ResourceGraphDefinition / Deployment equivalent which then would map to a set version of Resources through an instance.

Maybe a fitting terminology equivalence would be:

`ResourceGraphDefinition` -> `ResourceGraph` (you call this `ResourceGraphDefinitionReplica`, but notice that it manages multiple resources) -> `Resources`

Then the RGD controller would own multiple ResourceGraphs and can then map those to the (currently named) instance controller.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+**Mechanism**
+
+* Crucially, every modification to the RGD specification automatically triggers the creation of a new `ResourceGraphDefinitionReplica` by the KRO.   
+* Instance reconciliation picks current `ResourceGraphDefinitionReplica` during first reconciliation and is pinned to that replica (via annotation).  
+* We can have a mode where pinning is not done and all instances are updated to the latest replica always.  
+* To mimic existing behavior we can set the history depth to 1 and enable no pinning mode where instances are always updated to the latest replica.
+
+**Migration**
+
+* **User-Controlled Opt-in:** Users have control over when they migrate their instance to a new version. Instance can be moved to a different version/replica by changing the annotation.
```
jakobmoellerdev, line 82: How would this move look like? I still dont see how you actually ensure smooth moving between versions. Also I dont see how an annotation can simply change your instance? the instance is backed by a schema and that schema is based on the CRD. if a new rgd replica introduces a new field, then that needs to be mapped into the CRD backing the instance, or you cannot migrate.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+**Cons**
+
+* **Uncontrolled Rollouts:** All instances update simultaneously, which can be risky, especially for critical applications. Users cannot pin to a specific version to avoid a buggy rollout.  
+* **Irreversible Changes:** This model has a "unidirectional and irreversible" change model, which can be a drawback for rollbacks.  
+* **Breaking schema change handling:** Handling backward-incompatible schema changes is challenging and likely necessitates downtime or complex migrations.  
+* **No Version Pinning**: Since changes are immediately rolled out, there is no way to pin instances to a specific revision of RGD.
+
+## RGD with version sections
+
+This approach integrates versioning directly into a single RGD (Resource Group Definition) resource. A single RGD exists for multiple versions with sections within the RGD.
```
jakobmoellerdev, line 122: Im pretty sure this is a typo or AI :D resource graph* definition

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+     schema:
+       # schema for v2...
+     resources:
+       # resources for v2...
+  servedVersion: v2
+```
+
+* When a new version is added to the RGD, the KRO controller will create a new version of the corresponding Custom Resource Definition (CRD) (e.g., `MyWebApp/v2`).  
+* Instances are not automatically migrated. Users must explicitly change their instance's `apiVersion` (e.g., from `my.api/v1` to `my.api/v2`) to opt into the new version.  
+* This option requires supporting a "migrate" lifecycle for instances and an "adoption" functionality for new instances.
```
jakobmoellerdev, line 146: I believe all approaches I have seen proposed will need a migration tooling and lifecycling as well as adoption management. either automated or manual. so for me this is not a differentiator

jakobmoellerdev (commented):
Thanks for this writeup, I loved seeing all the sections side-by-side. 

Considering all the various ideas, I strongly favor `RGD with version sections` as it can cleanly integrate with existing Kubernetes upgrade / CRD behavior, can be extended with automated instance migration, offers phased rollouts, and is the least intrusive, without need for additional CLI tooling. It also keeps the user ability to edit their version(s) in place and is a rather easy change in terms of API compared to the rest.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+
+1. **Resource Adjustments:** Changes to `.spec.resources`:  
+   1. Resources are added.  
+   2. Resources are removed.  
+   3. Resources are modified (e.g., changing a container image, updating a label).  
+2. **Schema Evolution:** Changes to the `.spec.schema`:  
+   1. Backward-compatible changes:  
+      1. A new optional field is added.  
+      2. A new field with a default value is added.  
+      3. A default value is removed
```
matthchr, line 23: Maybe not the point of this document, but this is only backwards compatible if the field is optional after the removal? Worth being specific about that?

Also I suppose debatably backwards compatible in that resources I deployed before will not match resources I deploy in the future, which you could argue is breaking behaviorally.

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+   2. Resources are removed.  
+   3. Resources are modified (e.g., changing a container image, updating a label).  
+2. **Schema Evolution:** Changes to the `.spec.schema`:  
+   1. Backward-compatible changes:  
+      1. A new optional field is added.  
+      2. A new field with a default value is added.  
+      3. A default value is removed  
+   2. Backward-incompatible changes:  
+      1. A field is removed.  
+      2. Adding/Changing defaults.
```
matthchr, line 26: inconsistent with above - removing a default is changing defaults, isn't it?

docs/design/proposals/versioning.md
matthchr: I also don't have access to this

docs/design/proposals/versioning.md
matthchr: I tend to agree with @jakobmoellerdev that in a perfect world we would handle this in kro without requiring users to make a separate CRD and do a manual migration.

With that said, IMO it's OK to punt on solving that whole we solve the more pressing problem of nonbreaking rollouts. The way I see it:
- Rollouts of nonbreaking (at least _schema nonbreaking_ changes is a **MUST** for any sort of beta/stable kro.
- Rollouts of breaking (schema breaking or otherwise breaking changes) is a nice to have for any beta/stable kro, but could come later as well.

Do you agree @jakobmoellerdev or do you see these two being more coupled than I am thinking?

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+In Kubernetes, RGD is an object. When any change is made to an RGD, we lose the previous configuration. There is no notion of the history of an object. Only the current view exists. 
+
+To deal with this we have these options:
+
+* **Single canonical RGD**: Only one instance of RGD exists, latest.  
+* **Different RGD for each change**: Each change requires a new RGD and Instance CRD.  
+* **Replicaset pattern**: Replicas automatically created and instances pinned to a replica  
+* **Sections with RGD for each iteration number:** Schema and Resources are defined per iteration. Instances are pinned to a section.
+
+## Replicaset Pattern
```
matthchr, line 70: minor: Weird that the bullet-list above ordering doesn't match the order of the sections below. Maybe fix?

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+
+Introduce a new `ResourceGraphDefinitionReplica` CRD that is a point-in-time snapshot of a RGD.  This is automatically created and managed by KRO reconciler. Instances reconcile against a specific version of the replica.  
+**Mechanism**
+
+* Crucially, every modification to the RGD specification automatically triggers the creation of a new `ResourceGraphDefinitionReplica` by the KRO.
```
matthchr, line 75:
```suggestion
* Crucially, every modification to the RGD specification automatically triggers the creation of a new `ResourceGraphDefinitionReplica` by KRO.   
```

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+
+As the KRO project matures, platform teams (RGD authors) need a structured way to evolve their \`ResourceGraphDefinition\` (RGD) resources over time. Currently, any change to an RGD in a cluster is immediately propagated to all existing instances across all namespaces upon their next reconciliation. This "rollout-on-reconcile" behavior is fast but lacks the control necessary for some production environments. The goal is to allow RGD authors to introduce changes while giving control over how the changes are safely rolled out in the cluster. This aligns with GitOps principles, where version pinning and controlled automation are crucial for stability.
+
+# Background
+
+To understand the scope of the problems to be addressed, we need to understand a few things that will be used as inputs in the proposal.
+
+## What can change in an RGD
+
+When we say an RGD author makes a change, it could mean one or more of the following:
```
matthchr, line 13: I feel we are missing a section (here? below?) talking about the mechanics of changes a bit more. In particular there are two classes of changes rolling them out is very different (I think). You call the classes out here but then we move on to talking about options without addressing the differences in class that probably require different handling.

**Class 1: (your item 1):** Changes to `.spec.resources`. Most of the document focuses on this, and all of the pinning, etc makes sense in this context, at least as far as I understand it.

**Class 2 (your item 2):** Changes to `.spec.schema`. Lots of open questions here:
1. Do back-compatible changes happen in the same apiversion, or do we publish a new apiversion? Maybe better summarized as the question: "When do we publish a new api version of the CRD underlying the RGD?"
2. For "New field with a default value is added." We classify this as nonbreaking but I don't how this schema change could be rolled out in an incremental fashion. We update the default _in the crd_, and CRDs are global, so isn't every resource in the cluster immediately at risk of grabbing an update? I think kube-apiserver applies these defaults at admission time so the next time somebody updates/edits/etc an existing resource it will get this new default (though that default may not _go_ anywhere, if it's associated with a `.spec.resources` change as well, as will commonly be the case?)

docs/design/proposals/versioning.md
matthchr: I had a similar comment [here](https://github.com/kubernetes-sigs/kro/pull/935/changes#r2830847998) saying (I think) the same thing. It's not clear to me how we handle the CRD schema updates.

docs/design/proposals/versioning.md
matthchr, line 97: Is this where we are today? (I think yes? but maybe worth calling out explicitly that this option is: "no change from today"?

docs/design/proposals/versioning.md
matthchr: Is this assuming that `apiVersion` gets updated whenever there is any change (regardless of to `.spec.schema` or `.spec.resources`?)

docs/design/proposals/versioning.md
matthchr: Agree - for example comparing this option with `Replicaset Pattern`, we would need to write automation to allow auto-migration of instances in both cases I think, but if we can do it for one, I don't see a huge difference to doing it for the other. It's just a question of what the UX is, explicit or implicit?

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+
+* When a new version is added to the RGD, the KRO controller will create a new version of the corresponding Custom Resource Definition (CRD) (e.g., `MyWebApp/v2`).  
+* Instances are not automatically migrated. Users must explicitly change their instance's `apiVersion` (e.g., from `my.api/v1` to `my.api/v2`) to opt into the new version.  
+* This option requires supporting a "migrate" lifecycle for instances and an "adoption" functionality for new instances.  
+* Changes within a version are immediately applied across instances using that version.
+
+**Migration**
+
+* **User-Controlled Opt-in:** Users have control over when they migrate their instance to a new version.  
+* **Automated Migration:** A mechanism can be built to automate the migration process. This would involve creating a new instance object at the new version, copying relevant fields, and deleting the old one. This requires careful handling of resource adoption and state.
```
matthchr, line 152: Not sure I am following this. Why would we require deleting the old instance and creating a new one - can't I just apply the new version over the top of the old one (same as any Kubernetes CRD version migration?)

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+* **Controlled Rollouts:** Users can adopt new versions at their own pace. Provides fine-grained control over when instances adopt new versions, crucial for critical systems or phased rollouts.  
+* **Version Pinning**: Instances are by default pinned to a specific RGD. Migration is explicit.  
+* **Clear Separation:** Versions are clearly defined and managed within a single RGD.  
+* **GitOps Friendly:** Users can pin their instance to a specific version in their Git repository.  
+* **No Impact on Existing Instances:** Introducing a new version has no effect on consumers of older versions.
+
+**Cons**
+
+* **Increased Complexity:** The RGD resource becomes more complex.  
+* **Breaking schema change handling:** If the schema has a breaking change across versions,  it would require conversion webhooks to be defined. There is no mechanism to do it in KRO.
```
matthchr, line 165: I think this con is true for every option so it's not clear to me that it's worth calling out/repeating over and over.

We just don't have an answer for breaking changes and I don't think we're trying to come up with an answer here either, right?

docs/design/proposals/versioning.md
```diff
@@ -0,0 +1,268 @@
...
+* **Version Pinning**: Instances are by default pinned to a specific RGD. Migration is explicit.  
+* **Clear Separation:** Versions are clearly defined and managed within a single RGD.  
+* **GitOps Friendly:** Users can pin their instance to a specific version in their Git repository.  
+* **No Impact on Existing Instances:** Introducing a new version has no effect on consumers of older versions.
+
+**Cons**
+
+* **Increased Complexity:** The RGD resource becomes more complex.  
+* **Breaking schema change handling:** If the schema has a breaking change across versions,  it would require conversion webhooks to be defined. There is no mechanism to do it in KRO.  
+* **CRD Management:** The controller needs to manage the lifecycle of multiple CRD versions.
```
matthchr, line 166: See https://github.com/kubernetes-sigs/kro/pull/935/changes#r2830847998 - I don't think we're avoiding this with any options (except maybe the "do nothing option").

docs/design/proposals/versioning.md
matthchr: This maybe is also causing some confusion (at least for me) below, as when we talk about a new version of the RGD are we talking about:
1. A new version of the associated CRD?
2. Any update at all is a new RGD "version" (but may not actually produce a new CRD version?)

matthchr (commented):
Left some thoughts.

A lot of open questions here for me atm.

- [ ] How does versioning as discussed in this document relate to CRD versions?
- [ ] How do updates to CRDs "roll out", given the CRD itself is global? (The obvious answer is versions but we don't talk about that here)
- [ ] Is RGD versioning explicit, with the user choosing the version names and what constitutes a new version (sections in the RGD with version number), or implict, with kro choosing when a new version is created (replicasets approach)?

docs/design/proposals/versioning.md
jakobmoellerdev: Hmm I think I agree we can push out this but imho the proposal of creating a new RGD for breaking changes will also require setup of an adoption workflow so users can migrate between RGDs. for sure we could solve this though 👍

shivansh-gohem:
Hey @barney-s !!! Jakob pointed me here from my PR (#1051). I actually just built a lightweight versioning implementation that automatically merges CRD versions when the apiVersion is bumped. Would love to sync and see if this fits into the broader KREP design as a stepping stone!

allamand:
The versioning question comes up consistently from many organisations already using kro in production — here's the concrete use case:

  Scenario: A platform team manages ~80 EKS clusters provisioned via kro RGDs composing ACK resources (hub-and-spoke pattern, ~1,200 developers). They need to introduce a breaking change to their cluster-provisioning RGD (e.g., adding a required field, restructuring the resource graph).

  Desired workflow:

  1. Deploy a new version of the RGD with the breaking schema change
  2. Have kro serve both old and new CRD versions simultaneously
  3. Migrate existing cluster CRs one-by-one from old version to new version (progressive rollout via GitOps — dev → staging → prod)
  4. Once all CRs are migrated, deprecate the old version

  Why "new RGD for breaking changes" doesn't work at scale:

  - Managing two separate CRD kinds (eks-cluster-v1 / eks-cluster-v2) and manually migrating CRs between them is operationally heavy at 80+ clusters
  - It breaks GitOps workflows — you can't do an in-place migration, you need to delete and recreate
  - It forces direct cluster access (kro migrate CLI) instead of declarative GitOps-only operations

  Today's blocker: The RGD API version is immutable. You can't create v1alpha2 from the same RGD — the only workaround is renaming RGDs, which means managing parallel CRD kinds.

  I strongly align with @jakobmoellerdev's position on "RGD with version sections" — it maps cleanly to Kubernetes-native CRD multi-version serving, supports phased migration through
  conversion webhooks, and doesn't require CLI tooling or direct cluster access. This is the pattern enterprise customers expect from a Kubernetes-native tool.

  For non-breaking changes, the ReplicaSet pattern with version pinning makes sense. But for breaking changes, the "create a new RGD" recommendation is a significant adoption blocker
  for production organizations. Even if breaking change support comes in a later phase, the design should not preclude it — and ideally the chosen approach for non-breaking changes
  should be extensible to breaking changes without a fundamental architecture shift.

  This is not a single-customer ask — I've had this exact conversation with multiple organizations in the last few weeks.

docs/design/proposals/versioning.md
dudo: Given the following:

```yaml
spec:
  schema:
    apiVersion: v1alpha1
    kind: Application
    group: platform.acme.com
```

When you say new RDG, do you mean

```yaml
spec:
  schema:
    apiVersion: v1alpha2
    kind: Application
    group: platform.acme.com
```

or

```yaml
spec:
  schema:
    apiVersion: v1alpha1
    kind: TestApplication
    group: platform.acme.com
```

PR #936, opened 2026-01-06
feat: add Collections support + runtime/controller rewrite

a-hilaly:
Implements KREP-002 collections, but getting there required rewriting the
runtime and instance controller.

1/ collections (KREP-002)

Adds `forEach` to resource definitions, enabling a single resource block to
expand into N Kubernetes resources based on runtime data. Instead of writing
N blocks for N resources, users express "create N resources where N comes
from spec" and kro handles the rest.

Docs: https://a-hilaly.github.io/kro/next/docs/concepts/rgd/resource-definitions/collections

2/ `pkg/runtime` rewrite

The old `Synchronize` method tried to do everything in one pass. Now the
controller interacts with `runtime.Node` through four methods:
`GetDesired()`, `SetObserved()`, `IsIgnored()`, `IsReady()`. Graph and
runtime unified around a concrete `Node` type, replacing the
`ResourceDescriptor` interface abstraction.

3/ `pkg/applyset` refactor

Moved to `pkg/controller/instance/applyset/` and redesigned as stateless:
`Project()` computes metadata, `Apply()` handles SSA, `Prune()` cleans up
orphans.

4/ some bugs fixes found on the refactor way:

- CEL data pending errors use delayed requeue instead of exponential backoff
- String interpolation resolves correctly
- Instance status shrinks when expressions stop resolving

### DIFF
--- a/website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
+++ b/website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
@@ -456,6 +456,9 @@ expressions (AND semantics):
 In this example, `workerPods` is only considered ready when **every** pod in the
 collection has `status.phase == 'Running'`. Dependent resources wait for this
 condition before proceeding.
+:::important
+If the collection is empty (zero items), it is considered ready.
+:::
 
 :::tip
 Without `readyWhen`, a collection is considered ready once all resources are
@@ -533,26 +536,16 @@ array changes, kro creates, updates, or deletes resources to match.
 
 ### Scaling Up
 
-When items are added to the source array, kro creates new resources:
-
-```yaml
-# Instance spec change:
-# workers: ["alice", "bob"] → ["alice", "bob", "charlie"]
-#
-# Result: kro creates a new pod for "charlie"
-```
+When items are added to the source array, kro creates new resources.  
+Example: if `workers` changes from `["alice", "bob"]` to `["alice", "bob", "charlie"]`,
+it creates one new pod for **charlie**.
 
 ### Scaling Down
 
 When items are removed from the source array, kro automatically deletes the
-corresponding resources:
-
-```yaml
-# Instance spec change:
-# workers: ["alice", "bob", "charlie"] → ["alice"]
-#
-# Result: kro deletes pods for "bob" and "charlie"
-```
+corresponding resources.  
+Example: if `workers` changes from `["alice", "bob", "charlie"]` to `["alice"]`,
+it deletes the pods for **bob** and **charlie**.
 
 This cleanup happens automatically through kro's applyset mechanism - orphaned
 resources are pruned on each reconciliation.
@@ -560,14 +553,12 @@ resources are pruned on each reconciliation.
 ### Empty Collections
 
 An empty array results in zero resources - this is not an error:
-
-```yaml
-# workers: []
-# Result: zero pods created (collection exists but is empty)
-```
+for example, `workers: []` means no pods are created (the collection exists but is empty).
 
 This is useful for optional features that can be enabled by adding items to an
 initially empty array.
+
+Empty collections are considered ready, since there are no items to wait on.
 
 :::important
 If any iterator in a cartesian product is empty, zero resources are created.
--- a/website/docs/docs/concepts/rgd/02-resource-definitions/03-readiness.md
+++ b/website/docs/docs/concepts/rgd/02-resource-definitions/03-readiness.md
@@ -64,7 +64,8 @@ This ensures `${database.status.endpoint}` has a valid value when the app is cre
 
 ## What You Can Reference
 
-`readyWhen` expressions can only reference the resource itself (by its `id`):
+`readyWhen` expressions can only reference the resource itself (by its `id`).
+For collections, use the `each` keyword to reference the current item:
 
 ```kro
 # ✓ Valid - references the resource itself and returns boolean
@@ -72,6 +73,19 @@ This ensures `${database.status.endpoint}` has a valid value when the app is cre
   readyWhen:
     - ${deployment.status.availableReplicas > 0}
     - ${deployment.status.conditions.exists(c, c.type == "Available" && c.status == "True")}
+```
+
+```kro
+# ✓ Valid for collections - uses each for per-item readiness
+- id: workerPods
+  forEach:
+    - worker: ${schema.spec.workers}
+  readyWhen:
+    - ${each.status.phase == 'Running'}
+  template:
+    kind: Pod
+    metadata:
+      name: ${schema.metadata.name + '-' + worker}
 ```
 
 ```kro


Makefile
```diff
@@ -321,7 +321,7 @@ cli:
 
 .PHONY: test-e2e
 test-e2e: chainsaw ## Run e2e tests
-	$(CHAINSAW) test ./test/e2e/chainsaw
+	$(CHAINSAW) test --parallel 8 ./test/e2e/chainsaw
```
jakobmoellerdev: this feels awfully random of a parameter value :D

Makefile
```diff
@@ -126,7 +126,7 @@ vet: ## Run go vet against code.
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests. Use WHAT=unit or WHAT=integration
 ifeq ($(WHAT),integration)
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_VERSION) --bin-dir $(LOCALBIN) -p path)" go test -v ./test/integration/suites/... -coverprofile integration-cover.out -ginkgo.v
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_VERSION) --bin-dir $(LOCALBIN) -p path)" go run github.com/onsi/ginkgo/v2/ginkgo -p -v --cover --coverprofile=integration-cover.out ./test/integration/suites/...
```
jakobmoellerdev: maybe we wanna use go tool directive for this if we already run with ginkgo native. that said, any reason we really need to do this?

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+	Apply(ctx context.Context, resources []Resource, mode ApplyMode) (*ApplyResult, Metadata, error)
+
+	// Prune deletes orphaned resources (those with applyset label but not in KeepUIDs).
+	// Pass Project().PruneScope() to search both current batch locations AND parent memory.
+	Prune(ctx context.Context, opts PruneOptions) (*PruneResult, error)
+}
+
+// Resource is an input to Apply.
+type Resource struct {
+	// ID is a stable identifier provided by the controller (e.g., "deployment" or "workers-0").
```
jakobmoellerdev, line 70: the id is not provided by the controller but is standardized by the KEP and is calculated based on the parent: https://github.com/kubernetes-sigs/kro/pull/936/files#diff-085f40c5f3afbd6fd970437cd201e59ef39f6c9b32abef8eca544020d62cfe47R581

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+	// Pass the superset scope (union of batch + parent) to ensure
+	// prune finds all orphans.
+	Scope *PruneScope
+	// Concurrency limits parallel delete operations. 0 = len(candidates).
+	Concurrency int
+}
+
+// PruneScope defines the search space for orphan detection.
+type PruneScope struct {
+	GroupKinds sets.Set[schema.GroupKind]
```
jakobmoellerdev, line 104: i dont understand why we would want differing pruning scopes here. mind explaining a bit on when this would be useful?

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+	}
+	for ns := range parentNamespaces {
+		namespaces.Insert(ns)
+	}
+
+	return a.buildMetadata(gks, namespaces), nil
+}
+
+// Apply runs SSA on all resources and returns batch-only metadata.
+// Caller should call Prune separately after Apply succeeds.
```
jakobmoellerdev, line 238: what was the reasoning for splitting apply/prune into two method calls?

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+
+	// Work on a copy to avoid mutating caller's object
+	obj := r.Object.DeepCopy()
+
+	// Inject applyset membership label (required for prune to find managed resources)
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[ApplysetPartOfLabel] = a.applySetID
```
jakobmoellerdev, line 385: we should warn/fail here on a conflict with another apply set. i would argue fail is fine, but if options.Force is true this can become an issue

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+func (a *ApplySet) applyResource(
+	ctx context.Context,
+	r Resource,
+	mapping *meta.RESTMapping,
+	options metav1.ApplyOptions,
+) ApplyResultItem {
+	item := ApplyResultItem{ID: r.ID}
+
+	// Work on a copy to avoid mutating caller's object
+	obj := r.Object.DeepCopy()
```
jakobmoellerdev: do we really need a deepcopy here? this is quite expensive for kro. since youre using dyn client, i think applied is a different pointer no?

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+		}
+		if ns == "" {
+			ns = metav1.NamespaceDefault
+		}
+		return dynResource.Namespace(ns)
+	}
+	return dynResource
+}
+
+func (a *ApplySet) resolvedNamespace(mapping *meta.RESTMapping, obj *unstructured.Unstructured) (string, bool) {
```
jakobmoellerdev, line 434: nit: shared logic on namespace with resourceClient above, probably can be unified

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+}
+
+func (a *ApplySet) prune(
+	ctx context.Context,
+	mappings []*meta.RESTMapping,
+	namespaces sets.Set[string],
+	keepUIDs sets.Set[types.UID],
+	concurrency int,
+) ([]PruneResultItem, error) {
+	labelSelector := fmt.Sprintf("%s=%s", ApplysetPartOfLabel, a.applySetID)
```
jakobmoellerdev: nit: this should be a label selector that is built once statically when the apply set gets initialized

pkg/controller/instance/applyset/applyset.go
```diff
@@ -0,0 +1,594 @@
...
+	// Track candidates with their GVR for deletion
+	type pruneCandidate struct {
+		obj *unstructured.Unstructured
+		gvr schema.GroupVersionResource
+	}
+	var candidates []pruneCandidate
+
+	// List resources per (GVR, namespace) pair for namespaced resources,
+	// or once per GVR for cluster-scoped.
+	for _, mapping := range mappings {
```
jakobmoellerdev, line 471: this listing should not be done sequentially imho.

pkg/controller/instance/applyset/doc.go
```diff
@@ -0,0 +1,54 @@
...
+//
+// # Workflow
+//
+// 1. Project() computes union metadata (batch + parent annotations). Error = requeue.
+// 2. Controller patches parent with union metadata (before apply/prune)
+// 3. Apply() runs SSA and returns batch-only metadata
+// 4. Prune() deletes orphans using scope from Project().PruneScope()
+// 5. If prune completes, controller shrinks parent annotations to batch metadata
+//
+// ApplySet is stateless - all methods are pure functions of their inputs.
```
jakobmoellerdev, line 31: I think its a bit strange that we must always use Apply + Prune before parent annotations are shrunk. Why did you choose to split this out?

now we can have a scenario of apply + prune=>fail and then the controller parent annotations are not shrunk. Im not sure if this introduces side effects.

By definition, ApplySets are not stateless, so im not sure how you mean this

pkg/controller/instance/applyset/doc.go
```diff
@@ -0,0 +1,54 @@
...
+//
+// # Metadata Semantics
+//
+// Apply() returns batch-only metadata: just the GKs and namespaces of this batch.
+// Project() returns union metadata: batch + parent annotations (the "memory").
```
jakobmoellerdev, line 36: do I understand correctly that this is not real memory but instead a hard query on the REST Mapper?

pkg/cel/ast/inspector.go
```diff
@@ -107,10 +107,14 @@ type Inspector struct {
 }
 
 // knownFunctions contains the list of all CEL functions that are supported
+//
+// we need a better way to manage this list going forward... perhaps a Check
```
jakobmoellerdev, line 111: plus one on this. is it mandatory in this pr or can it be extracted?

pkg/controller/instance/const.go
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package instance
+
+const (
```
jakobmoellerdev: I know this wasnt there before but can we please add some docs on when and why these states exist and when they should be set? its honestly already probably too much tbh.

pkg/controller/instance/context.go
```diff
@@ -0,0 +1,98 @@
...
+	Runtime  runtime.Interface
+	Instance *unstructured.Unstructured
+	Config   ReconcileConfig
+
+	Mark         *ConditionsMarker
+	StateManager *StateManager
+}
+
+// NewReconcileContext constructs a new sequential reconciliation context.
+func NewReconcileContext(
```
jakobmoellerdev, line 58: oh my lord the one character variables are... something.

this constructor is imho warranting some more explanation

pkg/controller/instance/deletion.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return c.removeFinalizer(rcx)
+}
+
+func (c *Controller) deleteOne(rcx *ReconcileContext, node *runtime.Node) error {
+	rid := node.Spec.Meta.ID
+	desc := node.Spec.Meta
+
+	if desc.Type == graph.NodeTypeExternal {
+		rcx.StateManager.ResourceStates[rid] = &ResourceState{State: ResourceStateSkipped}
```
jakobmoellerdev: state manager is not safe for concurrent use, i think its important to note that. this is also different from using pruning for the apply set.

pkg/controller/instance/deletion.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	if len(list.Items) == 0 {
+		rcx.StateManager.ResourceStates[rid] = &ResourceState{State: ResourceStateDeleted}
+		return nil
+	}
+
+	// Delete each item
+	allDeleted := true
+	for _, item := range list.Items {
```
jakobmoellerdev: this should be concurrent as the collection is unordered.

pkg/graph/variable/variable.go
```diff
@@ -151,3 +162,29 @@ func (r ResourceVariableKind) IsDynamic() bool {
...
+
+// Priority returns the evaluation priority for this kind.
+// Lower values = higher priority (evaluated earlier).
+//
+// Priority order: Static/IncludeWhen (0) < Dynamic/ReadyWhen (1) < Iteration (2)
+//
+// This matters because the same expression can appear in multiple variables
+// with different kinds. We use the lowest priority so the expression is
+// evaluated early enough for all consumers.
+func (r ResourceVariableKind) Priority() int {
```
jakobmoellerdev: hmmm: imho this is not sufficient/accurate as mapping. there is a diff for dynamic and static iteration variables. if you have a forEach, then still the typing of this variable can be statically determined.

I think for example that if the base variable for the iteration variable is statically evaluatable, then it should be considered before dynamic, or readyWhen

pkg/metadata/labels.go
```diff
@@ -34,6 +35,11 @@ const (
 const (
 	NodeIDLabel = LabelKROPrefix + "node-id"
 
+	// Collection labels for tracking collection membership and position.
+	// These enable querying collection resources and understanding their position.
+	CollectionIndexLabel = LabelKROPrefix + "collection-index"
```
jakobmoellerdev, line 40: this label implies presence of exactly one collection, however Im not sure if we cannot have conflicting collection indices. is it possible to have 2 resources with collections that end up with the same id? im thinking probably not, just want to make sure here.

pkg/runtime/collection.go
```diff
@@ -0,0 +1,149 @@
...
+// evaluatedDimension pairs a forEach dimension name with its evaluated values.
+type evaluatedDimension struct {
+	name   string
+	values []any
+}
+
+// cartesianProduct computes the cartesian product of multiple dimensions.
+// Each dimension's Values are iterated over, producing all combinations.
+// Values can be any type - scalars, lists, maps - they are not flattened.
+func cartesianProduct(dimensions []evaluatedDimension) []map[string]any {
```
jakobmoellerdev, line 87: a few things on collections

we should probably set a maximum dimension number for now (arbitrary, maybe 5?)

if two dimensions share dim.name, later ones overwrite earlier values in the map. are we validating against this?

test/integration/environment/setup.go
```diff
@@ -157,7 +157,7 @@ func (e *Environment) setupController() error {
 	dc := dynamiccontroller.NewDynamicController(
 		zap.New(zap.WriteTo(e.ControllerConfig.LogWriter), zap.UseDevMode(true)),
 		dynamiccontroller.Config{
-			Workers:         3,
+			Workers:         10,
```
jakobmoellerdev: that looks very arbitrary, any specific reason?

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
jakobmoellerdev, line 1: im missing a cross reference to the collection labels you introduced when setting up items from collectiosn, I think theyre an important API.

Makefile
a-hilaly: Haha fair - it's not scientific, just what made my local test runs bearable after running them a hundred times for this PR. happy to make it configurable via env var or match nproc or similar if you prefere

Makefile
a-hilaly: Switched to `go tool` directive. 

>that said, any reason we really need to do this?

Ginkgo flags were just easier for the test setup I needed when i was iterating on the collections feature/rework.

pkg/controller/instance/applyset/applyset.go
a-hilaly: Mmmmmm I think there is a mix up here - the ID in this situation is a controller internal identifier for correlating apply results back to inputs. The KEP standardized ID is computed seperately in the `ID()` funciton in L581

pkg/controller/instance/applyset/applyset.go
a-hilaly: the new `ApplySet` is stateless - no state carried between method calls. and `Prune` needs to know where to look for orphans (which GKs which namespaces etc...) but it doesn't have access to the resources list. the thing is that now the responsability is delegated to the controller to call Project() to compute the scope (batch + parent memory) then passes it to `Prune()`. The parameter exists because Prune can't compute it internally without breaking statelessness.

That said my whole purpose was to 1/ simplify 2/ remove tracking from ApplySet and make it purely stateless, and haven't really focused on the signatures/interface. I believe we can build better abstractions on top of it (some sort `ApplySet` with tracking that could also hide the pruning/scope process). Unless this is a hard blocker, happy to re-iterate on the abstractions later.

pkg/controller/instance/applyset/applyset.go
a-hilaly: same reason as abov - statelessness. now Apply and Prune are pure functions with no shared state, they are reponsible of managing resources nothing more nothing less, we could build on top of this abstraction if we want to lighten the responsability of the instance reconciler.

pkg/controller/instance/applyset/applyset.go
a-hilaly: I agree that we should handle this, but I think it's part of a bigger picture, AFAIK kro doesn't handle child resource conflicts across instances/RGDs either. We need to track membership properly and handle conflicts holistically, not just at the applyset label level. I'd rather have us tackle this in a follow up PR that tackles conflicts in general, does that work for you?

pkg/controller/instance/applyset/applyset.go
a-hilaly: Good point - removing the DeepCopy.

pkg/controller/instance/applyset/doc.go
a-hilaly: You're right that ApplySet as a concept is stateful. This implementation intentionally doesn't try to achieve that I stripped it down to a minimal, stateless primitives (the instance controller is responsible of calling). `Apply`/`Prune` could be merged by building a tracking layer on top, which is fine. Tracking feels like a gray area between controller logic and applyset mechanics, and I'm not sure where it belongs yet. 

Regardless i'm not married to this design - we can address in a folow up or here if it's a hard blocker. just a lot of moving pieces in this PR already.

pkg/cel/ast/inspector.go
a-hilaly: I need this for the `lists.range()` support, it is used in the collections documentation https://a-hilaly.github.io/kro/next/docs/concepts/rgd/resource-definitions/collections#iterator-variables

pkg/controller/instance/applyset/doc.go
a-hilaly: The memory is stored in parent annotations as GK strings. When we prune, we use REST Mapper to convert those GKs to GVRs for the actual API calls. basically annotations = storage and RESTMapper == resolution at runtime.

pkg/controller/instance/deletion.go
a-hilaly: Collections are ordered (how it's implemented today) but since we don't have propagation control yet this is a valid optimization. Will defer to a follow up/future discussions.

pkg/graph/variable/variable.go
a-hilaly: This made me think a little bit and I believe this is more of a modeling (or even Naming) issue, not a bug/inaccuracy. A few thoughts on where i'm coming from:

1/ This PR "also" reworks runtime to evaluate expressions on demand (for example ForEach expression are only evaluated during expansion and never in advance, even if they are static), contrarily to what we did before (evaluate as many as possible) where priority mattered even more.
  more precisely, prior to this patch we need did evaluate ALL static expressions in advance https://github.com/kubernetes-sigs/kro/blob/main/pkg/runtime/runtime.go#L116-L121 - regardless of if they were needed for a reconciliation or not (which i've came to realise was a premature optimization)

2/ Priority is only used during expression deduplication in `getOrCreateExpr` - basically when the same expression appears in different contexes, we pick the lowest priority. [stops typing]

As i was typing this reply, i realized that we do not need a Kind priority system anymore because everything is evaluated on demand.

pkg/metadata/labels.go
a-hilaly: a few thoughts to clarify the utility here:
1/ collection items, will each have a unique index regardless of the output of ForEach products
2/ collection items will never have the same index, the controller guarantees items will have a stable ID in the collection (at least during one reconciliation). 
3/ Collection items are identified primarily by the combination of node-id + instance-id, the collection index is only useful when it comes to understanding the expansion ordering.

pkg/runtime/collection.go
a-hilaly: I like that, can we address this in a follow up PR?

>if two dimensions share dim.name, later ones overwrite earlier values in the map. are we validating against this?

yes. two ForEach entries cannot have the same dimension name + each dimension entry is evaluated on it's own.

test/integration/environment/setup.go
a-hilaly: At some point seconds started to matter, while working on this PR, i might've ran integration tests a good 200~ times. happy to revert

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
a-hilaly: Good catch

pkg/controller/instance/applyset/doc.go
jakobmoellerdev: acked, as i understand that means that as long as annotations are up to date and our rest mapper is reasonably fresh, no additional api calls. works for me :+1:

pkg/controller/instance/deletion.go
jakobmoellerdev: sure, works for me. would you mind adding a todo? im guessing this code will need a few passes anyhow later

pkg/graph/variable/variable.go
jakobmoellerdev: great discussion here actually, glad you elaborated. actually I agree. we should likely kill this in favor of lazily evaluating always

pkg/runtime/collection.go
jakobmoellerdev: of course, lets add todos wherever you feel like you wanna restrict PR scope

pkg/metadata/labels.go
jakobmoellerdev: I think my point rather stems from the argument of two collections targeting the same kubernetes resource and then overriding labels. (in the same instance that is).

pkg/controller/instance/const.go
a-hilaly: Added

pkg/controller/instance/context.go
a-hilaly: Added more godocs

pkg/metadata/labels.go
a-hilaly: I see what you mean, i believe that's a bigger problem with all resources, we need proper handling/ownership checks very applying resources, and report on conflicts. I'll craft a github issue for this topic

pkg/controller/instance/applyset/applyset.go
jakobmoellerdev: just finsihed review on it and I think contract is still satisfied, so gtg for me :+1:

pkg/controller/instance/applyset/applyset.go
jakobmoellerdev: hmm im wondering on this. I believe that this is contractually defined in apply set. still unsure if this is deferrable.

pkg/metadata/labels.go
jakobmoellerdev: do you mind linking that here?

a-hilaly: https://github.com/kubernetes-sigs/kro/issues/957

pkg/controller/instance/applyset/applyset.go
a-hilaly: addressed applysetID conflicts in https://github.com/kubernetes-sigs/kro/pull/936/changes/788116df7f0c2e20c74f3e953663acfe27af356c

website/static/crds/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -115,6 +115,26 @@ spec:
...
                       - metadata
                       type: object
+                    forEach:
+                      description: |-
+                        ForEach expands this resource into a collection of resources.
+                        Each entry binds a variable name to a CEL expression that evaluates to an array.
+                        kro creates one resource instance for each element in the array.
+                        With multiple entries, kro creates the cartesian product of all combinations.
+                        Use the variable directly in template expressions (e.g., ${region}).
+                        Example: [{"region": "${schema.spec.regions}"}, {"tier": "${schema.spec.tiers}"}]
```
jakobmoellerdev, line 125: i just noticed this, but I think we should fail or at least warn if not all variables in the forEach expression have been used. WDYT about that?

a-hilaly: Good call out, added in https://github.com/kubernetes-sigs/kro/pull/936/changes/788116df7f0c2e20c74f3e953663acfe27af356c..5d824d7fcfe36a18868a6edea72b271259e9ae11

a-hilaly: I believe we need to do something similar for other resources too. make sure there are CEL expressions in the identifier fields, and reject namespace setters for cluster scoped resoucres.

pkg/controller/instance/applyset/applyset_test.go
```diff
@@ -0,0 +1,1282 @@
...
+			expectedID := ID(parent)
+			if metadata.ID != expectedID {
+				t.Errorf("Project().ID = %q, want %q", metadata.ID, expectedID)
+			}
+		})
+	}
+}
+
+func TestPrune_ParentAnnotationsContributeToPruneScope(t *testing.T) {
+	ctx := context.Background()
```
jakobmoellerdev: nit: you can use t.Context() by now

pkg/controller/instance/status.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (u *unstructuredWrapper) GetConditions() []v1alpha1.Condition {
+	raw, found, err := unstructured.NestedSlice(u.Object, "status", "conditions")
+	if err != nil || !found {
+		return nil
+	}
+
+	b, _ := json.Marshal(raw)
+	var out []v1alpha1.Condition
+	_ = json.Unmarshal(b, &out)
```
jakobmoellerdev: should we really ignore errors here?

pkg/controller/instance/status.go
```diff
@@ -0,0 +1,175 @@
...
+	return out
+}
+
+func (u *unstructuredWrapper) SetConditions(conds []v1alpha1.Condition) {
+	b, _ := json.Marshal(conds)
```
jakobmoellerdev: ditto

pkg/controller/instance/status.go
```diff
@@ -0,0 +1,175 @@
...
+		return err
+	})
+}
+
+func (rcx *ReconcileContext) initialStatus() map[string]interface{} {
+	inst := rcx.Instance
+
+	conds := condSet.For(&unstructuredWrapper{inst}).List()
+
+	b, _ := json.Marshal(conds)
```
jakobmoellerdev: ditto

jakobmoellerdev:
while i appreciate continuous improvements this should be done in a separate PR imho. The last 2 commits alone were worth 2 prs.

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -500,6 +507,100 @@ expression:
...
+If you expect downstream resources to wait on actual items, ensure the
+collection is non-empty or use additional conditions in `includeWhen`.
+
+### Map Iteration Order Is Not Deterministic
+
+Map iteration order in Go is randomized. If you iterate over map keys or values,
+convert the map to a sorted array first to make resource ordering stable.
+Unstable ordering can lead to noisy diffs and unexpected reconciliation churn.
+
+### Identity Must Include All Iterator Dimensions
```
jakobmoellerdev: I think this should be fairly easy to check at compile. we can verify that name / namespace include all iterator dimensions by checking their variable in the appropriate fieldpath. WDYT?

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -500,6 +507,100 @@ expression:
...
+An empty array produces zero resources. The collection is still considered
+**ready**, because there are no items to wait on. This also applies to cartesian
+products where any iterator is empty (`2 × 0 = 0`).
+
+If you expect downstream resources to wait on actual items, ensure the
+collection is non-empty or use additional conditions in `includeWhen`.
+
+### Map Iteration Order Is Not Deterministic
+
+Map iteration order in Go is randomized. If you iterate over map keys or values,
```
jakobmoellerdev: I think this mutation is very unnice to do in CEL. if you have a "nice" way of doing that its probably worth adding here.

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
a-hilaly: already done in 4b35e62623f7c47d3c41b660b056c5af2b918f7a

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
a-hilaly: Yep. We do not have a better way that sort the keys for now. best we can do is provide custom CEL functions for that.

a-hilaly:
@jakobmoellerdev I jusst clipped the last two commits - sending them as seperate PRs.

a-hilaly:
opened https://github.com/kubernetes-sigs/kro/pull/970 and https://github.com/kubernetes-sigs/kro/pull/971

ross-grinvalds:
@a-hilaly, thank you for your work on this. It is very much appreciated!

Issue #907, opened 2025-12-17
No Doc about CEL HomogeneousAggregateLiterals | Bug?


AdamDelle:
### Documentation Issue

**Quick disclaimer**: I am unsure whether this is about missing documentation, a bug or a mistake on my end.

While getting to know kro I stumbled upon its use of CEL Expressions and I am wondering, how powerful they make kro. Especially whether you could use `.map(x,t)` to create mixed type list and map literal values.

I tried using a list of strings of keys of a secret that is already deployed on my cluster to provide those secrets to a deployment:

```
# spec.schema...
      deployment:
        secretName: string | default="my-secret"
        secretNames: '[]string | default=["SECRET_1","SECRET_2"]'

# resources...
      env: '${schema.spec.deployment.secretNames.map(s, { "name": s, "valueFrom": { "secretKeyRef": { "name": schema.spec.deployment.secretName, "key": s } } } ) }'

# goal (deployment)
    env:
        - name: SECRET_1
          valueFrom:
            secretKeyRef:
              name: my-secret
              key: SECRET_1
        - name: SECRET_2
          valueFrom:
            secretKeyRef:
              name: my-secret
              key: SECRET_2

```

However this results in the following error in the logs of the kro operator:

I was able to apply a **RGD** that contains the provided snippet and as far as I understand, it passed RGD validation, since it had `State: Active`. 

However the **instance** had `State: ERROR` and `Ready:Unknown` without a clear error message, only `Reason: Awaiting Reconciliation` and I found this error in the logs of the kro controller:

```
ERROR    dynamic-controller    Error syncing item, requeuing with rate limit    {"item": {"name":"api-kro-instance","namespace":"demo-test"}, "error": "failed to add resource to applyset: object {apps/v1, Kind=Deployment /api-kro} is not json marshallable: json: error calling MarshalJSON for type applyset.ApplyableObject: json: unsupported type: map[ref.Val]ref.Val"} 
```

[...]

Issue #905, opened 2025-12-17
dynamic-controller error: “failed to add resource to applyset … is not json marshallable” (unsupported type map[ref.al[]ref.Val)


kahirokunn:
### Description

**Observed Behavior**:
After applying the `ResourceGraphDefinition` and then creating an instance, the kro `dynamic-controller` keeps requeuing and logs this error. The `Cluster` resource is not reconciled successfully.

**Expected Behavior**:
The instance should reconcile successfully and create/update the referenced `cluster.x-k8s.io/v1beta1` `Cluster` resource.

**Reproduction Steps** (ResourceGraphDefinition and Instance):
1. Apply this `ResourceGraphDefinition`:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: k0smotrondockercluster.cluster.example.com
spec:
  schema:
    group: cluster.example.com
    apiVersion: v1alpha1
    kind: K0smotronDockerCluster
    spec:
      # ClusterClass
      clusterclassName: string | required=true

      # k0s version (e.g. v1.34.2+k0s.0)
      k0sVersion: string | required=true pattern="^v[0-9]+\.[0-9]+\.[0-9]+\+k0s\.[0-9]+$"

      # Replicas
      controlPlaneReplicas: integer | required=true
      machineDeploymentReplicas: integer | required=true

      # Cluster network configuration (with defaults)
      podCIDR: string | default="192.168.0.0/16" pattern="^([0-9]{1,3}\\.){3}[0-9]{1,3}/[0-9]{1,2}$"
      serviceCIDR: string | default="10.128.0.0/12" pattern="^([0-9]{1,3}\\.){3}[0-9]{1,3}/[0-9]{1,2}$"

      # Base cluster kubeconfig secret reference
      baseCluster:
        kubeconfigRef:
          name: string | required=true
          namespace: string | required=true

      template:
        metadata:
          annotations: "map[string]string | default={}"
          labels: "map[string]string | required=true"

    status:
      clusterStatus:
        conditions: ${cluster.status.conditions}
        controlPlaneReady: ${cluster.status.controlPlaneReady}
        infrastructureReady: ${cluster.status.infrastructureReady}
        observedGeneration: ${cluster.status.observedGeneration}
        phase: ${cluster.status.phase}
        v1beta2:
          conditions: ${cluster.status.v1beta2.conditions}

  resources:
    - id: clusterClass
      externalRef:
[...]

PR #940, opened 2026-01-09
fix: Convert CEL type to Go type recursively

cirias:
Fix the bug described in https://github.com/kubernetes-sigs/kro/issues/905 and https://github.com/kubernetes-sigs/kro/issues/907

The bug is, for nested lists and maps have more than one level, `GoNativeType` only converts the highest level, because `interface{}` is used in the reflect type description(`[]interface{}` for list, and `map[string]interface{}` for map), then all the original children will just be kept as its original type.

The fix is done by recursively calling `GoNativeType` for all the children.


jakobmoellerdev (commented):
/lgtm

Issue #845, opened 2025-11-21
Support referencing resources in includeWhen CEL expression


dev-marina:
### Feature Description

**Problem Statement**:

Current Behaviour supports only reference the "schema" variable.
I would like to be able to use externalRef in includeWhen statement. Here is example how it can look like:
 
    `apiVersion: kro.run/v1alpha1
     kind: ResourceGraphDefinition
     metadata:
     name: foo
     spec:
        schema:
        apiVersion: v1
        kind: Foo
        spec:
          crdName: string | required=true description="Name of the CRD"

    resources:
    - id: serviceAccount
      externalRef:
      apiVersion: v1
      kind: ServiceAccount
      metadata:
         name: service-account-name 
         namespace: ${schema.spec.namespace}
    
    - id: storageBucket
      template:
        apiVersion: storage.cnrm.cloud.google.com/v1beta1
        kind: StorageBucket
        metadata:
          name: ${schema.metadata.name}-bucket
          namespace: ${schema.metadata.namespace}
      includeWhen:
        - ${serviceAccount.metadata.exists(x, x.name == 'test')}`

**Proposed Solution**:

Users can use  externalRef resource in includeWhen cel expression to reference to externalRef resource

**Alternatives Considered**:

**Additional Context**:

It used to be possible to use externalRef in includeWhen statement before this [PR](https://github.com/kubernetes-sigs/kro/commit/14a86b96bd47426b151f58017f1e51344f7f86f5#diff-3393f5070e9756c4a6e99e84aa8e5b7c2061cdac97c143468f073d36c6d536f2) where behaviour was changed.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

conqquest:
i'm interested to working on this feature , please assign me

a-hilaly:
@guptapratykshh @tenminusthreeseven please do not assign/work on issues that haven't been discussed with maintainers or flagged as triage-accepted.

shivansh-source:
/assign    
 i will work on this .

PR #944, opened 2026-01-10
feat: enable `includeWhen` to have resource references

bschaatsbergen:
Before these changes, `includeWhen` only applied to user (instance) input. `includeWhen` expressions were schema-only, meaning they could only reference back to a user their instance spec via `schema`. This made it impossible to conditionally create resources based on the state or values of other resources in the graph.

This change makes it possible for `includeWhen` to reference any resource by ID:

  ```yaml
  - id: app
    template:
      kind: Deployment
      metadata:
        labels:
          environment: production

  - id: loadbalancer
    includeWhen:
      - ${app.metadata.labels.environment == 'production'} # Now works
```

It simply extracts dependencies through `includeWhen` expressions and adds them to the dependency graph when we're building graph. At runtime `includeWhen` expressions are now evaluated with access to all dependent resources.


shivansh-source:
is this going to solve #845

bschaatsbergen:
> is this going to solve #845 
> 
> 

Yes, it should.

a-hilaly:
@bschaatsbergen the runtime has changed significantly. can you please take a look again?

PR #949, opened 2026-01-16
test: improve unit test coverage for pkg/graph/variable

shivansh-gohem:
**Description**
This PR improves the unit test coverage for the `pkg/graph/variable` package. 

Previously, the package had ~43% coverage because the helper methods on `ResourceVariableKind` (`IsStatic`, `IsDynamic`, `IsIncludeWhen`, `String`) were untested.

I added a new test function `TestResourceVariableKind` to cover these cases.

**Changes:**
- Added `TestResourceVariableKind` to `pkg/graph/variable/variable_test.go`
- Coverage increased from **42.9%** to **100%**.

**Verification:**
Ran `go test ./pkg/graph/variable -cover` locally.


pkg/graph/variable/variable_test.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//      http://www.apache.org/licenses/LICENSE-2.0
```
jakobmoellerdev: looks like an accidental diff. is that intended?

shivansh-gohem: Fixed! Reverted the accidental header change in 4122046

shivansh-gohem:
Thanks for the review, @jakobmoellerdev!

I have reverted the accidental diff in the license header and ran go fmt to fix the linting error reported by the bot. The PR should be ready for another look once the checks pass!!

PR #951, opened 2026-01-17
feat: add bytes conversion to go native types

shabbskagalwala:
This PR adds conversion for bytes values back to Go native types for use in Kubernetes resources. Without this when using bytes in Kro resources and creating instances results in errors like 

```
ERROR	dynamic-controller	Error syncing item, requeuing with rate limit	{"item": {"name":"test-rgd-1","namespace":"kro"}, "error": "failed to create runtime resource graph definition: failed to evaluate static variables: unsupported type: bytes"}
```


a-hilaly (approved):
Neat! thank you @shabbskagalwala !

/ok-to-test
/lgtm

PR #952, opened 2026-01-17
chore: bump kro in kro command

tjamet:
This requires updating some references of the package name that was transitioned from kro-run to kubernetes-sigs in commit 2d6562b


a-hilaly (approved):
Thank you @tjamet !
/lgtm

PR #955, opened 2026-01-20
feat: add controller warmup support for improved leader failover

a-hilaly:
Add `--enable-controller-warmup` flag to start watches/informers before
leader election is won. This pre-populates caches and reduces leader
failover time. Requires `--leader-elect` to be set.


a-hilaly:
needs to be merged after #956 

/hold

Issue #895, opened 2025-12-12
0.7.0 Type check issue


shwetambry:
### Description

**Observed Behavior**: In versions 0.6.+ and 0.7.0, getting the type check error -- 
failed to validate node "rttest": failed to type-check template expression "has(schema.spec.dummytest) ? schema.spec.dummytest.dummya : dummyRef.spec[0].fieldblist" at path "spec.references": ERROR: <input>:1:28: found no matching overload for '_?_:_' applied to '(bool, list(__type_schema.spec.dummytest.dummya.@idx), list(__type_dummyRef.spec.@idx.fieldblist.@idx))'

In kro version 0.5.1, it works correctly.

**Reproduction Steps** :
Here is my test RGD
```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: test-dummy
spec:
  schema:
    apiVersion: v1
    kind: TestApp
    spec:
      name: string
      dummytest:
        dummya: '[]Custom'
    types:
      Custom:
        subfieldstr: string
        subfieldint: integer
  resources:
  - id: dummyRef
    externalRef:
      apiVersion: kro.bug/v1
      kind: Dummy
      metadata:
        name: dummycr
        namespace: test
  - id: rttest
    template:
      apiVersion: kro.bug/v1
      kind: Resourcetemplate
      metadata:
        name: test-rt
        namespace: test
      spec:
        references: "${has(schema.spec.dummytest) ? 
          schema.spec.dummytest.dummya : dummyRef.spec[0].fieldblist}"
        
```
Here are the test CRDs 
```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: dummies.kro.bug
spec:
  conversion:
    strategy: None
  group: kro.bug
  names:
    kind: Dummy
    listKind: DummyList
    plural: dummies
    shortNames:
    - dummy
    singular: dummy
  scope: Namespaced
  versions:
  - name: v1
    schema:
      openAPIV3Schema:
        properties:
          spec:
            items:
              properties:
                name: 
                  description: Name
                  type: string
                fieldaint:
                  description: Field a
                  type: integer
                fieldblist:
                  description: Field b
                  items:
                    properties:
                      subfieldstr:
                        description: Subfield 1
                        type: string
                      subfieldint:
                        description: Subfield 2
                        type: integer
                    required:
                    - subfieldstr
                    - subfieldint
                    type: object
                  type: array
[...]

PR #967, opened 2026-01-23
0.7.0 ternary expression type check issue solved

Mujib-Ahasan:
## Description
This PR improves how CEL ternary (condition ? l : r) expressions are handled during validation by detecting ternary expressions early and rewriting only when necessary using dyn(), avoiding hard CEL type checking failure while preserving existing behavior and schema integrity.

The goal is to unblock valid user expressions without changing CEL’s core typing rules or resource schemas.

## Previous Behavior
Today cel expressions are parsed and type checked against OpenAPI schemas using a typed CEL environment.
However CEL enforces a strict rule for ternary expressions that both branches of a ternary expression must resolve to the same static type and this causes valid, real-world expressions to fail validation.

## What this PR does

This PR introduces a non invasive preprocessing step that:
- detects ternary expressions in CEL ASTs (including nested and complex ones)
- identifies ternary branches that would cause type mismatches
- rewrites only the ternary branches using dyn() when needed.

and this allows  treat both branches as dyn, defer validation to runtime nd preserve strict typing everywhere else.

Resolves: #895

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,6 @@ require (
 	golang.org/x/exp v0.0.0-20240719175910-8a7402abbf56
 	golang.org/x/sync v0.12.0
 	golang.org/x/time v0.9.0
-	google.golang.org/genproto/googleapis/api v0.0.0-20250303144028-a0af3efb3deb
 	k8s.io/api v0.34.1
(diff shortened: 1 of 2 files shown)


jakobmoellerdev:
/close

closing due to above. feel free to revisit this if you like!

Issue #957, opened 2026-01-20
Detect and report resource membership conflicts


a-hilaly:
### Feature Description

When multiple sources produce resources with the same GVK/namespace/name, kro currently silently overrides the desired state, claims membership, and overwrites source identifying labels.

Conflict scenarios:
- different graph nodes in the same instance produce identical resources
- different collection items from the same collection producing identical resources
- different instances of the same RGD produce identical resources
- different RGDs producing identical resources

Before applying, kro should check existing resources and verify membership labels (`kro.run/resource-graph-definition-id`, `kro.run/instance-id`, `kro.run/node-id`, `kro.run/collection-index`) match. On mismatch, fail and surface the error in instance status.

Avr8:
I have been able to reproduce this bug in my local, Created a simple RGD with a deployment resource and created two instances with same deployment name. I could see the following membership labels changing
`applyset.kubernetes.io/part-of:`
` kro.run/instance-id: `
`kro.run/instance-name:`

Avr8:
I Would like to try adding a fix for this, if thats okay

a-hilaly:
@shivansh-gohem - We have not discussed this feature. Please do not claim conversations or agreements that didn't happen. 

@Avr8 - To clarify, I have not discussed this with anyone and the issue remains open. That said, this is still labeled `needs-triage` and requires design discussion before implementation. It's a non-trivial change.

shivansh-gohem:
@a-hilaly  
I want to sincerely apologize for my previous comment.

I am so sorry for discouraging the contributor but my intentions never were never meant to hurt someone , I misunderstood your comment . I am so sorry for this , I will not repeat the mistakes in future.

I am stepping back from this issue entirely.

NicholasBlaskey:
https://github.com/kubernetes-sigs/kro/pull/1244 should handle 

```
    different graph nodes in the same instance produce identical resources
    different collection items from the same collection producing identical resources
```

I believe these two situations are already handled by the applyset id label conflict check (which is the gvknn of the instance)
```
    different instances of the same RGD produce identical resources
    different RGDs producing identical resources
```

Different RGDs and different instances within an RGD will have both different applyset IDs

PR #968, opened 2026-01-25
fix: Prevent instances from overwriting resources owned by other instances

shivansh-gohem:
## Description
This PR addresses a resource conflict issue where an Instance could overwrite resources (e.g., ConfigMaps, Deployments) that are already managed by a different Instance.

Previously, if two instances claimed the same resource name (e.g., both trying to create a ConfigMap named `shared-config`), the second instance would silently overwrite the resource, potentially causing data loss or unexpected behavior for the first instance.

This change introduces a pre-flight check during reconciliation to ensure that:
1. If a resource already exists, it must be managed by Kro (`kro.run/owned=true`).
2. If it is managed by Kro, it must belong to the **current** Instance ID (`kro.run/instance-id`).

If there is a mismatch (i.e., the resource belongs to another Instance), the reconciliation now fails with a clear `resource conflict` error instead of overwriting the resource.

## Related Issue
Fixes (#957 )resource hijacking vulnerability where resources could be adopted by conflicting instances.

## How Has This Been Tested?
I reproduced the issue locally using `kind` and verified the fix:

1. **Reproduction:**
   - Created a `ResourceGraphDefinition` for `ConflictTest` that generates a ConfigMap based on `spec.targetName`.
   - Created `instance-a` pointing to `shared-resource-name`.
   - Created `instance-b` pointing to the same `shared-resource-name`.

2. **Verification:**
   - **Before fix:** `instance-b` successfully synced and overwrote the ConfigMap (taking ownership).
   - **After fix:** `instance-b` entered an `ERROR` state with the message:
     `resource conflict: ConfigMap/shared-resource-name is already owned by instance instance-a (ID: ...)`

3. **Regression Testing:**
   - Ran `make test` and confirmed all existing unit tests passed.

### DIFF
--- a/pkg/controller/instance/resources_test.go
+++ b/pkg/controller/instance/resources_test.go
@@ -1,0 +1,177 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package instance
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


shivansh-gohem:
@jakobmoellerdev !! Thanks for the review!

I have added the requested unit tests in pkg/controller/instance/resources_test.go.

a-hilaly:
/lgtm cancel
/close

Closing this PR for the following reasons:

1. The issue is still labeled `needs-triage`. While this might sound trivial when you use AI , this is a non triavial change and might require changes in different areas and different testing strategies. It is mandatory to discuss the approach and which part of the github issue you're addressing before working on it.
 
2. In #957, you stated you had "already been discussing this specific feature with me to discourage another contributor (@Avr8) from working on it. This was false - we never had any such discussion.

PR #969, opened 2026-01-25
applyset: detect KRO ownership conflicts (fixes #957)

### LINKED ISSUE #957: Detect and report resource membership conflicts

Avr8:
ApplySet performs ownership conflict detection using:

```
if r.Current != nil {
    currentApplySetID = r.Current.GetLabels()[ApplysetPartOfLabel]
}
```

This assumes the controller always provides the live object in Resource.Current.

However, in real reconcile paths this is not always true. When Current is nil:

- ApplySet performs no ownership validation
- SSA proceeds
- A second KRO instance (same RGD, different instance) can overwrite resources 
   already owned by another KRO instance
- This leads to instance label flipping and silent resource takeover

**Root Cause**

ApplySet trusted Current to represent cluster state and skipped validation
when it was not provided.

**Fix**

When `r.Current == nil`, ApplySet now performs a defensive GET from the cluster
before SSA to check existing ownership labels.

If an existing object is found and:
`kro.run/owned=true`
and ownership labels (instance-id, resource-graph-definition-id, node-id)
do not match the desired object

ApplySet returns an ownership conflict error instead of applying.

**Test Added** 

TestApply_DetectsOwnershipConflict

This test:

1) Creates an existing object owned by another KRO instance

2) Calls ApplySet with Current == nil

3) Verifies that ownership conflict is detected

This test fails without the fix and passes with it.

### DIFF
--- a/pkg/controller/instance/applyset/applyset.go
+++ b/pkg/controller/instance/applyset/applyset.go
@@ -389,7 +389,7 @@ func (a *ApplySet) applyResource(
 			// True ownership = instance level, not just RGD level
 			if existing[metadata.InstanceIDLabel] != desired[metadata.InstanceIDLabel] ||
 				existing[metadata.ResourceGraphDefinitionIDLabel] != desired[metadata.ResourceGraphDefinitionIDLabel] ||
-				existing[metadata.NodeIDLabel] != desired[metadata.NodeIDLabel] {
+				existing[metadata.NodeIDLabel] != desired[metadata.NodeIDLabel] || existing[metadata.CollectionIndexLabel] != desired[metadata.CollectionIndexLabel] {
 
 				item.Error = fmt.Errorf(
 					"kro ownership conflict for %s/%s: owned by different KRO instance",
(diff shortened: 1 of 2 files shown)


jakobmoellerdev:
could you please rebase? thx!

a-hilaly:
Closing for now. Before resubmitting:

1/ discuss your approach on #957 - specifically which conflict scenario(s) you want to address and your testing strategy
2/ rebase once we've aligned on the design

Avr8:
Hi @a-hilaly , due to my health issue i wasnt able to have a look at it, back in shape now...happy to discuss this and get it going!

Avr8:
Let me explain my solution and changes more clearly in this doc, should be helpful before we discuss

PR #970, opened 2026-01-25
refactor(instance): align node state tracking

a-hilaly:
Node processing is node-centric, but state tracking was still framed as
"resource" state with ad-hoc mutations, which made the controller flow
harder to reason about and document as it grows.

Rename resource state to node state, add NodeState helpers + NewNodeState,
type InstanceState, standardize node processing/deletion naming, move the
labeler field manager constant into the controller, and expand WHAT-focused
comments where behavior was ambiguous.

Also did some renames like `s/stage/process` and `s/plan/process` as I believe
planning will become a real thing when propagation control comes close to reality.


a-hilaly:
needs merge after #936 
/hold

pkg/controller/instance/state.go
```diff
@@ -16,44 +16,134 @@ package instance
...
 const (
-	InstanceStateInProgress = "IN_PROGRESS"
-	InstanceStateFailed     = "FAILED"
-	InstanceStateActive     = "ACTIVE"
-	InstanceStateDeleting   = "DELETING"
-	InstanceStateError      = "ERROR"
+	// InstanceStateInProgress means reconciliation is ongoing.
+	InstanceStateInProgress InstanceState = "IN_PROGRESS"
+	// InstanceStateFailed is a legacy state kept for compatibility.
+	InstanceStateFailed InstanceState = "FAILED"
```
jakobmoellerdev, line 26: nit: we should mark this as deprecated. maybe even worth removing in favor of error

pkg/controller/instance/state.go
```diff
@@ -16,44 +16,134 @@ package instance
...
+	InstanceStateActive InstanceState = "ACTIVE"
+	// InstanceStateDeleting means deletion workflow is running.
+	InstanceStateDeleting InstanceState = "DELETING"
+	// InstanceStateError means reconciliation hit an error.
+	InstanceStateError InstanceState = "ERROR"
 )
 
-type ResourceState struct {
+// NodeState values track the lifecycle of individual nodes during reconciliation.
+const (
```
jakobmoellerdev, line 36: it feels to me like these states should be part of the controller API actually since they are very important for understanding resource lifecycle (similar to upstream k8s) wdyt?

a-hilaly: Agreed - these should live in api/v1alpha1/ since they're part of the public interface. I was also thinking about adding clear transition semantics. Feel to me like enough scope for a seperate PR.

jakobmoellerdev:
@a-hilaly please create a follow up for the API level extraction as discussed

PR #971, opened 2026-01-25
docs: expand collections gotchas and cross-references

a-hilaly:
Add a Constraints & Gotchas section with the core pitfalls, & add
cross-references from the main sections to the new guidance.

### DIFF
--- a/website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
+++ b/website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
@@ -86,7 +86,8 @@ forEach:
   - key: ${schema.spec.labels.map(k, k).sort()}
 ```
 
-You can then access the value inside the template using the key.
+For ordering pitfalls and why this matters, see
+[Map Gotchas and pitfalls](#map-iteration-order-is-not-deterministic)
 
 ## Iterator Variables
 
@@ -219,6 +220,7 @@ deterministic order (arrays are ordered, maps are not - see
 If any iterator's collection is empty, zero resources are created. This follows
 standard cartesian product behavior: `2 × 0 = 0`.
 :::
+
 
 ## Collection Sources
 
@@ -460,6 +462,9 @@ condition before proceeding.
 If the collection is empty (zero items), it is considered ready.
 :::
 
+For empty-collection readiness, see
+[Constraints & Gotchas](#constraints--gotchas) below.
+
 :::tip
 Without `readyWhen`, a collection is considered ready once all resources are
 created. Add `readyWhen` when dependent resources need specific conditions to be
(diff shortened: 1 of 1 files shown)


a-hilaly:
/hold
needs merrge after #936

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -500,6 +508,100 @@ expression:
...
+### Empty Collections Are Ready
+
+An empty array produces zero resources. The collection is still considered
+**ready**, because there are no items to wait on. This also applies to cartesian
+products where any iterator is empty (`2 × 0 = 0`).
+
+If you expect downstream resources to wait on actual items, ensure the
+collection is non-empty or use additional conditions in `includeWhen`.
+
+### Map Iteration Order Is Not Deterministic
```
jakobmoellerdev, line 554: continuing from collections rework discussion I believe we should add a small example on how to do this. I feel like this is very hard to just "figure out". If we cant even put out a proper example, we should discourage maps entirely for now

a-hilaly: Added an example showing the full pattern with sorted keys and value access.

I don't think we need to discourage maps entirely - maybe could build some CEL helpers/macros to make this less painful (`x.sortedEntries()` that returns `[{key, value} ...])`

Fun fact I did some analysis on AWS APIs and arrays dominated by around 90% over maps. Not sure how representative that is globally, but it makes me think map iteration might be a niche use case anyway, and maybe it should be discouraged. I'll defer to you here to make a call.

jakobmoellerdev: *TLDR*: agree + lets proceed with suggesting preferral of arrays in apis (maybe in docs for collections as caveat) and see if folks have issues, if they do, lets introduce a macro / function later.

I have made similar checks in the past and turns out slices in serialized JSON/YAML are most of the time preferrable:

- patching — strategic merge patch on maps replaces the whole thing by default, lists with merge keys let you patch individual items cleanly
- extensibility — list items can grow new fields later, maps lock you into a fixed value type
- order — json objects are technically unordered, lists aren't
- tooling — kubectl diff, kustomize, helm all handle list merges more predictably

For now we should probably mention that APIs should likely always lean towards arrays anyhow as most serialized forms of maps can be just as well described by an array, but not the other way around:

```
- id: a
  key: value
```

can easily translate to `map[id]obj`.

on the docs note (caveat): Its important to realize that while we create CRDs ourselves, we basically orchestrate existing apis. so its pretty likely that we will face apis on maps. especially with internal developer apis that are not as much under public review / scrutiny, which is one of KROs main use cases.

a-hilaly: Fair, and I agreed. Added a :::tip::: section below

a-hilaly: Should we get rid of `### Iterating Over Maps` in L76?

PR #972, opened 2026-01-26
fix(schema): preserve nested array/object paths in status schema

a-hilaly:
The schema builder mutated a copy of map values, so updates to struct fields
like Type and Items were lost on nested array/object segments. This caused
paths such as `status.parents[0].child` to produce an object instead of an
array.

This patch rework the builder to update child schemas and write them back.

### DIFF
--- a/pkg/graph/schema/field_descriptor.go
+++ b/pkg/graph/schema/field_descriptor.go
@@ -139,62 +139,67 @@ func generateJSONSchemaFromFieldDescriptors(fieldDescriptors []fieldDescriptor) 
 	return rootSchema, nil
 }
 
-func addFieldToSchema(fieldDescriptor fieldDescriptor, schema *extv1.JSONSchemaProps) error {
-	segments, err := fieldpath.Parse(fieldDescriptor.Path)
+func addFieldToSchema(fd fieldDescriptor, schema *extv1.JSONSchemaProps) error {
+	segments, err := fieldpath.Parse(fd.Path)
 	if err != nil {
-		return fmt.Errorf("failed to parse path %s: %w", fieldDescriptor.Path, err)
+		return fmt.Errorf("failed to parse path %s: %w", fd.Path, err)
 	}
 
-	currentSchema := schema
-
-	for i, segment := range segments {
-		isLast := i == len(segments)-1
-
-		if segment.Index >= 0 {
-			// Handle array segment
-			if currentSchema.Type != "array" {
-				currentSchema.Type = "array"
-				currentSchema.Items = &extv1.JSONSchemaPropsOrArray{
-					Schema: &extv1.JSONSchemaProps{
-						Type:       "object",
-						Properties: make(map[string]extv1.JSONSchemaProps),
-					},
-				}
-			}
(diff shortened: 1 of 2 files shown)


pkg/graph/schema/field_descriptor.go
```diff
@@ -145,56 +145,81 @@ func addFieldToSchema(fieldDescriptor fieldDescriptor, schema *extv1.JSONSchemaP
...
-			// Handle array segment
-			if currentSchema.Type != "array" {
-				currentSchema.Type = "array"
-				currentSchema.Items = &extv1.JSONSchemaPropsOrArray{
-					Schema: &extv1.JSONSchemaProps{
-						Type:       "object",
-						Properties: make(map[string]extv1.JSONSchemaProps),
-					},
-				}
+	return addFieldToSchemaSegments(schema, segments, fieldDescriptor.Schema)
```
jakobmoellerdev: I think this reads very verbose. maybe im missing something but it could probably deserve some breaking up. as an inspiration:

Pulled out the nil-default to one place, DRY'd up the object init into a helper, dropped error returns that never fired, tidied up the slice indexing.

```go
func addFieldToSchema(fd fieldDescriptor, schema *extv1.JSONSchemaProps) error {
    segments, err := fieldpath.Parse(fd.Path)
    if err != nil {
        return fmt.Errorf("failed to parse path %s: %w", fd.Path, err)
    }
    
    leaf := fd.Schema
    if leaf == nil {
        leaf = &extv1.JSONSchemaProps{Type: "string"}
    }
    addFieldToSchemaSegments(schema, segments, leaf)
    return nil
}

func ensureObjectSchema(s *extv1.JSONSchemaProps) {
    if s.Type == "" {
        s.Type = "object"
    }
    if s.Properties == nil {
        s.Properties = make(map[string]extv1.JSONSchemaProps)
    }
}

func addFieldToSchemaSegments(
    current *extv1.JSONSchemaProps,
    segments []fieldpath.Segment,
    leaf *extv1.JSONSchemaProps,
) {
    if len(segments) == 0 {
        return
    }

    seg, rest := segments[0], segments[1:]

    // Array index segment
    if seg.Index >= 0 {
        current.Type = "array"
        if current.Items == nil || current.Items.Schema == nil {
            current.Items = &extv1.JSONSchemaPropsOrArray{Schema: &extv1.JSONSchemaProps{}}
        }
        if len(rest) == 0 {
            *current.Items.Schema = *leaf
            return
        }
        ensureObjectSchema(current.Items.Schema)
        addFieldToSchemaSegments(current.Items.Schema, rest, leaf)
        return
    }

    // Named field segment
    ensureObjectSchema(current)
    if len(rest) == 0 {
        current.Properties[seg.Name] = *leaf
        return
    }

    child := current.Properties[seg.Name]
    ensureObjectSchema(&child)
    addFieldToSchemaSegments(&child, rest, leaf)
    current.Properties[seg.Name] = child
}
```

a-hilaly: I like it! Fixed & co-authored.

PR #974, opened 2026-01-26
docs(cel): explain multiline expressions and YAML chomping

a-hilaly:
Add a CEL Expressions doc section that explains YAML block scalars for
standalone expresions and how `|-`/`>-` avoid trailing newlines.

### DIFF
--- a/website/docs/docs/concepts/rgd/03-cel-expressions.md
+++ b/website/docs/docs/concepts/rgd/03-cel-expressions.md
@@ -113,6 +113,37 @@ name: "app-${string(schema.spec.replicas)}"
 ```
 :::
 
+### Multiline Expressions and YAML Block Scalars
+
+YAML block scalars (`|` and `>`) often add a **trailing newline**. For fields that must be a *standalone* expression (like `includeWhen` and `readyWhen`), a trailing newline means the value is no longer exactly `${...}` and validation fails.
+
+Use the **chomp indicator** `-` to strip the final newline:
+
+```kro
+includeWhen:
+  - |-
+    ${
+      schema.spec.enabled &&
+      schema.spec.count > 0
+    }
+```
+
+If you prefer folded scalars (`>`), also use `>-` to avoid the trailing newline:
+
+```kro
+readyWhen:
+  - >-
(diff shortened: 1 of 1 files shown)


website/docs/docs/concepts/rgd/03-cel-expressions.md
```diff
@@ -113,6 +113,37 @@ name: "app-${string(schema.spec.replicas)}"
 ```
 :::
 
+### Multiline Expressions and YAML Block Scalars
+
+YAML block scalars (`|` and `>`) often add a **trailing newline**. For fields that must be a *standalone* expression (like `includeWhen` and `readyWhen`), a trailing newline means the value is no longer exactly `${...}` and validation fails.
```
jakobmoellerdev, line 118: IMHO we could check for this. couldnt we check if the expression contains new lines and then remove the following newline?

I know we could strip the new line but it feels like a gotcha thats easy enough to circumvent.

a-hilaly: I went back and forth on this for a while. The tricky part that was annoying me is that trimming isn't safe in all contexts, for `includeWhen/readyWhen` it would work fine, but i'm thinking of expressions can also appear in `ConfigMap` (mounted as files), IAM JSON blobs and string interpolation.

i just think can't think of a way to relyibly distinguish wants this new line vs YAML added this new line without heuristics that will surprise someone eventually.

a-hilaly: That said I'm open to trimming specifically for `includeWhen/readyWhen` since those are always standalone expressions.

jakobmoellerdev: hmmm yeah I get that. I guess since yaml has an inbuilt way of doing this, this is fine.
all right, fine for me!

jakobmoellerdev: I dont think we should change behavior based on which expression type we have as much as possible. it would be even more confusing if it comes to it.

PR #976, opened 2026-01-26
fix(graph): reject cluster-scoped resources with namespace set

a-hilaly:
While many tools allow this, it is technically incorrect. Cluster scoped
resources do not have nor should set a namespace.

That said, in collections, item identity matching relies on namespace +
name and assumes cluster scoped objects have an empty namespace.
Namespace normalization only runs for namespaced resources, so any
namespace set on a cluster-scoped template persists and breaks
desired<->observed matching in collections.

Add `validateTemplateConstraints` to catch this at RGD build time

### DIFF
--- a/pkg/graph/validation.go
+++ b/pkg/graph/validation.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"regexp"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/graph/validation.go
```diff
@@ -258,3 +259,22 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
 	}
 	return nil
 }
+
+// validateTemplateConstraints enforces template-level constraints before parsing expressions.
+// Keep this small and focused on invariants that must hold regardless of CEL.
+//
+// TODO: We need more constraints here, e.g, you cannot set kro owned labels/annotations, and
```
jakobmoellerdev: hmm regarding status fields, shouldnt it be technically allowed?

pkg/graph/validation.go
```diff
@@ -258,3 +259,22 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
...
+// maybe you cannot set status fields?
+func validateTemplateConstraints(rgResource *v1alpha1.Resource, resourceObject map[string]interface{}, namespaced bool) error {
+	if !namespaced {
```
jakobmoellerdev, line 268: to me it feels like we are duplicating validation logic from the API Server or schema here. shouldnt we be able to validate on the resource level instead? or is this a special case?

pkg/graph/validation.go
a-hilaly: Good question - I've never seen anyone do it, and I've always assumed kro won't touch status fields. Let me remove that from to the right now, to be explored.

pkg/graph/validation.go
a-hilaly: Very suprisingly the API server doesn't reject Cluster scoped resources with a `metadata.namespace` field set - it silently ignores it. I bet thats because the schema for `v1.ObjectMeta` is shared between all resources and always includes namespace.

This causes our templates to have a namespace but what we observe from the API server doesn't which makes it tricky to map collection items back to observed states using their identity. Not impossible, just a question of when/where we want to solve this. I thought stopping it as early as possible would be best.

PR #978, opened 2026-01-26
fix(website): version CRD with docs snapshots

a-hilaly:
Move the RGD CRD yaml from `static/` to `docs/api/` so each versioned
doc release includes its own CRD snapshot. Previously all doc versions
pointed to the same (in main) CRD in `static/.`

Backfilled CRD for `v0.7.0` and `v0.7.1`.


a-hilaly:
working preview in https://a-hilaly.github.io/kro/api/crds/resourcegraphdefinition

/unhold

PR #982, opened 2026-01-28
fix(release): capture `GIT_VERSION` once to prevent -dirty suffix in LDFAGS

a-hilaly:
GIT_VERSION was re-evaluated on each make invocation. When package-helm
modified helm/Chart.yaml between build-image and publish-image, the second
evaluation of `git describe --dirty` detected uncommitted changes.

Export `GIT_VERSION` once at script start so all make commands use the same
frozen value.


jakobmoellerdev (commented):
/lgtm

PR #986, opened 2026-01-29
fix: filter child events by parent instance GVK

shabbskagalwala:
This PR aims to solve the issue with multiple ResourceGraphDefinitions watch the same child resource types, child resource events would trigger reconciliation attempts across all RGDs, not just the RGD that owns the instance.

Kro would log errors when controllers tried to reconcile instances with the same name but different types.

```
"instance not found (likely deleted)" 
```

What I aimed to do with this PR is add parent instance GVK labels to all child resources and filters child events in the dynamic controller to only enqueue reconciliations when the child's parent instance GVK matches the controller's expected GVK.


pkg/metadata/labels.go
```diff
@@ -49,6 +50,9 @@ const (
 	InstanceIDLabel        = LabelKROPrefix + "instance-id"
 	InstanceLabel          = LabelKROPrefix + "instance-name"
 	InstanceNamespaceLabel = LabelKROPrefix + "instance-namespace"
+	InstanceGroupLabel     = LabelKROPrefix + "instance-group"
```
jakobmoellerdev, line 53: please change this to 2 labels:

1 for apiVersion (group+version)
1 for kind

I would like to avoid having to introduce too many labels

pkg/dynamiccontroller/dynamic_controller_test.go
```diff
@@ -169,6 +169,9 @@ func TestDynamicController_WatchBehavior(t *testing.T) {
 		metadata.OwnedLabel:             "true",
 		metadata.InstanceLabel:          deploy.GetName(),
 		metadata.InstanceNamespaceLabel: deploy.GetNamespace(),
+		metadata.InstanceGroupLabel:     deployGVK.Group,
```
jakobmoellerdev, line 172: you are only testing that these labels are set but I was expecting a test that confirms there is no conflict. could you add one?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -599,6 +599,25 @@ func (dc *DynamicController) handlerForChildGVR(parent, child schema.GroupVersio
...
 		}
 
+		// This prevents controllers from different RGDs from reconciling instances with the same name
+		// but different types.
+		instanceGroup := lbls[metadata.InstanceGroupLabel]
+		instanceVersion := lbls[metadata.InstanceVersionLabel]
+		instanceKind := lbls[metadata.InstanceKindLabel]
+
+		if instanceGroup != parentGVK.Group || instanceVersion != parentGVK.Version || instanceKind != parentGVK.Kind {
+			dc.log.V(2).Info("Skipping child event - parent instance GVK mismatch",
```
jakobmoellerdev: imho there is no need to log if we have a GVK mismatch. this is a case that should never be considered and can spam => please remove the log

shabbskagalwala: Should we just return here? I can remove the log

pkg/metadata/labels.go
shabbskagalwala: Ah thats a nice suggestion, I can update the labels

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: yes please!

pkg/metadata/labels.go
shabbskagalwala: Updated labels :)

pkg/dynamiccontroller/dynamic_controller.go
shabbskagalwala: removed log and added a return

pkg/dynamiccontroller/dynamic_controller_test.go
shabbskagalwala: Need to figure out tests for this case :) i'll have on shortly

pkg/metadata/labels_test.go
```diff
@@ -240,12 +242,27 @@ func TestNewInstanceLabeler(t *testing.T) {
...
+
 		labeler := NewInstanceLabeler(obj)
 		assert.Equal(t, GenericLabeler{
-			InstanceLabel:          name,
-			InstanceNamespaceLabel: namespace,
-			InstanceIDLabel:        string(uid),
+			InstanceLabel:           name,
+			InstanceNamespaceLabel:  namespace,
+			InstanceIDLabel:         string(uid),
+			InstanceAPIVersionLabel: group + "/" + version,
```
jakobmoellerdev: theres a method on gvk to convert it to apiversion kind i believe (or there is a function somewhere)

shabbskagalwala: I don't think one label for apiVersion will work

```
a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

```
metadata.labels: Invalid value: \"kro.run/v1alpha1\"
```

jakobmoellerdev: ack, then lets fall back to your original version. sorry for the back and forth :/

shabbskagalwala: reverted back to 3 labels. Just missing the test. I will get that out soon

test/integration/suites/core/instance_conflict_test.go
```diff
@@ -0,0 +1,272 @@
...
+				HaveKeyWithValue(metadata.InstanceVersionLabel, "v1alpha1"),
+				HaveKeyWithValue(metadata.InstanceKindLabel, "TestConflictApp1"),
+			), "Recreated ConfigMap should have correct instance GVK labels for TestConflictApp1")
+
+			g.Expect(cfgMap1.Data).To(SatisfyAll(
+				HaveKeyWithValue("source", "app1"),
+			))
+		}, 30*time.Second, time.Second).WithContext(ctx).Should(Succeed())
+
+		By("Modify instance2 ConfigMap data to check if the data is changed back to original state")
```
a-hilaly, line 246: Thank you for adding these!

PR #994, opened 2026-01-30
Fix cmd/kro by updating Kro dependency to v0.8.1

chrisdoherty4:
# Summary

Updates `cmd/kro` to use Kro v0.8.1 to fix the compile error below.

```
# github.com/kubernetes-sigs/kro/cmd/kro/commands/generate
commands/generate/diagram.go:114:39: resource.Meta undefined (type *graph.Resource has no field or method Meta)
```


a-hilaly:
I just realized that I merged two PRs having the same diff, i didn't know Git can do that :)

PR #995, opened 2026-01-30
feat(metrics): add CEL metrics for compilation and evaluation durations

### LINKED ISSUE #195: expose CEL execution prometheus metrics
### Feature Description

Currently kro lacks observability around CEL expression execution. we need to add prometheus metrics similar to the Kubernetes API server to track CEL evaluation performance and failures. These metrics will help operators monitor the health and performance of ResourceGroups/I reconcilers.

We should follow the pattern from: 
https://github.com/kubernetes/apiserver/blob/2241ea00526fdd746871375dfca871627f6a511a/pkg/cel/metrics/metrics.go#L32-L62

Denyme24:
## What Changed
- **New metrics package** (`pkg/cel/metrics.go`): Defines two histograms to track CEL compilation and evaluation timing
- **Instrumented CEL execution** (`pkg/runtime/eval.go`): Records metrics during actual CEL operations
- **Comprehensive tests** (`pkg/cel/metrics_test.go`): Validates metric collection and registration
- **Uses controller-runtime registry**: Metrics registered with `metrics.Registry` for consistency with existing kro metrics

## Metrics Exposed
- `kro_cel_compilation_duration_seconds` - CEL compilation time with success/error labels
- `kro_cel_evaluation_duration_seconds` - CEL evaluation time with success/error labels

Both use exponential buckets optimized for CEL's typical performance characteristics (10µs to ~10ms). Durations are tracked as float64 seconds, preserving microsecond precision for subsecond operations.
## Design
Follows the Kubernetes apiserver pattern as referenced in the issue. Metrics are:
- Minimal overhead - only timing critical paths
- Non-breaking - no API changes
- Labeled for observability - tracks both success and error cases

Fixes #195

### DIFF
--- a/pkg/graph/validation.go
+++ b/pkg/graph/validation.go
@@ -17,11 +17,13 @@ package graph
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
+	"github.com/kubernetes-sigs/kro/pkg/metadata"
 )
 
 var (
@@ -262,8 +264,6 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
 
 // validateTemplateConstraints enforces template-level constraints before parsing expressions.
 // Keep this small and focused on invariants that must hold regardless of CEL.
-//
-// TODO: We need more constraints here, e.g, you cannot set kro owned labels/annotations...
 func validateTemplateConstraints(rgResource *v1alpha1.Resource, resourceObject map[string]interface{}, namespaced bool) error {
 	if !namespaced {
 		_, found, err := unstructured.NestedFieldNoCopy(resourceObject, "metadata", "namespace")
@@ -272,6 +272,33 @@ func validateTemplateConstraints(rgResource *v1alpha1.Resource, resourceObject m
 		}
 		if found {
 			return fmt.Errorf("resource %q is cluster-scoped and must not set metadata.namespace", rgResource.ID)
+		}
+	}
+
+	// Validate that users don't set KRO-owned labels
+	if err := validateNoKROOwnedLabels(rgResource.ID, resourceObject); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validateNoKROOwnedLabels enforces that the resource template doesn't define any label with
+// LabelKROPrefix (kro.run/). These labels are reserved for internal use ONLY.
+func validateNoKROOwnedLabels(resourceID string, resourceObject map[string]interface{}) error {
+	labelsRaw, found, err := unstructured.NestedFieldCopy(resourceObject, "metadata", "labels")
+	if err != nil || !found {
+		return nil
+	}
+
+	labelsMap, ok := labelsRaw.(map[string]interface{})
+	if !ok {
+		return nil
+	}
+
+	for key := range labelsMap {
+		if strings.HasPrefix(key, metadata.LabelKROPrefix) {
+			return fmt.Errorf("invalid label for resource %q. labels with prefix %q are reserved for internal use", resourceID, metadata.LabelKROPrefix)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Denyme24:
hey @a-hilaly @jakobmoellerdev, PTAL.

pkg/cel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+
+// MustRegister registers the metrics with the given Prometheus registry.
+func (m *CelMetrics) MustRegister(registry prometheus.Registerer) {
+	registry.MustRegister(m.compilationTime)
+	registry.MustRegister(m.evaluationTime)
+}
+
+// For now, register with the default registry
+//
+// TODO(a-hilaly): rework all kro custom metrics to use a custom registry, and
```
jakobmoellerdev: why are you adding a todo from amine?

pkg/runtime/eval.go
```diff
@@ -101,17 +102,32 @@ func evalListExpr(env *cel.Env, expr *expressionEvaluationState, ctx map[string]
 // CEL errors are returned as-is; callers should use isCELDataPending() to check
 // if the error indicates data is pending and should be retried.
 func evalRawCEL(env *cel.Env, expr string, ctx map[string]any) (any, error) {
+	// Measure compilation time
+	compileStart := time.Now()
 	ast, issues := env.Compile(expr)
+	compileDuration := time.Since(compileStart).Seconds()
```
jakobmoellerdev, line 108: why are we rounding to seconds?

go.mod
```diff
@@ -57,6 +57,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
```
jakobmoellerdev, line 60: whats the reason this shows up as indirect now? looks a bit off

pkg/cel/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	namespace = "kro"
+	subsystem = "cel"
+)
+
+// Metrics provides access to CEL metrics.
+var Metrics = newCelMetrics()
```
jakobmoellerdev, line 28: why are we not using the default metrics registry from controller runtime here?

pkg/cel/metrics.go
Denyme24: removed, we're now using the correct registry directly.

pkg/runtime/eval.go
Denyme24: not actually rounding - `time.Since().Seconds()` returns float64 with full precision.
eg: 
- 23µs → `0.000023` seconds
- 1.5ms → `0.0015` seconds  
- 100µs → `0.0001` seconds

go.mod
Denyme24: Transitive dependency from `prometheus/client_golang` - expected and necessary per `go mod graph`.

pkg/cel/metrics.go
Denyme24: updated to use `metrics.Registry` from `sigs.k8s.io/controller-runtime/pkg/metrics` instead of `prometheus.DefaultRegisterer` for consistency with the rest of kro's metrics.

Denyme24:
> A few conceptual questions and one thing on accuracy. IMHO I dont think we should track CEL expressions on second level accuracy since most cel expressions are subsecond evals.

No precision loss - Prometheus stores float64 seconds, so 23µs = `0.000023`s. Our buckets (10µs-10ms) are optimized for subsecond CEL evals per Prometheus convention.

Denyme24:
hey @jakobmoellerdev, thanks for reviewing the PR, have made some changes to address your review comments. PTAL.

a-hilaly:
This PR is introducing metrics into a constantly changing and prone to change area in the code base. i'd put on hold for now.

/hold

jakobmoellerdev:
Hmmm, not sure I agree. we should not be afraid to introduce metrics for execution latency on expressions. Instead they should be ported later to be valid.

a-hilaly:
merged in a seperate PR

PR #997, opened 2026-01-30
chore: remove managed-by label from kro instances

shabbskagalwala:
Removing the `app.kubernetes.io/managed-by` from Kro managed instances. This is a well known label that is used by Helm charts - https://helm.sh/docs/chart_best_practices/labels/ and is always set to the Release Service. When Kro instances are deployed via Helm charts through ArgoCD - this results in a constant out of sync -> sync -> out of sync loop due to the label mismatch. 

Kro already uses `kro.run/owned: 'true'` label on instances.

Technically, `ignoreDifferences` could be added to Argo to ignore differences on this label but that would not be an ideal scenario.

### DIFF
--- a/test/integration/suites/core/annotation_label_test.go
+++ b/test/integration/suites/core/annotation_label_test.go
@@ -140,7 +140,7 @@ var _ = Describe("Labels and Annotations", func() {
 			), "default kro labels should also be present")
 
 			g.Expect(instance.GetLabels()).ToNot(HaveKey(metadata.ManagedByLabelKey),
-				"instance should not have app.kubernetes.io/managed-by label as it may be used by other application lifecycle tooling")
+				"instance should not have app.kubernetes.io/managed-by label it may be used by other application lifecycle tooling")
 
 		}, 10*time.Second, time.Second).WithContext(ctx).Should(Succeed())
 


test/integration/suites/core/annotation_label_test.go
```diff
@@ -139,6 +139,9 @@ var _ = Describe("Labels and Annotations", func() {
 				HaveKeyWithValue(metadata.ResourceGraphDefinitionNameLabel, rgd.GetName()),
 			), "default kro labels should also be present")
 
+			g.Expect(instance.GetLabels()).ToNot(HaveKey(metadata.ManagedByLabelKey),
+				"instance should not have app.kubernetes.io/managed-by label to avoid conflicts with Helm")
```
jakobmoellerdev: nit
```suggestion
				"instance should not have app.kubernetes.io/managed-by label as it may be used by other application lifecycle tooling")
```

shabbskagalwala: Updated :)

shabbskagalwala:
apologies, fixed the linter and the label got removed :/

michaelhtm (approved):

a-hilaly (approved):
Thank you @shabbskagalwala !
/lgtm

PR #998, opened 2026-01-30
refactor(simpleschema): rework type system for easier extension

a-hilaly:
This refactor removes duplicated walks and split “parse vs handle”
branches that made extending the simple schema type system hard to
extended and reason about.

This commit introduces a single `Type` interface that now owns dependency
discovery and schema generation, and `ParseField` is the one place
types and markers are interpreted.

With types moved into a subpackage and the transformer simplified,
adding new type kinds no longer requires touching multiple parsing
branches or traversal passes.

Also add extensive unit tests that cover 100% of `pkg/simpleschema` and
e2e tests to protect from future regressions.

### DIFF
--- a/pkg/simpleschema/markers.go
+++ b/pkg/simpleschema/markers.go
@@ -15,6 +15,7 @@
 package simpleschema
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 	"strconv"
@@ -24,6 +25,9 @@ import (
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/utils/ptr"
 )
+
+// ErrUnknownMarker is returned when an unrecognized marker is encountered.
+var ErrUnknownMarker = errors.New("unknown marker")
 
 // This package is used find extra markers in the NeoCRD schema that maps to
 // something in OpenAPI schema. For example, the `required` marker in the NeoCRD
@@ -93,7 +97,7 @@ func markerTypeFromString(s string) (MarkerType, error) {
 		MarkerTypeMaxItems:
 		return MarkerType(s), nil
 	default:
-		return "", fmt.Errorf("unknown marker type: %s", s)
+		return "", fmt.Errorf("%w: %s", ErrUnknownMarker, s)
 	}
 }
 
@@ -104,10 +108,9 @@ type Marker struct {
 	Value      string
 }
 
-// parseMarkers parses a marker string and returns a `Marker` struct.
+// ParseMarkers parses a string of markers and returns a slice of Marker structs.
 // The marker string should be in the format `marker=value`.
-// parseMarkers parses a string of markers and returns a slice of Marker structs
-func parseMarkers(markers string) ([]*Marker, error) {
+func ParseMarkers(markers string) ([]*Marker, error) {
 	var result []*Marker
 	var currentMarker *Marker
 	var inQuotes bool
@@ -124,7 +127,7 @@ func parseMarkers(markers string) ([]*Marker, error) {
 			}
 			markerType, err := markerTypeFromString(key)
 			if err != nil {
-				return nil, fmt.Errorf("invalid marker key '%s': %v", key, err)
+				return nil, fmt.Errorf("invalid marker key %q: %w", key, err)
 			}
 			currentMarker = &Marker{MarkerType: markerType, Key: key}
 			buffer.Reset()
@@ -251,6 +254,8 @@ func applyMarker(schema *extv1.JSONSchemaProps, marker *Marker, key string, pare
 		return applyMinItemsMarker(schema, marker)
 	case MarkerTypeMaxItems:
 		return applyMaxItemsMarker(schema, marker)
+	default:
+		return fmt.Errorf("%w: %s", ErrUnknownMarker, marker.MarkerType)
 	}
 	return nil
 }
--- a/pkg/simpleschema/parse.go
+++ b/pkg/simpleschema/parse.go
@@ -21,22 +21,26 @@ import (
 	"github.com/kubernetes-sigs/kro/pkg/simpleschema/types"
 )
 
+// ParseTypeString parses the type portion of a field string (before any "|").
+func ParseTypeString(s string) (types.Type, error) {
+	typeStr := strings.SplitN(s, "|", 2)[0]
+	typeStr = strings.TrimSpace(typeStr)
+	if typeStr == "" {
+		return nil, fmt.Errorf("empty type")
+	}
+	return parseType(typeStr)
+}
+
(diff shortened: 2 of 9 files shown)


pkg/simpleschema/types/atomic.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Atomic represents primitive types: string, integer, boolean, float.
+type Atomic string
+
+const (
+	String  Atomic = "string"
+	Integer Atomic = "integer"
+	Boolean Atomic = "boolean"
+	Float   Atomic = "float"
+)
```
a-hilaly, line 29: Supporting `bytes` type will be as easy as adding it as constant here. rest is history.

pkg/simpleschema/markers.go
```diff
@@ -193,3 +207,222 @@ func unescapeString(s string) string {
...
+	case MarkerTypePattern:
+		return applyPatternMarker(schema, marker)
+	case MarkerTypeUniqueItems:
+		return applyUniqueItemsMarker(schema, marker)
+	case MarkerTypeMinItems:
+		return applyMinItemsMarker(schema, marker)
+	case MarkerTypeMaxItems:
+		return applyMaxItemsMarker(schema, marker)
+	}
+	return nil
```
jakobmoellerdev, line 260: why do we not return an error if we dont find the marker type?

pkg/simpleschema/simpleschema.go
```diff
@@ -29,11 +29,15 @@ import (
 // the type name and the value its specification. These custom types will be
 // available as predefined types in the transformer.
 func ToOpenAPISpec(obj map[string]interface{}, customTypes map[string]interface{}) (*extv1.JSONSchemaProps, error) {
-	tf := newTransformer()
-	if err := tf.loadPreDefinedTypes(customTypes); err != nil {
+	t := &transformer{
+		customTypes: make(map[string]customType),
+	}
+
+	if err := t.loadCustomTypes(customTypes); err != nil {
```
jakobmoellerdev: this constructor looks off. Why dont we simply call newTransformer(customTypes)?

pkg/simpleschema/types/atomic.go
jakobmoellerdev: I dont see why we would need to support a bytes atomic type. IMHO this would be a string that is base64 formatted. what am I missing here?

pkg/simpleschema/types/types.go
```diff
@@ -0,0 +1,30 @@
...
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// Resolver resolves custom type names to their schemas.
+type Resolver interface {
+	Resolve(name string) (*extv1.JSONSchemaProps, error)
+}
+
+// Type represents a parsed type that can provide dependencies and build schemas.
+type Type interface {
```
jakobmoellerdev, line 32: I think this type interface deserves a bit of love on the method doc, especially since its so central now.

pkg/simpleschema/parse.go
```diff
@@ -0,0 +1,114 @@
...
+			return nil, nil, err
+		}
+	}
+
+	return typ, markers, nil
+}
+
+// parseSpec parses a spec into a Type.
+// Handles both type strings and nested object maps.
+// Note: Markers from string specs are discarded; they're handled separately in buildSchema.
```
jakobmoellerdev: Why not split ParseField into ParseTypeString(s string) (types.Type, error) and ParseMarkers(s string) ([]*Marker, error), so parseSpec can call just ParseTypeString with no ambiguity?

pkg/simpleschema/parse.go
```diff
@@ -0,0 +1,114 @@
...
+		valType, err := parseType(val)
+		if err != nil {
+			return nil, err
+		}
+		return types.Map{Value: valType}, nil
+	}
+	if strings.HasPrefix(s, "map[") {
+		return nil, fmt.Errorf("map key must be string")
+	}
+	return types.Custom(s), nil
```
jakobmoellerdev, line 117: at this layer, parseType can't validate custom type existence (it doesn't have the registry). That's fine. But the function name and signature don't communicate that unrecognized strings become custom types. 

At minimum, probably worth a comment. Ideally, maybe we return something like types.Unresolved(s) that forces the caller to resolve it?, rather than types.Custom(s) which sounds like a valid, resolved type.

pkg/simpleschema/parse.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func parseMapString(s string) (val string, ok bool) {
```
jakobmoellerdev: parseMapString rejects `map[string]` but the error is silent

When someone writes map[string] with no value type, parseMapString returns 
false, so parseType falls through to:

```go
if strings.HasPrefix(s, "map[") {
    return nil, fmt.Errorf("map key must be string")
}
```

map[string] does have a string key — the problem is a missing value type. The error should be "map value type is required" or similar right?

pkg/simpleschema/parse.go
```diff
@@ -0,0 +1,114 @@
...
+			fields[name] = t
+		}
+		return types.Struct{Fields: fields}, nil
+	default:
+		return nil, fmt.Errorf("unexpected type: %T", spec)
+	}
+}
+
+// parseType parses a type string like "[]map[string]Person" into a Type.
+func parseType(s string) (types.Type, error) {
```
jakobmoellerdev, line 79: parseType doesn't guard against recursive depth

Pathological input like [][][][][][]...string (deeply nested slices) would recurse unboundedly. In practice this probably doesn't matter since input comes from RGD YAML, but maybe worth to do sth like

```
func parseType(s string) (types.Type, error) {
    return parseTypeWithDepth(s, 0)
}

func parseTypeWithDepth(s string, depth int) (types.Type, error) {
    if depth > 16 {
        return nil, fmt.Errorf("type nesting too deep: %q", s)
    }
    // ...
}
```

pkg/simpleschema/parse.go
a-hilaly: Good catch

pkg/simpleschema/types/atomic.go
a-hilaly: For some reason i always thought that bytes was a type... my way of saying that we can easily add new atomic types or new composite types easily via the Type/Resolver interface.

pkg/simpleschema/markers.go
a-hilaly: Unreachable code from a Package API perspective. But added it.

pkg/simpleschema/parse.go
a-hilaly: Added comments to explain when this is actually getting resolved

PR #1003, opened 2026-01-31
fix: applying resources with null fields

michaelhtm:
When applying fields that evaluate to `null`, we get an error denying the request.
This issue was not allowing us to create resources with fields that might evaluate to null.

With these changes, if a field evaluates to null, we remove it from the manifest before attempting to apply it.

example:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: bucket.kro.run
spec:
  schema:
    apiVersion: v1alpha1
    kind: S3Bucket
    spec:
      name: string | required=true
      policy: string
  resources:
  - id: policyBucket
    template:
      apiVersion: s3.services.k8s.aws/v1alpha1
      kind: Bucket
      metadata:
        name: ${schema.spec.name}
      spec:
        name: ${schema.spec.name}
        policy: '${schema.spec.?policy}'
```

### DIFF
--- a/pkg/runtime/resolver/resolver.go
+++ b/pkg/runtime/resolver/resolver.go
@@ -204,7 +204,12 @@ func (r *Resolver) setValueAtPath(path string, value interface{}) error {
 			if i == len(segments)-1 {
 				array := current.([]interface{})
 				if value == nil {
+					// Nil array elements are removed rather than set to null. SSA's structured-merge-diff
+					// library sets omitted fields to null, but most CRDs don't define nullable:true.
+					// Removing the element avoids SSA conflicts and aligns with K8s API conventions.
+					// See: kubernetes-sigs/structured-merge-diff#171, kubernetes/kubernetes#103011
 					array = append(array[:segment.Index], array[segment.Index+1:]...)
+					updateParent(parent, parentKey, parentIndex, array)
 				} else {
 					array[segment.Index] = value
 				}
@@ -221,8 +226,13 @@ func (r *Resolver) setValueAtPath(path string, value interface{}) error {
 			}
 
 			if i == len(segments)-1 {
-				// if the expression value evaluates to nil
-				// remove field (specifically in cases where we have optionals)
+				// Nil values trigger field deletion instead of setting null. This works around an
+				// SSA limitation where structured-merge-diff sets omitted fields to null during Apply.
+				// Most CRD providers (ACK, ASO, KCC) don't mark optional fields as nullable:true,
+				// causing SSA conflicts when null is applied. Deleting the field releases ownership
+				// and follows K8s API conventions (field omission > null). kro will not re-manage
+				// this field until the expression evaluates to non-nil.
+				// See: kubernetes-sigs/structured-merge-diff#171, kubernetes/kubernetes#103011
 				if value == nil {
 					delete(currentMap, segment.Name)
 				} else {
--- a/pkg/runtime/resolver/resolver_test.go
+++ b/pkg/runtime/resolver/resolver_test.go
@@ -187,6 +187,15 @@ func TestSetValueAtPath(t *testing.T) {
 			},
 		},
 		{
+			name: "set top level field to nil",
+			resource: map[string]interface{}{
+				"name": "${someexpression}",
+			},
+			path:  "name",
+			value: nil,
+			want:  map[string]interface{}{},
+		},
+		{
 			name: "set nested field",
 			resource: map[string]interface{}{
 				"spec": map[string]interface{}{},
@@ -200,6 +209,19 @@ func TestSetValueAtPath(t *testing.T) {
 			},
 		},
 		{
+			name: "set nested field to nil",
+			resource: map[string]interface{}{
+				"spec": map[string]interface{}{
+					"replicas": "${someexpression}",
+				},
+			},
+			path:  `spec.replicas`,
+			value: nil,
+			want: map[string]interface{}{
+				"spec": map[string]interface{}{},
+			},
+		},
+		{
 			name:     "create intermediate structures",
 			resource: map[string]interface{}{},
 			path:     `spec.template.metadata.name`,
@@ -210,6 +232,19 @@ func TestSetValueAtPath(t *testing.T) {
 						"metadata": map[string]interface{}{
 							"name": "my-pod",
 						},
+					},
+				},
+			},
+		},
+		{
+			name:     "create intermediate structures with nil value",
+			resource: map[string]interface{}{},
+			path:     `spec.template.metadata.name`,
+			value:    nil,
+			want: map[string]interface{}{
+				"spec": map[string]interface{}{
+					"template": map[string]interface{}{
+						"metadata": map[string]interface{}{},
 					},
 				},
 			},
@@ -248,6 +283,22 @@ func TestSetValueAtPath(t *testing.T) {
 			},
 		},
 		{
+			name: "set array element to nil",
+			resource: map[string]interface{}{
+				"containers": []interface{}{
+					map[string]interface{}{"name": "container1"},
+					"${someexpression}",
+				},
+			},
+			path:  "containers[1]",
+			value: nil,
+			want: map[string]interface{}{
+				"containers": []interface{}{
+					map[string]interface{}{"name": "container1"},
+				},
+			},
+		},
+		{
 			name:     "create array and set element",
 			resource: map[string]interface{}{},
 			path:     `spec.containers[0].ports[0].containerPort`,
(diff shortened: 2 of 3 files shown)


pkg/runtime/resolver/resolver.go
```diff
@@ -232,6 +232,12 @@ func (r *Resolver) setValueAtPath(path string, value interface{}) error {
 			}
 
 			if i == len(segments)-1 {
+				// if the expression value evaluates to nil
+				// remove field (specifically in cases where we have optionals)
+				if value == nil {
```
jakobmoellerdev, line 236: do you mind writing a test for this, please? 
thanks!

In general im wondering if the null field is something that we should actually allow and have a special case for this in our syntax. unsure. for now this is fine I believe

a-hilaly: I do have two questions:
1/ why doesn't SSA accept null here - is it failing at schema validation before SSA even runs because the field isn't marked `nullable: true`?

2/ Should ACK be marking these fields as `nullable` in their CRDs instead of kro stripping nulls, or is field omission the preferred Kubernetes convention for "no value"?

michaelhtm: Adding tests...

> In general im wondering if the null field is something that we should actually allow and have a special case for this in our syntax

I'm not sure if there are other modifications we'd want to do..but you're right, some sort of templating would be nice to have if there are other use cases besides making a field optional

michaelhtm: 1. The failure is happening as a result of SSA apply (i got the same error when using kubectl SSA as well)
2. You're right, making fields nullable on ACK side does also fix this issue. The different between both solutions is,
  a. In the implemented solution, the field would no longer be considered a managed field, so if the customer manually changes the field that is supposed to be null, the controller will not change it back
  b. if we expect CRDs that support `nullable` field property for optional fields, we would continue detecting drift detection for the specific field, but i noticed a lot of CRDs do not support make their fields nullable (checked ASO, KCC, and ACK)

a-hilaly: >The failure is happening as a result of SSA apply (i got the same error when using kubectl SSA as well)

why? does it only happen for that specific field? does it only happen for ACK resources? what's the behaviour for non-ACK resources?

I really think that this is a one way door, and it will be breaking if we ever decide to revert this behaviour back. Hence why we need to be careful here.

michaelhtm: it happens for fields that are not nullable. In ACK we do not set the nullable property, but i noticed other CRDs don't either (ASO/KCC). 
[Kubectl client side apply makes sure to remove null fields](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go#L691-L695), while SSA doesn't. This causes a failure with the following error: (this is the same error i got when using `kubectl --server-side apply`
```
The Bucket "bucket" is invalid: 
* spec.policy: Invalid value: "null": spec.policy in body must be of type string: "null"
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation 
```

jakobmoellerdev: My question to you would be: If we by default remove the field if the expression resolves to null, how would we enforce actually setting a field to null on purpose?

To me it sounds like this is a special case that is very hard to get right and the current implementation just changes from one opinionated behavior to another.

michaelhtm: If the expression resolves to null, kro does remove/omit the field. The only time it doesn't is when the user manually changes the field. In that case, kro does not enforce the removal

jakobmoellerdev: TLDR: I vote to accept the PR with still needed docs changes, but only as part of the next minor release and with major communication attached

Okay I went into the rabbit hole now:
The fundamental issue lives in the structured-merge-diff (SMD) library that powers SSA's merge algorithm. When a field manager previously owned a field and then omits it in a subsequent Apply, SMD sets the field to null in the merged object to signal deletion: https://github.com/kubernetes-sigs/structured-merge-diff/issues/171, https://github.com/kubernetes/kubernetes/issues/103011, https://github.com/grafana/mimir/issues/11200 as some examples where this popped up

The major infrastructure CRD providers (ACK, ASO, KCC) generally don't mark optional fields as nullable: true. This is partly convention (Kubernetes API conventions prefer field omission over null), and partly because nullable has its own complications — null values can get dropped from etcd and disappear from managedFields metadata when switching between null and non-null values.

Adding nullable: true to upstream CRDs is not something kro can control, and it would be unreasonable to gate kro's usability on CRD authors doing so.

Field omission is the Kubernetes convention for "no value." The Kubernetes API conventions have always preferred omitting fields over setting them to null. should kro's CEL expressions resolving to null map to "this field is not part of my intent" — which in SSA terms means "don't include it in the Apply, and release ownership.", or should it rather mean "this field should be null and that is my intent"?  I would argue it should be field omission. kubectl has stripped nulls before sending to the API server since forever. kro doing the same pre-Apply is thus not novel.

To align with CSA behavior I suggest:

- Marking the change as breaking and communicating clearly
- Document clearly that a CEL expression resolving to null means "omit this field from the Apply" (release ownership), not "set this field to null."
- If a user explicitly wants to set null (for CRDs that do support nullable), that should require a distinct expression/syntax in the future (e.g., an explicit nullValue() helper or similar). Don't conflate "expression returned nothing" with "I want the JSON null value."
- Add a comment in the code explaining the upstream SSA limitation

cheeseandcereal: > If the expression resolves to null, kro does remove/omit the field. The only time it doesn't is when the user manually changes the field. In that case, kro does not enforce the removal

It's worth noting, that this will permanently break some use-cases. For example, from an issue I posted in kubernetes 5 years ago: https://github.com/kubernetes/kubernetes/issues/46861#issuecomment-731700805

In that case, if you want to switch an env var value in a pod spec between `value` <-> `valueRef`, you do _have_ to explicitly null one of the fields. If kro doesn't support this "explicit null" in some way, it will break use-cases like this.

If we want to go forward with this change, there also needs to be a way to indicate "explicit null" for gotcha use-cases like I mentioned above. I think this probably needs to be explicitly included with the same release if this change is accepted.

---

It's also worth noting that if we _stop_ using implicit null, the behavior of SSA can get "unexpected" or "unintuitive" if there are multiple field owners for things kro is trying to manage (even when using things like the "force" option for resolving conflicts with the SSA request).

I can pretty much guarantee users will see unexpected behavior without implicit null if they have workflows where other things are touching the fields that kro is trying to manage due to SSA/field management behavior.

The alternative to this would be to keep the current behavior, and allow for distinct expression syntax that is "explicit undefined" instead. This would also prevent the need from making a breaking change here, while still supporting use-cases to change fields in APIs where explicit nulls are not allowed.

> When a field manager previously owned a field and then omits it in a subsequent Apply, SMD sets the field to null in the merged object to signal deletion

It's actually even more complicated than this. If the previously owned field was owned by _more than one_ field owner (which SSA "supports"), excluding that field in a subsequent apply will actually lead SSA to generate a patch excluding the field entirely (not null it). If you explicitly set to null in the apply, then SSA's patch will always include the 'null' for that field regardless of owners, and this will be a conflict that you have to force. Sometimes this is the only way to ensure an expected field is actually removed when applying with SSA if you have multiple, or other entity field ownership (although yes, the API also must support the explicit null).

shabbskagalwala:
Apologies for the comment on this - but i think this might become more needed. It practically breaks most major CRDs (ACK etc). So far AWS S3, ElastiCache, Kafka and many more of the major services run into this issue. 

```
- lastTransitionTime: "2026-02-25T16:00:11Z"
    message: 'resource reconciliation failed: apply results contain errors: Bucket.s3.services.k8s.aws
      "my-test-bucket-uw2" is invalid: [spec.lifecycle:
      Invalid value: "null": spec.lifecycle in body must be of type object: "null",
      <nil>: Invalid value: "null": some validation rules were not checked because
      the object was invalid; correct the existing errors to complete validation]'
```

Kro validations fail when setting the field to `{}` but the same field when added to the S3 CR object works fine (without using ssa). 

Maybe another option would be to use some explicit cel magic/helper to drop such fields - like maybe `omit()` or something similar, such that it doesn't interfere with SSA behavior as such


Once again, apologies for asking but is this fix maybe planned for anytime soon? With resources that can have a dozen permutation combinations of such fields, its not possible to use `includeWhen` on Kro either.

pkg/runtime/resolver/resolver.go
a-hilaly: Opened https://github.com/kubernetes-sigs/kro/pull/1121 for explicit field omission control

michaelhtm:
This issue has been addressed in https://github.com/kubernetes-sigs/kro/pull/1139
/close

PR #1014, opened 2026-02-03
Compile CEL programs at build time and reuse at runtime 

a-hilaly:
CEL Programs are stateless and thread safe, designed to compile once and
evaluate many times. Previously, we created and discarded CEL environments
on every expression evaluation at runtime: parse the string, compile to
AST, create Program, evaluate, then throw everything away. At build time,
we performed type checking but discarded the compiled Programs. This
wasted work that could be preserved.

This patch introduces an `Expression` struct holding both the original
CEL string and its compiled `Program`. The parser creates `Expression`
objects with only `Original` set. The builder uses a lightweight inspector
env for dependency extraction (just identifier names, no schemas), then a
single `TypedEnvironment` with all schemas for compilation. Following
Kubernetes CEL patterns: create one env at init, compile all expressions
against it, use AST inspection for scope validation. Compiled Programs
are preserved in `Expression.Program` and reused directly at runtime,
skipping parse and compile phases entirely.

### DIFF
--- a/pkg/cel/expression.go
+++ b/pkg/cel/expression.go
@@ -1,0 +1,81 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cel
+
+import (
+	"fmt"
+
+	"github.com/google/cel-go/cel"
+)
+
+// Expression wraps a CEL expression with its compiled program and metadata.
+// Programs are compiled once at graph build time and reused across reconciliations.
+// The struct is immutable and thread-safe after construction. It is save to use
+// by multiple runtimes and reconciliations in parallel - thanks to Program being
+// thread-safe.
+//
+// Lifecycle:
+//   - Parser: creates with Original set (References and Program nil)
+//   - Builder: populates References during dependency extraction (via ast.Inspector)
+//   - Builder: populates Program during compilation (after type validation)
+//   - Runtime: calls Eval() with context containing values for References
+type Expression struct {
+	// Original is the raw CEL expression string, preserved for error messages
+	// and debugging. Set by parser.
+	Original string
+
+	// References lists all identifiers this expression accesses (e.g., "schema", "vpc").
+	// These are the keys that must be present in the context passed to Eval.
+	// Set by builder during dependency extraction.
+	//
+	// Note: References includes "schema" if used, but schema is NOT a DAG dependency.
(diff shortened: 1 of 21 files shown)


pkg/cel/expression.go
```diff
@@ -0,0 +1,81 @@
...
+// The struct is immutable and thread-safe after construction. It is save to use
+// by multiple runtimes and reconciliations in parallel - thanks to Program being
+// thread-safe.
+//
+// Lifecycle:
+//   - Parser: creates with Original set (References and Program nil)
+//   - Builder: populates References during dependency extraction (via ast.Inspector)
+//   - Builder: populates Program during compilation (after type validation)
+//   - Runtime: calls Eval() with context containing values for References
+type Expression struct {
```
jakobmoellerdev, line 34: Im generally looking at this PR right now and believe this might take it too far. Programmatic caching of the CEL expression imho is achievable with much less effort than replacing all references to expressions strings. I believe the approach that Kubernetes takes in that there is a separate data structure that is referenced as a cache is more maintainable than pulling individual Expression structs through the codebase all the way from parser to runtime.

Also regarding thread-safety: while it may be true that a program is thread-safe, the context in which the program was created may have changed. For example, typing information may differ in a second execution due to the underlying schema having changed that results in the decl type which is the baseline for the type check.

a-hilaly: >Im generally looking at this PR right now and believe this might take it too far. Programmatic caching of the CEL expression imho is achievable with much less effort than replacing all references to expressions strings. I believe the approach that Kubernetes takes in that there is a separate data structure that is referenced as a cache is more maintainable than pulling individual Expression structs through the codebase all the way from parser to runtime.

[let me know if miss-understood your comment] I want to clarify the framing a bit - this isn't really a cache in the traditional sense (no lookups, no invalidation). It's pre compilation: we compile expressions once when the RGD graph is built, and after that point, the source string and its compiled program "travel" together as one unit. They're inseparable - the Expression struct inthe `cel` package is just the natural home for both. I kept the original string for craft comprehensive error messages that shows up logs/conditions.

Adding to that, the same expression string can compile differently depending on context - `${item.name}` has different type information depending on which `forEach` it's in. Hence why we can't "cache" based on expression strings.

> Also regarding thread-safety: while it may be true that a program is thread-safe, the context in which the program was created may have changed. For example, typing information may differ in a second execution due to the underlying schema having changed that results in the decl type which is the baseline for the type check.

when an RGD is updated, the entire graph is rebuilt from scratch - old compiled programs are discarded and new ones are compiled against the updated schema. There's no scenario where a program outlives its schema.

jakobmoellerdev: We synced on this and I want to summarize this in public domain as well as its quite valuable design-wise:

The main issue I had with this PR is that `Expression` now owns multiple parts of the State of a KRO Expression Eval. What do I mean by State?

1. The raw expression, evaluated by their dependencies
2. The compiled expression, represented by the CEL program
3. The evaluated expression, represented by a method on the expression

While the PR is functionally correct, it binds the codebase in a way that becomes harder to maintain. Ideally we want to achieve a "pipeline" of sorts that is also reflected in kro packages:

1. A package purely for dependency introspection (our current inspector)
2. A package purely for compilation of expressions (the CEL program cache, think of it as a `compiler`)
3. A package purely for runtime evaluation (this is our current graph runtime and builder)

We agreed to slowly move this into that direction to make the codebase more accessible for contributors as well, but I wont block this PR as my take is also that we should not re-program cel expression every run

PR #1015, opened 2026-02-03
fix(dynamiccontroller): fix ctx race condition, achieve 100% test coverage

a-hilaly:
Fix race condition where ctx field could be read concurrently during
handler registration while Start() was still initializing. Add
startedLock mutex to protect ctx access in `Start()` and `getContext()`

Achieves 100% test coverage for dynamiccontroller package.

### DIFF
--- a/pkg/dynamiccontroller/dynamic_controller_test.go
+++ b/pkg/dynamiccontroller/dynamic_controller_test.go
@@ -23,9 +23,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/kubernetes-sigs/kro/pkg/dynamiccontroller/internal"
-	"github.com/kubernetes-sigs/kro/pkg/metadata"
-	"github.com/kubernetes-sigs/kro/pkg/requeue"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
@@ -44,6 +41,10 @@ import (
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"github.com/kubernetes-sigs/kro/pkg/dynamiccontroller/internal"
+	"github.com/kubernetes-sigs/kro/pkg/metadata"
+	"github.com/kubernetes-sigs/kro/pkg/requeue"
 )
 
 // NOTE(a-hilaly): I'm just playing around with the dynamic controller code here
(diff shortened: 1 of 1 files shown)


a-hilaly:
```
go test ./pkg/dynamiccontroller/... -cover -race -count=1                                                                                                                                                                                                                                       9s 
# github.com/kubernetes-sigs/kro/pkg/dynamiccontroller.test
ok      github.com/kubernetes-sigs/kro/pkg/dynamiccontroller    5.203s  coverage: 100.0% of statements
ok      github.com/kubernetes-sigs/kro/pkg/dynamiccontroller/internal   2.844s  coverage: 93.8% of statements
```

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -158,9 +158,12 @@ type registration struct {
 //     (handler map). This allows concurrent updates to
 //     different GVRs without blocking each other.
 type DynamicController struct {
+	// startedLock protects the ctx field.
+	startedLock sync.Mutex
```
jakobmoellerdev: my personal opinion is that this is not a race. It is perfectly expected that Start is not safe for concurrent use because it is started once at controller start. Thus for me it is a bit strange to add here. WDYT?

pkg/dynamiccontroller/internal/lazy_informer.go
```diff
@@ -180,3 +183,11 @@ func (w *LazyInformer) Shutdown() {
 	w.informer = nil
 	w.handlers = make(map[string]cache.ResourceEventHandlerRegistration)
 }
+
+// SetInformerForTesting replaces the internal informer for testing purposes.
+// This should only be used in tests to inject mock informers.
+func (w *LazyInformer) SetInformerForTesting(inf cache.SharedIndexInformer) {
```
jakobmoellerdev, line 189: this should not be an exported method no?

pkg/dynamiccontroller/internal/lazy_informer.go
```diff
@@ -110,9 +110,12 @@ func (w *LazyInformer) AddHandler(ctx context.Context, id string, h cache.Resour
 
 	// Start informer if first handler
 	if len(w.handlers) == 1 {
-		go w.run()
+		// Capture references to avoid race when sync fails and we nil out w.informer
+		inf := w.informer
+		done := w.done
```
jakobmoellerdev, line 115: the capture of done here is quite interesting. we have a mutex that is locked so I dont see how a failure would race out here? do you mind explaining?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -160,8 +161,8 @@ type registration struct {
 type DynamicController struct {
 	// Parent run context, inherited by all informer stop contexts.
 	// It is set by Start and meant to be used to register the controller context with a global handler
-	// such as the controller-runtime manager.
-	ctx context.Context
+	// such as the controller-runtime manager. Thread-safe.
+	ctx atomic.Pointer[context.Context]
```
a-hilaly, line 165: Cool, thanks @jakobmoellerdev !

PR #1018, opened 2026-02-04
fix(runtime): handle empty forEach collections and their dependents

a-hilaly:
A bug reported in https://github.com/kubernetes-sigs/kro/issues/17#issuecomment-3843929563 showed that when a collections node produce zero
items, resources that depend on it through CEL expressions like `size(entries)`
fail with "no such attribute" errors. The instance stays stuck in `IN_PROGRESS`
forever because the dependent can't evaluate its expressions.

The problem was in `buildContext()`, which skips dependencies where
`len(observed) == 0`. This made sense for resources that haven't been observed
yet, but it also excluded empty collections that had been legitimately
observed as having zero items. The fix is to check `observed == nil` instead,
so an empty slice `[]` still gets included in the CEL context while a nil
slice (meaning not yet observed for collections) continues to be skipped.

However, fixing this broke existing tests. [The deep chaining test](https://github.com/kubernetes-sigs/kro/blob/main/test/integration/suites/core/collection_test.go#L1007) started
failing because Pods were being created with stale data. Investigating this
led to a second bug: `IsReady()` was returning true for any node without
`readyWhen` expressions, even if that node had never been observed from the
cluster. Before our `buildContext` fix, this didn't matter because unbsoserved
nodes were excluded from the context anyway. Basically the two bugs were
cancelling each other out.

Once `buildContext` started including empty observations, the readiness bug
became visible. A collection with zero desired items would report ready
before it had been observed, causing its dependents to render with
`entries = []` before the collection was actually processed.

The fix (for the second bug) is done by moving the `readyWhen` len check after
the observation check in both `isSingleResourceReady()` and `isCollectionReady()`.
A node must be observed before it can be considered ready, regardless of
whether it has readyWhen expressions.

### DIFF
--- a/pkg/controller/instance/resources.go
+++ b/pkg/controller/instance/resources.go
@@ -215,12 +215,6 @@ func (c *Controller) processNode(
 		return nil, "", err
 	}
 
-	// Nothing to apply (empty desired)
-	if len(desired) == 0 {
-		state.SetSkipped()
-		return nil, "", nil
-	}
-
 	switch node.Spec.Meta.Type {
 	case graph.NodeTypeExternal:
 		if err := c.processExternalRefNode(rcx, node, state, desired); err != nil {
@@ -303,12 +297,6 @@ func (c *Controller) processCollectionNode(
 
 	collectionSize := len(expandedResources)
 
-	// Empty collection is already handled - state would be Ready
-	if collectionSize == 0 {
-		state.SetReady()
-		return nil, nil
-	}
-
 	// LIST all existing collection items with single call (more efficient than N GETs)
 	existingItems, err := c.listCollectionItems(rcx, gvr, id)
 	if err != nil {
@@ -319,6 +307,12 @@ func (c *Controller) processCollectionNode(
 	// Pass unordered observed items to runtime; it will align them to desired
 	// order by identity.
 	node.SetObserved(existingItems)
+
+	// Empty collection: observed is set (possibly with orphans to prune), mark ready.
+	if collectionSize == 0 {
+		state.SetReady()
+		return nil, nil
+	}
 
 	// Build lookup map for current items keyed by namespace/name.
 	existingByKey := make(map[string]*unstructured.Unstructured, len(existingItems))
--- a/pkg/runtime/node.go
+++ b/pkg/runtime/node.go
@@ -513,9 +513,6 @@ func (n *Node) IsReady() (bool, error) {
 		return true, nil
 	}
 
-	if len(n.readyWhenExprs) == 0 {
-		return true, nil
-	}
 	if n.Spec.Meta.Type == graph.NodeTypeCollection {
 		return n.isCollectionReady()
 	}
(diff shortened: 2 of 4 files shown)


pkg/controller/instance/resources.go
```diff
@@ -215,8 +215,9 @@ func (c *Controller) processNode(
 		return nil, "", err
 	}
 
-	// Nothing to apply (empty desired)
-	if len(desired) == 0 {
+	// Nothing to apply (empty desired). Collections still need to proceed
+	// to SetObserved so empty collections are visible to dependents.
+	if len(desired) == 0 && node.Spec.Meta.Type != graph.NodeTypeCollection {
```
jakobmoellerdev: can we introduce a method to say `IsCollection()` to simplify this a bit?
also im wondering:

wouldnt it be smarter so make a check like this pseudocode 

```
if len(desired) == 0
  if isCollection == true
     state.SetReady()
  else
     state.SetSkipped()
  return
```

imo this would make the whole code path easier to understand. let me know what you think of that

pkg/runtime/node.go
```diff
@@ -513,9 +513,6 @@ func (n *Node) IsReady() (bool, error) {
 		return true, nil
 	}
 
-	if len(n.readyWhenExprs) == 0 {
-		return true, nil
-	}
```
michaelhtm, line 518: why is this check moving?

pkg/runtime/node_test.go
```diff
@@ -362,6 +381,45 @@ func TestNode_IsReady_Collection(t *testing.T) {
...
+				withDesired(
+					newUnstructured("v1", "ConfigMap", "ns", "cm-1"),
+					newUnstructured("v1", "ConfigMap", "ns", "cm-2"),
+				).
+				withObservedUnstructured(newUnstructured("v1", "ConfigMap", "ns", "cm-1")).
+				build(),
+			wantReady: false,
+		},
+		{
+			name: "collection without readyWhen, observed >= desired - ready",
```
michaelhtm, line 410: nit: should we test collection with readyWhen?

pkg/runtime/node.go
a-hilaly: it's moving inside `isSingleResourceReady` and `isCollectionReady`:
> A node must be observed before it can be considered ready, regardless of
whether it has readyWhen expressions.

pkg/runtime/node_test.go
a-hilaly: I have patch locally making graph builder and runtime.go test coverage 100% i'll ship it after this.

pkg/controller/instance/resources.go
```diff
@@ -303,12 +297,6 @@ func (c *Controller) processCollectionNode(
 
 	collectionSize := len(expandedResources)
 
-	// Empty collection is already handled - state would be Ready
-	if collectionSize == 0 {
-		state.SetReady()
-		return nil, nil
-	}
```
a-hilaly, line 310: Completely nuked.

a-hilaly: The shortcut is not actually needed. All subsequent functions handle this case.

jakobmoellerdev (commented):
/lgtm

thats now a much smoother path through the processNode. thanks for doing another iteration

PR #1023, opened 2026-02-04
fix: validate status schema

michaelhtm:
Currently, when the RGD instance status schema is processed, the CRD it
generates only contains the status fields that contain a CEL expression.
This can mislead users who define typed fields as instance status.

With this PR, we return an error if an RGD has status fields with no CEL
expressions


a-hilaly:
@michaelhtm looks linter isn't happy. can you fix it?

a-hilaly:
@michaelhtm can we also have integration tests?

michaelhtm:
Sure, I can add some tests

a-hilaly (approved):
👍

jakobmoellerdev (commented):
/lgtm

thanks for that fix!

Issue #648, opened 2025-08-27
Should apiVersion and group be separated in spec.schema?


ellistarn:
I'm playing around a bit with group versions for KRO and noticed that apiVersion and group are separated.

```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: my-object
spec:
  schema:
    apiVersion: v1alpha1
    group: my.group
    kind: MyObject
```

As opposed to :
```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: my-object
spec:
  schema:
    apiVersion: my.group/v1alpha1
    kind: MyObject
```

I noticed this when trying to instantiate MyObject by just copy pasting the schema and swapping out the schema values for real values. 

Is this intentionally separated?

jakobmoellerdev:
Hi! Thanks for this. I do agree that this is a bad misnomer on KRO side. apiVersion is standardized and what kro is trying to achieve more matches "Group, Version, Kind". Thus we should probably rename apiVersion to "Version" for the next iteration of RGD. Thanks for bringing this up. Interested on your takes @barney-s @a-hilaly even though im certain you guys are aware of this already

a-hilaly:
@ellistarn @jakobmoellerdev for context i was thinking of managed RGD versions where users do not provide a "Version" but still provide a group => https://github.com/kro-run/kro/pull/176#discussion_r1913930057 . However yeah i do believe that we need a better UX around this.

jakobmoellerdev:
I think on this side we might wanna go for either apiVersion/kind or (group)+version+kind, to accomodate the forms

ellistarn:
I think there is value in modeling the object in the schema as close as possible to the instantiation. This would suggest:
```
spec:
  schema:
    apiVersion: my.group/v1alpha1
    kind: MyObject
```

> where users do not provide a "Version" but still provide a group

I'm not sure I understand the thinking here. I expect RGD authors would want to maintain control over versioning. I'm not sure this is something that KRO can solve (beyond basic validation mutations, etc).

a-hilaly:
@ellistarn @jakobmoellerdev Yeah, I agree the current separation is confusing. Let's go with the standard `apiVersion: my.group/v1alpha1` format - it's what everyone expects from Kubernetes resources anyway. For the managed versions use case I mentioned we can figure that out separately. The UX improvement here is worth it.

Added a label for `v1beta1` and tagging this issue

[...]

PR #1025, opened 2026-02-05
combine group and apiversion in the apiVersion field

olamilekan000:
```
change combines the group and apiversion fields into the apiVersion to match 
k8s naming convention for apiVersion. 
This also deprecates the group field.
```

Fixes [648](https://github.com/kubernetes-sigs/kro/issues/648)

Below shows the result of the changes:
- setting `apiVersion` without a group being set
image

- setting `apiVersion` and also setting the group field
image

- setting apiVersion value with both group and apiversion
image

### DIFF
--- a/examples/apigateway/vpc-lattice/rg-lattice.yaml
+++ b/examples/apigateway/vpc-lattice/rg-lattice.yaml
@@ -4,7 +4,7 @@ metadata:
   name: lattice.kro.run
 spec:
   schema:
-    apiVersion: v1alpha1
+    apiVersion: kro.run/v1alpha1
     kind: Lattice
     spec:
       name: string
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -52,14 +52,15 @@ type Schema struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="kind is immutable"
 	Kind string `json:"kind,omitempty"`
 	// APIVersion is the version identifier for the generated CRD.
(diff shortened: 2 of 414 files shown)


olamilekan000:
@jakobmoellerdev can you kindly review

jakobmoellerdev:
Hi there! Thanks for this. I think my main issue with this is that it requires a dedicated migration path and would be a breaking change. I'm not sure if we can really afford to have everyone adjust their apiVersion like this. Do you think it would make sense to allow only apiVersion and assume a sane default?

/ok-to-test

olamilekan000:
> Hi there! Thanks for this. I think my main issue with this is that it requires a dedicated migration path and would be a breaking change. I'm not sure if we can really afford to have everyone adjust their apiVersion like this. Do you think it would make sense to allow only apiVersion and assume a sane default?
> 
> /ok-to-test

so something like, If `apiVersion` doesn't contain "/" or a group, we assume group is kro.run (or use the group field if set)

examples/apigateway/vpc-lattice/rg-lattice.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: lattice.kro.run
 spec:
   schema:
-    apiVersion: v1alpha1
+    apiVersion: kro.run/v1alpha1
```
ellistarn, line 7: This category of change could be made in isolation and rapidly merged. Separate from the interface change you are making which I suspect will take a bit longer.

a-hilaly:
Thanks for working on this.

A change like this needs a migration plan, backwards compatibility, and closer collaboration as it is developed, since it changes part of the RGD contract for existing users.

As written, this PR is too broad to take without that in place first. I think the idea should stay tracked in #648, and we can revisit it later with a backwards-compatible design.

Issue #1024, opened 2026-02-05
Empty arrays are replaced with null


artyom-p:
### Description

**Observed Behavior**:

```yaml
  spec:
    schema:
      group: mygroup
      apiVersion: v1
      kind: MyCR   
      types:
        MyType:
          name: string | required="true"
    spec:
      items: '[]MyCR | required="true"'
  resources:
    - id: resource
       template:
         ...
         arrayValue: ${schema.spec.items} # this is replaced with null and valdiation of that resource fails
  ```

When defined array property is empty - `[]`, KRO seem to replace it with `null` that causes furhter validation issues.
 
**Expected Behavior**:
Empty array should be forwarded furthere and not replace it with null

**Versions**:
- kro version: 0.8.4
- Kubernetes Version (`kubectl version`): 1.34

PR #1026, opened 2026-02-06
Fix empty arrays being converted to null instead of []

a-hilaly:
resolves https://github.com/kubernetes-sigs/kro/issues/1024

Empty CEL lists were converted to nil slices in `convertList` because
the result variable was declared with no-value- `var`instead of `make`.
When serialized to JSON, nil slices become `null` rather than empty
arrays, causing downstream validation failures when users pass empty
arrays through schema fields

### DIFF
--- a/pkg/cel/conversions.go
+++ b/pkg/cel/conversions.go
@@ -73,7 +73,7 @@ func convertList(v ref.Val) (interface{}, error) {
 	if !ok {
 		return v.ConvertToNative(reflect.TypeOf([]interface{}{}))
 	}
-	var result []interface{}
+	result := make([]interface{}, 0)
 	it := lister.Iterator()
 	for it.HasNext() == types.True {
 		elem := it.Next()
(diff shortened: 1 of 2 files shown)


pkg/cel/conversions.go
```diff
@@ -73,7 +73,7 @@ func convertList(v ref.Val) (interface{}, error) {
 	if !ok {
 		return v.ConvertToNative(reflect.TypeOf([]interface{}{}))
 	}
-	var result []interface{}
+	result := make([]interface{}, 0)
```
jakobmoellerdev, line 76: this is a change from default cel behavior in that we now have the Zero value set as an empty slice. I think this is fine but needs to be documented. I think we should open a section around CEL behavior in KRO in the docs.

a-hilaly: Doesn't null already get handled before we reach this code? If we're in `convertList,` CEL gave us an actual list - the fix just preserves that distinction.

my understanding is that by the time you're in `convertList`  CEL already said this can't be nul.

jakobmoellerdev: true, agree actually.

PR #1029, opened 2026-02-08
Handle CRD deletion so RGDs don’t get stuck in Active

Aman-Cool:
While testing cleanup scenarios, I ran into a case where deleting a KRO-managed CRD externally leaves the owning RGD permanently in Active, even though the CRD (and all its instances) are gone.

It turns out the CRD metadata watch was filtering out delete events, so the controller never re-reconciled after the CRD disappeared. The mapper already handles delete events correctly — they just weren’t reaching it.

This change allows CRD delete events through so the RGD can reconcile and recover via the existing `crdManager.Ensure()` logic.

Small change, but it restores correct state and self-healing when CRDs are removed outside the controller.

### DIFF
--- a/test/integration/suites/core/crd_test.go
+++ b/test/integration/suites/core/crd_test.go
@@ -476,5 +476,86 @@ var _ = Describe("CRD", func() {
 
 			Expect(env.Client.Delete(ctx, rgd)).To(Succeed())
 		})
+
+		It("should detect externally deleted CRD and recreate it", func(ctx SpecContext) {
+			rgdName := "test-crd-ext-deletion"
+			rgd := generator.NewResourceGraphDefinition(rgdName,
+				generator.WithSchema(
+					"TestExtDeletion", "v1alpha1",
+					map[string]interface{}{
+						"field1": "string",
+						"field2": "integer | default=42",
+					},
+					nil,
+				),
+				generator.WithResource("res1", map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "${schema.spec.field1}",
+					},
+					"data": map[string]interface{}{
+						"key":  "value",
+						"key2": "${string(schema.spec.field2)}",
+					},
+				}, nil, nil),
+			)
+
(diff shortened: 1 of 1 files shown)


Aman-Cool:
@jakobmoellerdev @a-hilaly ,This PR fixes a case where deleting a KRO-managed CRD externally leaves the owning ResourceGraphDefinition stuck in `Active`. By allowing CRD delete events to trigger reconciliation, the existing self-healing path is restored.
Please let me know if any changes are needed.

chrisdoherty4:
This probably needs an integration test. 

As an aside, I'm wondering, given the CRD is completely managed by the controller, why we don't configure an `Owns()` on the CRD? Edit: probably because we don't configure OwnerReferences?

Aman-Cool:
@chrisdoherty4 ,Thanks for the suggestion, that makes sense.

I’ve added an integration test under the existing CRD Watch Reconciliation suite that deletes a KRO-managed CRD out-of-band and checks that the controller recreates it and the owning RGD recovers to Active.

It fails on the old behavior and passes with this change. Happy to adjust the test or move it if you think it belongs elsewhere.

jakobmoellerdev:
> This probably needs an integration test.
> 
> As an aside, I'm wondering, given the CRD is completely managed by the controller, why we don't configure an `Owns()` on the CRD? Edit: probably because we don't configure OwnerReferences?

Owner References with RGDs have been very much under discussion. Currently the status quo is that none are maintained as the Kubernetes based GC is considered too dangerous as it will trigger a CRD delete which in turn can break users unintentionally. There are some designs out there, but ultimately we have not pushed it out yet.

Aman-Cool:
@jakobmoellerdev @chrisdoherty4  ,Thanks for the review and for taking the time to explain the OwnerReferences background, that context was really helpful. Appreciate the guidance.

a-hilaly:
@Aman-Cool looks like the integration test in here is flaky and causes the integrations tests to fail in several other PRs. Can you take a look?

Aman-Cool:
@a-hilaly thanks for the heads-up! Opened a follow-up PR(#1037) to fix the flaky integration test and make the cleanup deterministic.

Issue #806, opened 2025-11-12
Support for RGDs to manage Cluster Scoped CRDs


a-buck:
### Feature Description

**Problem Statement**:
Currently, kro hard codes the generated CRD from a RGD to be namespaced scoped only. It would be great to be able to generate cluster scoped CRDs as well.

**Proposed Solution**:

A new field `namespaced` (bool) can be added to ResourceGraphDefinition.spec.schema.

**Alternatives Considered**:

**Additional Context**:

This is particularly motivated for us by RGDs that create a mixture of cluster and namespace scoped resources

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

dhiraj-143r:
will you plz assign me this issue , I am interested to contribute in it

conqquest:
i am interested to work on it

jakobmoellerdev:
@a-hilaly i believe there were quite a bit of discussions around this whole topic a few weeks back in some community call. Do you recall what our outcome here was?

a-hilaly:
IIRC I'no one was opposed to the idea of adding an immutable `spec.schema.scope` that lets users define the scope of their instances. I still don't know the implications and tradeoffs of supporting this, but i can see valid use cases for it for sure.

antcybersec:
hey @a-hilaly , From the discussion above, it sounds like the direction might be to introduce something like spec.schema.scope (immutable) instead of just a namespaced bool. I’m happy to put together a KREP first to outline the design, possible impacts, and how cluster-scoped CRDs would be handled.

a-hilaly:
Go for it, thank you @antcybersec !

PR #1030, opened 2026-02-09
KREP-10 for cluster-scoped instance CRDs

antcybersec:
This PR adds a KREP document for supporting cluster-scoped instance CRDs in kro. This is a design proposal only - implementation will follow in a separate PR after design review and approval. Refs: #806

### DIFF
--- a/docs/design/proposals/cluster-scoped-instances.md
+++ b/docs/design/proposals/cluster-scoped-instances.md
@@ -21,6 +21,58 @@ Users have expressed the need for this capability, particularly for RGDs that cr
 
 2. **Mixed Scope Resources**: Create RGDs that manage both cluster-scoped and namespaced resources together, where the instance itself needs to be cluster-scoped to properly represent the cluster-level component.
 
+   **Example**: An RGD that creates a `ClusterRole` (cluster-scoped) along with a `RoleBinding` (namespaced) that binds it to a `ServiceAccount` in a specific namespace. The instance itself must be cluster-scoped because `ClusterRole` is cluster-scoped, even though the `RoleBinding` and `ServiceAccount` are namespaced:
+
+   ```yaml
+   apiVersion: kro.run/v1alpha1
+   kind: ResourceGraphDefinition
+   metadata:
+     name: cluster-role-with-binding
+   spec:
+     schema:
+       apiVersion: v1alpha1
+       kind: ClusterAccessPolicy
+       scope: Cluster  # Instance is cluster-scoped
+       spec:
+         roleName: string
+         namespace: string  # Where to create the RoleBinding
+         serviceAccountName: string
+         rules: list<map>
+     resources:
+       # Cluster-scoped resource
+       - id: clusterrole
+         template:
+           apiVersion: rbac.authorization.k8s.io/v1
+           kind: ClusterRole
+           metadata:
+             name: ${schema.spec.roleName}
+           rules: ${schema.spec.rules}
+       # Namespaced resource
+       - id: serviceaccount
+         template:
+           apiVersion: v1
+           kind: ServiceAccount
+           metadata:
+             name: ${schema.spec.serviceAccountName}
+             namespace: ${schema.spec.namespace}
+       # Namespaced resource that references cluster-scoped ClusterRole
+       - id: rolebinding
+         template:
+           apiVersion: rbac.authorization.k8s.io/v1
+           kind: RoleBinding
+           metadata:
+             name: ${schema.spec.roleName}-binding
+             namespace: ${schema.spec.namespace}
+           roleRef:
+             apiGroup: rbac.authorization.k8s.io
+             kind: ClusterRole
+             name: ${clusterrole.metadata.name}
+           subjects:
+             - kind: ServiceAccount
+               name: ${serviceaccount.metadata.name}
+               namespace: ${schema.spec.namespace}
+   ```
+
 3. **Operator Patterns**: Support operator-like patterns where the instance CRD represents a cluster-wide configuration that manages multiple namespaced resources.
 
 ## Proposal


docs/design/proposals/cluster-scoped-instances.md
```diff
@@ -0,0 +1,226 @@
...
+- Define infrastructure components that are inherently cluster-wide
+- Create resources that need to exist at the cluster level for operational reasons
+
+Users have expressed the need for this capability, particularly for RGDs that create a mixture of cluster and namespace-scoped resources (see [issue #806](https://github.com/kubernetes-sigs/kro/issues/806)).
+
+### Use Cases
+
+1. **Cluster Infrastructure Management**: Define RGDs that manage cluster-level infrastructure components like cluster-wide policies, cluster issuers, or cluster-scoped storage classes.
+
+2. **Mixed Scope Resources**: Create RGDs that manage both cluster-scoped and namespaced resources together, where the instance itself needs to be cluster-scoped to properly represent the cluster-level component.
```
chrisdoherty4, line 22: Out of curiosity, do you have a concrete example of a combination RGD that specifies both namespace and cluster scoped resources?

antcybersec: > Out of curiosity, do you have a concrete example of a combination RGD that specifies both namespace and cluster scoped resources?

 I've added this example to the KREP document.

docs/design/proposals/cluster-scoped-instances.md
```diff
@@ -0,0 +1,278 @@
...
+        rules: ${schema.spec.rules}
+```
+
+### Backward Compatibility
+
+- **Default Behavior**: When `scope` is omitted, it defaults to `Namespaced`, preserving existing behavior
+- **Existing RGDs**: All existing RGDs continue to work without modification
+- **Immutable Field**: Once set, scope cannot be changed, preventing migration issues
+
+### Validation
```
jakobmoellerdev, line 161: I believe this is a good summary. I am missing validation that if we have a cluster scoped instance, the creation of namespaced resources is impossible without defining a namespace (or we need to define a default namespace that without explicitly being set is `default`). This needs a decision I believe.

docs/design/proposals/cluster-scoped-instances.md
```diff
@@ -0,0 +1,278 @@
...
+### Risks and Mitigations
+
+1. **Risk**: Users might accidentally create cluster-scoped CRDs
+   - **Mitigation**: Default to `Namespaced`, requires explicit opt-in
+
+2. **Risk**: Scope immutability might be confusing
+   - **Mitigation**: Clear documentation and validation error messages
+
+3. **Risk**: Controller bugs with cluster-scoped instances
+   - **Mitigation**: Comprehensive testing, especially for status updates
```
jakobmoellerdev, line 270: I think for testing I would like to avoid multiplying out tests for namespaced and cluster scoped. how could we achieve this?

docs/design/proposals/cluster-scoped-instances.md
```diff
@@ -0,0 +1,278 @@
...
+```go
+type ResourceScope string
+
+const (
+	ScopeNamespaced ResourceScope = "Namespaced"
+	ScopeCluster    ResourceScope = "Cluster"
+)
+```
+
+#### Implementation Changes
```
NickKeller, line 113: Should there also be validation that rejects an RGD where the scope is Namespaced, but it contains resources that are cluster scoped?

docs/design/proposals/cluster-scoped-instances.md
```diff
@@ -0,0 +1,278 @@
...
+
+Users have expressed the need for this capability, particularly for RGDs that create a mixture of cluster and namespace-scoped resources (see [issue #806](https://github.com/kubernetes-sigs/kro/issues/806)).
+
+### Use Cases
+
+1. **Cluster Infrastructure Management**: Define RGDs that manage cluster-level infrastructure components like cluster-wide policies, cluster issuers, or cluster-scoped storage classes.
+
+2. **Mixed Scope Resources**: Create RGDs that manage both cluster-scoped and namespaced resources together, where the instance itself needs to be cluster-scoped to properly represent the cluster-level component.
+
+   **Example**: An RGD that creates a `ClusterRole` (cluster-scoped) along with a `RoleBinding` (namespaced) that binds it to a `ServiceAccount` in a specific namespace. The instance itself must be cluster-scoped because `ClusterRole` is cluster-scoped, even though the `RoleBinding` and `ServiceAccount` are namespaced:
```
chrisdoherty4, line 24: Doesn't this example mean there's only a single instance of the RGD that can be created since you can't re-create the cluster role? 

Its not clear to me that there's a use-case for mixed scope RGDs. As others have said I'm wondering if we should prevent this to start.

docs/design/proposals/cluster-scoped-instances.md
chrisdoherty4: I have a similar thought (https://github.com/kubernetes-sigs/kro/pull/1030/changes#r2795734923).

jakobmoellerdev: why is that? is it not valid to have cluster scoped rbac for a namespace scoped RGD? I currently dont see a need to artificially limit this. Maybe im missing some nuance here

docs/design/proposals/cluster-scoped-instances.md
jakobmoellerdev: Mixed Scope RGDs are a reality and are used. If we enforce such a limit you would need to create multiple RGDs to split cluster and namespace scoped resources. Im not sure we really want this.

NickKeller: I think we would need to consider the case where 2 RGDs define a cluster scoped resource, but the configuration is different(i.e. 2 RGDs define a ClusterRole `myclusterrole`, but each RGD defines different permissions). I think this would cause unexpected results/errors, so at the very least this scenario should be rejected at validation

docs/design/proposals/cluster-scoped-instances.md
NickKeller: My thought is that if we're requiring the instance CR to be cluster scoped if there are cluster scoped resources in the RGD, then it's an error if the user specifies `scope:Namespaced` or leaves it blank, and it should be rejected

docs/design/proposals/cluster-scoped-instances.md
jakobmoellerdev: this is already a handled case and KRO is able to detect such a conflict and provide an error in the runtime. I believe this is expected and okay.

docs/design/proposals/cluster-scoped-instances.md
jakobmoellerdev: how can scope be Namespaced if the instance CR is cluster scoped? how is that even possible? am I missing something?

NickKeller: Using a minimal example(from the KREP), I'm saying that this should be invalid and rejected, since `spec.scope` == "Namespaced" (the example has spec.scope == Cluster)

```yaml
apiVersion: kro.run/v1alpha1
   kind: ResourceGraphDefinition
   metadata:
     name: cluster-role-with-binding
   spec:
     schema:
       apiVersion: v1alpha1
       kind: ClusterAccessPolicy
       scope: **Namespaced**
       spec:
         roleName: string
         namespace: string  # Where to create the RoleBinding
         serviceAccountName: string
         rules: list<map>
     resources:
       # Cluster-scoped resource
       - id: clusterrole
         template:
           apiVersion: rbac.authorization.k8s.io/v1
           kind: ClusterRole
           metadata:
             name: ${schema.spec.roleName}
           rules: ${schema.spec.rules}
```

Unless this is already rejected (sorry I'm new here :) )

chrisdoherty4: Having thought on it more I agree with @jakobmoellerdev, no validation is needed. I thought through some use-cases - there are several but the most obvious to me is the operator pattern that often combines a Deployment (namespace scoped) with a ClusterRole and *WebhookConfiguration, both of which are cluster scoped.

I think there are debatable best practices, and then there's what Kro should enforce and I don't think Kro should enforce anything here.

Obviously if you have a cluster scoped RGD with namespace scoped resources Kro will need to be opinionated on behavior.

docs/design/proposals/cluster-scoped-instances.md
chrisdoherty4: Replied in the other thread. This can be resolved.

antcybersec:
> @antcybersec Can you please mark this as KREP 13?

hey @a-hilaly I will do this asap

a-hilaly:
@antcybersec are you still planning on implementing this? this krep was accepted for the 0.9 release which is happening next week.

a-hilaly:
cc @antcybersec friendly ping

antcybersec:
> cc @antcybersec friendly ping

hey @a-hilaly is there anything i need to do ...

a-hilaly:
@antcybersec i see some commits tagged here referencing this KREP, mind opening a PR?

PR #1033, opened 2026-02-10
Add json.marshal and json.unmarshal cel functions

NicholasBlaskey:
Add a `json` cel library to allow easier integration from / to json in RGDs

initially has
- ```json.marshal```
- ```json.unmarshal```

Tested with example.

Also tested error case of 
```
                    - name: JSON_UNMARSHAL_ERROR_EXAMPLE
                      value: "${json.unmarshal('{\"exampleJSONKey\": \"exampleJSONValue\"}').exampleJSONKey}"
```

which causes a reconciliation failure 
```
    Last Transition Time:  2026-02-10T19:45:47Z
    Message:               resource reconciliation failed: node "deployment": eval "json.unmarshal('not valid json at all').key": json.unmarshal failed to parse JSON: invalid character 'o' in literal null (expecting 'u')
    Observed Generation:   1
    Status:                False
    Type:                  Ready
```

### DIFF
--- a/pkg/cel/library/json.go
+++ b/pkg/cel/library/json.go
@@ -16,6 +16,7 @@ package library
 
 import (
 	"encoding/json"
+	"math"
 	"reflect"
 
 	"github.com/google/cel-go/cel"
@@ -58,11 +59,28 @@ import (
 //	json.marshal({"name": "test", "count": 42})
 //	json.marshal([1, 2, 3])
 //	json.marshal(schema.spec.config)
-func JSON() cel.EnvOption {
-	return cel.Lib(&jsonLibrary{})
+func JSON(options ...JSONOption) cel.EnvOption {
+	lib := &jsonLibrary{version: math.MaxUint32}
+	for _, o := range options {
+		lib = o(lib)
+	}
+	return cel.Lib(lib)
 }
 
-type jsonLibrary struct{}
+type jsonLibrary struct {
+	version uint32
+}
+
+// JSONOption is a functional option for configuring the json library.
+type JSONOption func(*jsonLibrary) *jsonLibrary
+
+// JSONVersion configures the version of the json library.
+func JSONVersion(version uint32) JSONOption {
+	return func(lib *jsonLibrary) *jsonLibrary {
+		lib.version = version
+		return lib
+	}
+}
 
 func (l *jsonLibrary) LibraryName() string {
 	return "json"
--- a/pkg/cel/library/json_test.go
+++ b/pkg/cel/library/json_test.go
@@ -211,3 +211,9 @@ func TestJSONMarshal(t *testing.T) {
 		})
 	}
 }
+
+func TestJSONVersion(t *testing.T) {
+	lib := &jsonLibrary{}
+	JSONVersion(1)(lib)
+	assert.Equal(t, uint32(1), lib.version)
+}


pkg/cel/conversion/conversions.go
a-hilaly, line 1: I agree that the cel package deserves a better organization. Can we make this change in a seperate PR?

a-hilaly (commented):
Can you please update the CEL library docs? https://kro.run/docs/concepts/rgd/cel-expressions#available-cel-libraries

pkg/cel/conversion/conversions.go
jakobmoellerdev: yes please, this should be focused to the functions only.

pkg/cel/library/json.go
```diff
@@ -0,0 +1,123 @@
...
+
+type jsonLibrary struct{}
+
+func (l *jsonLibrary) LibraryName() string {
+	return "json"
+}
+
+func (l *jsonLibrary) CompileOptions() []cel.EnvOption {
+	return []cel.EnvOption{
+		cel.Function("json.unmarshal",
```
jakobmoellerdev, line 91: this should be versioned internally to v1alpha1

pkg/cel/library/json.go
```diff
@@ -0,0 +1,123 @@
...
+		),
+	}
+}
+
+func (l *jsonLibrary) ProgramOptions() []cel.ProgramOption {
+	return nil
+}
+
+func unmarshalJSON(jsonString ref.Val) ref.Val {
+	if jsonString.Type() != types.StringType {
```
jakobmoellerdev: I think you could use the native default type adapter here to get a string.

pkg/cel/conversion/conversions.go
NicholasBlaskey: Yep can do

I moved these functions into a new package to resolve a circular import error I was getting

NicholasBlaskey: created https://github.com/kubernetes-sigs/kro/pull/1043

NicholasBlaskey:
> Can you please update the CEL library docs? https://kro.run/docs/concepts/rgd/cel-expressions#available-cel-libraries

updated

a-hilaly:
@jakobmoellerdev @matthchr shall we pin this for 0.9.0? or are we good with a 0.8.6? personally i'm fine with either

jakobmoellerdev:
I would favor a pin but won't block on 0.8.6

pkg/cel/library/json.go
NicholasBlaskey: Could you clarify what you mean for versioned internally here? 

I see some other libraries have integer versioning like this but not seeing any use a string version like `v1alpha1`. The random library also doesn't seem to have any versioning currently. 
https://github.com/google/cel-go/blob/v0.27.0/ext/lists.go#L213

jakobmoellerdev: mabye take a look at https://github.com/google/cel-go/blob/v0.27.0/ext/lists.go#L175-L189 or https://github.com/google/cel-go/blob/v0.27.0/ext/sets.go#L88-L97. Its basically about introducing a version so we can control it later. i dont really mind if its v1alpha1 or an integer tbh.

NicholasBlaskey: added support for setting a version 

currently the version integer is unused but can be referenced for future updates to the library

jakobmoellerdev:
On another note: https://github.com/google/cel-go/pull/1261/changes introduced a JSON Value compatibility. Do you think this is relevant here?

NicholasBlaskey:
> On another note: https://github.com/google/cel-go/pull/1261/changes introduced a JSON Value compatibility. Do you think this is relevant here?

doesn't seem relevant for `unmarshal`

For ```marshalJSON``` we could use `value.ConvertToNative(types.JSONValueType)` instead of our helper `conversion.GoNativeType(value)`.

As far as I can tell they should be equivalent but don't have a strong opinion either way, open to changing it after https://github.com/kubernetes-sigs/kro/pull/1047 merges

a-hilaly:
@NicholasBlaskey can you please rebase this one?

a-hilaly:
cc @jakobmoellerdev

PR #1035, opened 2026-02-10
KREP-011: RGD Variables

michaelhtm:
KREP-011 introduces a type of resource in the RGD defined using `data`.
Unlike the `template` and `externalRef` resources, this new type does
not map to a kubernetes resource, but instead it holds complex data that 
can be referenced throughout the RGD

### DIFF
--- a/docs/design/proposals/variables.md
+++ b/docs/design/proposals/variables.md
@@ -1,0 +1,495 @@
+# KREP-011: RGD Variables
+
+## Problem statement
+
+kro does not support intermediate computed values in ResourceGraphDefinitions.
+This creates two problems:
+
+1. **No reusable expressions.** When a complex CEL expression like
+   `${schema.metadata.name + '-' + schema.spec.environment}` is needed across
+   multiple resources, users must copy-paste it into every resource that uses it.
+   Updating the expression requires changes in every location, creating
+   maintenance burden and risk of inconsistency.
+
+2. **No computed lists inside a resource.** If a user wants to define a list of
+   containers as a template and reference it inside a Pod spec, there is no way
+   to do so today. `forEach` creates multiple *resources*, but there is no
+   mechanism to produce a computed *value* (like an array) that can be embedded
+   within a single resource's template.
+
+Refs: 
+- https://github.com/kubernetes-sigs/kro/issues/1008
+- https://github.com/kubernetes-sigs/kro/issues/781
+
+## Proposal
+
+This proposal introduces **variable nodes** — a new resource node type in the RGD
+that computes values without creating Kubernetes resources. Variable nodes
+participate in the DAG alongside template and externalRef nodes, enabling
+reusable expressions and computed lists that other resources can reference.
+
+### Overview
+
+Extend the RGD `resources` field to support a new node type: **variable nodes**.
+Today, each entry in `resources` is either:
+
+- A **template** node — kro creates and manages a Kubernetes resource.
+- An **externalRef** node — kro reads an existing Kubernetes resource (read-only).
+
+This proposal adds:
+
+- A **variable** node — kro computes values without creating any Kubernetes resource.
+
+Variable nodes participate in the DAG like any other node. They can reference
+`schema.spec` fields and other resources, and downstream resources can reference
+them. They support `forEach`, `includeWhen`, and `readyWhen`.
+
+### Design details
+
+#### Variable node structure
+
+A variable node uses a `variable` field (mutually exclusive with `template` and
+`externalRef`). The `variable` field must be a map (key-value pairs). Values
+within the map can be constants, CEL expressions, arrays, or nested objects.
+
+Values within a variable node can be:
+
+- **Constants:** literal values the user wants to reuse throughout the RGD.
+- **Static variables:** CEL expressions referencing `schema.spec` — the type is
+  known at compile time (RGD reconciler) and the value is provided at runtime
+  (instance reconciler).
+- **Dynamic variables:** CEL expressions referencing other resources in the RGD,
+  creating dependencies in the DAG.
+
+```yaml
+- id: vars
+  variable:
+    prefix: constant-value
+    name: ${schema.spec.name}
+    podName: ${myPod.metadata.name}
+```
+
+#### Referencing variable nodes
+
+Variable nodes can be referenced by any resource or other variable nodes in the RGD,
+as long as it's not cyclic. References work in two ways:
+
+- **By path:** `${vars.prefix}`, `${vars.name}` — accesses a specific field
+  within the variable node's map.
+- **By id:** `${vars}` — receives the entire map structure.
+
+Standard kro type validation applies at compile time: the type of the field
+where the variable is used must match the type of the referenced value.
+
+A variable node's top-level keys must not match its own id. Since variable node fields
+are referenced as `${<id>.<key>}`, a key matching the id would create confusion
+about whether `${naming.naming}` refers to the node itself or a nested field.
+For example, the following is invalid:
+
+```yaml
+- id: naming
+  variable:
+    naming: ${schema.spec.name}  # error: top-level key "naming" matches node id
+    prefix: ${schema.spec.name + '-' + schema.spec.environment}
+```
+
+#### forEach with variable nodes
+
+Variable nodes support `forEach`, producing an array of computed values rather than
+multiple Kubernetes resources. This addresses problem #2 — computed lists that
+can be embedded within a single resource template.
+
+**Example: shared container list across Deployment and StatefulSet**
+
+```yaml
+apiVersion: kro.run/v1alpha1
+kind: ResourceGraphDefinition
+metadata:
+  name: shared-pod-template
+spec:
+  schema:
+    apiVersion: v1alpha1
+    kind: MyApp
+    spec:
+      appImage: string
+      workers: "[]string"
+  resources:
+    # Variable node: compute a list of containers using forEach
+    - id: containers
+      forEach:
+        - worker: ${schema.spec.workers}
+      variable:
+        name: ${worker}
+        image: ${schema.spec.appImage}
+        command: ["process", "${worker}"]
+
+    # Both Deployment and StatefulSet reference the same container list
+    - id: deployment
+      template:
+        apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: ${schema.metadata.name + '-deploy'}
+        spec:
+          template:
+            spec:
+              containers: ${containers}
+
+    - id: statefulset
+      template:
+        apiVersion: apps/v1
+        kind: StatefulSet
+        metadata:
+          name: ${schema.metadata.name + '-sts'}
+        spec:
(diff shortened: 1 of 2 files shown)


docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+  - Circular dependency detection involving data nodes
+- **E2E tests:**
+  - Full lifecycle: RGD with data nodes -> instance creation -> resource
+    creation with correct computed values -> update -> deletion
+
+## Discussion and notes
+
+#### Questions for reviewers
+
+1. **Data-to-data references:** Should one data node be able to reference
```
NickKeller: Hi! I'm new here. 

My thought on this is that disallowing this would not really accomplish the goal of resuable expressions. Using your `naming.prefix` example, it would require the user to put `prefix-` on each of the full labels, which is still management burden. 

I think the bar would be "If I was looking at this in a PR, would I leave a comment to group multiple instances of the same value into a constant or shared variable"

NickKeller: at the very least I think data nodes should be allowed to reference data nodes with constant or static values. I'm not sure about dynamic values

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+structural flexibility that template nodes have for defining Kubernetes
+manifests.
+
+Values within a data node can be:
+
+- **Constants:** literal values the user wants to reuse throughout the RGD.
+- **Static variables:** CEL expressions referencing `schema.spec` — the type is
+  known at compile time (RGD reconciler) and the value is provided at runtime
+  (instance reconciler).
+- **Dynamic variables:** CEL expressions referencing other resources in the RGD,
```
NickKeller, line 61: I think by allowing dynamic variables in a data node, you're leaving the possibility of an inifinte reconciliation loop?

I'm imagining a data node `mydatanode` that references `deployment.status`. Then the deployment references `mydatanode` in some way. Or would that be caught by either existing or yet-to-be built cycle detection?

(New here so I don't have a lot of context on what already exists :) )

docs/design/proposals/variables.md
michaelhtm: Hey! Welcome
My thinking was, if someone can define everything they need in one `Variable` node, would there be a need to create a second node that references the first?

There would be no actual issue with implementing it, the main concern was its necessity.

docs/design/proposals/variables.md
michaelhtm: There is a system that detects cyclic graphs at compile time (RGD reconciler)

docs/design/proposals/variables.md
NickKeller: Thanks for the clarification. I think whether all variables are in one big mega node, or split up over multiple data nodes can be left up to the user's desires

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+- A **template** node — kro creates and manages a Kubernetes resource.
+- An **externalRef** node — kro reads an existing Kubernetes resource (read-only).
+
+This proposal adds:
+
+- A **data** node — kro computes values without creating any Kubernetes resource.
+
+Data nodes participate in the DAG like any other node. They can reference
+`schema.spec` fields and other resources, and downstream resources can reference
+them. They support `forEach` and `includeWhen`. `readyWhen` is not applicable
```
michaelhtm: After thinking more about includeWhen on data nodes there may be some concerns to discuss.
Currently, when includeWhen is false for a template node, all its dependents are also excluded. Should data nodes follow the same cascading behavior?

For template nodes this makes sense: if you skip the VPC, you skip the subnets. But for data nodes the semantics are less obvious. If a naming data node is excluded, should every resource referencing ${naming.prefix} also be excluded? That seems like surprising behavior for what's essentially a variable holder.

Should we disallow includeWhen on data nodes entirely (like readyWhen), or is there a use case where conditionally computing a data node is meaningful?

jakobmoellerdev: Why would the semantics for a data node need to be different behavior than for a resource node? For me - maybe contrary to what you point out - it would be much less intuitive to have differing behavior for both types of nodes. kro loves to do things uniformly, and I think for includeWhen this would be the same. 

That said, I agree - arguably there are not many use cases for includeWhen on variable/data nodes at all. But we are building primitives in kro, so I don't see a reason why it shouldn't be possible

a-hilaly: I believe we should allow all of them: includeWhen/readyWhen/forEach

cheeseandcereal: > readyWhen is not applicable

Although I agree the general use-case probably doesn't need this, is there a specific reason to exclude it? I could see a scenario where maybe some data node is looking at the status of some other resource, and is only ready when, for example, deployment a has at least 1 replica and deployment b also has at least 1 replica, or something like that.

Maybe a bit strange, but is there a reason we would want to explicitly disallow any of these even if they don't align closely with the generic use-case?

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+
+##### Referencing data nodes
+
+Other resources reference data node values in two ways:
+
+- **By path:** `${vars.prefix}`, `${vars.name}` — accesses a specific field
+  within a map-structured data node.
+- **By id:** `${vars}` — receives the entire data structure. If the data node
+  is a scalar, this gives the scalar value. If it's a map, this gives the whole
+  map.
```
a-hilaly: 2 thoughts:
1/ i think it confusing to call them `data` and reference them as `vars`
2/ why not just reference them by their name? as in their id

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+  compiler, which infers the return type from the expression and the known types
+  of referenced nodes.
+- **Identify constant fields** and infer their types from the parsed YAML values
+  (`5` is an integer, `"hello"` is a string, `true` is a boolean, `[1, 2, 3]`
+  is an integer array, nested maps preserve their structure).
+
+This reuse means data node type inference requires no new parsing infrastructure
+— the same code path that handles instance status fields applies directly.
+
+## Other solutions considered
```
NicholasBlaskey, line 258: My initial instinct for this would be to create some kind of new field called variables that is a map under each resource

Did we consider this?

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+them. They support `forEach` and `includeWhen`. `readyWhen` is not applicable
+to data nodes since they don't create Kubernetes resources — a data node is
+considered ready as soon as its expressions evaluate successfully. Specifying
+`readyWhen` on a data node is a validation error.
+
+#### Design details
+
+##### Data node structure
+
+A data node uses a `data` field (mutually exclusive with `template` and
```
NicholasBlaskey: I think my one concern with the design is introducing a new concept / node type is adding to the surface area for users to learn about `Kro`, its possible this is the right move but this does feel like there is a decent cost to complexity and interactions with other directions and so on. 

It feels like trying to treat a Kubernetes resource and a abstract store of data as the same thing can run into a bunch of oddities like validation errors for `readyWhen`

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+      template:
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: ${variableResource.spec.complexStringVar}-pod
+```
+
+Issues with this approach:
+
+- **RGD proliferation:** Users need a new RGD for every distinct variable
```
a-hilaly, line 339: https://kro.run/docs/advanced/rgd-chaining

docs/design/proposals/variables.md
michaelhtm: No specific reason. I mainly couldn't find a use case for it. 
In the deployment's case, I would expect deployment to have the readyWhen condition.
I also thought it would be simpler to not allow the use in the beginning, and later decide to support it if there's a use case for it

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+- A **data** node — kro computes values without creating any Kubernetes resource.
+
+Data nodes participate in the DAG like any other node. They can reference
+`schema.spec` fields and other resources, and downstream resources can reference
+them. They support `forEach` and `includeWhen`. `readyWhen` is not applicable
+to data nodes since they don't create Kubernetes resources — a data node is
+considered ready as soon as its expressions evaluate successfully. Specifying
+`readyWhen` on a data node is a validation error.
+
+#### Design details
```
a-hilaly: 1/ should a data node have a declared type it's validated against?

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+
+Extend the RGD `resources` field to support a new node type: **data nodes**.
+Today, each entry in `resources` is either:
+
+- A **template** node — kro creates and manages a Kubernetes resource.
+- An **externalRef** node — kro reads an existing Kubernetes resource (read-only).
+
+This proposal adds:
+
+- A **data** node — kro computes values without creating any Kubernetes resource.
```
a-hilaly: what are the naming options we have here? `variable`, `declaration` ?

docs/design/proposals/variables.md
```diff
@@ -0,0 +1,418 @@
...
+
+This proposal introduces **data nodes** — a new resource node type in the RGD
+that computes values without creating Kubernetes resources. Data nodes
+participate in the DAG alongside template and externalRef nodes, enabling
+reusable expressions and computed lists that other resources can reference.
+
+#### Overview
+
+Extend the RGD `resources` field to support a new node type: **data nodes**.
+Today, each entry in `resources` is either:
```
a-hilaly, line 34: what about the option of putting data nodes under `spec.variables`?

docs/design/proposals/variables.md
a-hilaly: > In the deployment's case, I would expect deployment to have the readyWhen condition.

Maybe a data node aggregate many resource status fields and sets a readyWhen that waits on all of them

HansG89:
@michaelhtm what is the status on this? We have a use case that depends on KREP-011 (see #1322). Happy to help move this forward if needed.

Skarlso:
@michaelhtm 
Hello.

This is `def` right? Or am I mistaken.

This will be possible with the Graph PR here: https://github.com/kubernetes-sigs/kro/pull/1355

Something like:
```yaml
apiVersion: kro.run/v1alpha1
kind: Graph
metadata:
  name: hello-naming
  namespace: gtest
spec:
  nodes:
    - id: naming
      def:
        prefix: ${config.data.env}-app
```

Which allows the usage of `${naming.prefix}`. If so, can this and the PR be closed?

Skarlso:
Maybe it's different because this is for RGDs.

jakobmoellerdev:
def nodes are our new intended way of declaring variables in Graphs. This proposal would need to encompass a similar idea just for RGDs.

PR #1036, opened 2026-02-11
Limit collection size to 1000 to protect cluster performance

NicholasBlaskey:
Limit collection size to 1000 to protect cluster performance.

Check this before we create the Cartesian product to avoid potential performance impacts of many dimensions.

One callout, this is hardcoded to 1000 initially, open to adding a configuration option / CLI flag to allow more control over this.

Tested with creating an RGD/instance over limit
kro log
```
2026-02-10T23:58:57Z    ERROR   dynamic-controller      Error syncing item, requeuing with rate limit   {"item": {"name":"my-instance-will-cel-functions","namespace":"default"}, "error": "node \"configmaps\": collection size of 1250 is over the maximum collection size of 1000"}
```

status
```
  Type:                  GraphResolved
    Last Transition Time:  2026-02-11T00:00:27Z
    Message:               resource reconciliation failed: node "configmaps": collection size of 1250 is over the maximum collection size of 1000
    Observed Generation:   1
    Status:                False
    Type:                  ResourcesReady
    Last Transition Time:  2026-02-11T00:00:27Z
    Message:               resource reconciliation failed: node "configmaps": collection size of 1250 is over the maximum collection size of 1000

```


pkg/runtime/collection.go
```diff
@@ -81,21 +81,26 @@ type evaluatedDimension struct {
...
-		return nil
+		return nil, nil
 	}
 
 	total := 1
 	for _, dim := range dimensions {
 		if len(dim.values) == 0 {
-			return nil
+			return nil, nil
 		}
```
a-hilaly, line 96: Should we check for `total > maxCollectionSize` here? it's practically absurd but this can overflow

NicholasBlaskey: good point, updated to check inside the loop now

pkg/runtime/collection.go
```diff
@@ -81,20 +81,25 @@ type evaluatedDimension struct {
 	values []any
 }
 
+const maxCollectionSize = 1000
```
jakobmoellerdev: to me this feels like an arbitrary number. Just because its named maxCollectionSize doesnt make it less magical for me. Can you explain your reasoning? if it is not arbitrary, the reasoning should be written here.

if it is arbitrary, I would argument to at least make the variable configurable and reduce it.

NicholasBlaskey: 1000 is arbitrary, updated the to a configurable parameter

currently defaults to 1000 still but I am happy to change 1000 to a lower default if you prefer

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -533,6 +533,22 @@ and increases cluster churn:
...
+Collections are limited to **1000 resources** per collection by default. If a `forEach`
+expression or cartesian product would create more than the limit resources, the
+reconciliation will fail with an error.
+
+This limit helps prevent accidental resource explosion and protects cluster
+performance. This limit is configurable through the CLI flag `--rgd-max-collection-size`
+or the Helm value `config.rgd.maxCollectionSize`.
+
+:::warning Deletion Limitation
+Currently, collections that exceed this limit will get stuck during deletion. If you
```
jakobmoellerdev, line 547: regarding this: I think this is not acceptable for merge. whats the reason collections will get stuck in deletion?

NicholasBlaskey: Currently if there are errors trying to generate the RGD collection it will fail to delete because calling deletion requires us to evaluate it. 

We can see this right now if you add a cel expression that will error to a `forEach` statement.
```
- id: secrets
      forEach:
        - badValue: ${[10 / 0]}
      template:
        apiVersion: v1
        kind: Secret
        metadata:
          name: ${schema.spec.name + '-' + string(badValue)}
        stringData:
          value: ${string(badValue)}
```

I think we should eventually address this in both cases. Not sure if this is completely blocking though.

NicholasBlaskey: thoughts @a-hilaly?

a-hilaly: This behaviour isn't introduced in this PR. If the graph doesn't compile or runtime fails with a non-data-pending error, the instance reconciler will fail as well.

a-hilaly: We can address this problem in a seperate PR. For now users need to make sure that their graph compile and that their collections are in bound before triggering the deletion.

jakobmoellerdev (commented):
/lgtm

PR #1037, opened 2026-02-11
Fix/stabilize crd deletion test

Aman-Cool:
Follow-up to #1029 to fix flakiness in the integration test and address an underlying controller bug.

## Problem
The integration test added in #1029 was flaky because of a race condition in how the RGD controller handles deletion events. When an RGD is deleted, Kubernetes sets `DeletionTimestamp` (a metadata-only change that doesn't increment `generation`). The controller's `GenerationChangedPredicate` was filtering out these events, so finalizers wouldn't run and CRDs wouldn't get cleaned up.

### Controller Fix (`pkg/controller/resourcegraphdefinition/controller.go`)
Modified the event filter to use `predicate.Or()` combining:
- `GenerationChangedPredicate` (for normal spec changes)
- Custom predicate that allows Update events where `DeletionTimestamp` is set (for deletions)

This ensures the reconciler sees deletion events and runs finalizers, similar to the CRD watch fix in #1029.

### Test Stabilization (`test/integration/suites/core/crd_test.go`)
- Wait for RGD Active state and finalizer before deletion
- Verify both RGD and CRD reach `NotFound` after deletion
- Removed any workarounds - tests actual finalizer behavior

This fixes both the test flakiness and the production bug where manual RGD deletions wouldn't trigger proper CRD cleanup.


a-hilaly:
@Aman-Cool can you please rebase this PR?

a-hilaly:
@Aman-Cool looks like this is still a problem:
```
   [FAIL] CRD CRD Watch Reconciliation [It] should detect externally deleted CRD and recreate it 
```
Waiting doesn't actually change anything because the next Expect statement is Eventual as well. 
```go
Eventually(func(g Gomega, ctx SpecContext) {
  // ...
}, 30*time.Second, 2*time.Second).WithContext(ctx).Should(Succeed())
```

Aman-Cool:
@a-hilaly ,I refactored the test so each `Eventually` now waits on a concrete, observable state change (including CRD recreation via a new UID), and removed the redundant/nested waits.  
This should make the test deterministic now, happy to adjust further if I missed anything

test/integration/suites/core/crd_test.go
```diff
@@ -524,38 +524,43 @@ var _ = Describe("CRD", func() {
 			// Externally delete the managed CRD to simulate an out-of-band deletion.
 			Expect(env.Client.Delete(ctx, crd)).To(Succeed())
 
-			// The controller's CRD metadata watch should detect the deletion event and
-			// trigger a reconciliation of the owning RGD, which calls crdManager.Ensure()
-			// to recreate the CRD.
+			// Signal: CRD recreated with a new UID AND RGD back to Active.
+			// A different UID proves the controller detected the deletion (via its
+			// CRD watch) and called Ensure() to create a replacement. The RGD
+			// returning to Active confirms reconciliation completed successfully.
```
a-hilaly: not sure the last statement is correct. Does an RGD become Inactive if you delete a CRD?

test/integration/suites/core/crd_test.go
```diff
@@ -524,38 +524,43 @@ var _ = Describe("CRD", func() {
...
 			//
-			// Without the fix (DeleteFunc returning false in the CRD watch predicate),
-			// this would timeout because the delete event would be silently filtered
-			// out, leaving the CRD permanently deleted.
+			// Without the fix (DeleteFunc returning false), this times out because
+			// the delete event is filtered and the controller never re-reconciles.
 			recreatedCRD := &apiextensionsv1.CustomResourceDefinition{}
 			Eventually(func(g Gomega, ctx SpecContext) {
-				err := env.Client.Get(ctx, types.NamespacedName{Name: crdName}, recreatedCRD)
-				g.Expect(err).ToNot(HaveOccurred())
```
a-hilaly, line 537: unnecessary change.

test/integration/suites/core/crd_test.go
```diff
@@ -524,38 +524,43 @@ var _ = Describe("CRD", func() {
...
-			}, 30*time.Second, 2*time.Second).WithContext(ctx).Should(Succeed())
 
-			// Verify the RGD recovers to Active state after CRD recreation
-			Eventually(func(g Gomega, ctx SpecContext) {
-				err := env.Client.Get(ctx, types.NamespacedName{Name: rgd.Name}, rgd)
-				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(env.Client.Get(ctx, types.NamespacedName{Name: rgd.Name}, rgd)).To(Succeed())
 				g.Expect(rgd.Status.State).To(Equal(krov1alpha1.ResourceGraphDefinitionStateActive))
-			}, 10*time.Second, time.Second).WithContext(ctx).Should(Succeed())
+			}, 30*time.Second, time.Second).WithContext(ctx).Should(Succeed())
```
a-hilaly, line 566: i'm still not sure if this is really fixing the flaky test, it looks quiet the same test we had before.

Aman-Cool:
@a-hilaly ,Thanks for the detailed review, I’ve tightened this up.

- Removed the RGD `Active` `Eventually` (RGD does not transition to `Inactive` during CRD recreation).
- Reverted the unnecessary change to keep the diff focused.

The key fix is the cleanup gate: the test now waits for **both the CRD and RGD to be `NotFound`** before exiting. Previously, the RGD finalizer could still be running while the next integration test started, causing cross-test interference. Gating on full resource removal ensures the controller and finalizer have completely finished, eliminating that suite-level flake.

Happy to adjust further if needed.

a-hilaly:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kro/1037/presubmits-integration-tests/2021445473042173952 is another flaky test we have. unrelated to this one.

Aman-Cool:
@a-hilaly I took a look at the logs from the failed run. The actual failure is in `CRD CRD Creation - should delete CRD when ResourceGraphDefinition is deleted` at `crd_test.go:192`  it times out after 20s with `Expected an error, got nil`. There's also a `handler for gvr no longer exists, dropping item` log for `test-recovery` just before it, which suggests a race condition where the controller cleans up before the test assertion can observe the expected state. Should I go ahead and fix it?

a-hilaly:
>@a-hilaly I took a look at the logs from the failed run. The actual failure is in CRD CRD Creation - should delete CRD when ResourceGraphDefinition is deleted at crd_test.go:192 it times out after 20s with Expected an error, got nil. There's also a handler for gvr no longer exists, dropping item log for test-recovery just before it, which suggests a race condition where the controller cleans up before the test assertion can observe the expected state. Should I go ahead and fix it?

@Aman-Cool  If it's a simple fix, yes please.

a-hilaly:
@Aman-Cool can you fix the other flaky test in a seperate PR please?

Aman-Cool:
@a-hilaly Dug into the logs and it turns out it's not just flakiness, there's an actual bug here. The controller uses `GenerationChangedPredicate` as a global event filter. When an RGD is deleted, the API server sets `DeletionTimestamp` which is a metadata-only change that doesn't increment `Generation`, so the predicate filters the event out and the reconciler never sees the deletion. The CRD is never cleaned up, and the test times out waiting for something that can never happen.

Also the test wasn't waiting for the RGD to become Active before deleting, meaning the finalizer might not have been set yet.

Fix has 2 parts:
1. Wait for RGD Active + CRD exists before deleting, ensures initial reconciliation is complete and the finalizer is in place
2. Bump the spec after deletion, increments Generation so the `GenerationChangedPredicate` lets the event through, the reconciler fetches the latest object with `DeletionTimestamp`, enters the cleanup path, deletes the CRD, and removes the finalizer

Aman-Cool:
@a-hilaly Sure, I'll open a separate PR for that.

Aman-Cool:
@a-hilaly Done, opened a separate PR to fix the other flaky test.

Aman-Cool:
@jakobmoellerdev ,whenever you get a chance, would love to get your thoughts on this one.

a-hilaly:
@Aman-Cool looks like your PR is fixing two tests at once, you make it 2 PRs instead?

test/integration/suites/core/crd_test.go
```diff
@@ -174,15 +174,32 @@ var _ = Describe("CRD", func() {
...
 			}, 10*time.Second, time.Second).WithContext(ctx).Should(Succeed())
 
-			// Delete ResourceGraphDefinition
+			// Delete ResourceGraphDefinition and bump spec to trigger
+			// reconciliation. The controller's GenerationChangedPredicate
+			// filters the DeletionTimestamp metadata-only update (no
+			// generation change), so a spec update is needed to let the
+			// reconciler observe the pending deletion.
 			Expect(env.Client.Delete(ctx, rgd)).To(Succeed())
+			Eventually(func(g Gomega, ctx SpecContext) {
```
jakobmoellerdev, line 189: this does not make sense to me. why would we need to update the RGD right after deleting it? we should probably react to that delete in the reconcile now anyhow right? (see our rgd reconcile watches)

Aman-Cool:
@jakobmoellerdev,You're absolutely right, the spec bump was a workaround that pointed to a real controller bug. 

I've fixed the root cause now. The issue was that `GenerationChangedPredicate` was filtering out RGD deletion events (since `DeletionTimestamp` is a metadata-only change that doesn't increment generation). 

**Changes in this updated PR:**

1. **Controller fix** (`pkg/controller/resourcegraphdefinition/controller.go`):
   - Modified the event filter to use `predicate.Or()` combining `GenerationChangedPredicate` with a custom predicate that allows Update events where `DeletionTimestamp` is set
   - Now the reconciler properly sees deletion events, just like we did for CRD watches in #1029

2. **Test cleanup** (`test/integration/suites/core/crd_test.go`):
   - Removed the spec-bump workaround entirely
   - Wait for finalizer to be set before deleting (ensures cleanup will happen)
   - Verify both RGD and CRD reach `NotFound` (proper cleanup gate)

This fixes both the test flakiness and the production bug where RGD deletions wouldn't trigger finalizer cleanup. The approach mirrors what we did in #1029 for CRD deletion detection.

Aman-Cool:
@a-hilaly @jakobmoellerdev , The PR description has been changed accordingly.

test/integration/suites/core/crd_test.go
```diff
@@ -174,19 +174,36 @@ var _ = Describe("CRD", func() {
...
 				g.Expect(env.Client.Get(ctx, types.NamespacedName{Name: crdName},
 					&apiextensionsv1.CustomResourceDefinition{})).To(Succeed())
 			}, 10*time.Second, time.Second).WithContext(ctx).Should(Succeed())
 
-			// Delete ResourceGraphDefinition
+			// Wait for the finalizer to be set before deleting
+			Eventually(func(g Gomega, ctx SpecContext) {
+				current := &krov1alpha1.ResourceGraphDefinition{}
+				g.Expect(env.Client.Get(ctx, types.NamespacedName{Name: rgd.Name}, current)).To(Succeed())
+				g.Expect(current.Finalizers).To(ContainElement("kro.run/finalizer"))
```
a-hilaly, line 192: I do not understand why we're asserting finalizer existence in here

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -102,7 +102,14 @@ func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) e
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("ResourceGraphDefinition").
 		For(&v1alpha1.ResourceGraphDefinition{}).
-		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		WithEventFilter(predicate.Or(
+			predicate.GenerationChangedPredicate{},
+			predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					return !e.ObjectNew.GetDeletionTimestamp().IsZero()
```
a-hilaly, line 109: needs a nil guard? how about checking object old as well?

PR #1038, opened 2026-02-11
applyset: prevent stale deletes during prune using UID preconditions

WHOIM1205:
# Fix TOCTOU race in ApplySet prune to prevent deleting recreated resources

## Summary

This PR fixes a time-of-check to time-of-act (TOCTOU) race condition in the ApplySet prune logic that could lead to silent deletion of newly recreated resources in production Kubernetes clusters.

When a resource is listed for pruning and then recreated with a new UID before the DELETE request is issued, the current code can delete the replacement object. This change adds a UID precondition to prune DELETE calls so that only the originally listed object can be deleted.

---

## Problem Description

The ApplySet prune flow currently:

1. Lists resources that are candidates for pruning
2. Deletes each resource by name

Between the LIST and DELETE steps, Kubernetes allows a resource to be deleted and recreated with the same name but a different UID. Without a UID precondition, the DELETE request may remove the new object, resulting in silent data loss.

This race can occur during controller restarts, high object churn, or ApplySet-driven reconciliation, and is especially relevant in CNCF / LFX multi-tenant clusters.

---

## What Was Fixed

- Captured the UID of each prune candidate at LIST time
- Added `DeleteOptions.Preconditions{UID}` to all prune DELETE requests
- Ensured the API server rejects stale deletes when the resource UID changes
- Preserved existing behavior in the steady-state path

---

## Code Changes

**Section modified:**

**Change summary:**
- Added UID preconditions to prune DELETE calls for both namespaced and cluster-scoped resources

This is a minimal, targeted correctness fix with no refactors or API changes.

---

## Test Verification

**New test added:**

**Test name:** `TestPrune_UIDPreconditionPreventsStaleDelete`
- test case is passed locally 
image

**Verification steps:**
```bash
go test ./pkg/controller/instance/applyset/... \
  -v \
  -run TestPrune_UIDPreconditionPreventsStaleDelete \
  -count=1
go test ./pkg/controller/instance/applyset/... -v

### CODE REVIEW

**Comment by WHOIM1205** [comment]
Hey @a-hilaly 
This PR fixes a small TOCTOU race in ApplySet prune where a resource recreated between LIST and DELETE could be deleted by mistake. Added a UID precondition to make deletes safe. Includes a focused test. Thanks for the review!

File: pkg/controller/instance/applyset/applyset.go [hunk 1]
```diff
@@ -541,11 +541,17 @@ func (a *ApplySet) prune(
...
 			var err error
 			if c.obj.GetNamespace() != "" {
-				err = a.client.Resource(c.gvr).Namespace(c.obj.GetNamespace()).Delete(egCtx, c.obj.GetName(), metav1.DeleteOptions{})
+				err = a.client.Resource(c.gvr).Namespace(c.obj.GetNamespace()).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			} else {
-				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), metav1.DeleteOptions{})
+				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			}
 
 			if err != nil && !apierrors.IsNotFound(err) {
```
a-hilaly: I believe we also need to tolerate 409 conflict errors?

I believe that's what .Delete returns when there a UID mismatch

pkg/controller/instance/applyset/applyset_test.go
```diff
@@ -590,6 +592,147 @@ func TestPrune(t *testing.T) {
...
+	}
+
+	_, err = applier2.Prune(ctx, PruneOptions{
+		KeepUIDs: result2.ObservedUIDs(),
+		Scope:    projectMeta2.PruneScope(),
+	})
+	if err == nil {
+		t.Fatal("Prune() expected error from UID precondition mismatch, got nil")
+	}
+}
```
a-hilaly: I think what we should assert here is the Delete is safely skipped (original object is gone)

PR #1039, opened 2026-02-11
fix: reject reserved status field names at build time

WHOIM1205:
This PR adds validation to reject reserved top-level status field names
`state` and `conditions` in ResourceGraphDefinition.

These fields are managed internally by the instance controller for lifecycle
tracking. Previously, users could define them in the status schema, but their
values were silently overwritten at runtime.

The graph builder now fails early if these reserved names are used.

Changes:
- Added validation in buildStatusSchema
- Simplified check using strings.HasPrefix
- Added unit tests covering both reserved names
- Updated existing tests to avoid reserved fields

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -727,6 +727,20 @@ func buildStatusSchema(
 		return nil, nil, nil, fmt.Errorf("status fields without expressions are not supported: %v", noExpressionFields)
 	}
 
+	// Reject reserved status field names. The instance controller uses "state"
+	// and "conditions" for lifecycle tracking and will silently overwrite any
+	// user-defined values for these fields.
+	reservedStatusFields := map[string]bool{"state": true, "conditions": true}
+	for _, fd := range fieldDescriptors {
+		topLevel := strings.SplitN(fd.Path, ".", 2)[0]
+		if reservedStatusFields[topLevel] {
+			return nil, nil, nil, fmt.Errorf(
+				"status field %q uses reserved name %q: kro reserves \"state\" and \"conditions\" for instance lifecycle tracking",
+				fd.Path, topLevel,
+			)
+		}
+	}
+
 	// Instance status expressions can ONLY reference resources, not schema.
 	// At runtime, status is populated after resources are created.
 
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -727,16 +727,9 @@ func buildStatusSchema(
 		return nil, nil, nil, fmt.Errorf("status fields without expressions are not supported: %v", noExpressionFields)
 	}
 
-	// Reject reserved status field names. The instance controller uses "state"
-	// and "conditions" for lifecycle tracking and will silently overwrite any
-	// user-defined values for these fields.
-	for _, fd := range fieldDescriptors {
-		if strings.HasPrefix(fd.Path, "state") || strings.HasPrefix(fd.Path, "conditions") {
-			return nil, nil, nil, fmt.Errorf(
-				"status field at path %q uses a reserved status field managed by kro",
-				fd.Path,
-			)
-		}
+	// Reject reserved status field names.
+	if err := validateReservedStatusFields(fieldDescriptors); err != nil {
+		return nil, nil, nil, err
 	}
 
 	// Instance status expressions can ONLY reference resources, not schema.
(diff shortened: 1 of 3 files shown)


pkg/graph/builder.go
```diff
@@ -727,6 +727,20 @@ func buildStatusSchema(
 		return nil, nil, nil, fmt.Errorf("status fields without expressions are not supported: %v", noExpressionFields)
 	}
 
+	// Reject reserved status field names. The instance controller uses "state"
+	// and "conditions" for lifecycle tracking and will silently overwrite any
+	// user-defined values for these fields.
+	reservedStatusFields := map[string]bool{"state": true, "conditions": true}
```
a-hilaly: Can we put these two paths near their friends https://github.com/kubernetes-sigs/kro/blob/main/pkg/graph/node.go#L38-L44

pkg/graph/builder.go
```diff
@@ -727,6 +727,20 @@ func buildStatusSchema(
...
+	for _, fd := range fieldDescriptors {
+		topLevel := strings.SplitN(fd.Path, ".", 2)[0]
```
a-hilaly: I think you only need to test with `strings.HasPrefix`

pkg/graph/builder.go
```diff
@@ -727,6 +727,20 @@ func buildStatusSchema(
...
+		if reservedStatusFields[topLevel] {
+			return nil, nil, nil, fmt.Errorf(
+				"status field %q uses reserved name %q: kro reserves \"state\" and \"conditions\" for instance lifecycle tracking",
```
a-hilaly:
```suggestion
				"status field at path %q uses a reserved status field that is managed by kro: %q",
```

a-hilaly (commented):
Thank you @WHOIM1205 !
left a few comments in line
/ok-to-test

a-hilaly:
@WHOIM1205 can you please provide a concise and short PR description of the changes? no need to paste all the AI generated output there.

WHOIM1205:
Thanks for the review @a-hilaly 

I’ve addressed the feedback:

moved the validation closer to the related path checks for better grouping.
simplified the logic to use strings.HasPrefix instead of splitting and using a map.
updated the error message to be shorter and clearer, indicating the field is managed by KRO.
also shortened the PR description as suggested.

Please take another look when you get a chance. Happy to make further adjustments if needed.

pkg/graph/builder.go
a-hilaly: You haven't address this one @WHOIM1205

a-hilaly:
@WHOIM1205 first comment haven't been addressed yet

WHOIM1205:
> @WHOIM1205 first comment haven't been addressed yet

Thanks for pointing that out.

I’ve reorganized the validation so it sits alongside the existing path-related checks in node.go for consistency. I also simplified the logic and kept the behavior unchanged — it still rejects top-level `state` and `conditions` fields with the same error.
All graph tests are passing locally. Please let me know if you’d prefer this kept inline instead of using the small helper.

WHOIM1205:
@a-hilaly , @jakobmoellerdev is there anything i can change in this pr

pkg/graph/node.go
```diff
@@ -41,6 +41,12 @@ const (
 	MetadataNamePath = "metadata.name"
 	// MetadataNamespacePath is the path to the resource namespace field.
 	MetadataNamespacePath = "metadata.namespace"
+	// ReservedStatusFieldState is the status field reserved by the instance
+	// controller for lifecycle state tracking.
+	ReservedStatusFieldState = "state"
+	// ReservedStatusFieldConditions is the status field reserved by the
+	// instance controller for condition reporting.
+	ReservedStatusFieldConditions = "conditions"
```
michaelhtm, line 49: should these be `state.` and `conditions.`?
If not, cases where someone defines something like `conditionsz` or `stateValue` would be marked as invalid

michaelhtm (commented):
Thanks @WHOIM1205 
left one question below

pkg/graph/node.go
WHOIM1205: > should these be `state.` and `conditions.`? If not, cases where someone defines something like `conditionsz` or `stateValue` would be marked as invalid

Good catch — thanks for pointing that out.

I’ve updated the validation to only reject the exact fields `state` and `conditions`, as well as their subpaths (e.g. `state.*`, `conditions.*`). Fields like `stateValue` or `conditionsz` are no longer affected.

The change is minimal and all graph tests are passing locally.

jakobmoellerdev (commented):
/lgtm

a-hilaly:
I'm still thinking about this, i'm wondering if there are cases where we want to allow this actually. Putting a soft hold on this.

/hold

jakobmoellerdev:

/close

PR #1040, opened 2026-02-11
Fix/applyset prune uid precondition

WHOIM1205:
# applyset: prevent stale deletes during prune using UID preconditions

## Summary

This PR fixes a time-of-check to time-of-act (TOCTOU) race in the ApplySet prune logic.

Between the LIST and DELETE steps, a resource can be deleted and recreated with the same name but a different UID. Without a UID precondition on DELETE, the newly recreated object could be incorrectly deleted.

This change adds a UID precondition to prune DELETE calls and correctly handles 409 Conflict responses.

---

## Problem

The prune flow:

1. LIST candidate resources.
2. DELETE each resource by name.

If a resource is recreated between these steps, the DELETE may remove the replacement object because it only matches by name.

When a UID precondition is used, the API server returns a 409 Conflict if the UID no longer matches. This is expected and safe behavior, but it must not cause prune to fail.

---

### applyset.go

In the `prune` method DELETE section:

- Capture UID from the listed object.
- Pass it via `DeleteOptions.Preconditions.UID`.
- Treat both `IsNotFound` and `IsConflict` as safe-to-ignore.
- Only append to `Pruned` results when `err == nil` (actual delete occurred).

This ensures:
- Stale deletes are prevented.
- Recreated resources are not deleted.
- Prune does not fail on expected 409 Conflict.

---

### applyset_test.go

Added `TestPrune_UIDPrecondition` with two subtests:

| Subtest | Scenario | Expected Result |
|----------|----------|----------------|
| `uid matches - resource pruned` | Normal prune | 1 pruned, no error |
| `uid mismatch - resource not pruned` | DELETE returns 409 Conflict | 0 pruned, no error, resource still exists |

The test verifies that:
- UID match behaves normally.
- UID mismatch safely skips delete.
- No error is returned on 409.
- The recreated resource remains intact.

---

## Verification
Screenshot 2026-02-11 090315

Run:

```bash
go test ./pkg/controller/instance/applyset/... -v -count=1

### DIFF
--- a/pkg/controller/instance/applyset/applyset.go
+++ b/pkg/controller/instance/applyset/applyset.go
@@ -541,26 +541,36 @@ func (a *ApplySet) prune(
 
 	for _, c := range candidates {
 		eg.Go(func() error {
+			uid := c.obj.GetUID()
+			deleteOpts := metav1.DeleteOptions{
+				Preconditions: &metav1.Preconditions{UID: &uid},
+			}
 			var err error
 			if c.obj.GetNamespace() != "" {
-				err = a.client.Resource(c.gvr).Namespace(c.obj.GetNamespace()).Delete(egCtx, c.obj.GetName(), metav1.DeleteOptions{})
+				err = a.client.Resource(c.gvr).Namespace(c.obj.GetNamespace()).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			} else {
-				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), metav1.DeleteOptions{})
+				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			}
 
-			if err != nil && !apierrors.IsNotFound(err) {
+			switch {
+			case err != nil && !apierrors.IsNotFound(err) && !apierrors.IsConflict(err):
 				return fmt.Errorf("delete %s/%s: %w", c.obj.GetNamespace(), c.obj.GetName(), err)
+			case apierrors.IsConflict(err):
+				a.log.V(2).Info("skipped prune due to UID mismatch (resource recreated)",
+					"name", c.obj.GetName(),
+					"namespace", c.obj.GetNamespace(),
+					"gvr", c.gvr.String(),
+				)
+			case err == nil:
+				mu.Lock()
+				results = append(results, PruneResultItem{Object: c.obj})
+				mu.Unlock()
+				a.log.V(2).Info("pruned resource",
+					"name", c.obj.GetName(),
+					"namespace", c.obj.GetNamespace(),
+					"gvr", c.gvr.String(),
+				)
 			}
-
-			mu.Lock()
-			results = append(results, PruneResultItem{Object: c.obj})
-			mu.Unlock()
-
-			a.log.V(2).Info("pruned resource",
-				"name", c.obj.GetName(),
-				"namespace", c.obj.GetNamespace(),
-				"gvr", c.gvr.String(),
-			)
 			return nil
 		})
 	}
--- a/pkg/controller/instance/applyset/applyset_test.go
+++ b/pkg/controller/instance/applyset/applyset_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by WHOIM1205** [comment]
/assign @a-hilaly 
Closed the previous PR after cleanup. This PR contains the corrected fix per review (409 handling + updated test).

File: pkg/controller/instance/applyset/applyset.go [hunk 1]
```diff
@@ -541,26 +541,32 @@ func (a *ApplySet) prune(
...
+				err = a.client.Resource(c.gvr).Namespace(c.obj.GetNamespace()).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			} else {
-				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), metav1.DeleteOptions{})
+				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			}
 
-			if err != nil && !apierrors.IsNotFound(err) {
+			if err != nil && !apierrors.IsNotFound(err) && !apierrors.IsConflict(err) {
 				return fmt.Errorf("delete %s/%s: %w", c.obj.GetNamespace(), c.obj.GetName(), err)
 			}
```
a-hilaly: Can we add a debug log in here?
```
if apierrors.IsConflict(err) {
    a.log.V(2).Info("skipped prune: UID mismatch (resource recreated)", ...)
}
```

pkg/controller/instance/applyset/applyset_test.go
```diff
@@ -590,6 +591,140 @@ func TestPrune(t *testing.T) {
...
+		orphan.SetUID(types.UID("original-uid"))
+
+		// Simulate: between LIST and DELETE, the resource was recreated with a new UID.
+		// The fake client has the recreated object (new UID), so when DELETE comes in
+		// with precondition for the old UID, it should fail with 409 Conflict.
+		recreated := newConfigMap("orphan-cm", "default")
+		recreated.SetLabels(map[string]string{
+			ApplysetPartOfLabel: applySetID,
+		})
+		recreated.SetUID(types.UID("new-uid"))
```
a-hilaly, line 661: looks like this resource isn't used

WHOIM1205:
@a-hilaly 
addressed the review comments:
- Added a V(2) debug log for 409 Conflict (UID mismatch) cases.
- Fixed the test to properly use the recreated object when simulating UID mismatch.

Tests passing locally.

pkg/controller/instance/applyset/applyset.go
```diff
@@ -541,26 +541,39 @@ func (a *ApplySet) prune(
...
+					"gvr", c.gvr.String(),
+				)
+			}
+			if apierrors.IsConflict(err) {
+				a.log.V(2).Info("skipped prune due to UID mismatch (resource recreated)",
+					"name", c.obj.GetName(),
+					"namespace", c.obj.GetNamespace(),
+					"gvr", c.gvr.String(),
+				)
+			}
```
a-hilaly: nit:
```suggestion
			if apierrors.IsConflict(err) {
				a.log.V(2).Info("skipped prune due to UID mismatch (resource recreated)",
					"name", c.obj.GetName(),
					"namespace", c.obj.GetNamespace(),
					"gvr", c.gvr.String(),
				)
			}
			if err == nil {
				mu.Lock()
				results = append(results, PruneResultItem{Object: c.obj})
				mu.Unlock()

				a.log.V(2).Info("pruned resource",
					"name", c.obj.GetName(),
					"namespace", c.obj.GetNamespace(),
					"gvr", c.gvr.String(),
				)
			}
```

a-hilaly:
@WHOIM1205 you haven't addressed all the provided feedback

WHOIM1205:
> @WHOIM1205 you haven't addressed all the provided feedback

Refactored the post-delete handling into a switch so the conflict, success, and error paths are clearly mutually exclusive. No behavior change just cleaner structure per feedback. All tests passing.

WHOIM1205:
@a-hilaly , @jakobmoellerdev is there anything i can change in this pr

pkg/controller/instance/applyset/applyset.go
```diff
@@ -541,26 +541,36 @@ func (a *ApplySet) prune(
...
 			if c.obj.GetNamespace() != "" {
-				err = a.client.Resource(c.gvr).Namespace(c.obj.GetNamespace()).Delete(egCtx, c.obj.GetName(), metav1.DeleteOptions{})
+				err = a.client.Resource(c.gvr).Namespace(c.obj.GetNamespace()).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			} else {
-				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), metav1.DeleteOptions{})
+				err = a.client.Resource(c.gvr).Delete(egCtx, c.obj.GetName(), deleteOpts)
 			}
 
-			if err != nil && !apierrors.IsNotFound(err) {
+			switch {
```
jakobmoellerdev, line 555: this switch case is a bit hard to understand. My understanding is that if the err != nil and it is not found or conflict, we are no longer appending the item to prune result. is that correct?

jakobmoellerdev:
@WHOIM1205 friendly reminder

a-hilaly:
friendly reminder @WHOIM1205

a-hilaly:
Opened https://github.com/kubernetes-sigs/kro/pull/1081

PR #1041, opened 2026-02-11
Fix flaky validate RGD status integration test

Aman-Cool:
Two tests in validation_test.go were using the same RGD name `test-k8s-valid` and schema kind `TestK8sValidation`, causing a conflict when they ran in parallel or randomized order — the second test would fail trying to create an RGD that was still terminating from the first.

Renamed the second test's RGD to `test-k8s-invalid-status` and schema kind to `TestK8sInvalidStatus` to fix the collision.

### DIFF
--- a/test/integration/suites/core/validation_test.go
+++ b/test/integration/suites/core/validation_test.go
@@ -218,7 +218,7 @@ var _ = Describe("Validation", func() {
 		It("should reject RGDs with plain fileds (no expression)", func(ctx SpecContext) {
 			rgd := generator.NewResourceGraphDefinition("test-k8s-invalid-status",
 				generator.WithSchema(
-					"TestK8sValidation", "v1alpha1",
+					"TestK8sInvalidStatus", "v1alpha1",
 					map[string]interface{}{
 						"name": "string",
 					},


Aman-Cool:
@a-hilaly opened a separate PR for the other flaky test. Two tests in `validation_test.go` were sharing the same RGD name and schema kind, causing conflicts when running in parallel or randomized order. Fixed by renaming them to avoid the collision.

a-hilaly:
@Aman-Cool this has been already fixed in https://github.com/kubernetes-sigs/kro/pull/1034

Aman-Cool:
@a-hilaly Thanks for pointing that out! Looking at #1034, it only renames the RGD name while this PR also renames the schema kind `TestK8sValidation` to `TestK8sInvalidStatus`, which fully eliminates the collision. Happy to close this and let #1034 proceed if you prefer, just wanted to flag that this fix is a bit more complete.

jakobmoellerdev:
Youre correct here. I think this just needs a rebase with your CRD name change and then we should be good.

/ok-to-test

Aman-Cool:
@jakobmoellerdev , Done

PR #1045, opened 2026-02-12
chore: limit collection dimension to 5

michaelhtm:
limit collection dimension to 5


a-hilaly:
@michaelhtm can you also remove https://github.com/kubernetes-sigs/kro/blob/main/pkg/graph/validation.go#L167

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -201,6 +201,7 @@ type Resource struct {
 	// Example: [{"region": "${schema.spec.regions}"}, {"tier": "${schema.spec.tiers}"}]
 	//
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxItems=5
```
jakobmoellerdev: why explicitly 5? this feels very arbitrary imho. I also think it should be configurable and not be hardcoded into the CRD.

a-hilaly: @jakobmoellerdev mainly from https://github.com/kubernetes-sigs/kro/pull/936#discussion_r2702517432 :)

a-hilaly: I'm fine with a runtime-level validation as well

jakobmoellerdev: I would favor this for now. For me this is something most users wont see anyhow and I do actually have some (admittedly) cases here that multiply out with more than 5 dimensions.

a-hilaly: Should we make it a controller flag? default to 10?

jakobmoellerdev: sure that would work for me.

a-hilaly: I think same should apply for the collection size limit after expansion

michaelhtm: just to reiterate, 
This validation should no longer be at the CRD level, but somewhere here? https://github.com/kubernetes-sigs/kro/blob/main/pkg/graph/validation.go#L167

a-hilaly: Correct. We can catch that at RGD build time.

jakobmoellerdev:
@michaelhtm friendly reminder

a-hilaly:
cc @michaelhtm

a-hilaly (approved):

PR #1050, opened 2026-02-15
fix(docs): Wrong schema apiVersion definition in RGD chaining

gagahsyuja:
Hi!

This PR fixed the wrong use of apiVersion in `advanced/rgd-chaining` page. 

In schema, apiVersion should only mention the version and not the group, such:
`apiVersion: v1alpha1`

And not:
`apiVersion: kro.run/v1alpha1`


michaelhtm (approved):
great catch @gagahsyuja

Issue #482, opened 2025-04-04
Allow ResourceGraphDefinition authors maintain serveral versions


acll19:
### Feature Description

**Problem Statement**:

Users of kro are not likely to start with an stable version of their Resource Graph Definitions (RGD) which means that they could update their RGD spec several times and the way they have to inform their users that a change has been made (and what kind of change even) is by releasing a new version of the resource group.

Currently this is not possible with kro because the `apiVersion` of an RGD is immutable.

**Proposed Solution**:

As I see it, two things need to be considered 1) managing RGD versions in the controller and 2) user experience (UX), as in how to offer an API that's simple to use yet simple to maintain.

Let's start with the second one as I believe the UX will influence the implementation the most.  Given the many ways users could define and evolve their APIs, there is no simple way to solve this problem unless some conventions are in place:
      
- `spec.schema.apiVersion` will always be the latest version of the underlying CRD
- Users can keep old versions under `spec.previousSchemas[*]`
- A previous schema  MUST declare `apiVersion` which MUST be older than `spec.schema.apiVersion`
- A previous schema's `spec` will ONLY declare fields that have been deprecated, the field name has changed or even the field's structure have changed. An example of a field whose structure has changed could be in a previous schema `spec.image` and in the next schema: `spec.image.repository` and `spec.image.tag`. Converting from one to another would not be too hard (this one could be stretch).
- The same goes for a previous schema's status
- Version priority is the same as described [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority) for CRDs

This way, for people interacting with kro for the first time will not need to deal with boilerplate code which will only be needed later if ever. At the same time it is still a simple way for RGD authors who do need to maintain several versions.

Here's an example

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: web-application
spec:
  schema:
    apiVersion: v1beta1
    kind: WebApplication
    spec:
      name: string | required=true description="My Name"
      replicas: integer | default=1 minimum=1 maximum=100
      image: # new way of setting image details
        repository: string | required=true description="Path to the image repository"
        tag: string | required=true description="Image tag"

      ingress:
        enabled: boolean | default=false
        host: string | default="example.com"
        path: string | default="/"

      ports: "[]integer"

      env: "map[string]string"

    status:
      availableReplicas: ${deployment.status.availableReplicas}
[...]

PR #1051, opened 2026-02-15
feat: allow RGD authors to maintain multiple versions (#482)

shivansh-gohem:
## Description
This PR implements the **core versioning logic** for `ResourceGraphDefinition`, allowing users to update the `apiVersion` schema without breaking existing resources or being blocked by validation.

## Why this approach? (Addressing Complexity)
While explicitly listing `previousSchemas` (as originally proposed) offers visibility, it requires a complex schema migration engine. This PR enables versioning **now** using native Kubernetes CRD mechanics. This does not preclude adding `previousSchemas` in a future PR for better visibility, but unblocks the core feature today.

fixes(#482)

### DIFF
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -53,12 +53,12 @@ type Schema struct {
 	Kind string `json:"kind,omitempty"`
 	// APIVersion is the version identifier for the generated CRD.
 	// Must follow Kubernetes versioning conventions (v1, v1alpha1, v1beta1, etc.).
-	// This field is immutable after creation.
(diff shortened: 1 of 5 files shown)


shivansh-gohem:
ptal @jakobmoellerdev whenever you'll have time

shivansh-gohem:
Hey @jakobmoellerdev , I actually haven't synced with @barney-s yet! I just read through KREP-009 though, and I noticed you preferred a simpler in-place CRD upgrade path rather than the heavy ReplicaSet pattern. That's exactly what this PR does (it merges the versions in the background). I'll drop a note on his KREP so we can figure out how this fits into the bigger picture!

PR #1052, opened 2026-02-16
Membership conflicts 957

Avr8:
ApplySet performs ownership conflict detection using:

`if r.Current != nil {
    currentApplySetID = r.Current.GetLabels()[ApplysetPartOfLabel]
}
 `

ApplySet performs ownership conflict detection using:

```
if r.Current != nil {
    currentApplySetID = r.Current.GetLabels()[ApplysetPartOfLabel]
}
```
This assumes the controller always provides the live object in Resource.Current.

However, in real reconcile paths this is not always true. When Current is nil:

ApplySet performs no ownership validation
SSA proceeds
A second KRO instance (same RGD, different instance) can overwrite resources
already owned by another KRO instance
This leads to instance label flipping and silent resource takeover
Root Cause

ApplySet trusted Current to represent cluster state and skipped validation
when it was not provided.

Fix

When `r.Current == nil`, ApplySet now performs a defensive GET from the cluster
before SSA to check existing ownership labels.

If an existing object is found and:
`kro.run/owned=true`
and ownership labels (instance-id, resource-graph-definition-id, node-id)
do not match the desired object

ApplySet returns an ownership conflict error instead of applying.

Test Added

TestApply_DetectsOwnershipConflict

This test:

Creates an existing object owned by another KRO instance

Calls ApplySet with Current == nil

Verifies that ownership conflict is detected

This test fails without the fix and passes with it.


Avr8:
This is in continuation to: [membership-conflict](https://github.com/kubernetes-sigs/kro/pull/969)

I understand the issue requires triage, i have added the scenarios considered,test added in the description..i would love to discuss on how to proceed for the next steps.

pkg/controller/instance/applyset/applyset.go
```diff
@@ -354,9 +355,52 @@ func (a *ApplySet) applyResource(
 
 	// Conflict check using observed state (from controller GET), if provided.
 	var currentApplySetID string
+
+	// If controller did not provide Current, we must defensively check
+	// whether the resource already exists in the cluster to avoid
+	// silently overwriting resources owned by other KRO instances.
+	if r.Current == nil {
```
jakobmoellerdev, line 362: in which case would current be nil here?

a-hilaly:
Closing due to inactivity

Issue #1058, opened 2026-02-18
Kro controller metrics report errors when objects are deleted


tomasaschan:
### Description

**Observed Behavior**:
When deleting an instance of an RGD, the Kro controller will emit a few log lines like this
```
2026-02-18T16:20:07Z    INFO    applications-controller instance not found (likely deleted)     {"controller": "applications", "controllerGroup": "kro.run", "controllerKind": "Application", "namespace": "foo", "name": "my-app"}
2026-02-18T16:20:10Z    INFO    applications-controller instance not found (likely deleted)     {"controller": "applications", "controllerGroup": "kro.run", "controllerKind": "Application", "namespace": "foo", "name": "my-app"}
```
Those are in itself not an issue, but it also increments the `dynamic_controller_handler_errors_total` metric, which confuses observability tooling that something is wrong when it actually isn't, causing false alarms and, in the long run, alarm fatigue.

**Expected Behavior**:
Deleting RGD instances does not emit indications of failures unless something actually goes wrong.

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):
Run these steps, for example, in a local [kind](https://kind.sigs.k8s.io/) cluster:

1. [Install Kro](https://kro.run/docs/getting-started/Installation/), for example using the raw manifest installation method:
   ```sh
   kubectl apply -f https://github.com/kubernetes-sigs/kro/releases/download/v0.8.5/kro-core-install-manifests.yaml
   ```

2. Install the `Application` RGD from [the first example in the docs](https://kro.run/docs/concepts/rgd/overview#example). Also create the sample `Application` from the same page (perhaps in a separate namespace):
   ```sh
   kubectl apply -f rgd.yaml
   kubectl create namespace foo
   kubectl apply -f app.yaml -n foo
   ```

3. Observe the `Application` become healthy, and the pods starting up in the `foo` namespace. Then, delete the `Application` again:
   ```sh
   kubectl delete application -n foo my-app
   ```
   Observe that the `Application` disappears, as do the pods and deployment in the `foo` namespace.

4. Observe logs in the Kro controller emitting the "applications-controller instance not found (likely deleted)" message as stated above (e.g. with `kubectl logs -n kro-system kro-<tab>`)

5. Observe that the counter has incremented to a nonzero value, by opening a connection to the controller pod:
   ```sh
   kubectl port-forward -n kro-system kro-<tab> 8078
   ```
[...]

PR #1059, opened 2026-02-18
Fix: Kro controller metrics report errors when objects are deleted 

Avr8:
# Fix: Kro controller metrics report errors when objects are deleted #1058

## Problem

When deleting an RGD instance (e.g. `kubectl delete application -n foo my-app`), the Kro controller correctly processes the deletion but was incrementing the `dynamic_controller_handler_errors_total` metric. This caused:

- **False alarms** in observability/alerting (e.g. Prometheus alerts firing on normal deletes)
- **Alarm fatigue** over time

### Root cause

During deletion and during normal reconciliation (e.g. “resource deletion in progress”), the instance controller returns **requeue control-flow errors** (`requeue.RequeueNeeded` or `requeue.RequeueNeededAfter`) so the item is processed again. These are expected and not failures.

In `syncFunc`, we were incrementing `handlerErrorsTotal` for **any** non-`NotFound` error. Requeue types are errors from the type system’s perspective but represent “please requeue,” not “something went wrong.” So they were incorrectly counted as handler errors.

## Solution

1. **`isRequeueSignal(err)`**  
   New helper that returns `true` when `err` is `*requeue.RequeueNeeded` or `*requeue.RequeueNeededAfter`. No new dependencies; `requeue` is already used in this package.

2. **Metric increment in `syncFunc`**  
   We only increment `handlerErrorsTotal` when the handler returns an error that is:
   - not `nil`,
   - not `apierrors.IsNotFound` (already excluded),
   - and **not** a requeue signal (`!isRequeueSignal(err)`).

So:

- **Instance deleted** → handler returns `nil` → metric not incremented.
- **Requeue (e.g. deletion in progress)** → handler returns requeue error → metric not incremented.
- **Real failures** (e.g. apply errors, API errors) → metric still incremented.

## Implementation details

- **File:** `pkg/dynamiccontroller/dynamic_controller.go`
- **New function:** `isRequeueSignal(err error) bool`  
  Uses a type switch on `*requeue.RequeueNeeded` and `*requeue.RequeueNeededAfter` (no `errors.As` needed; we only care about these two concrete types).
- **Change in `syncFunc`:**  
  Condition updated from  
  `if err != nil && !apierrors.IsNotFound(err)`  
  to  
  `if err != nil && !apierrors.IsNotFound(err) && !isRequeueSignal(err)`  
  with a short comment explaining that instance-deleted and requeue outcomes are not counted as errors.

## Testing

- **With fix:** Delete an Application instance → `dynamic_controller_handler_errors_total` does not increase (verified via `curl :8078/metrics`).
- **Without fix (reverted):** Same delete → counter increments (e.g. 2 → 3), confirming the bug and that the fix addresses it.

### DIFF
--- a/pkg/dynamiccontroller/dynamic_controller.go
+++ b/pkg/dynamiccontroller/dynamic_controller.go
@@ -309,6 +309,20 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
 	return true
 }
 
+// isRequeueSignal returns true if err is a requeue control-flow error (RequeueNeeded
+// or RequeueNeededAfter). These are not failures and should not be counted as handler errors.
+func isRequeueSignal(err error) bool {
+	if err == nil {
+		return false
+	}
+	switch err.(type) {
+	case *requeue.RequeueNeeded, *requeue.RequeueNeededAfter:
+		return true
+	default:
+		return false
+	}
+}
+
 func (dc *DynamicController) syncFunc(ctx context.Context, oi ObjectIdentifiers, handler Handler) error {
 	gvrKey := keyFromGVR(oi.GVR)
 	dc.log.V(1).Info("Syncing object", "gvr", gvrKey, "key", oi.NamespacedName)
@@ -323,7 +337,9 @@ func (dc *DynamicController) syncFunc(ctx context.Context, oi ObjectIdentifiers,
 	}()
 
 	err := handler(ctx, ctrl.Request{NamespacedName: oi.NamespacedName})
-	if err != nil {
+	// Do not count expected outcomes as handler errors: instance deleted (NotFound)
+	// and requeue signals (RequeueNeeded/RequeueNeededAfter) are normal control flow.
+	if err != nil && !apierrors.IsNotFound(err) && !isRequeueSignal(err) {
 		handlerErrorsTotal.WithLabelValues(gvrKey).Inc()
 	}
 	return err


pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -309,6 +309,20 @@ func (dc *DynamicController) processNextWorkItem(ctx context.Context) bool {
...
 }
 
+// isRequeueSignal returns true if err is a requeue control-flow error (RequeueNeeded
+// or RequeueNeededAfter). These are not failures and should not be counted as handler errors.
+func isRequeueSignal(err error) bool {
+	if err == nil {
+		return false
+	}
+	switch err.(type) {
+	case *requeue.RequeueNeeded, *requeue.RequeueNeededAfter:
```
jakobmoellerdev, line 319: im not sure it is valid to consider requeue needed as a non error signal. after all there can be failures that will induce a retry and still be a valid error. WDYT?

tomasaschan:
The proposed solution seems like it (even more than before) tries to get away with using error handling as a control flow mechanism. I think a better (but more intrusive) approach would be to mimic what `controller-runtime` does and pass [`reconcile.Result`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.1/pkg/reconcile#Result)`, err` as the return values where needed here.

Semantics in the controller-runtime project are to respect `reconcile.Result`'s requeue behavior if `err == nil`, and requeue immediately (respecting rate limiting) otherwise. I haven't looked into whether this needs an even bigger refactor, but it superficially seems like that could work here too.

jakobmoellerdev:
> The proposed solution seems like it (even more than before) tries to get away with using error handling as a control flow mechanism. I think a better (but more intrusive) approach would be to mimic what `controller-runtime` does and pass [`reconcile.Result`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.1/pkg/reconcile#Result)`, err` as the return values where needed here.
> 
> Semantics in the controller-runtime project are to respect `reconcile.Result`'s requeue behavior if `err == nil`, and requeue immediately (respecting rate limiting) otherwise. I haven't looked into whether this needs an even bigger refactor, but it superficially seems like that could work here too.

I feel like this is a bit of a moot point. specifically in controller runtime there is a terminal error that also defines a custom control flow switch no?

shivansh-gohem:
It looks like PR #1059 stalled out a few weeks ago after the discussion around refactoring to use `reconcile.Result, err` instead of pure error handling. 

I agree with @tomasaschan that mimicking `controller-runtime` semantics is the cleanest way to fix the metric reporting without losing visibility into actual errors. 

@jakobmoellerdev If @Avr8 isn't actively working on the refactor, I'd love to pick this up and implement the `(reconcile.Result, error)` pattern for the handlers!

pkg/dynamiccontroller/dynamic_controller.go
a-hilaly: My two cents: for this metric, I would not treat requeue as a real error. the metric here reads to me as “the dynamic controller handler failed”, not “the handler returned any non nil Go error”. requeue errors are explicit control flow signals to retry so counting them feels misleading.

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -323,7 +337,9 @@ func (dc *DynamicController) syncFunc(ctx context.Context, oi ObjectIdentifiers,
 	}()
 
 	err := handler(ctx, ctrl.Request{NamespacedName: oi.NamespacedName})
-	if err != nil {
+	// Do not count expected outcomes as handler errors: instance deleted (NotFound)
+	// and requeue signals (RequeueNeeded/RequeueNeededAfter) are normal control flow.
+	if err != nil && !apierrors.IsNotFound(err) && !isRequeueSignal(err) {
```
a-hilaly, line 342: how about extracting this to `shouldCountHandlerError(err error) bool` function?

a-hilaly (approved):
This is super annoying in dashboard, shows absurd amount of errors when everything is just fine

cc @jakobmoellerdev

Issue #1055, opened 2026-02-17
kro.run/node-id label exceeds 63-character Kubernetes limit for long resource IDs


fErten:
### Overview

 `kro.run/node-id` label exceeds 63-character Kubernetes limit for long resource IDs

## Description

**Observed Behavior**:

Starting with kro v0.8.0, the `kro.run/node-id` label is applied to **all** child resources during instance reconciliation, using the resource `id` from the RGD as the label value. When the `id` exceeds 63 characters, Kubernetes rejects the resource creation because label values are limited to 63 characters by the Kubernetes API spec. This causes instance reconciliation to fail with an error like:

```
ConfigMap "long-id-configmap" is invalid: metadata.labels: Invalid value:
"thisIsAnExtremelyLongResourceIdentifierThatExceedsSixtyThreeCharactersLimit":
must be no more than 63 characters
```

The resource `id` is not validated at RGD creating time, so the RGD is accepted and becomes Active, but instances fail with reconciliation.

**Expected Behavior**:

Either:
1. Kro should validate the `id` field length at RGD admission/reconciliation time and reject RGDs with IDs exceeding 63 characters with a clear error message, OR
2. Kro should truncate/hash the `kro.run/node-id` label value to fit within the 63-character Kubernetes label value limit while keeping the full ID internally

**Reproduction Steps**:

1. Install kro v0.8.0 or later

2. Apply this ResourceGraphDefinition with a resource ID exceeding 63 characters:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: long-id-bug-repro
spec:
  schema:
    apiVersion: v1alpha1
    kind: LongIdBugRepro
    spec:
      appName: string
  resources:
    - id: shortId
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: short-id-configmap
          namespace: default
        data:
          key: ${schema.spec.appName}
    - id: thisIsAnExtremelyLongResourceIdentifierThatExceedsSixtyThreeCharactersLimit
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: long-id-configmap
          namespace: default
        data:
          key: ${schema.spec.appName}
```

3. Wait for the RGD to become Active (it will be accepted without error)

4. Create an instance:

The instance will be stuck in `IN_PROGRESS` with:

```
Conditions:
  Type:     ResourcesReady
  Status:   False
  Reason:   NotReady
[...]

PR #1061, opened 2026-02-19
fix: hash kro.run/node-id label to respect 63-char limit

fErten:
Proposal for: https://github.com/kubernetes-sigs/kro/issues/1055

### Summary:
The kro.run/node-id label is set from the RGD resource id. When that ID is longer than 63 characters, Kubernetes rejects the resource because label values are limited to 63 characters. RGDs with long IDs become Active, but instance reconciliation fails at apply time.

### Change:
 Setting the label to a 16-character hex digest of the node ID (SHA-256), and store the full original ID in the annotation `kro.run/node-id-original`. This keeps labels within the 63-char limit and preserves the original ID for debugging.

Note: (first time contributor) My understanding is that this change does not introduce a breaking change, basically on the next reconciliation cycle, the controller applies by resource name and updates the label (and adds the annotation) on existing objects. With two reconcile cycles everything should be consistent.

**Migration fallback for collection children during upgrade:**

When the controller upgrades from non-hashing to hashing on node-ids, existing collection children still carry the raw (unhashed) kro.run/node-id label. listCollectionItems now performs a two-step lookup: it first queries by the hashed label value, and if no items are found, falls back to querying by the raw value. This ensures the deletion path doesn't miss pre-upgrade children and leave orphans. The fallback is safe to remove (in later major versions) once all clusters have completed at least one reconciliation cycle after the upgrade.

### DIFF
--- a/pkg/controller/instance/resources_collection.go
+++ b/pkg/controller/instance/resources_collection.go
@@ -71,7 +71,7 @@ func (c *Controller) processCollectionNode(
 	// selector listCollectionItems uses to LIST them.
 	selector := labels.SelectorFromSet(labels.Set{
 		metadata.InstanceIDLabel: string(rcx.Instance.GetUID()),
-		metadata.NodeIDLabel:     id,
+		metadata.NodeIDLabel:     metadata.SafeNodeID(id),
 	})
 	requestCollectionWatch(rcx, id, gvr, rcx.Instance.GetNamespace(), selector)
 
@@ -125,7 +125,7 @@ func (c *Controller) listCollectionItems(
 	instanceUID := string(rcx.Instance.GetUID())
 	selector := fmt.Sprintf("%s=%s,%s=%s",
 		metadata.InstanceIDLabel, instanceUID,
-		metadata.NodeIDLabel, nodeID,
+		metadata.NodeIDLabel, metadata.SafeNodeID(nodeID),
 	)
 
 	// List across all namespaces - collection items may span namespaces
--- a/pkg/controller/instance/resources_collection.go
+++ b/pkg/controller/instance/resources_collection.go
@@ -121,21 +121,45 @@ func (c *Controller) listCollectionItems(
 	gvr schema.GroupVersionResource,
 	nodeID string,
 ) ([]*unstructured.Unstructured, error) {
-	// Filter by both instance UID and node ID for precise matching
 	instanceUID := string(rcx.Instance.GetUID())
+
+	items, err := c.listByNodeIDLabel(rcx, gvr, instanceUID, metadata.SafeNodeID(nodeID))
+	if err != nil {
+		return nil, err
+	}
+	if len(items) > 0 {
+		return items, nil
+	}
+
+	// Fallback: children created before the node-id hashing change still carry
+	// the raw node ID as their label value. Without this second lookup the
+	// deletion path would miss them and leave orphans behind.
+	items, err = c.listByNodeIDLabel(rcx, gvr, instanceUID, nodeID)
+	if err != nil {
+		return nil, err
+	}
+	if len(items) > 0 {
+		rcx.Log.V(1).Info("found collection items with raw node-id label from pre-hashing controller version",
+			"nodeID", nodeID, "count", len(items))
+	}
+	return items, nil
+}
+
+func (c *Controller) listByNodeIDLabel(
+	rcx *ReconcileContext,
+	gvr schema.GroupVersionResource,
+	instanceUID, nodeIDValue string,
+) ([]*unstructured.Unstructured, error) {
 	selector := fmt.Sprintf("%s=%s,%s=%s",
 		metadata.InstanceIDLabel, instanceUID,
-		metadata.NodeIDLabel, metadata.SafeNodeID(nodeID),
+		metadata.NodeIDLabel, nodeIDValue,
 	)
-
-	// List across all namespaces - collection items may span namespaces
 	list, err := rcx.Client.Resource(gvr).List(rcx.Ctx, metav1.ListOptions{
 		LabelSelector: selector,
 	})
 	if err != nil {
 		return nil, err
 	}
-
 	items := make([]*unstructured.Unstructured, len(list.Items))
 	for i := range list.Items {
 		items[i] = &list.Items[i]
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -35,8 +35,8 @@ const (
 )
 
 const (
-	NodeIDLabel      = LabelKROPrefix + "node-id"
-	NodeIDAnnotation = LabelKROPrefix + "node-id-original"
+	NodeIDLabel    = LabelKROPrefix + "node-id"
+	NodeAnnotation = LabelKROPrefix + "node"
 
 	// Collection labels for tracking collection membership and position.
 	// These enable querying collection resources and understanding their position.
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -15,10 +15,9 @@
 package metadata
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"strconv"
 	"strings"
 
@@ -222,11 +221,12 @@ func NewCollectionItemLabeler(nodeID string, index, size int) GenericLabeler {
 	}
 }
 
-// SafeNodeID returns a deterministic fix length (16 chars) nodeId.
-// k8s label safe representation due to char limit
+// SafeNodeID returns a deterministic fixed-length (16 chars) nodeId
+// using FNV-64a, safe for K8s label values due to the 63-char limit.
 func SafeNodeID(nodeID string) string {
-	hash := sha256.Sum256([]byte(nodeID))
-	return hex.EncodeToString(hash[:])[:16]
+	h := fnv.New64a()
+	h.Write([]byte(nodeID))
+	return fmt.Sprintf("%016x", h.Sum64())
 }
 
 func safeVersion(version string) string {
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -15,6 +15,8 @@
 package metadata
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"strconv"
@@ -34,7 +36,8 @@ const (
 )
 
 const (
-	NodeIDLabel = LabelKROPrefix + "node-id"
+	NodeIDLabel      = LabelKROPrefix + "node-id"
+	NodeIDAnnotation = LabelKROPrefix + "node-id-original"
 
 	// Collection labels for tracking collection membership and position.
 	// These enable querying collection resources and understanding their position.
@@ -213,10 +216,17 @@ func NewKROMetaLabeler() GenericLabeler {
 // - collection-size: the total number of items in the collection (e.g "3")
 func NewCollectionItemLabeler(nodeID string, index, size int) GenericLabeler {
 	return map[string]string{
-		NodeIDLabel:          nodeID,
+		NodeIDLabel:          SafeNodeID(nodeID),
 		CollectionIndexLabel: strconv.Itoa(index),
 		CollectionSizeLabel:  strconv.Itoa(size),
 	}
+}
+
+// SafeNodeID returns a deterministic fix length (16 chars) nodeId.
+// k8s label safe representation due to char limit
+func SafeNodeID(nodeID string) string {
+	hash := sha256.Sum256([]byte(nodeID))
+	return hex.EncodeToString(hash[:])[:16]
 }
 
 func safeVersion(version string) string {
--- a/pkg/metadata/labels.go
+++ b/pkg/metadata/labels.go
@@ -32,11 +32,14 @@ import (
 const (
 	// LabelKROPrefix is the label key prefix used to identify KRO owned resources.
 	LabelKROPrefix = v1alpha1.KRODomainName + "/"
+	// InternalLabelKROPrefix is the prefix for internal-only metadata that
+	// should not be treated as part of kro's public API surface.
+	InternalLabelKROPrefix = "internal.kro.run/"
 )
 
 const (
 	NodeIDLabel    = LabelKROPrefix + "node-id"
-	NodeAnnotation = LabelKROPrefix + "node"
+	NodeAnnotation = InternalLabelKROPrefix + "node"
 
 	// Collection labels for tracking collection membership and position.
 	// These enable querying collection resources and understanding their position.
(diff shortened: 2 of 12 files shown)


a-hilaly (commented):
Thank you @fErten!

I'm split on this. Not opposed to hashing just wondering what else could break. Hashing makes the labels unreadable and now we need a side annotation to recover the original value. What's the trade off vs just enforcing node IDs to be shorter than 63 chars at admission layer? That feels simpler and keeps labels debuggable. Is there a reason ids need to be longer than 63 chars?

fErten:
> Is there a reason ids need to be longer than 63 chars?

Hi @a-hilaly, thanks for the reply! In practice, I think hitting this limit is more likely. For example, we use a pipeline to generate RGDs from raw K8s manifests (with some internal templating). We generate deterministic resource IDs using kind + metadata.name from each resource, which tends to exceed the length limit.

Having predictable resource IDs is crucial for us to apply changes before and after RGD creation, and to have correct mapping between resourceId/raw-manfest. This might be specific to our workflow, but we really appreciate the flexibility here.

I definitely do agree the plain mapping is more straightforward. However, my thinking was that since node-ids are just for internal traversal, we can effectively use an annotation to back up the original ID. Overall I think it shouldn't majorly impact traceability for both the end user and the controller.

Note: also verified the migration path locally. Testing with both short and long resourceIds, the reconciliation loop successfully adopted the resources. No manual intervention was needed, and I confirmed that no resources were orphaned in the process.

fErten:
fyi @jakobmoellerdev adjusted a single e2e test which was overlooked and all checks seem to pass now.

pkg/metadata/labels.go
```diff
@@ -200,12 +203,19 @@ func NewKROMetaLabeler() GenericLabeler {
...
+		NodeIDLabel:          SafeNodeID(nodeID),
 		CollectionIndexLabel: strconv.Itoa(index),
 		CollectionSizeLabel:  strconv.Itoa(size),
 	}
 }
 
+// SafeNodeID returns a deterministic fix length (16 chars) nodeId.
+// k8s label safe representation due to char limit
+func SafeNodeID(nodeID string) string {
+	hash := sha256.Sum256([]byte(nodeID))
```
jakobmoellerdev: whats your reasoning for shoosing sha256 opposed to something like fnv?

fErten: Hi, i've used it since lib was already available in the project and the applyset was already using sha256 for similar purpose, wanted to keep consistent. 

https://github.com/kubernetes-sigs/kro/blob/main/pkg/controller/instance/applyset/applyset.go#L639

jakobmoellerdev: I believe we have a non-cryptographic application here so i think im opposed to using a crypto hash if im honest. I think for applyset we just chose it due to the upstream specification (which probably also doesnt this imho).

fErten: fair enough, please see ce19e4ffef83b54420cb2d77a065507be9db52b3

a-hilaly (changes requested):
@fErten Thanks for the explanation, this clarified the use case and I agree we need better handling for long node IDs.

My remaining concern is collection migration safety. Changing `kro.run/node-id` semantics (raw -> hashed) can cause existing collection children to stop matching selector-based discovery after upgrade. In practice that can lead to replacement behavior (old items pruned, new ones created) which is risky for production workloads with long provisioning times like cloud resources.

How about a partial/auto migration path to support this, and can we make that explicit in code/tests/docs?

fErten:
> How about a partial/auto migration path to support this, and can we make that explicit in code/tests/docs?

@a-hilaly Thanks for the feedback. You're right that the label change causes list collection to return empty on upgrade, which can lead to a downstream CEL failure for a conciliation cycle.

On replacement behavior, prune seems to use `applyset.kubernetes.io/part-of` with UID tracking, and not rely on the kro.run/node-id. So existing items are updated in place and stay in the keep set. Am I missing a path where replacement could happen? I've check this migration for my project locally as well, didn't capture any delete events.

To mitigate the CEL blip, we can add a fallback in listCollectionItems query with hashed label first, if empty retry with raw label. First reconcile finds items via fallback, apply updates labels, subsequent reconciles match directly. But i am not sure this if is worth maintaining since the system can self heal there short, if there isn't a prune/re-create cycle i am missing. wdyt?

jakobmoellerdev:
@a-hilaly I think we chatted about this before and I was also fairly certain that no pruning should happen. can you give an example of a collection aggregation that would break?

a-hilaly:
@fErten  @jakobmoellerdev you both are right, i should stop reviewing PRs late at night :)

fErten:
@a-hilaly @jakobmoellerdev Perfect, thank you both for clarifying! Is there any other open topic for the provided fix?

jakobmoellerdev (commented):
Overall impl wise this is good but we need a documentation on this in the website on our labels. Ideally the behavior is described and outlined. (same as the annotation)

fErten:
> Overall impl wise this is good but we need a documentation on this in the website on our labels. Ideally the behavior is described and outlined. (same as the annotation)

thanks @jakobmoellerdev, please see 83e06dc1ccf078b3a5d530892c55137a9cbbe9c7, hope that is sufficient.

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -766,26 +766,43 @@ creates multiple resources at runtime, but dependencies treat it as one unit.
...
+
+**Annotations:**
+
+| Annotation | Description | Example |
+|------------|-------------|---------|
+| `kro.run/node-id-original` | The original human-readable resource ID from the RGD | `workerPods` |
+
+The `kro.run/node-id` label stores a deterministic hash rather than the raw
+resource ID because Kubernetes limits label values to 63 characters. The
+original ID is preserved in the `kro.run/node-id-original` annotation for
```
jakobmoellerdev: this could probably just be named kro.run/node now that I look at this no?

fErten: I assume you refer to the annotation, correct? It might be a good idea I agree, less verbose and `original` kinda implies something has changed although this is a just new annotation

fErten: @jakobmoellerdev please see cc3761f802bf2f52a802c3cad0b37edb261d63ab

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -769,26 +769,43 @@ creates multiple resources at runtime, but dependencies treat it as one unit.
...
 | `kro.run/instance-id` | UID of the instance that owns this resource | `a1b2c3...` |
 
-These labels enable:
-- **Querying**: Find all items in a collection with `kubectl get pods -l kro.run/node-id=workerPods`
+
+**Annotations:**
+
+| Annotation | Description | Example |
+|------------|-------------|---------|
+| `kro.run/node` | The human-readable resource ID from the RGD | `workerPods` |
```
jakobmoellerdev:
```suggestion
| `kro.run/node` | A human-readable node identifier from the RGD | `workerPods` |
```

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -769,26 +769,43 @@ creates multiple resources at runtime, but dependencies treat it as one unit.
...
+kro automatically applies labels and annotations to each resource in a collection to track
+membership and ordering. These are part of kro's public API and can be 
 used for querying or debugging:
 
+**Labels:**
+
 | Label | Description | Example |
 |-------|-------------|---------|
-| `kro.run/node-id` | The resource ID from the RGD | `workerPods` |
+| `kro.run/node-id` | FNV-64a hash of the resource ID from the RGD (16-char hex) | `39edeed848ad0ddc` |
```
jakobmoellerdev:
```suggestion
| `kro.run/node-id` | FNV-64a hash of the node identifier from the RGD (16-char hex) | `39edeed848ad0ddc` |
```

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
```diff
@@ -769,26 +769,43 @@ creates multiple resources at runtime, but dependencies treat it as one unit.
...
+| Annotation | Description | Example |
+|------------|-------------|---------|
+| `kro.run/node` | The human-readable resource ID from the RGD | `workerPods` |
+
+The `kro.run/node-id` label stores a deterministic hash rather than the raw
+resource ID because Kubernetes limits label values to 63 characters. The
+original ID is preserved in the `kro.run/node` annotation for
+traceability.
+
+These labels and annotations enable:
```
jakobmoellerdev: IMHO this is impractical. This was the very concern amine brought up. How would one even get the correct node identifier? IMHO the lookups described below are not really something we expect people to use manually

website/docs/docs/concepts/15-instances.md
```diff
@@ -123,7 +123,13 @@ Resources created by kro (Deployments, Services, ConfigMaps, etc.) receive label
 | `kro.run/collection-index` | Position in the collection (0-indexed) |
 | `kro.run/collection-size` | Total number of items in the collection |
 
-These labels allow you to identify exactly which instance owns each managed resource, which is essential when multiple instances of the same RGD exist in a cluster. For collection resources, see [Collection Labels](./rgd/02-resource-definitions/04-collections.md#collection-labels) for more details.
+**Annotations:**
+
+| Annotation | Description |
+|------------|-------------|
+| `kro.run/node` | The human-readable resource ID from the RGD (e.g., `workerPods`). This preserves the readable ID since `kro.run/node-id` stores a fixed-length hash to comply with the Kubernetes 63-character label value limit. |
```
jakobmoellerdev:
```suggestion
| `kro.run/node` | The human-readable node identifier based on the RGD (e.g., `workerPods`). This preserves a readable identifier since `kro.run/node-id` stores a fixed-length hash to comply with the Kubernetes 63-character label value limit. |
```

website/docs/docs/concepts/rgd/02-resource-definitions/04-collections.md
fErten: removed aecee970fcbd8255653feccfe650d58c2cfa9c50

jakobmoellerdev (commented):
/lgtm

fErten:
> /lgtm

thanks for the review @jakobmoellerdev, looks like PR only pending on the approval check now.

jakobmoellerdev:
This is normal don't worry. We have a 4-eye principle so I purposefully am waiting for another maintainer to approve of this change

a-hilaly (commented):
I think there is still one upgrade bug here for collections. Since the deletion path doesn't call prune, there is a problem is on collection discovery during deletion. after this change we only look up collection children using the hashed `kro.run/node-id` but old collection childs created before the upgrade will still have the old value until they get reconciled once.. so if an instance is deleted before the upgrade happens, we can miss those existing collection children during deletion and leave them behind.

I think we need a small migration fallback in collection lookup for that case + a test

a-hilaly:
One other thing: can we align this with [#1094](https://github.com/kubernetes-sigs/kro/pull/1094)? should we make at least the new annotation use `internal.kro.run/` instead?

fErten:
> I think we need a small migration fallback in collection lookup for that case + a test

hi @a-hilaly, I believe you are referring to this cel blip issue we've talked before (and if deletion happens before it), are you onboard with the suggestion below described previously.

>we can add a fallback in listCollectionItems query with hashed label first, if empty retry with raw label. First reconcile finds items via fallback, apply updates labels, subsequent reconciles match directly. But i am not sure this if is worth maintaining since the system can self heal there short, if there isn't a prune/re-create cycle i am missing. wdyt?

There is still an edge case if the controller is interrupted mid-reconcile after updating 2 of 3 children's labels to hashed then the fallback wouldn't run.

fErten:
Hi @jakobmoellerdev @a-hilaly, sorry for the break! I've synced the latest changes from main, added the migration path as discussed and also migrated the new annotation to `internal` pattern as suggested. Feel free to take a another look! Thanks in advance!

togrulseyid:
Hi @fErten @jakobmoellerdev  @a-hilaly, any updates on the timeline for merging this? We’re currently facing the same issue and are looking forward to the official fix. Thanks!

jakobmoellerdev (commented):
I am in general for fixing this but still not entirely sold on the label now not being human readable. Admittedly I dont have a better solution right now to be honest. very on the fence here

fErten:
> I am in general for fixing this but still not entirely sold on the label now not being human readable.

Hi, I am happy to support if there is an alternative idea. Maybe just for me to understand better: what is the use case you have in mind for the human-readable label? We are very early adopters of Kro and we did not use or need label itself for any traversing. From a code POV, that also all seems to be internal processing.

fErten:
Hi @jakobmoellerdev @a-hilaly just following up on our lastest discussion. Regarding the label human-readability, looks like there isn't a clear alternative at the moment, would you be open for merging the current mitigation plan? This issue has us completely stuck on v0.7 for a while and we'd be very happy to finally upgrade and use newer cool features like `omit` functions. Since it seems to be impacting other users as well? Wdyt, also happy to iterate if there is any open points. Thanks in advance!

jakobmoellerdev (commented):
@jlbutler @a-hilaly @michaelhtm can you confirm we are okay with this? would like to move ahead here

pkg/controller/instance/resources_collection.go
```diff
@@ -121,21 +121,45 @@ func (c *Controller) listCollectionItems(
...
+		return nil, err
+	}
+	if len(items) > 0 {
+		return items, nil
+	}
+
+	// Fallback: children created before the node-id hashing change still carry
+	// the raw node ID as their label value. Without this second lookup the
+	// deletion path would miss them and leave orphans behind.
+	items, err = c.listByNodeIDLabel(rcx, gvr, instanceUID, nodeID)
```
cheeseandcereal, line 137: I'm curious if we're worried about performance implications here. Now any time we can't find something, we are doubling the list api calls to kubernetes.

I'm not sure I have a better solution, since we need to be able to maintain some amount of backwards compatibility, but at the same time, this may permanently double our api usage for this not found edge case.

Curious what others think. Happy to accept if people don't think this is a concern.

cheeseandcereal:
Yeah doesn't seem like there's a perfect solution here, so this seems like a reasonable approach. My 2 questions:

1. Do we want to use the `internal.kro.run` prefix for the annotation and imply that this value is internal and shouldn't be relied upon? Or should we use the same regular prefix we do for the labels? I would vote for a "non-internal" prefix so users can actually assume they can rely on this annotation, but happy to defer.
2. Are we worried about the performance implications of the backwards compatibility branch, which will require us to double our lists in any case where the collection items are not found?

PR #1062, opened 2026-02-19
feat: disable instance reconciliation when a kro.run/reconcile:"disabled" label is present

### LINKED ISSUE #1042: Add ability to disable automatic reconciliation per instance (like Argo auto-sync toggle)
## Problem Statement

kro continuously reconciles instances by design. During a production incident, it may be necessary to apply immediate manual changes without kro reverting them.

Currently, I’m not aware of a supported way to temporarily disable reconciliation for a specific instance. Removing the label `kro.run/owned: "true"` seems to stop the loop, but it’s unclear whether this is an intended or safe approach.

A clear, documented mechanism for suspending reconciliation per instance would improve incident response workflows.

---

## Proposed Solution

Introduce an official per-instance toggle to suspend reconciliation, for example:

- `spec.suspend: true`
- or an annotation like `kro.run/reconcile: disabled`

Default behavior would remain unchanged.

---

## Alternatives Considered

- Removing the `kro.run/owned` label (appears to work but feels like a workaround)
- Scaling down the controller (too broad)
- Blocking updates via RBAC (not practical)

---

## Additional Context

Similar to disabling auto-sync in Argo CD, this would allow controlled manual intervention without affecting other instances.

---

* Please vote on this issue by adding a 👍 reaction  
* If interested in working on this feature, please leave a comment

NickKeller:
resolves #1042

This change modifies the instance controller to ignore updates to the child nodes when a `kro.run/reconcile: "disabled"` label is present.
This will also place a "Suspended" condition on the instance, to be clear to the user what is happening

This only affects reconciliation on create and update. Delete is always allowed, so that way resources are not orphaned.

### DIFF
--- a/website/docs/docs/concepts/15-instances.md
+++ b/website/docs/docs/concepts/15-instances.md
@@ -45,6 +45,10 @@ This continuous loop ensures your resources stay in sync with your desired state
 - Consistent state management
 - Rich status tracking
 
+:::tip
+To suspend active reconciliation of an instance for debugging purposes, apply a label with the key `kro.run/reconcile` and the value `disabled`. See [Debugging Specific Labels](./15-instances.md#labels-and-ownership) for more details
+:::
+
 ### Reactive Reconciliation
 
 kro automatically watches all resources managed by an instance and triggers reconciliation when any of them change:
@@ -218,7 +222,7 @@ High-level status showing what the instance is doing:
 
 ### 2. Conditions
 
-Detailed status information structured hierarchically. kro provides a top-level `Ready` condition that reflects overall instance health, supported by three sub-conditions that track different phases:
+Detailed status information structured hierarchically. kro provides a top-level `Ready` condition that reflects overall instance health, supported by four sub-conditions that track different phases:
 
 - **`InstanceManaged`** - Instance finalizers and labels are properly set
   - Ensures the instance is under kro's management
@@ -235,14 +239,19 @@ Detailed status information structured hierarchically. kro provides a top-level 
   - Monitors the health of resources in topological order
   - Reports when all resources have reached their ready state
 
+- **`ReconciliationSuspended`** - Instance Reconciliation is Suspended
+  - Set to True when the label `kro.run/reconcile: "disabled"` label key and value are present. Otherwise set to false
+  - Explicitly shows if kro is actively reconciling an instance or not
+
 - **`Ready`** - Instance is fully operational (top-level condition)
   - Aggregates the state of all sub-conditions
   - Only becomes True when all sub-conditions are True
   - **The primary condition to monitor for instance health**
   - Use this condition in automation, CI/CD, and health checks
 
+
 :::tip
-Always use the `Ready` condition to determine instance health. The sub-conditions (`InstanceManaged`, `GraphResolved`, `ResourcesReady`) are provided for debugging purposes and may change in future versions. kro reserves the right to add, remove, or modify sub-conditions without breaking compatibility as long as the `Ready` condition behavior remains stable.
+Always use the `Ready` condition to determine instance health. The sub-conditions (`InstanceManaged`, `GraphResolved`, `ResourcesReady`, `ReconciliationSuspended`) are provided for debugging purposes and may change in future versions. kro reserves the right to add, remove, or modify sub-conditions without breaking compatibility as long as the `Ready` condition behavior remains stable.
 :::
 
 Each condition includes:


NickKeller:
right now this only disables instance reconciliation, do we also want to have this logic for RGD reconciliation?

pkg/controller/instance/controller.go
```diff
@@ -170,12 +171,19 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
...
 	//--------------------------------------------------------------
-	// 7. Reconcile nodes (SSA + prune) and update runtime state
+	// 7. Reconcile nodes (SSA + prune) and update runtime state, only if the suspend label is not present.
 	//--------------------------------------------------------------
-	if err := c.reconcileNodes(rcx); err != nil {
-		rcx.Mark.ResourcesNotReady("resource reconciliation failed: %v", err)
-		_ = c.updateStatus(rcx)
-		return err
+	labels := inst.GetLabels()
+	reconcileState, ok := labels[metadata.InstanceSuspendReconciliationLabel]
```
NickKeller: this check here is only required if we want to have an explicit status condition that the reconciliation is suspended. Otherwise this gets a lot simpler, adn we can implement this entirely in the dynamiccontrollers updatefunc.

I however, like having this here, since we can be explicit about the status.

pkg/metadata/labels.go
```diff
@@ -47,12 +47,13 @@ const (
...
-	InstanceGroupLabel     = LabelKROPrefix + "instance-group"
-	InstanceVersionLabel   = LabelKROPrefix + "instance-version"
-	InstanceKindLabel      = LabelKROPrefix + "instance-kind"
+	InstanceIDLabel                    = LabelKROPrefix + "instance-id"
+	InstanceLabel                      = LabelKROPrefix + "instance-name"
+	InstanceNamespaceLabel             = LabelKROPrefix + "instance-namespace"
+	InstanceGroupLabel                 = LabelKROPrefix + "instance-group"
+	InstanceVersionLabel               = LabelKROPrefix + "instance-version"
+	InstanceKindLabel                  = LabelKROPrefix + "instance-kind"
+	InstanceSuspendReconciliationLabel = LabelKROPrefix + "reconcile"
```
jakobmoellerdev: we call the label suspend in code but call it reconcile as a value. why not have the value be "suspend"?

NickKeller: I think it's more clear to have a label of `kro.run/reconcile: "disabled"` than `kro.run/suspend: "true"`, because it clearly states that reconciliation is disabled.

So I can change the code to have a more clear variable name

NickKeller: changed variable name to "InstanceReconcileLabel"

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -358,9 +358,14 @@ func (dc *DynamicController) updateFunc(parentGVR schema.GroupVersionResource, o
 		return
 	}
 	if newMeta.GetGeneration() == oldMeta.GetGeneration() {
-		dc.log.V(2).Info("Skipping update due to unchanged generation",
-			"name", newMeta.GetName(), "namespace", newMeta.GetNamespace(), "generation", newMeta.GetGeneration())
-		return
+		// normally skip, but we should enqueue if the oldMeta had the reconcile disabled label, and the newMeta doesn't
+		oldLbls := oldMeta.GetLabels()
+		newLbls := newMeta.GetLabels()
+		if !(oldLbls[metadata.InstanceReconcileLabel] == "disabled" && newLbls[metadata.InstanceReconcileLabel] != "disabled") {
```
a-hilaly: Do we want to use `strings.EqualFold` here as well?

pkg/dynamiccontroller/dynamic_controller_test.go
```diff
@@ -463,6 +463,13 @@ func TestUpdateFunc_GenerationFiltering(t *testing.T) {
 
 	dc.updateFunc(parentGVR, oldObj, struct{}{})
 	assert.Equal(t, 0, dc.queue.Len(), "invalid new object should not enqueue")
+
+	oldObj.SetLabels(map[string]string{
+		metadata.InstanceReconcileLabel: "disabled",
+	})
```
a-hilaly, line 488: oldObj and newObj still have different generations. we need a seperate unit test for this.

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -358,9 +358,14 @@ func (dc *DynamicController) updateFunc(parentGVR schema.GroupVersionResource, o
...
+			dc.log.V(2).Info("Skipping update due to unchanged generation",
+				"name", newMeta.GetName(), "namespace", newMeta.GetNamespace(), "generation", newMeta.GetGeneration())
+			return
+		}
```
a-hilaly, line 369: maybe let's pull this entire logic in a seperate function that we can unit test?

a-hilaly (commented):
Thanks @NickKeller - `kro.run/reconcile: "disabled"` looks reasonable to me and not off pattern; no strict Kubernetes standard exists here, and this wording is clear. I’d just ensure pause/unpause checks are case insensitive in both controller and dynamic-controller paths.

edit: just an FYI.

looks like crossplane and cluster api both use `/paused: "true"`  cc @jakobmoellerdev

pkg/dynamiccontroller/dynamic_controller.go
NickKeller: ah right - I use EqualFold elsewhere, I should be consistent

a-hilaly (commented):
Can we also update the latest docs of `Instance` to mention this annotation?

NickKeller:
> Can we also update the latest docs of `Instance` to mention this annotation?

I was just thinking that :)

pkg/dynamiccontroller/dynamic_controller_test.go
NickKeller: Moved to a dedicated function and added a dedicated unit test

pkg/dynamiccontroller/dynamic_controller.go
NickKeller: done

website/docs/docs/concepts/15-instances.md
a-hilaly, line 250: I think we need to mention the new condition in here https://kro.run/docs/concepts/instances#2-conditions

a-hilaly, line 60: Maybe a `:::tip` here stating that to suspend the reconciliation of an instance... you can set this annotation, a link to other section?

NickKeller: added :)

It made more sense to put it in the section above, since that talks about the specific instance reconciliation rather than the resources

NickKeller: done :)

a-hilaly (approved):
Awesome, thanks @NickKeller !
/approve

PR #1064, opened 2026-02-21
KREP-012: Cluster Targets for multi-cluster orchestration

JamesGAWS:
KREP-012 introduce cluster targets to the RGD schema allowing RGD authors to to orchestrate resources across multiple clusters from a single cluster.


docs/design/proposals/cluster-targets.md
```diff
@@ -0,0 +1,663 @@
...
+}
+
+// Target specifies the destination for resource creation.
+// Currently supports Cluster targeting with future extensibility for other protocols.
+type Target struct {
+    // Cluster specifies a remote Kubernetes cluster as the target.
+    // Exactly one target type must be specified.
+    //
+    // +kubebuilder:validation:Optional
+    Cluster *ClusterProtocol `json:"cluster,omitempty"`
```
NickKeller, line 80: doesn't this mean that any instances of this RGD will only be shipped to 1 cluster? What if multiple clusters are desired?

JamesGAWS: Example 3 covers the use case of orchestrating across multiple clusters with nested RGDs. Future directions covers extending the `resource` schema to include a `target` field.

docs/design/proposals/cluster-targets.md
```diff
@@ -0,0 +1,663 @@
...
+resource management.
+
+## Proposal
+
+Add a `Target` field to the RGD spec that specifies where all resources in the RGD should be created. The `Target` contains a `Cluster` field with kubeconfig credentials for accessing remote clusters.
+
+### Key Features
+
+1. **Cluster-scoped targeting**: Resources in an RGD can specify a target cluster
+2. **Kubeconfig-based authentication**: Cluster credentials stored as Kubernetes Secrets
```
reshnm, line 37: I know this may be a stretch. But a lot of kubernetes controllers that support multi-cluster, "only" support Kubernetes configs.
Kubernetes has native support for JWT and OIDC. One thing that would be really helpful is that one coud specify the remote cluster API endoint, CA and an audience. Kro would then use its own service account to create a token on the fly to authenticate at the remote cluster.
The first version of the multi cluster support wouldn't need to support that, but maybe it can be considered in the API that there may be different types of authentication in the future.

PR #1074, opened 2026-02-26
refactor(graph): rewrite builder as a multi stage compiler pipeline

a-hilaly:
The old Builder was a single ~2k+ line file that parsed, validated,
resolved, linked, type-checked, compiled, and assembled a graph all in
one pass. This made it hard to test individual concerns, hard to produce
good error messages, and hard to reason about what data was available at
which point in the process. This commit replaces it with a proper staged
pipeline where each phase has a clear input, a clear output, and no
hidden coupling to the others.

The stages are `Parse`, `Validate`, `Resolve`, `Link`, `TypeCheck`, `Generate`,  and
`Assemble`. Each one produces a typed intermediate representation that the
next stage consumes, so the builder itself becomes a short function that
chains them together. Errors are now classified as terminal or retriable
at the source, which lets the controller decide retry behavior without
string matching.

The outcome is better local reasoning, safer evolution, and clearer runtime
behavior. Errors are produced at the phase where they originate and carry
retry semantics from the source, allowing controllers to make deterministic
requeue decisions without string matching.

d2(1)


jakobmoellerdev:
closing this one out for now because it got stale too quickly but lets please revisit this again for 0.10 and try to make it happen. I still love this idea.

/close

PR #1075, opened 2026-02-26
fix: remove the rgd labels from managed resources

jakobmoellerdev:
- Updated child resource labeling logic to exclude RGD-specific labels, aligning ownership tracking with ApplySet and avoiding conflicts and "flip-flopping" with ownership when using instances from other RGDs inside the resources
- Update logic in controllers and integration tests to prevent unnecessary RGD labels on child resources.
- Also updates our docs with our actual label values set.

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller_reconcile.go
+++ b/pkg/controller/resourcegraphdefinition/controller_reconcile.go
@@ -51,15 +51,17 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(
 	}
 	mark.ResourceGraphValid()
 
-	// Setup metadata labeling
-	graphExecLabeler, err := r.setupLabeler(rgd)
+	// Build instance labeler: kro metadata + RGD-specific labels.
+	// This is applied to CRDs and instances. Child resources only get r.metadataLabeler.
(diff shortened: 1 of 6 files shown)


pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -74,10 +74,19 @@ func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(
...
 	}
 
+	// Build a child resource labeler that excludes RGD-specific labels.
+	// RGD labels are only meaningful on CRDs (for ownership detection) and
+	// instances (for user visibility). Child resources use ApplySet labels
+	// for GC/pruning instead.
+	childResourceLabeler := metadata.Exclude(graphExecLabeler,
+		metadata.ResourceGraphDefinitionIDLabel,
+		metadata.ResourceGraphDefinitionNameLabel,
+	)
```
michaelhtm: nit: would childResourceLabeler be requal to r.metadataLabeler?
Exclude is a good addition, but i wonder if it is necessary here

jakobmoellerdev: thanks for catching this!

PR #1077, opened 2026-02-27
feat: implement standard Kubernetes feature gates

### LINKED ISSUE #510: Feature flags
### Feature Description

Need a way to flag features as alpha/beta/GA, so we can introudce a "stable" version of KRO while still introducing new experimental features.

shivansh-gohem:
Fixes #510 

### Description
As suggested by @ramonvermeulen, this implements a Kubernetes-style feature gate registry using `k8s.io/component-base/featuregate` (the same package used by Kubernetes core). This provides standard `--feature-gates=MyFeature=true` CLI parsing, lifecycle tracking (Alpha/Beta/GA), and default states.

### Changes Made
* **`pkg/features/features.go`**: Implemented the global feature registry (`FeatureGate`), starting with `ParallelDagWalking` as the first Alpha feature gate placeholder.
* **`pkg/features/features_test.go`**: Added a test suite covering default states, overriding via `Set()`, and error handling for unknown features.
* **`cmd/controller/main.go`**: Wired the `-feature-gates` string directly into KRO's existing `flag` setup, successfully bridging it to `featuregate.Set()`. Added startup logging for known feature gates.
* **`go.mod`**: Promoted `k8s.io/component-base` to a direct dependency.

### Verification
* Verified `go test ./pkg/features/...` passes.
* Verified `go run ./cmd/controller/main.go --help` properly displays the newly registered `-feature-gates` flag.
* Verified controller boots successfully against a live `kind` cluster with the flag enabled.


jakobmoellerdev:
I dont really see any features that need gating like this. what were you thinking of when writing this?

shivansh-gohem:
@jakobmoellerdev as in the Issue #510. In that thread, @ramonvermeulen suggested implementing the standard `k8s.io/component-base/featuregate` registry for future use, and explicitly mentioned `ParallelDagWalking` as a potential first candidate, which is why I used it as the initial Alpha placeholder here.  if you would suggest I'll remove the `ParallelDagWalking` placeholder and just leave an empty registry, or if the core team has decided feature gates aren't needed anymore, just let me know! I am happy to adjust the PR

jakobmoellerdev:
what would parallel dag walking even be? I feel like this can be brought back in when we actually want any gating to happen. currently I dont see a need for gating and we should stabilize and work with what we have. happy to revisit this anytime though

/close

shivansh-gohem:
Understood! That makes complete sense regarding stabilization. I'll leave this closed for now, but the implementation is here in the commit history whenever the project is ready for it down the road. I'll take a look at the active issues and see where else I can help out!

ramonvermeulen:
I mentioned the two options mainly to open up the discussion about requirements and usage of feature gates, to figure out if there is really a need / use-case for this. That's also why I didn't start implementing right away.

In the example I mentioned parallel dag walking because I know it is a feature for which there is an open issue, and with feature gates it could theoretically be added as experimental / "non-GA" feature first. It has nothing to do specifically with feature gates.

Issue #1071, opened 2026-02-25
Status Condition Regression: Pending Sub‑Object No Longer Reported


tbarizien:
### Description

When an RGD has a topological order ["A","B","C"] where C depends on A and B. If a CRD instantiates that RGD and has one of A or B pending (not ready, erroring or whatever other reason), kro no longer points to the node that is pending but rather on the next node waiting for resolution. 

This makes debugging quite frustrating since kro indicates "C cannot be resolved" or "awaiting resource readiness" instead of "A is not ready" via the CRD conditions, making the user have to know the RGD to be able to pinpoint the source of the problem.

This wasnt a problem before `8.0.0` (even though the message was far from perfect, possibly signaling some erroring when there wasn't any), from what I can read in the code, some refractoring done to introduce collection support introduced this regression. 

**Observed Behavior**:

When one of the nodes is not ready, the `ResourcesReady` and `Ready` conditions report on the next node and not on the pending node with the following message: `resource reconciliation failed: waiting for unresolved resource "{next_node_name}"`, or with `awaiting resource readiness`. 

**Expected Behavior**:

When one of the nodes is not ready, the `ResourcesReady` and `Ready` conditions should indicate which node/resource is pending and awaiting to be ready.  
Using kro version 7.0.0, the message would have been following: `at least one resource reports an error: resource not ready: expression {pending_node_readyWhen_condition} evaluated to false: %!w(<nil>)`. This is not perfect but definetely more helpful.

**Reproduction Steps** (Please include `ResourceGraphDefinition` and `Instances` files):

Use the [job-example](https://kro.run/examples/kubernetes/job-example) RGD and instantiate it with default values. While the job is sleeping for 30 seconds, check the status of the instantiated CRD. It simply reports `awaiting resource readiness` without providing information that it is the sleep job that is pending. 

**Versions**:
- kro version: 8.5.0
- Kubernetes Version (`kubectl version`): 1.34

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Which option describes the most your issue?

Instance (Create, Update, Deletion)

tbarizien:
[...]

PR #1078, opened 2026-02-27
fix: pass up expressive error

fabianburth:
This PR addresses the regression reported in https://github.com/kubernetes-sigs/kro/issues/1071

### DIFF
--- a/pkg/controller/instance/resources.go
+++ b/pkg/controller/instance/resources.go
@@ -78,6 +78,7 @@ func (c *Controller) reconcileNodes(rcx *ReconcileContext) error {
 	// ---------------------------------------------------------
 	// 4. Prune orphans (when desired is fully resolved)
 	// ---------------------------------------------------------
+	pruneNeedsRetry := false
 	//
 	// Prune is intentionally gated by two independent conditions:
 	//   1) prune == true  -> all desired objects were resolvable (no ErrDataPending)
@@ -88,11 +89,12 @@ func (c *Controller) reconcileNodes(rcx *ReconcileContext) error {
 	// omitted from the apply set. Keeping both checks visible prevents  regressions
 	// where one gate gets removed and prune becomes unsafe.
 	if prune && result.Errors() == nil {
-		pruned, err := c.pruneOrphans(rcx, applier, result, supersetPatch, batchMeta)
+		pruned, needsRetry, err := c.pruneOrphans(rcx, applier, result, supersetPatch, batchMeta)
 		if err != nil {
 			return err
 		}
 		clusterMutated = clusterMutated || pruned
+		pruneNeedsRetry = pruneNeedsRetry || needsRetry
 	}
 
 	// ---------------------------------------------------------
@@ -109,6 +111,9 @@ func (c *Controller) reconcileNodes(rcx *ReconcileContext) error {
 
 	if lastUnresolvedErr != nil {
 		return rcx.delayedRequeue(fmt.Errorf("waiting for unresolved resource: %w", lastUnresolvedErr))
+	}
+	if pruneNeedsRetry {
+		return rcx.delayedRequeue(fmt.Errorf("prune encountered UID conflicts; retrying"))
 	}
 	if clusterMutated {
 		return rcx.delayedRequeue(fmt.Errorf("cluster mutated"))
@@ -145,28 +150,37 @@ func (c *Controller) processNodes(
 }
 
 // pruneOrphans deletes previously managed resources that are not in the current
-// apply set and then shrinks parent applyset metadata.
+// apply set. It shrinks parent applyset metadata only when prune completes
+// without UID conflicts.
 func (c *Controller) pruneOrphans(
 	rcx *ReconcileContext,
 	applier *applyset.ApplySet,
 	result *applyset.ApplyResult,
 	supersetPatch applyset.Metadata,
 	batchMeta applyset.Metadata,
-) (bool, error) {
+) (bool, bool, error) {
 	pruneScope := supersetPatch.PruneScope()
 	pruneResult, err := applier.Prune(rcx.Ctx, applyset.PruneOptions{
 		KeepUIDs: result.ObservedUIDs(),
 		Scope:    pruneScope,
 	})
 	if err != nil {
-		return false, rcx.delayedRequeue(fmt.Errorf("prune failed: %w", err))
+		return false, false, rcx.delayedRequeue(fmt.Errorf("prune failed: %w", err))
+	}
+
+	// Keep superset metadata and retry prune on UID conflicts.
+	if pruneResult.HasConflicts() {
+		rcx.Log.V(1).Info("prune skipped resources due to UID conflicts; keeping superset applyset metadata for retry",
+			"conflicts", pruneResult.Conflicts,
+		)
+		return pruneResult.HasPruned(), true, nil
 	}
 
 	// Prune succeeded (errors return directly), safe to shrink metadata
 	if err := c.patchInstanceWithApplySetMetadata(rcx, batchMeta); err != nil {
 		rcx.Log.V(1).Info("failed to shrink instance annotations", "error", err)
 	}
-	return pruneResult.HasPruned(), nil
+	return pruneResult.HasPruned(), false, nil
 }
 
 // createApplySet constructs an applyset configured for the current instance.
--- a/NOTICE
+++ b/NOTICE
@@ -14,7 +14,6 @@ Submitted on behalf of a third-party: @AnushkaaBansal
 Submitted on behalf of a third-party: @AshwinSriram11
 Submitted on behalf of a third-party: @candonov
 Submitted on behalf of a third-party: @chkp-dvirpa
-Submitted on behalf of a third-party: @codesenju
 Submitted on behalf of a third-party: @feliperubo
 Submitted on behalf of a third-party: @HeeManSu7
 Submitted on behalf of a third-party: @iamahgoub
--- a/pkg/controller/instance/applyset/applyset.go
+++ b/pkg/controller/instance/applyset/applyset.go
@@ -336,12 +336,12 @@ func (a *ApplySet) Prune(ctx context.Context, opts PruneOptions) (*PruneResult, 
 	}
 
 	// List and delete orphans
-	pruned, err := a.prune(ctx, pruneMappings, scopeNamespaces, opts.KeepUIDs, opts.Concurrency)
+	pruned, conflicts, err := a.prune(ctx, pruneMappings, scopeNamespaces, opts.KeepUIDs, opts.Concurrency)
 	if err != nil {
 		return nil, err
 	}
 
-	return &PruneResult{Pruned: pruned}, nil
+	return &PruneResult{Pruned: pruned, Conflicts: conflicts}, nil
 }
 
 func (a *ApplySet) applyResource(
@@ -454,7 +454,7 @@ func (a *ApplySet) prune(
 	namespaces sets.Set[string],
 	keepUIDs sets.Set[types.UID],
 	concurrency int,
-) ([]PruneResultItem, error) {
+) ([]PruneResultItem, int, error) {
 	// Track candidates with their GVR for deletion
 	type pruneCandidate struct {
 		obj *unstructured.Unstructured
@@ -522,7 +522,7 @@ func (a *ApplySet) prune(
 		})
 	}
 	if err := listGroup.Wait(); err != nil {
-		return nil, err
+		return nil, 0, err
 	}
 
 	// Delete candidates concurrently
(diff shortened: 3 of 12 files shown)


pkg/controller/instance/resources.go
```diff
@@ -119,24 +124,24 @@ func (c *Controller) reconcileNodes(rcx *ReconcileContext) error {
...
-			return nil, "", err
+			if !runtime.IsDataPending(err) {
+				return nil, err
+			}
+			unresolvedErr = errors.Join(unresolvedErr, err)
 		}
 		resources = append(resources, resourcesToAdd...)
-		if unresolvedID != "" {
-			lastUnresolved = unresolvedID
-		}
```
fabianburth: Not sure yet if accumulating the errors might lead to overwhelmingly long errors.

pkg/controller/instance/state.go
```diff
@@ -133,10 +133,10 @@ func (s *StateManager) NewNodeState(id string) *NodeState {
 }
 
 // NodeErrors aggregates errors across all node states.
-func (s *StateManager) NodeErrors() error {
+func (s *StateManager) NodeErrors(filter func(e error) bool) error {
```
fabianburth: Bit of a hack, maybe we should add a separate `WaitingForReadinessErr` to the node state.

pkg/controller/instance/resources.go
```diff
@@ -102,8 +107,8 @@ func (c *Controller) reconcileNodes(rcx *ReconcileContext) error {
 	// (including WaitingForReadiness) before the controller checks it.
 	rcx.StateManager.Update()
 
-	if lastUnresolved != "" {
-		return rcx.delayedRequeue(fmt.Errorf("waiting for unresolved resource %q", lastUnresolved))
+	if unresolvedErr != nil {
+		return rcx.delayedRequeue(fmt.Errorf("waiting for unresolved resource: %w", unresolvedErr))
 	}
```
fabianburth: I've also been wondering, why is it required to enqueue another event if we are not able to resolve a node? Is there a change an expression becomes resolvable without our controller receiving another watch event?

pkg/controller/instance/resources.go
```diff
@@ -576,8 +581,14 @@ func (c *Controller) processApplyResults(
...
 
+	// we don't want to keep reconciling with exponential backoff if the error
+	// is waiting for readiness
+	// we will notice the readiness through a watch event
+	filter := func(err error) bool {
+		return !runtime.IsWaitingForReadiness(err)
+	}
 	// Aggregate all node errors
-	if err := rcx.StateManager.NodeErrors(); err != nil {
+	if err := rcx.StateManager.NodeErrors(filter); err != nil {
```
jakobmoellerdev: as far as I can see this is the only place where you actually do set a filter. IMHO this interface is too complex. Why cant we check with errors.Is on the readiness error? there are a few things on simplification:

1. NodeErrors is a multi-return that is just abstracted behind error. If you wanna check on the nested errors you could use Unwrap here or return []error
2. I think filter should be a vararg, and not passing anything will be the same as your return true filter that you use everywhere else.

pkg/runtime/errors.go
```diff
@@ -35,6 +37,10 @@ func IsDataPending(err error) bool {
 	return errors.Is(err, ErrDataPending)
 }
 
+func IsWaitingForReadiness(err error) bool {
+	return errors.Is(err, ErrWaitingForReadiness)
```
jakobmoellerdev: This is really not necessary as a function, its just a one liner.

pkg/controller/instance/state.go
jakobmoellerdev: node state currently only contains cel state. maybe we can not introduce an error that contains all of the states and has some methods?

pkg/runtime/node.go
```diff
@@ -409,7 +405,7 @@ func (n *Node) evaluateExprsFiltered(exprs map[string]struct{}, continueOnPendin
 					if continueOnPending {
 						continue
 					}
-					return nil, true, ErrDataPending
+					return nil, true, fmt.Errorf("failed to evaluate expression: %q (%w)", err.Error(), ErrDataPending)
```
jakobmoellerdev: something is definetly off here. one should never unwrap an error to wrap it again. you can use %w multiple times now I believe in newer go versions but even if not, you should not deref the error string

pkg/runtime/node.go
```diff
@@ -525,30 +521,30 @@ func (n *Node) isSingleResourceReady() (bool, error) {
...
-		return false, nil
+		return fmt.Errorf("collection not computed (%w)", ErrWaitingForReadiness)
 	}
 	if len(n.desired) == 0 {
-		return true, nil
+		return nil
 	}
 	if len(n.observed) < len(n.desired) {
-		return false, nil
+		return fmt.Errorf("collection not ready (%w)", ErrWaitingForReadiness)
```
jakobmoellerdev: maybe we should write the collection parameters into the error?

pkg/runtime/errors.go
fabianburth: Well, I included it for consistency with 
```
// IsDataPending returns true if the error indicates data is pending and
// evaluation should be retried later.
func IsDataPending(err error) bool {
	return errors.Is(err, ErrDataPending)
}
```

jakobmoellerdev: also this is probably moot :D wouldnt do this again.

pkg/runtime/node.go
fabianburth: I think I had this. The issue is that only the first one is being considered by `Unwrap` which leads to issues. I could also "double wrap" with `fmt.Errorf` ;)

pkg/controller/instance/state.go
fabianburth: Removed the filter again. I am now manually filtering for that one specific call in `resources.go`.

pkg/controller/instance/resources.go
fabianburth: > Removed the filter again. I am now manually filtering for that one specific call in resource.go.

pkg/runtime/errors.go
fabianburth: removed

pkg/runtime/node.go
fabianburth: nvm, tried it again, I guess I just got confused back then.

pkg/runtime/node.go
fabianburth: Added node id and observed vs desired to the error message. Is that about what you expected?

pkg/controller/instance/resources.go
jakobmoellerdev: I think due to external resources this can always happen. if we watch them maybe not. I think the risk is quite high that we might drop a change in state.

pkg/runtime/node.go
```diff
@@ -492,30 +489,30 @@ func (n *Node) SetObserved(observed []*unstructured.Unstructured) {
...
 	}
-	return n.isSingleResourceReady()
+	return n.checkSingleResourceReadiness()
 }
 
-func (n *Node) isSingleResourceReady() (bool, error) {
+func (n *Node) checkSingleResourceReadiness() error {
 	if len(n.observed) == 0 {
-		return false, nil
+		return fmt.Errorf("node %q: no observed state for node %q: %w", n.Spec.Meta.ID, n.Spec.Meta.ID, ErrWaitingForReadiness)
```
a-hilaly: nit: redundant `node: %q` in the error mesage

pkg/controller/instance/resources.go
```diff
@@ -576,8 +581,16 @@ func (c *Controller) processApplyResults(
...
+	// we will notice the readiness through a watch event
+	var errs []error
+	for _, state := range rcx.StateManager.NodeStates {
+		if state.Err != nil && !errors.Is(state.Err, runtime.ErrWaitingForReadiness) {
+			errs = append(errs, state.Err)
+		}
+	}
+	if err := errors.Join(errs...); err != nil {
 		return fmt.Errorf("apply results contain errors: %w", err)
 	}
```
a-hilaly: i'm not sure this is correct. We do want to requeue for at least two reasons:
1/ the watches can be missed in faulty/intense environements
2/ status changes are ignored (skipped when on unchanged generation)

pkg/runtime/node.go
```diff
@@ -492,30 +489,30 @@ func (n *Node) SetObserved(observed []*unstructured.Unstructured) {
 	n.observed = observed
 }
 
-// IsReady evaluates readyWhen expressions using observed state.
+// CheckReadiness evaluates readyWhen expressions using observed state.
```
a-hilaly, line 492: I’m worried this PR went out of scope. The PR goal looked like improving error clarity but now it change the reconciliation behaviour in meaningful ways:
- readiness handling moved to new control flow (`CheckReadiness` / `ErrWaitingForReadiness`)
- processApplyResults now suppresses waiting errors
- reconcile requeue behavior appears different

fabianburth: If I did, it was by accident. 
I actually tried very hard not to change the reconcile behaviour at all.
I'll check again whether I still ended up changing it though.

fabianburth: Yeah, so I debugged it again, and I believe I did not change the reconcile behaviour. 

I set up an RGD with a `readyWhen` condition evaluating to `false` and reconciled it with the controller on the KRO main branch version and this version.
Now, in both cases, we reach this code in resource.go
```
	if err := errors.Join(errs...); err != nil {
		return fmt.Errorf("apply results contain errors: %w", err)
	}
```
with empty errors. 

In my implementation of course only because I actively filter the `ErrWaitingForReadiness` before reaching this code. But I only did this to preserve the current reconcile behaviour. 
And after running into that point about 4 times, we never reach the breakpoint again (in both implementations). 

---

And the comment I put into the code there 
```
	// we don't want to keep reconciling with exponential backoff if the error
	// is waiting for readiness
	// we will notice the readiness through a watch event
```
was actually just my assumption at that point, and I thought I'd comment it, because I thought this is kind of special behaviour (although I recognise now, that assumption might have been wrong too).

pkg/controller/instance/resources.go
fabianburth: See comment below. TLDR; I think, the behaviour on main is equivalent.

pkg/controller/instance/resources.go
a-hilaly: I believe this is become more verbose yes. we only want to report on the last unresolved one.

pkg/controller/instance/resources.go
```diff
@@ -477,16 +482,16 @@ func (c *Controller) processExternalRefNode(
 
 	node.SetObserved([]*unstructured.Unstructured{actual})
 
-	ready, err := node.IsReady()
-	if err != nil {
+	if err := node.CheckReadiness(); err != nil {
+		if errors.Is(err, runtime.ErrWaitingForReadiness) {
+			state.SetWaitingForReadiness(fmt.Errorf("waiting for external reference %q: %w", id, err))
+			return nil
```
a-hilaly, line 502: Shall we return a requeue error in here? i believe this is a regression a introduced with the runtime refactor

pkg/controller/instance/resources.go
```diff
@@ -576,8 +581,16 @@ func (c *Controller) processApplyResults(
...
+	// we don't want to keep reconciling with exponential backoff if the error
+	// is waiting for readiness
+	// we will notice the readiness through a watch event
+	var errs []error
+	for _, state := range rcx.StateManager.NodeStates {
+		if state.Err != nil && !errors.Is(state.Err, runtime.ErrWaitingForReadiness) {
+			errs = append(errs, state.Err)
+		}
+	}
+	if err := errors.Join(errs...); err != nil {
```
a-hilaly, line 604: AFAIK we'll always have one node error, are there any cases where we can observed errors on multiple nodes?

pkg/runtime/node.go
a-hilaly: You’re right, and thanks for double-checking this. For some reason i thought we did requeue when waiting for readiness, but it looks like I did introduce that regression when reworking the runtime...

I took another pass and i don't see a real reconcile behaviour change. So i'm good with it as is.

pkg/runtime/node.go
```diff
@@ -492,30 +489,30 @@ func (n *Node) SetObserved(observed []*unstructured.Unstructured) {
...
 // Ignored nodes are treated as ready for dependency gating purposes.
-func (n *Node) IsReady() (bool, error) {
+func (n *Node) CheckReadiness() error {
```
a-hilaly, line 494: The reason why this function returned a boolean and an error was to distinguish from CEL errors and truly not-ready resources. I understand that now we're delegating this to error checks instead. But the idea was to prepare ground for propgataion control for collections in the future.

pkg/controller/instance/resources.go
a-hilaly: We can address this in a separate PR

pkg/controller/instance/resources.go
a-hilaly: cc @jakobmoellerdev

pkg/controller/instance/resources.go
fabianburth: I think `processNodes` is calling `processNode()` (which sets errors on the node state) in a loop that is not interrupted by error returns (see `lastUnresolvedErr` vs aggregation).
So, I'd expect that potentially observing errors on multiple (up to all) nodes is the current behaviour here.

pkg/controller/instance/resources.go
fabianburth: Changed back to `lastUnresolvedErr` :)

pkg/controller/instance/resources.go
fabianburth: Happy to do it, but are we sure we'd only want to do this for external nodes? (we are in the `processExternalRefNode` function)

Intuitively, after your comments, I'd have expected that we instead essentially remove the error filtering for loop  at the bottom of `processApplyResults()`.
```
	for _, state := range rcx.StateManager.NodeStates {
		if state.Err != nil && !errors.Is(state.Err, runtime.ErrWaitingForReadiness) {
			errs = append(errs, state.Err)
		}
	}
```

PR #1079, opened 2026-02-27
feat: external ref watches and external ref collections

jakobmoellerdev:
Introduce a layered watch architecture that enables instance-level watch
registration so changes to managed, external, and external collection
resources trigger immediate re-reconciliation instead of relying on the
requeue timer.

Scalar external ref — watches a single named resource:

```yaml
apiVersion: kro.run/v1alpha1                                                         
kind: ResourceGraphDefinition
metadata:                                                                            
  name: app-with-external-config                                                   
spec:
  schema:
    kind: App
    apiVersion: v1alpha1
    spec: {}
    status: {}
  resources:
    - id: extcm
      externalRef:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ext-config        # watched by name
    - id: managed
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: managed-config
        data:
          replicas: ${extcm.data.replicas}   # updates reactively when ext-config
changes
```

Collection external ref — watches all resources matching a label selector:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: app-with-external-collection
spec:
  schema:
    kind: AppCollection
    apiVersion: v1alpha1
    spec: {}
    status:
      configCount: ${string(size(extconfigs))}   # reactive count of matching resources
  resources:
    - id: extconfigs
      externalRef:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          selector:               # watched via label selector
            matchLabels:
              team: alpha
```

## flow
                                                                  
1. **Register** — instance reconciler calls `requestWatch(nodeID, GVR, name, namespace)` *before* reading the resource
2. **Index** — coordinator adds entry: `scalarIndex[GVR][namespace/name] → {instanceKey}`  
3. **Informer** — `WatchManager.EnsureWatch(GVR)` starts a shared informer if none exists (ref-counted)
4. **Event** — informer fires → `RouteEvent()` does O(1) map lookup → enqueues matching instances
5. **Cleanup** — `Done()` diffs current vs previous cycle; stale entries removed, orphaned informers stopped

## changes

- Add WatchManager (reference-counted informer lifecycle per GVR) and
  WatchCoordinator (reverse-index event routing from instance reconcilers)
- Replace the monolithic LazyInformer with the new coordinator, simplifying
  DynamicController to only manage parent watches and handler dispatch
- Register watches from instance reconcilers for managed resources,
  external refs, and external collections (label-selector-based)
- Fix external collection CEL typing: wrap ExternalCollection schemas as
  lists so size()/map()/filter() work in status expressions
- Add integration tests for external ref watch reactivity (scalar + collection)

This is a partial implementation of https://github.com/kubernetes-sigs/kro/pull/738

Most important diff:
- Namespace selectors are not supported, because fetching + ordering items from multiple namespaces into a collection is undefined (as in, would namespace objects a come before or after namespace objects b in the collection result)
- Label Selectors are part of the metadata section, and only for external refs, because they impact our watches only.
- List Kinds (as in `DeploymentList`) are not enforced on the API Spec because most users inherently assume that a `Deployment` with a selector would result in a list.

### DIFF
--- a/pkg/controller/instance/resources.go
+++ b/pkg/controller/instance/resources.go
@@ -500,16 +500,25 @@ func (c *Controller) processExternalRefNode(
 	// Register watch BEFORE reading the external resource.
 	requestWatch(rcx, id, node.Spec.Meta.GVR, desired.GetName(), desired.GetNamespace())
 
-	// External refs are read-only here: fetch and push into runtime for dependency/readiness.
-	actual, err := c.readExternalRefNode(rcx, node, desired)
+	// External refs are read-only: fetch and push into runtime for dependency/readiness.
+	ri := resourceClientFor(rcx, node.Spec.Meta, desired.GetNamespace())
+	actual, err := ri.Get(rcx.Ctx, desired.GetName(), metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			state.SetWaitingForReadiness(fmt.Errorf("waiting for external reference %q: %w", id, err))
 			return nil
 		}
-		state.SetError(err)
-		return err
+		state.SetError(fmt.Errorf("external ref get %s %s/%s: %w",
+			desired.GroupVersionKind().String(), desired.GetNamespace(), desired.GetName(), err))
+		return state.Err
 	}
+
+	rcx.Log.V(2).Info("External reference resolved",
+		"id", id,
+		"gvk", desired.GroupVersionKind().String(),
+		"namespace", actual.GetNamespace(),
+		"name", actual.GetName(),
+	)
 
 	node.SetObserved([]*unstructured.Unstructured{actual})
 
@@ -524,33 +533,6 @@ func (c *Controller) processExternalRefNode(
 	state.SetReady()
 
 	return nil
-}
-
-// readExternalRefNode fetches the referenced object for an external node.
-func (c *Controller) readExternalRefNode(
-	rcx *ReconcileContext,
-	node *runtime.Node,
-	desired *unstructured.Unstructured,
-) (*unstructured.Unstructured, error) {
-	nodeMeta := node.Spec.Meta
-	ri := resourceClientFor(rcx, nodeMeta, desired.GetNamespace())
-
-	name := desired.GetName()
-	obj, err := ri.Get(rcx.Ctx, name, metav1.GetOptions{})
-	if err != nil {
-		return nil, fmt.Errorf("external ref get %s %s/%s: %w",
-			desired.GroupVersionKind().String(), desired.GetNamespace(), name, err,
-		)
-	}
-
-	rcx.Log.V(2).Info("External reference resolved",
-		"id", node.Spec.Meta.ID,
-		"gvk", desired.GroupVersionKind().String(),
-		"namespace", obj.GetNamespace(),
-		"name", obj.GetName(),
-	)
-
-	return obj, nil
 }
 
 // processApplyResults updates runtime observations and node states from apply results.
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -25,6 +25,7 @@ import (
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/apiserver/pkg/cel/openapi"
 	"k8s.io/apiserver/pkg/cel/openapi/resolver"
@@ -320,45 +321,12 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
 // The selector (if any) is embedded directly in the template so that ParseSchemalessResource
 // can extract CEL expressions from the entire resource in a single pass.
 func (b *Builder) buildExternalRefResource(
-	externalRef *v1alpha1.ExternalRef) map[string]interface{} {
-	resourceObject := map[string]interface{}{}
-	resourceObject["apiVersion"] = externalRef.APIVersion
-	resourceObject["kind"] = externalRef.Kind
-	metadata := map[string]interface{}{}
-	if externalRef.Metadata.Name != "" {
-		metadata["name"] = externalRef.Metadata.Name
+	externalRef *v1alpha1.ExternalRef) (map[string]interface{}, error) {
+	result, err := runtime.DefaultUnstructuredConverter.ToUnstructured(externalRef)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert ExternalRef to unstructured: %w", err)
 	}
-	if externalRef.Metadata.Namespace != "" {
-		metadata["namespace"] = externalRef.Metadata.Namespace
-	}
-	if externalRef.Metadata.Selector != nil {
-		sel := map[string]interface{}{}
-		if len(externalRef.Metadata.Selector.MatchLabels) > 0 {
-			ml := map[string]interface{}{}
-			for k, v := range externalRef.Metadata.Selector.MatchLabels {
-				ml[k] = v
-			}
-			sel["matchLabels"] = ml
-		}
-		if len(externalRef.Metadata.Selector.MatchExpressions) > 0 {
-			exprs := make([]interface{}, len(externalRef.Metadata.Selector.MatchExpressions))
-			for i, req := range externalRef.Metadata.Selector.MatchExpressions {
-				vals := make([]interface{}, len(req.Values))
-				for j, v := range req.Values {
-					vals[j] = v
-				}
-				exprs[i] = map[string]interface{}{
-					"key":      req.Key,
-					"operator": string(req.Operator),
-					"values":   vals,
-				}
-			}
-			sel["matchExpressions"] = exprs
-		}
-		metadata["selector"] = sel
-	}
-	resourceObject["metadata"] = metadata
-	return resourceObject
+	return result, nil
 }
 
 // buildRGResource builds a node from the given resource definition.
@@ -383,7 +351,10 @@ func (b *Builder) buildRGResource(
 			return nil, nil, fmt.Errorf("failed to unmarshal resource %s: %w", rgResource.ID, err)
 		}
 	} else {
-		resourceObject = b.buildExternalRefResource(rgResource.ExternalRef)
+		var err error
+		if resourceObject, err = b.buildExternalRefResource(rgResource.ExternalRef); err != nil {
+			return nil, nil, fmt.Errorf("failed to build external ref resource %s: %w", rgResource.ID, err)
+		}
 	}
 
 	// 3. Check if it looks like a valid Kubernetes resource.


ellistarn:
w00t! Did you explore adding namespace selectors as proposed in https://github.com/kubernetes-sigs/kro/pull/738. Open minded to adding this as a second pass when KRO support both global and namespaced resources.

jakobmoellerdev:
In my understanding KRO already supports cluster and namespaced resources. Namespace selectors are not supported here, because fetching + ordering items from multiple namespaces into a collection is undefined (as in, would namespace objects a come before or after namespace objects b in the collection result, is it even deterministic?)

ellistarn:
> In my understanding KRO already supports cluster and namespaced resources. Namespace selectors are not supported here, because fetching + ordering items from multiple namespaces into a collection is undefined (as in, would namespace objects a come before or after namespace objects b in the collection result, is it even deterministic?)

Interesting -- how do we order collection objects in a namespace? I could imagine ordering the namespaces alphanumerically by ID. Perhaps we can take the same approach in both cases?

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -112,17 +112,25 @@ type CRDMetadata struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="(has(self.name) && size(self.name) > 0) || has(self.selector)",message="exactly one of name or selector must be provided"
+// +kubebuilder:validation:XValidation:rule="!((has(self.name) && size(self.name) > 0) && has(self.selector))",message="name and selector are mutually exclusive"
```
a-hilaly: How about `(has(self.name) && !has(self.selector)) || (!has(self.name) && has(self.selector))` and MinLenght=1 on both name and namespace?

pkg/controller/instance/context.go
```diff
@@ -43,6 +44,10 @@ type ReconcileContext struct {
 
 	Mark         *ConditionsMarker
 	StateManager *StateManager
+
+	// Watcher is the per-instance watch handle from the coordinator.
+	// May be nil if the coordinator is not configured.
```
a-hilaly: Is there a reason the coordinator might not be configured? i'm still at the beginning of this review but this strikes me as it should never be nil

pkg/metadata/labels.go
```diff
@@ -206,6 +206,18 @@ func NewCollectionItemLabeler(nodeID string, index, size int) GenericLabeler {
...
+	result := make(map[string]string, len(l.Labels()))
+	for k, v := range l.Labels() {
+		result[k] = v
+	}
+	for _, key := range keys {
+		delete(result, key)
+	}
+	return GenericLabeler(result)
+}
+
```
a-hilaly: I guess left over from previous branch, needs a rebase.

pkg/runtime/node.go
```diff
@@ -139,6 +140,10 @@ func (n *Node) GetDesired() ([]*unstructured.Unstructured, error) {
 		// External refs resolve like resources (for name/namespace CEL),
 		// but the caller reads instead of applies.
 		result, err = n.hardResolveSingleResource(n.templateVars)
+	case graph.NodeTypeExternalCollection:
+		// External collections don't produce desired objects; the caller
+		// uses the selector to LIST resources. Return empty slice.
+		result = []*unstructured.Unstructured{}
```
a-hilaly: I think we should allow users to dynamically set label selectors using CEL.

This would be handled with the case above.

pkg/runtime/node.go
```diff
@@ -504,10 +516,12 @@ func (n *Node) IsReady() (bool, error) {
 		return true, nil
 	}
 
-	if n.Spec.Meta.Type == graph.NodeTypeCollection {
+	switch n.Spec.Meta.Type {
+	case graph.NodeTypeCollection, graph.NodeTypeExternalCollection:
 		return n.isCollectionReady()
+	default:
+		return n.isSingleResourceReady()
 	}
```
a-hilaly: Probably a miss on my end, we need to panic on instance type?

test/integration/suites/core/externalref_watch_test.go
```diff
@@ -0,0 +1,270 @@
...
+			err = env.Client.Get(ctx, types.NamespacedName{
+				Name:      "managed-config",
+				Namespace: namespace,
+			}, managedCM)
+			g.Expect(err).ToNot(HaveOccurred())
+			g.Expect(managedCM.Data).To(HaveKeyWithValue("replicas", "1"))
+		}, 20*time.Second, time.Second).WithContext(ctx).Should(Succeed())
+
+		By("sleeping to let the watch settle")
+		time.Sleep(5 * time.Second)
```
a-hilaly: Is the sleep needed?

a-hilaly: I'm assuming that if the instance is Active the watch must be settled ?

pkg/dynamiccontroller/internal/lazy_informer.go
a-hilaly, line 1: Good call :)

pkg/dynamiccontroller/watch.go
```diff
@@ -0,0 +1,211 @@
...
+			if e := toEvent(obj, EventAdd); e != nil {
+				onEvent(*e)
+			}
+		},
+		UpdateFunc: func(_, newObj interface{}) {
+			if e := toEvent(newObj, EventUpdate); e != nil {
+				onEvent(*e)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
```
a-hilaly, line 234: I think we never did this before, but do we want to handle tombstones?

pkg/dynamiccontroller/watch.go
```diff
@@ -0,0 +1,211 @@
...
+		AddFunc: func(obj interface{}) {
+			if e := toEvent(obj, EventAdd); e != nil {
+				onEvent(*e)
+			}
+		},
+		UpdateFunc: func(_, newObj interface{}) {
+			if e := toEvent(newObj, EventUpdate); e != nil {
+				onEvent(*e)
+			}
+		},
```
a-hilaly: Do we want to handle the case where an object changes from matching to non matching? that would mean readong oldObj

pkg/dynamiccontroller/watch.go
```diff
@@ -0,0 +1,211 @@
...
+func (m *WatchManager) ActiveWatchCount() int {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return len(m.watches)
+}
+
+// WatchCount returns the number of active watches. Alias for metrics.
+func (m *WatchManager) WatchCount() int {
+	return m.ActiveWatchCount()
+}
```
a-hilaly: Curious, why do we need this alias? naming?

pkg/dynamiccontroller/watch.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if _, ok := m.watches[gvr]; ok {
+		return
+	}
+
+	w := m.newWatch(gvr)
+	m.watches[gvr] = w
+
+	go w.informer.Run(w.stopCh)
+	m.log.V(1).Info("Informer started", "gvr", gvr)
```
a-hilaly, line 95: I'm wondering if we should wait for a sync and surface the error here?

pkg/dynamiccontroller/watch_test.go
a-hilaly, line 1: Can we keep close to 100% test coverage like we did for lazy informer

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -329,317 +311,167 @@ func (dc *DynamicController) syncFunc(ctx context.Context, oi ObjectIdentifiers,
...
-	}
-	oldMeta, err := meta.Accessor(oldObj)
-	if err != nil {
-		dc.log.Error(err, "failed to access old object meta")
-		return
-	}
-	if newMeta.GetGeneration() == oldMeta.GetGeneration() {
-		dc.log.V(2).Info("Skipping update due to unchanged generation",
-			"name", newMeta.GetName(), "namespace", newMeta.GetNamespace(), "generation", newMeta.GetGeneration())
-		return
```
a-hilaly: I believe we do still need this?

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -329,317 +311,167 @@ func (dc *DynamicController) syncFunc(ctx context.Context, oi ObjectIdentifiers,
...
+			UpdateFunc: func(_, newObj interface{}) {
+				dc.enqueueFromInformer(parent, newObj, EventUpdate)
+			},
+			DeleteFunc: func(obj interface{}) {
+				dc.enqueueFromInformer(parent, obj, EventDelete)
+			},
+		}
+		reg, err := inf.AddEventHandler(parentHandler)
+		if err != nil {
+			dc.handlers.Delete(parent)
```
a-hilaly, line 383: do we wanna also shutdown the informer?

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+			c.log.V(1).Info("Stopped orphaned child watch", "gvr", gvr)
+		}
+	}
+}
+
+func (c *WatchCoordinator) removeCollectionIndexLocked(key instanceKey, req *WatchRequest) {
+	entries := c.collectionIndex[req.GVR]
+	filtered := entries[:0]
+	for _, e := range entries {
+		if e.key == key && e.selector.String() == req.Selector.String() && e.namespace == req.Namespace {
```
a-hilaly: to be confirmed but I think matching on `selector.String()` can miss match `a=b,c=d` and `c=d,a=b`

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+	// Collection matches (selector scan).
+	for _, entry := range c.collectionIndex[event.GVR] {
+		if entry.namespace != "" && event.Namespace != entry.namespace {
+			continue
+		}
+		if entry.selector.Matches(labels.Set(event.Labels)) {
+			c.enqueue(entry.key.parentGVR, entry.key.instance)
+			matched = true
+		}
+	}
```
a-hilaly, line 321: I believe we should dedup here. An instance can be requeued twice if it matches both a scalar and collection entry

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+func (c *WatchCoordinator) ForInstance(parentGVR schema.GroupVersionResource, instance types.NamespacedName) InstanceWatcher {
+	return &instanceWatcher{
+		coordinator: c,
+		parentGVR:   parentGVR,
+		instance:    instance,
+	}
+}
+
+// addWatch registers a watch request for the given instance.
+func (c *WatchCoordinator) addWatch(key instanceKey, req WatchRequest) error {
```
a-hilaly, line 139: Still wrapping my head around this but it looks like Collection requests are not idempotant across reconciles. we only compare GVR/Name/Namespace here and selectors are ignored + getting their index appended each time.

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+		c.addCollectionIndexLocked(key, req)
+	} else {
+		c.addScalarIndexLocked(key, req)
+	}
+
+	return nil
+}
+
+// doneInstance finalizes the reconciliation cycle for an instance.
+func (c *WatchCoordinator) doneInstance(key instanceKey) {
```
a-hilaly, line 188: Same comment here

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+	// Call Watch() BEFORE operating on the resource to avoid event gaps.
+	Watch(req WatchRequest) error
+
+	// Done signals that all Watch() calls for this reconciliation cycle
+	// are complete. Any watch requests from the previous cycle that were
+	// NOT re-requested are automatically cleaned up.
+	//
+	// If reconciliation fails before Done(), previous requests stay active.
+	Done()
+}
```
a-hilaly, line 46: good design, current/previous + Done() is a solid model.

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+	// NodeID is the graph node ID (for debugging/metrics).
+	NodeID string
+	// GVR is the GroupVersionResource to watch.
+	GVR schema.GroupVersionResource
+	// Name is the specific resource name (scalar watches).
+	Name string
+	// Namespace is the resource namespace. Empty for cluster-scoped resources.
+	Namespace string
+	// Selector is a label selector for collection watches. nil means scalar watch.
+	Selector labels.Selector
```
a-hilaly, line 62: Selectors do also have `matchExpressions` we're not handling that as well.

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+		if old.GVR != req.GVR || old.Name != req.Name || old.Namespace != req.Namespace {
+			c.removeRequestFromIndexesLocked(key, old)
+		}
+	}
+
+	// Add to current cycle.
+	state.current[req.NodeID] = &req
+
+	// Ensure an informer is running for this GVR (non-blocking).
+	c.watches.EnsureWatch(req.GVR)
```
a-hilaly: Can we call EnsureWatch outside the lock?

pkg/dynamiccontroller/coordinator.go
a-hilaly, line 188: Same comment we probably want to stop the watches outside the lock, to let other instances register their watches

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	matched := false
+
+	// Scalar matches (O(1) per GVR+name).
+	if byName, ok := c.scalarIndex[event.GVR]; ok {
+		key := types.NamespacedName{Name: event.Name, Namespace: event.Namespace}
+		for instKey := range byName[key] {
+			c.enqueue(instKey.parentGVR, instKey.instance)
```
a-hilaly: Same comment enqueue can move outside the RLock

pkg/dynamiccontroller/coordinator.go
```diff
@@ -0,0 +1,427 @@
...
+
+	// Collection matches (selector scan).
+	for _, entry := range c.collectionIndex[event.GVR] {
+		if entry.namespace != "" && event.Namespace != entry.namespace {
+			continue
+		}
+		if entry.selector.Matches(labels.Set(event.Labels)) {
+			c.enqueue(entry.key.parentGVR, entry.key.instance)
+			matched = true
+		}
```
a-hilaly, line 320: what if the oldObject loses the labels, do we want to enqueue here?

pkg/controller/instance/resources.go
```diff
@@ -646,3 +662,98 @@ func setStateFromReadiness(node *runtime.Node, state *NodeState) {
...
+		state.SetError(fmt.Errorf("invalid label selector for %s: %w", id, err))
+		return state.Err
+	}
+
+	// Register collection watch with the coordinator.
+	requestCollectionWatch(rcx, id, nodeMeta.GVR, rcx.Instance.GetNamespace(), selector)
+
+	// Determine namespace for LIST.
+	ns := rcx.Instance.GetNamespace()
+
```
a-hilaly: External collections can also be scoped to specific namespace.

pkg/controller/instance/resources.go
```diff
@@ -646,3 +662,98 @@ func setStateFromReadiness(node *runtime.Node, state *NodeState) {
...
+	// LIST external resources matching the selector.
+	var ri = rcx.Client.Resource(nodeMeta.GVR)
+	list, err := ri.Namespace(ns).List(rcx.Ctx, metav1.ListOptions{
+		LabelSelector: selector.String(),
+	})
```
a-hilaly: External collections can be cluster scoped too

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -112,17 +112,25 @@ type CRDMetadata struct {
...
 	// If empty, the instance's namespace will be used.
 	//
 	// +kubebuilder:validation:Optional
 	Namespace string `json:"namespace,omitempty"`
+	// Selector is a label selector for collection external references.
+	// When set, all resources matching the selector are included.
+	// Mutually exclusive with Name.
+	//
+	// +kubebuilder:validation:Optional
+	Selector *metav1.LabelSelector `json:"selector,omitempty"`
```
a-hilaly, line 133: LabelSelector struct also has `MatchExpressions`, but i'm fine not supporting it to start with.

pkg/dynamiccontroller/watch.go
jakobmoellerdev: I will revisit this

pkg/controller/instance/context.go
jakobmoellerdev: only during testing, added a comment on that

pkg/runtime/node.go
jakobmoellerdev: this is now possible

pkg/dynamiccontroller/watch.go
jakobmoellerdev: added a handle for this

pkg/dynamiccontroller/watch.go
jakobmoellerdev: done

pkg/dynamiccontroller/watch.go
jakobmoellerdev: removed

pkg/dynamiccontroller/watch_test.go
jakobmoellerdev: Im currently at 90% package coverage, will need some more work to get to 100

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: added back in

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: now using match from label selector directly

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: expression handling is now included everywhere I believe

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: updated lock management.

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: updated lock management.

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: why is that?

pkg/controller/instance/resources.go
jakobmoellerdev: updated that. does that solve your concern?

pkg/controller/instance/resources.go
jakobmoellerdev: updated that. does that solve your concern?

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: added matchexpression support to the best of my ability :D

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: added a retrigger on old labels

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: I will think about this, unsure. only considered this for deregister

jakobmoellerdev: if the event handler cannot be added and we delete the handler, my assumption is the informer is anyhow broken and not started

jakobmoellerdev:
> > In my understanding KRO already supports cluster and namespaced resources. Namespace selectors are not supported here, because fetching + ordering items from multiple namespaces into a collection is undefined (as in, would namespace objects a come before or after namespace objects b in the collection result, is it even deterministic?)
> 
> Interesting -- how do we order collection objects in a namespace? I could imagine ordering the namespaces alphanumerically by ID. Perhaps we can take the same approach in both cases?

My take is that we will rely on the CEL macro sortBy for sorting and by default pass through the LIST call as is. The reason for that is that it is significantly faster for operations such as aggregations that dont require a sort. (think of something like count all configmaps in a namespace)

ellistarn:
> > > In my understanding KRO already supports cluster and namespaced resources. Namespace selectors are not supported here, because fetching + ordering items from multiple namespaces into a collection is undefined (as in, would namespace objects a come before or after namespace objects b in the collection result, is it even deterministic?)
> > 
> > 
> > Interesting -- how do we order collection objects in a namespace? I could imagine ordering the namespaces alphanumerically by ID. Perhaps we can take the same approach in both cases?
> 
> My take is that we will rely on the CEL macro sortBy for sorting and by default pass through the LIST call as is. The reason for that is that it is significantly faster for operations such as aggregations that dont require a sort. (think of something like count all configmaps in a namespace)

 By that same argument, could you just do the same for ordering the namespaces on a namesapce selector?

---

edit:

Looking into this a bit more, WATCH doesn't support namespace selector -- clients are forced to do it client side. At this point, I think we're probably better off just using CEL to filter as needed, since it doesn't give us any perf benefit to do partial watches.

pkg/controller/instance/context.go
a-hilaly: Are the nil checks worth it in our implementation to accommodate testing? how instantiating a dead or fake coordinator instead?

pkg/dynamiccontroller/dynamic_controller.go
a-hilaly: I'm just worried about leaked watches here. Shutting down the informer doesn't hurt IMO

pkg/dynamiccontroller/watch.go
jakobmoellerdev: I have now added a wait on sync

pkg/dynamiccontroller/watch_test.go
jakobmoellerdev: 100% coverage on watch now present

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: sure, killed the watch here to burn the world

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: TestCollectionWatch_SelectorChange shows that this works as intended in my view. The doneInstance comparison skipping selector is intentionally safe because addCollectionIndexLocked already replaced the entry during the current cycle's Watch() calls.

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: The matched map is a map[instanceKey]struct{} — it's a set. Both scalar and 
  collection matches write to the same set, so duplicates are impossible. an event matching both a scalar watch and a collection selector for the same instance produces exactly 1 enqueue.

jakobmoellerdev:
In general not opposed to implementing a namespace selector logic later, just want to keep the PR reviewable

pkg/controller/instance/context.go
jakobmoellerdev: Updated

pkg/graph/builder.go
```diff
@@ -457,15 +465,67 @@ func (b *Builder) buildRGResource(
...
+		IncludeWhen:  includeWhen,
+		ReadyWhen:    readyWhen,
+		ForEach:      forEachDimensions,
 	}
 	return node, resourceSchema, nil
 }
 
+// parseSelectorExpressions extracts CEL expressions from matchLabels values.
+// Static values (no ${...}) are skipped. Returns nil if all values are static.
+func parseSelectorExpressions(sel *metav1.LabelSelector) ([]*variable.ResourceField, error) {
```
a-hilaly: do we really need this? feels like we should just extend `buildExternalRefResource` and use Schemaless parser?

pkg/graph/builder.go
```diff
@@ -626,6 +693,35 @@ func extractForEachDependencies(
 	return allDeps, nil
 }
 
+// extractSelectorDependencies extracts dependencies from CEL expressions
+// in SelectorVars (matchLabels values containing ${...}).
+func extractSelectorDependencies(
```
a-hilaly: this is not needed if we do the above

pkg/graph/node.go
```diff
@@ -125,6 +125,11 @@ type Node struct {
 	// Variables holds the CEL expression fields found in the template.
 	Variables []*variable.ResourceField
 
+	// SelectorVars holds CEL expression fields extracted from
+	// Selector.MatchLabels values (ExternalCollection nodes only).
+	// When non-empty, the selector must be evaluated at runtime.
+	SelectorVars []*variable.ResourceField
```
a-hilaly: scalar ExternalRefs just set the template, we can do the same here as well

pkg/runtime/node.go
```diff
@@ -743,8 +749,112 @@ func (n *Node) contextDependencyIDs(iterCtx map[string]any) (singles, collection
...
 
-// GetResolvedSelector returns the label selector from the node's graph spec.
-// This is only valid for NodeTypeExternalCollection nodes.
-func (n *Node) GetResolvedSelector() *metav1.LabelSelector {
-	return n.Spec.Meta.Selector
+// GetResolvedSelector returns the resolved label selector for this
+// ExternalCollection node. If the selector contains CEL expressions,
+// they are evaluated against the node's dependencies. Returns nil if
+// the node has no selector. Returns an error if evaluation fails.
+func (n *Node) GetResolvedSelector() (*metav1.LabelSelector, error) {
```
a-hilaly: This becomes just resolve template if we store the whole thing in .Template

pkg/controller/instance/resources.go
```diff
@@ -669,3 +685,106 @@ func setStateFromReadiness(node *runtime.Node, state *NodeState) {
...
+	}
+
+	// Determine namespace for LIST and watch.
+	// Cluster-scoped resources (e.g. Namespaces, ClusterRoles) use empty
+	// namespace so the LIST is not scoped to a single namespace.
+	ns := rcx.Instance.GetNamespace()
+	if !nodeMeta.Namespaced {
+		ns = ""
+	}
+
```
a-hilaly: What if externalRef is:
```yaml
externalRef:
    metadata:
       namespace: ${....}
       selectors:
            matchLabels: ${....}
```

It is selecting a resources that match certain labels in a "user specified namespace" that is different from the instance namespace

pkg/controller/instance/resources.go
```diff
@@ -669,3 +685,106 @@ func setStateFromReadiness(node *runtime.Node, state *NodeState) {
...
+	labelSelector, err := node.GetResolvedSelector()
+	if err != nil {
+		state.SetError(fmt.Errorf("failed to resolve selector for %s: %w", id, err))
+		return state.Err
+	}
+	if labelSelector == nil {
+		state.SetSkipped()
+		return nil
+	}
```
a-hilaly: Why is it skipped? doesn't it just mean select everything?

pkg/dynamiccontroller/coordinator.go
a-hilaly: freeing the lock for other events

jakobmoellerdev: its a read lock so it can be held by N without there being a write lock. wouldnt cause contention here

pkg/controller/instance/resources.go
jakobmoellerdev: what do you mean select anything. you mean like a non filtered list?
sure can change!

pkg/graph/builder.go
jakobmoellerdev: removed this. thanks for the pointer

pkg/controller/instance/resources.go
jakobmoellerdev: youre completely right. made it so it use the node namespace first if set, if not the instance namespace

pkg/graph/node.go
```diff
@@ -91,6 +97,9 @@ type NodeMeta struct {
 	Namespaced bool
 	// Dependencies lists the IDs of nodes this node depends on.
 	Dependencies []string
+	// Selector is the label selector for ExternalCollection nodes.
+	// nil for all other node types.
+	Selector *metav1.LabelSelector
```
a-hilaly: Do we still need this?

jakobmoellerdev: no we dont! good catch

pkg/dynamiccontroller/dynamic_controller.go
a-hilaly: This wasn't added back @jakobmoellerdev

PR #1080, opened 2026-03-02
fix(runtime): schema-aware CEL value conversion

jakobmoellerdev:
Before this change, observed Kubernetes resources were passed to the CEL evaluation context as raw map[string]interface{} values. CEL treated all fields as dyn, which meant type mismatches between compile-time declarations (derived from OpenAPI schemas) and runtime values. For example, a Secret's data field declared as map(string,bytes) at compile time would arrive as map(string,string) at runtime, causing CEL evaluation errors that could not get circumvented (for example using base64 functions on configmap data was a pain)

This commit wraps observed resources with OpenAPI schema information using Kubernetes' own UnstructuredToVal conversion (adapted from k8s.io/apiserver/pkg/cel/common/values.go) in a more lenient form. This ensures CEL runtime values match their compile-time types — Secret data evaluates as bytes, timestamps as proper CEL timestamps, integers as ints rather than floats, etc.

Key changes:

- Add valunstructured.go: schema-aware CEL value types (unstructuredMap, unstructuredList) that provide typed field access during evaluation while preserving the original raw data for output. This is basically a port from kubernetes upstream, but with more lenient availability on allow unknown fields.

- Update GoNativeType to return original raw data for schema-aware maps via Value() instead of reconstructing through Iterator/Get, which drops fields without schemas and round-trips types unnecessarily (string → Timestamp → string). Computed timestamps and durations are converted to JSON-safe strings since all CEL results ultimately end up in Kubernetes unstructured objects where DeepCopyJSONValue only handles JSON primitives. (we may want to move GoNativeType to be closer to places needing it and use ref.Val by default to avoid this)

- Wire resourceSchema into runtime nodes so buildContext can wrap dependencies with schema-aware types.

### DIFF
--- a/pkg/cel/conversion/conversions.go
+++ b/pkg/cel/conversion/conversions.go
@@ -24,6 +24,8 @@ import (
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // ErrUnsupportedType is returned when the type is not supported.
@@ -48,23 +50,12 @@ func GoNativeType(v ref.Val) (interface{}, error) {
 	case types.BytesType:
 		return v.Value().([]byte), nil
 	case types.DurationType:
-		// Convert to string for JSON-safe unstructured objects.
-		return v.Value().(time.Duration).String(), nil
+		return v1.Duration{Duration: v.Value().(time.Duration)}.ToUnstructured(), nil
 	case types.TimestampType:
-		// Convert to RFC3339 string for JSON-safe unstructured objects.
-		// CEL timestamp arithmetic happens within evaluation; by the time
-		// GoNativeType is called, the result needs to be JSON-compatible.
-		return v.Value().(time.Time).Format(time.RFC3339), nil
+		return v1.Time{Time: v.Value().(time.Time)}.ToUnstructured(), nil
 	case types.ListType:
 		return convertList(v)
 	case types.MapType:
-		// If the underlying value is already a JSON-safe string-keyed map
-		// (e.g., from schema-aware unstructuredMap wrapping Kubernetes data),
-		// use it directly. This avoids lossy reconstruction through Iterator/Get
-		// which can drop fields without schemas and introduce type mismatches.
-		if m, ok := v.Value().(map[string]interface{}); ok {
-			return m, nil
-		}
 		return convertMap(v)
 	case types.OptionalType:
 		opt := v.(*types.Optional)
@@ -103,8 +94,18 @@ func convertList(v ref.Val) (interface{}, error) {
 func convertMap(v ref.Val) (interface{}, error) {
 	mapper, ok := v.(traits.Mapper)
 	if !ok {
-		return v.ConvertToNative(reflect.TypeOf(map[string]interface{}{}))
+		return v.ConvertToNative(reflect.TypeOf(map[string]any{}))
 	}
+
+	// Fast path: if the underlying value is already a raw Go map, return it
+	// directly. This matches ConvertToNative behavior and avoids having to iterate
+	// over already correctly present maps.
+	// Raw map values (from Kubernetes unstructured data) are already in the
+	// correct Go form, so no recursive conversion is needed.
+	if rawMap, ok := v.Value().(map[string]interface{}); ok {
+		return runtime.DeepCopyJSON(rawMap), nil
+	}
+
 	result := make(map[string]interface{})
 	it := mapper.Iterator()
 	for it.HasNext() == types.True {
--- a/pkg/cel/unstructured/valunstructured.go
+++ b/pkg/cel/unstructured/valunstructured.go
@@ -14,7 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package runtime
+// Package unstructured provides schema-aware conversion of Kubernetes
+// unstructured data to CEL ref.Val values. It is a port of the upstream
+// k8s.io/apiserver/pkg/cel/common.UnstructuredToVal with adjustments for kro.
+package unstructured
 
 import (
 	"fmt"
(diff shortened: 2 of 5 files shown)


test/integration/suites/core/schema_aware_cel_test.go
```diff
@@ -0,0 +1,138 @@
...
+		rgd := generator.NewResourceGraphDefinition(
+			"test-schema-cel",
+			generator.WithSchema(
+				"TestSchemaCel",
+				"v1alpha1",
+				map[string]any{
+					"clientId": "string",
+				},
+				map[string]any{
+					"decodedClientId": `${string(secret.data.clientId)}`,
```
a-hilaly, line 62: shouldn't this be using `base64.decode`?

a-hilaly: nvm i see that the conversation happens natively now.

pkg/cel/unstructured/valunstructured.go
a-hilaly, line 1: This file strikes me as belonging to `pkg/cel/**`

pkg/cel/unstructured/valunstructured.go
```diff
@@ -0,0 +1,676 @@
...
+		case string:
+			return types.String(v)
+		case int:
+			return types.Int(v)
+		case int32:
+			return types.Int(v)
+		case int64:
+			return types.Int(v)
+		}
+		return types.NewErr("invalid data, expected XIntOrString value to be either a string or integer")
```
a-hilaly, line 59: do we really need to do this kind validation in runtime?

a-hilaly: if feels like we're never supposed to run into these, they are guaranteed to be ccorrect from the api serveR?

pkg/cel/unstructured/valunstructured.go
a-hilaly, line 1: Do we want high test coverage for this file?

pkg/cel/unstructured/valunstructured.go
```diff
@@ -0,0 +1,676 @@
...
+// UnstructuredToVal converts a Kubernetes unstructured data element to a CEL Val.
+// The root schema of custom resource schemas is expected to contain type meta and object meta schemas.
+// If Embedded resources do not contain type meta and object meta schemas, they will be added automatically.
+// nolint:gocyclo // this should stay close to upstream [common.UnstructuredToVal]
+func UnstructuredToVal(unstructured interface{}, schema common.Schema) ref.Val {
+	if unstructured == nil {
+		if schema.Nullable() {
+			return types.NullValue
+		}
+		return types.NewErr("invalid data, got null for schema with nullable=false")
```
a-hilaly, line 46: Before we used to treat CEL expressions that resolve to `nil` as data pending. It kinda looks like a breaking change to me, e.g
```yaml
readyWhen:
- ${job.status.completionTime != null}
```
with job
```yaml
status:
  completionTime: null
```

before this change, the readyWhen expression would resolve to false (not an error). Now this would error out with `invalid data, got null for schema with nullable=false`

pkg/graph/graph.go
```diff
@@ -41,4 +42,9 @@ type Graph struct {
 
 	// CRD is the generated CustomResourceDefinition for the instance.
 	CRD *extv1.CustomResourceDefinition
+
+	// ResourceSchemas maps node ID to the OpenAPI schema for that resource.
+	// Includes resource schemas (keyed by resource ID) and the instance schema
+	// (keyed by InstanceNodeID). Used at runtime for schema-aware CEL value conversion.
+	ResourceSchemas map[string]*spec.Schema
```
a-hilaly, line 49: I wonder how this impacts our performance. Different RGDs built in different times might refer to identical schemas (that won't be GC-ed)

a-hilaly: We might wanna rethink our caching strategies (LRU and friends) in the near future.

pkg/runtime/node.go
```diff
@@ -612,6 +618,10 @@ func (n *Node) evaluateForEach() ([]map[string]any, error) {
 // buildContext builds the CEL activation context from node dependencies.
 // If only is provided, only those dependency IDs are included in the context.
 // If only is empty/nil, all dependencies are included.
+//
+// When a dependency has a resourceSchema, its observed objects are wrapped using
+// Kubernetes' UnstructuredToVal for schema-aware type conversion. This ensures
+// CEL runtime values match their compile-time types (e.g., Secret data as bytes).
 func (n *Node) buildContext(only ...string) map[string]any {
```
a-hilaly, line 645: Now that this is becoming expensive we might want to call `filterContext` before `buildContext`

a-hilaly (commented):
Thank you @jakobmoellerdev ! I left a few questions and comments in line.

pkg/graph/graph.go
jakobmoellerdev: Whats your suggestion here? I will need to pass schemas to runtime otherwise we will not have a chance to have proper value instances in CEL.

pkg/cel/unstructured/valunstructured.go
jakobmoellerdev: Ack, still have this in draft as the port is in progress

jakobmoellerdev: I think so. Just unsure if I find a better way for a port yet.

pkg/cel/unstructured/valunstructured.go
jakobmoellerdev: Hmmm I think you might be right, I will double check this

pkg/cel/unstructured/valunstructured.go
jakobmoellerdev: I didnt consider it a validation and more of a case that should never happen tbh. its based on the return parsed in unstructured

pkg/runtime/node.go
jakobmoellerdev: I have included various context optimizations on this now.

pkg/cel/unstructured/valunstructured.go
jakobmoellerdev: added some more coverage

pkg/graph/graph.go
jakobmoellerdev: My suggestion here is that we can provide a schemaresolver interface with a second pr. this one is getting out of hand already

pkg/cel/unstructured/valunstructured.go
jakobmoellerdev: I think this is fine. Null values on object properties are treated as absent, not as null

`job.status.completionTime → "no such key" error → classified as data pending →
  returns (false, nil)`

There is now a difference between map data and object data absence. field == null checks are impossible in upstream:
  - Before: null == null → true → ready
  - After: "no such key" → data pending → never ready (silent infinite retry)

That being said, we can choose to diverge in downstream from how kubernetes evaluates the expression. It all comes down to our "absent vs null" problem again

pkg/cel/conversion/conversions.go
```diff
@@ -89,12 +93,25 @@ func convertList(v ref.Val) (interface{}, error) {
...
 	}
+
+	// Fast path: if the underlying value is already a raw Go map, return it
+	// directly. This matches ConvertToNative behavior and avoids having to iterate
+	// over already correctly present maps.
+	// Raw map values (from Kubernetes unstructured data) are already in the
+	// correct Go form, so no recursive conversion is needed.
+	if rawMap, ok := v.Value().(map[string]interface{}); ok {
+		return rawMap, nil
+	}
```
a-hilaly, line 107: Now that context values are schema wrapped, wouldn't it cause mutations of the original objects unintentionally? thinking we need `deepCopyMap(rawMap)`

jakobmoellerdev: I think we were currently not running into this but have decided to add a deep copy anyhow for separation of model safety in the heap.

a-hilaly:
shipit

PR #1083, opened 2026-03-03
Document escaping Bash '${VAR}' syntax for CEL

mwarkentin:
Added section on escaping Bash variable syntax in CEL expressions.

### DIFF
--- a/website/docs/docs/concepts/rgd/03-cel-expressions.md
+++ b/website/docs/docs/concepts/rgd/03-cel-expressions.md
@@ -144,9 +144,9 @@ readyWhen:
 - The `-` is what removes the final newline.
 :::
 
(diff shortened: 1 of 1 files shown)


website/docs/docs/concepts/rgd/03-cel-expressions.md
```diff
@@ -144,6 +144,40 @@ readyWhen:
 - The `-` is what removes the final newline.
 :::
 
+### Escaping Bash `${VAR}` Syntax
```
a-hilaly: This applies to any shell (sh, zsh ...) and tools like envsubst and kro :). Might be worth generalizing the title to something like: `"Escaping ${...} Syntax"` ?

PR #1085, opened 2026-03-04
Implementation of Graph Revisions (KREP-013)

a-hilaly:
When an RGD spec changed, kro hot-swapped the compiled graph with no durable
record of what ran before. That made rollback and debugging harder, and left no
stable revision identity for controlled migration features. This commit adds
the base revision layer described in KREP-013.

It introduces a cluster-scoped GraphRevision snapshot and a dedicated
GraphRevision controller that compiles each revision once and stores results in
a shared in-memory registry. RGD reconcile now issues revisions from a
canonical spec hash, tracks latest observed and active revisions, persists a
monotonic lastIssuedRevision, and applies bounded retention GC. Instance
controllers now resolve the latest active compiled graph from the registry.

This also wires the feature end to end: API/CRD types, status fields, hashing,
registry semantics (Pending/Active/Failed), controller flags, Helm config,
RBAC, and integration/unit coverage. Issued GraphRevisions now carry
ownerReferences to their source ResourceGraphDefinition so lifecycle cleanup is
explicit and consistent with the design.

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller_reconcile.go
+++ b/pkg/controller/resourcegraphdefinition/controller_reconcile.go
@@ -483,12 +483,15 @@ func (r *ResourceGraphDefinitionReconciler) getLatestGraphRevisionView(
 	}, true
 }
 
-// createGraphRevision creates a new GraphRevision object and seeds the
-// in-memory registry entry as Pending.
+// createGraphRevision creates a new GraphRevision object.
 //
 // GraphRevisions are immutable snapshots of the RGD spec. The revision number
 // is monotonically increasing per RGD name, and the spec hash is stored
 // for duplicate-detection on subsequent reconciles.
+//
+// Runtime revision state is owned by the GraphRevision controller. The RGD
+// controller only creates the API object and waits for the GR controller to
+// warm the shared in-memory registry from there.
 func (r *ResourceGraphDefinitionReconciler) createGraphRevision(
 	ctx context.Context,
 	rgd *v1alpha1.ResourceGraphDefinition,
@@ -530,12 +533,6 @@ func (r *ResourceGraphDefinitionReconciler) createGraphRevision(
 	if err := r.Create(ctx, graphRevision); err != nil {
 		return nil, fmt.Errorf("creating graph revision %q: %w", graphRevision.Name, err)
 	}
-	r.revisionsRegistry.Put(revisions.Entry{
-		RGDName:  rgd.Name,
-		Revision: revision,
-		SpecHash: specHash,
-		State:    revisions.RevisionStatePending,
-	})
 	return graphRevision, nil
 }
 
--- a/pkg/controller/resourcegraphdefinition/controller_reconcile_test.go
+++ b/pkg/controller/resourcegraphdefinition/controller_reconcile_test.go
@@ -35,6 +35,7 @@ import (
 
 	internalv1alpha1 "github.com/kubernetes-sigs/kro/api/internal.kro.run/v1alpha1"
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
+	"github.com/kubernetes-sigs/kro/pkg/graph"
 	graphhash "github.com/kubernetes-sigs/kro/pkg/graph/hash"
 	"github.com/kubernetes-sigs/kro/pkg/graph/revisions"
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
@@ -215,9 +216,8 @@ func TestCreateGraphRevision_SetsOwnerReferences(t *testing.T) {
 	assert.Empty(t, stored.Labels[metadata.ResourceGraphDefinitionIDLabel])
 	assert.Equal(t, "hash-3", stored.Labels[metadata.GraphRevisionHashLabel])
 
-	entry, ok := registry.Get("demo", 3)
-	require.True(t, ok)
-	assert.Equal(t, revisions.RevisionStatePending, entry.State)
+	_, ok := registry.Get("demo", 3)
+	assert.False(t, ok, "RGD controller should not seed runtime revision state")
 }
 
 func TestCreateGraphRevision_AlreadyExistsReturnsError(t *testing.T) {
@@ -262,6 +262,48 @@ func TestCreateGraphRevision_AlreadyExistsReturnsError(t *testing.T) {
 
 	_, ok := registry.Get("demo", 3)
 	assert.False(t, ok)
+}
+
+func TestCreateGraphRevision_DoesNotOverwriteExistingRegistryEntry(t *testing.T) {
+	t.Parallel()
+
+	scheme := runtime.NewScheme()
+	require.NoError(t, internalv1alpha1.AddToScheme(scheme))
+	require.NoError(t, v1alpha1.AddToScheme(scheme))
+
+	cl := newFakeClientBuilder().WithScheme(scheme).Build()
+	registry := revisions.NewRegistry()
+	compiledGraph := &graph.Graph{TopologicalOrder: []string{"deploy"}}
+	registry.Put(revisions.Entry{
+		RGDName:       "demo",
+		Revision:      3,
+		SpecHash:      "hash-3",
+		State:         revisions.RevisionStateActive,
+		CompiledGraph: compiledGraph,
+	})
+
+	reconciler := &ResourceGraphDefinitionReconciler{
+		Client:            cl,
+		apiReader:         cl,
+		metadataLabeler:   metadata.NewKROMetaLabeler(),
+		revisionsRegistry: registry,
+	}
+
+	rgd := &v1alpha1.ResourceGraphDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "demo",
+			UID:  types.UID("new-uid"),
+		},
+	}
+
+	created, err := reconciler.createGraphRevision(context.Background(), rgd, 3, "hash-3")
+	require.NoError(t, err)
+	assert.Equal(t, graphRevisionName("demo", 3), created.Name)
+
+	entry, ok := registry.Get("demo", 3)
+	require.True(t, ok)
+	assert.Equal(t, revisions.RevisionStateActive, entry.State)
+	assert.Same(t, compiledGraph, entry.CompiledGraph)
 }
 
 func TestCreateGraphRevision_HashLabelMergeError(t *testing.T) {
@@ -1264,13 +1306,8 @@ func TestReconcileResourceGraphDefinitionRevisionPaths(t *testing.T) {
 				require.Len(t, revisionList.Items, 2)
 				assert.Contains(t, []int64{revisionList.Items[0].Spec.Revision, revisionList.Items[1].Spec.Revision}, int64(10))
 
-				currentSpecHash, hashErr := graphhash.Spec(rgd.Spec)
-				require.NoError(t, hashErr)
-
-				entry, ok := reconciler.revisionsRegistry.Get(rgd.Name, 10)
-				require.True(t, ok)
-				assert.Equal(t, currentSpecHash, entry.SpecHash)
-				assert.Equal(t, revisions.RevisionStatePending, entry.State)
+				_, ok := reconciler.revisionsRegistry.Get(rgd.Name, 10)
+				assert.False(t, ok, "RGD controller should wait for the GR controller to warm the registry")
 				assert.True(t, conditionFor(t, rgd, GraphRevisionsResolved).IsUnknown())
 			},
 		},
@@ -1439,9 +1476,8 @@ func TestReconcileResourceGraphDefinition_RecreateWithEmptyLiveListClearsStaleRe
 	require.Len(t, revisionList.Items, 1)
 	assert.Equal(t, int64(1), revisionList.Items[0].Spec.Revision)
 
-	latest, ok := reconciler.revisionsRegistry.Latest(rgd.Name)
-	require.True(t, ok)
-	assert.Equal(t, int64(1), latest.Revision)
+	_, ok := reconciler.revisionsRegistry.Latest(rgd.Name)
+	assert.False(t, ok, "newly issued revisions should not appear in the registry until the GR controller processes them")
 }
 
 func TestReconcileResourceGraphDefinition_TerminatingRevisionsBlockReconcile(t *testing.T) {
@@ -1574,11 +1610,8 @@ func TestReconcileResourceGraphDefinition_TerminatingRevisionsBlockReconcile(t *
 				assert.Equal(t, int64(1), revisionList.Items[0].Spec.Revision,
 					"new revision should start from 1 after stale cleanup")
 
-				// The new revision 1 should be in the registry as Pending
-				// (the GR controller will compile and promote it to Active).
-				entry, ok := registry.Get(rgd.Name, 1)
-				require.True(t, ok, "new revision 1 should be in registry")
-				assert.Equal(t, revisions.RevisionStatePending, entry.State)
+				_, ok := registry.Get(rgd.Name, 1)
+				assert.False(t, ok, "RGD issuance should not seed runtime registry state")
 			}
 		})
 	}
--- a/test/integration/suites/core/graphrevision_test.go
+++ b/test/integration/suites/core/graphrevision_test.go
@@ -84,7 +84,10 @@ var _ = Describe("GraphRevision Lifecycle", func() {
 			currentGRs := listGraphRevisions(ctx, rgdName)
 			g.Expect(currentGRs).To(HaveLen(1))
 
-			graphVerified := findGRCondition(currentGRs[0].Status.Conditions, krov1alpha1.GraphRevisionConditionTypeGraphVerified)
+			graphVerified := findGRCondition(
+				currentGRs[0].Status.Conditions,
+				krov1alpha1.GraphRevisionConditionTypeGraphVerified,
+			)
 			g.Expect(graphVerified).ToNot(BeNil())
 			g.Expect(graphVerified.Status).To(Equal(metav1.ConditionTrue))
 


docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,458 @@
...
+instances exist across the cluster. The team updates the RGD to add a sidecar
+container to the deployment template.
+
+Without GraphRevisions: all 200 instances pick up the new graph on their next
+reconcile. If the sidecar has a bug, all 200 instances are affected. The team
+cannot see what the previous graph looked like, cannot roll back without
+reverting the RGD spec (and hoping they remember the exact previous state), and
+cannot test the change on a subset first.
+
+With GraphRevisions: the RGD controller creates revision 4. The team can diff
```
ellistarn: You can even put a "manual review" propogation control blocked in place.

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,458 @@
...
+(created, not yet compiled), Active (compiled, graph available), or Failed
+(compilation error). These states are internal scheduling signals between
+controllers, not exposed in the GraphRevision API status.
+
+### API Changes
+
+#### New CRD: GraphRevision
+
+```yaml
+apiVersion: kro.run/v1alpha1
```
michaelhtm: nit: should we make the apiVersion `internal.kro.run`?

api/v1alpha1/conditions.go
```diff
@@ -38,6 +38,13 @@ const (
 	// reconciler for that resource. This condition indicates the state of the
 	// reconciler.
 	ResourceGraphDefinitionConditionTypeReconcilerReady ConditionType = "ReconcilerReady"
+
+	// GraphRevisionConditionTypeGraphVerified indicates the graph snapshot has
+	// been compiled and validated successfully.
+	GraphRevisionConditionTypeGraphVerified ConditionType = "GraphVerified"
```
jakobmoellerdev, line 44: we should keep terminology clean with our RGD. In RGDs we call graphs that are successfully setup "Active". Shouldn't this be the same here?

api/v1alpha1/conditions.go
```diff
@@ -38,6 +38,13 @@ const (
...
+	// GraphRevisionConditionTypeReady indicates the GraphRevision is ready to be
+	// consumed.
+	GraphRevisionConditionTypeReady ConditionType = "Ready"
```
jakobmoellerdev: Is this just our standard ready condition?

api/internal.kro.run/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,98 @@
...
+	ResourceGraphDefinitionUID types.UID `json:"resourceGraphDefinitionUID,omitempty"`
+	// Revision is a monotonic revision number assigned per ResourceGraphDefinition name.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=1
+	Revision int64 `json:"revision"`
+	// SpecHash is the canonical hash of the source ResourceGraphDefinition spec.
+	//
+	// +kubebuilder:validation:Required
+	SpecHash string `json:"specHash"`
```
jakobmoellerdev: can we make this hash a struct with algo and value? or alternatively specify algorithm used here in spec

api/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,98 @@
...
+	// ResourceGraphDefinitionName identifies the source ResourceGraphDefinition by name.
+	// This field is the authoritative identity for matching/adoption decisions.
+	//
+	// +kubebuilder:validation:Required
+	ResourceGraphDefinitionName string `json:"resourceGraphDefinitionName"`
+	// ResourceGraphDefinitionUID is the source ResourceGraphDefinition UID at creation time.
+	// It is informational only and must not be used as the primary ownership key.
+	//
+	// +kubebuilder:validation:Optional
+	ResourceGraphDefinitionUID types.UID `json:"resourceGraphDefinitionUID,omitempty"`
```
jakobmoellerdev: why is this part of spec if it is informational? shouldn't this be part of status or be enforced?

im also thinking that if a RGD is recreated with the same name, we must be able to distinguish revisions from old and new right? so that would only work with uid

api/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,98 @@
...
+	// +kubebuilder:validation:Required
+	DefinitionSpec ResourceGraphDefinitionSpec `json:"definitionSpec"`
+}
+
+// GraphRevisionStatus defines the observed state of GraphRevision.
+type GraphRevisionStatus struct {
+	// TopologicalOrder is the ordered list of resource IDs based on dependencies.
+	TopologicalOrder []string `json:"topologicalOrder,omitempty"`
+	// Conditions represent the latest available observations of the GraphRevision state.
+	Conditions Conditions `json:"conditions,omitempty"`
```
jakobmoellerdev: why dont you use the standard conditions array here from metav1?

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -239,6 +239,13 @@ type ResourceGraphDefinitionStatus struct {
 	// Resources provides detailed information about each resource in the graph,
 	// including their dependencies.
 	Resources []ResourceInformation `json:"resources,omitempty"`
+	// LatestObservedGV points to the newest GraphRevision observed from the current RGD spec.
+	LatestObservedGV *GraphRevisionReference `json:"latestObservedGV,omitempty"`
```
jakobmoellerdev: GV is overlapping with GroupVersion.

what about having a struct for GraphRevision and then a diff for that?

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -239,6 +239,13 @@ type ResourceGraphDefinitionStatus struct {
...
+	// LatestActiveGV points to the newest GraphRevision currently considered active.
```
jakobmoellerdev: we need a better description on what active means for sure. the switching condition is not sufficiently explained in the CRD imho

api/v1alpha1/graphrevision_types.go
jakobmoellerdev: why is this part of spec if it is informational? shouldn't this be part of status or be enforced?

im also thinking that if a RGD is recreated with the same name, we must be able to distinguish revisions from old and new right? so that would only work with uid

api/internal.kro.run/v1alpha1/graphrevision_types.go
jakobmoellerdev: On another note, we might want to inject the hash as a label or make one of these identifier fields indexable so we can issue an effective lookup from the rgd controller

cmd/controller/main.go
```diff
@@ -59,6 +61,7 @@ func main() {
 		leaderElectionNamespace                     string
 		probeAddr                                   string
 		allowCRDDeletion                            bool
+		graphRevisionConcurrentReconciles           int
```
jakobmoellerdev, line 72: note to us during impl: we need to limit the history buildup on revisions somehow, but this might not be adjustable after it has been set. currently in spec this is not touched at all (like in deployments), so there could be orphaned revisions if we have an adjustment of this flag (i.e. from 10 to 5) => needs cleanup

api/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,98 @@
...
+	// Resources provides detailed information about each resource in the graph.
+	Resources []ResourceInformation `json:"resources,omitempty"`
```
jakobmoellerdev: Here i think we have an issue for collections and external references.
It is currently impossible from the node ID alone to derive the actual objects put into the collections. This means that combination of RGD Spec + Resource Information is insufficient to determine a previous state. => Do we want to preserve this or are we okay with this being lossy?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,458 @@
...
+**Immutable snapshots.** A GraphRevision's spec is immutable (enforced by a CEL
+validation rule: `self == oldSelf`). Once created, it never changes. A new RGD
+spec change produces a new GraphRevision, never an update to an existing one.
+
+**Monotonic revision numbers.** Each GraphRevision has a `revision` field,
+monotonically increasing per RGD name. The RGD persists its high-water mark in
+`status.lastIssuedRevision` so revision numbers never go backwards, even across
+GC cycles.
+
+**Deterministic hashing.** A canonical SHA-256 hash of the RGD spec determines
```
jakobmoellerdev: why is this a cryptographic hash?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,458 @@
...
+canonical JSON, sorted map keys) before hashing. If the hash matches the latest
+GraphRevision, no new revision is created.
+
+**Name-based ownership.** GraphRevisions are grouped by RGD name (via label),
+not RGD UID. This means a deleted-and-recreated RGD with the same name can find
+and reuse existing GraphRevisions. OwnerReferences use the current RGD's UID for
+garbage collection, but the name label drives discovery.
+
+**Three-state registry.** The in-memory registry tracks each revision as Pending
+(created, not yet compiled), Active (compiled, graph available), or Failed
```
jakobmoellerdev: like these states but the conditions are non synced I think (see my comment on there). You write they are not exposed but the status still has a Ready state for compile no?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,458 @@
...
+GC cycles.
+
+**Deterministic hashing.** A canonical SHA-256 hash of the RGD spec determines
+whether a new revision is needed. RawExtension payloads are normalized (YAML to
+canonical JSON, sorted map keys) before hashing. If the hash matches the latest
+GraphRevision, no new revision is created.
+
+**Name-based ownership.** GraphRevisions are grouped by RGD name (via label),
+not RGD UID. This means a deleted-and-recreated RGD with the same name can find
+and reuse existing GraphRevisions. OwnerReferences use the current RGD's UID for
```
jakobmoellerdev: Why is this desired? Also if we really want this the UID should disappear from Revision Spec since its immutable in its entirety

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,458 @@
...
+3. Compile the graph from the snapshot spec.
+4. On success: Put Active (with compiled graph), set GraphVerified=True.
+5. On failure: Put Failed, set GraphVerified=False.
+6. Update status (topological order, resources, conditions).
+7. On deletion: evict from registry, remove finalizer.
+
+### Naming Convention
+
+GraphRevision names follow the pattern `{rgdName}-r{revision}`. If the RGD name
+would cause the total to exceed 253 characters, the RGD name prefix is
```
jakobmoellerdev: this truncation could lead to revisions being shared for two rgds that are truncated to the same base. are we sure about this here?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,458 @@
...
+
+**OwnerReference cascade.** Each GraphRevision has an OwnerReference pointing to
+its source RGD. When the RGD is deleted, Kubernetes garbage collection deletes
+all its GraphRevisions. The GV controller's finalizer ensures registry entries
+are evicted before the API object disappears.
+
+**Retention floor.** The RGD controller runs GC at the end of each reconcile. If
+the number of GraphRevisions exceeds `--rgd-max-graph-revisions` (default 20),
+it deletes the oldest revisions beyond the retention limit. The registry is
+pruned in parallel via `DeleteBelow`.
```
jakobmoellerdev: nit: probably deserves a rename to DeleteRevisionsBefore()

pkg/controller/graphrevision/controller.go
```diff
@@ -0,0 +1,191 @@
...
+			"controllerKind", "GraphRevision",
+		)
+		if req != nil {
+			log = log.WithValues("name", req.Name)
+		}
+		return log
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("GraphRevision").
```
jakobmoellerdev: why is this explicitly named?

pkg/controller/graphrevision/controller.go
```diff
@@ -0,0 +1,191 @@
...
+		For(&v1alpha1.GraphRevision{}).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
jakobmoellerdev: nit: the generation can never change in the current spec. we should also probably still reconcile on label changes

pkg/controller/graphrevision/controller.go
```diff
@@ -0,0 +1,191 @@
...
+		r.registry.Put(revisions.Entry{
+			RGDName:  revision.Spec.ResourceGraphDefinitionName,
+			Revision: revision.Spec.Revision,
+			SpecHash: revision.Spec.SpecHash,
+			State:    revisions.RevisionStateFailed,
+		})
+		return nil, nil, err
+	}
+
+	mark.GraphVerified()
```
jakobmoellerdev, line 178: why do we differentiate between active and verified?

pkg/controller/graphrevision/controller.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (r *GraphRevisionReconciler) reconcileGraphRevision(
+	ctx context.Context,
+	revision *v1alpha1.GraphRevision,
+) ([]string, []v1alpha1.ResourceInformation, error) {
+	mark := NewConditionsMarkerFor(revision)
+	// Mark pending before compile so RGD/instance reconcilers can see that this
+	// revision exists but is not ready to serve yet.
+	r.registry.Put(revisions.Entry{
```
jakobmoellerdev: here im a bit torn. 

wouldn't this mean that while the revision is reconciled, it will always issue a pending state until the revision has compiled even if it was already successful once? 

would it make sense to only put it on pending if there was no initial state?

pkg/controller/instance/controller.go
```diff
@@ -56,6 +58,12 @@ type ReconcileConfig struct {
 	RGDConfig graph.RGDConfig
 }
 
+// GraphRevisionResolver resolves the newest issued graph revision for an RGD.
+type GraphRevisionResolver interface {
```
jakobmoellerdev, line 71: methods and interface deserve some more detail docs

pkg/controller/instance/controller.go
```diff
@@ -242,6 +257,33 @@ func (c *Controller) ensureManaged(rcx *ReconcileContext) error {
...
+	}
+
+	switch latest.State {
+	case revisions.RevisionStateActive:
+		// Active implies the newest issued revision compiled successfully and its
+		// graph is present; this is guaranteed by the GV reconciler and registry
+		// invariant.
+		return latest.CompiledGraph, nil
+	case revisions.RevisionStatePending:
+		return nil, requeue.NeededAfter(
```
jakobmoellerdev, line 331: im assuming this should be an instant requeue and not a default requeue no?

pkg/controller/instance/controller.go
```diff
@@ -242,6 +257,33 @@ func (c *Controller) ensureManaged(rcx *ReconcileContext) error {
...
+			fmt.Errorf("latest issued graph revision %d is pending", latest.Revision),
+			c.reconcileConfig.DefaultRequeueDuration,
+		)
+	case revisions.RevisionStateFailed:
+		return nil, requeue.None(fmt.Errorf("latest issued graph revision %d failed", latest.Revision))
+	default:
+		return nil, requeue.NeededAfter(
```
jakobmoellerdev, line 338: same here, this should probably be an exponential backoff

pkg/controller/instance/controller.go
```diff
@@ -242,6 +257,33 @@ func (c *Controller) ensureManaged(rcx *ReconcileContext) error {
 	return nil
 }
 
+func (c *Controller) resolveCompiledGraph() (*graph.Graph, error) {
```
jakobmoellerdev, line 318: can we parameterize this to the kind of the revision to fetch (i.e. Latest, Active, Previous), so it will be easier for us down the line to touch up the instance reconciler again?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -35,26 +41,89 @@ import (
 func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinition(
 	ctx context.Context,
 	rgd *v1alpha1.ResourceGraphDefinition,
-) ([]string, []v1alpha1.ResourceInformation, error) {
+) (ctrl.Result, []string, []v1alpha1.ResourceInformation, error) {
 	log := ctrl.LoggerFrom(ctx)
 	mark := NewConditionsMarkerFor(rgd)
+	defer func() {
+		err := r.garbageCollectGraphRevisions(ctx, rgd)
+		if err != nil {
```
jakobmoellerdev: shouldnt this error propagate to status?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -35,26 +41,89 @@ import (
...
+		}
+	}()
+
+	graphRevisions, err := r.listGraphRevisions(ctx, rgd)
+	if err != nil {
+		return ctrl.Result{}, nil, nil, fmt.Errorf("listing graph revisions: %w", err)
+	}
+	cacheWarmed := r.isRevisionCacheWarmed(rgd.Name, graphRevisions)
+	if !cacheWarmed {
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, rgd.Status.TopologicalOrder, rgd.Status.Resources, nil
```
jakobmoellerdev: random magic requeue time spotted! needs adjustment

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -35,26 +41,89 @@ import (
...
+	latestRevision := getLatestGraphRevision(graphRevisions)
+
+	var processedRGD *graph.Graph
+	var resourcesInfo []v1alpha1.ResourceInformation
+
+	if latestRevision != nil && latestRevision.Spec.SpecHash == currentSpecHash {
+		rgd.Status.LatestObservedGV = graphRevisionReference(*latestRevision)
+		rgd.Status.LastIssuedRevision = maxInt64(rgd.Status.LastIssuedRevision, latestRevision.Spec.Revision)
+
+		cachedEntry, ok := r.revisionsRegistry.Get(rgd.Name, latestRevision.Spec.Revision)
```
jakobmoellerdev: quick question: shouldnt this be part of the latestRevision fetching? its a bit strange we explicitly have to call into the registry now

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -35,26 +41,89 @@ import (
...
+		if !ok {
+			return ctrl.Result{RequeueAfter: 2 * time.Second}, rgd.Status.TopologicalOrder, rgd.Status.Resources, nil
```
jakobmoellerdev: yet another magic requeue

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -35,26 +41,89 @@ import (
...
+		log.V(1).Info("reconciling resource graph definition graph")
+		processedRGD, resourcesInfo, err = r.reconcileResourceGraphDefinitionGraph(ctx, rgd)
+		if err != nil {
+			mark.ResourceGraphInvalid(err.Error())
+			return ctrl.Result{}, nil, nil, err
+		}
+		mark.ResourceGraphValid()
+
+		revision := maxInt64(rgd.Status.LastIssuedRevision, latestRevisionNumber(latestRevision)) + 1
+		createdGV, createErr := r.createGraphRevision(ctx, rgd, revision, currentSpecHash)
```
jakobmoellerdev: can the rgd runs conflict with each other if we have multiple concurrent workers of the RGD controller? I think not, just want to make sure

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -194,3 +269,151 @@ func (e *microControllerError) Unwrap() error { return e.err }
 func newGraphError(err error) error           { return &graphError{err} }
 func newCRDError(err error) error             { return &crdError{err} }
 func newMicroControllerError(err error) error { return &microControllerError{err} }
+
+func (r *ResourceGraphDefinitionReconciler) listGraphRevisions(ctx context.Context, rgd *v1alpha1.ResourceGraphDefinition) ([]v1alpha1.GraphRevision, error) {
+	labelSelector := labels.SelectorFromSet(map[string]string{
```
jakobmoellerdev: I was hoping for a field selector based on spec instead of labels here, but meh :D

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -194,3 +269,151 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+
+	revisionList := &v1alpha1.GraphRevisionList{}
+	if err := r.List(ctx, revisionList, &client.ListOptions{LabelSelector: labelSelector}); err != nil {
+		return nil, err
+	}
+
+	return revisionList.Items, nil
+}
+
+func (r *ResourceGraphDefinitionReconciler) isRevisionCacheWarmed(rgdName string, revisions []v1alpha1.GraphRevision) bool {
```
jakobmoellerdev: is this warming really even necessary? im trying to understand why we need to wait for all revisions

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -194,3 +269,151 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+
+func graphRevisionReference(revision v1alpha1.GraphRevision) *v1alpha1.GraphRevisionReference {
+	return &v1alpha1.GraphRevisionReference{
+		Name:     revision.Name,
+		Revision: revision.Spec.Revision,
+		SpecHash: revision.Spec.SpecHash,
+	}
+}
+
+func maxInt64(a, b int64) int64 {
```
jakobmoellerdev: math.Max is not a thing?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -194,3 +269,151 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+	latestIdx := 0
+	for i := 1; i < len(revisions); i++ {
+		if revisions[i].Spec.Revision > revisions[latestIdx].Spec.Revision {
+			latestIdx = i
+		}
+	}
+	return &revisions[latestIdx]
+}
+
+func latestRevisionNumber(latest *v1alpha1.GraphRevision) int64 {
```
jakobmoellerdev: this feels like it should go directly on the revision

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -194,3 +269,151 @@ func (e *microControllerError) Unwrap() error { return e.err }
 func newGraphError(err error) error           { return &graphError{err} }
 func newCRDError(err error) error             { return &crdError{err} }
 func newMicroControllerError(err error) error { return &microControllerError{err} }
+
+func (r *ResourceGraphDefinitionReconciler) listGraphRevisions(ctx context.Context, rgd *v1alpha1.ResourceGraphDefinition) ([]v1alpha1.GraphRevision, error) {
```
jakobmoellerdev: a lot of the behavior from the KREP could live in this code here to make it more understandable on why things are the way they are

pkg/graph/hash/hash.go
```diff
@@ -0,0 +1,112 @@
...
+	"encoding/json"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+
+	"github.com/kubernetes-sigs/kro/api/v1alpha1"
+)
+
+// Spec computes a deterministic SHA-256 hash for an RGD spec.
```
jakobmoellerdev: should be fnv or similar if its not for security purposes no?

pkg/graph/hash/hash.go
```diff
@@ -0,0 +1,112 @@
...
+//
+// The hash is based on a canonicalized form of the spec:
+// - Struct field order is stable via JSON marshaling.
+// - Map key order is canonicalized.
+// - RawExtension payloads are normalized into canonical JSON bytes.
+func Spec(spec v1alpha1.ResourceGraphDefinitionSpec) (string, error) {
+	normalized, err := normalizeSpec(spec)
```
jakobmoellerdev, line 50: why cant we use the inbuilt converters here to give us a json representation? we should run JCS normalization on this json: https://github.com/gowebpki/jcs (in encoding/json/v2 in stdlib this exists also)

api/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,98 @@
...
+	Revision int64 `json:"revision"`
+	// SpecHash is the canonical hash of the source ResourceGraphDefinition spec.
+	//
+	// +kubebuilder:validation:Required
+	SpecHash string `json:"specHash"`
+	// DefinitionSpec is an immutable snapshot of the source ResourceGraphDefinition spec.
+	// This includes user-authored schema and resource templates.
+	//
+	// +kubebuilder:validation:Required
+	DefinitionSpec ResourceGraphDefinitionSpec `json:"definitionSpec"`
```
michaelhtm: RGD may reach the maximum object size in etcd. What should that scenario look like for Graph Revisions if we're copying the RGD in graph revisions

docs/design/proposals/graph-revisions.md
a-hilaly: added

api/v1alpha1/conditions.go
a-hilaly: `status.state` and `status.conditions` are different, RGD has a `GraphVerified` condition which we are keeping for GRs too. (same for Active state)

api/v1alpha1/conditions.go
a-hilaly: changed

api/internal.kro.run/v1alpha1/graphrevision_types.go
a-hilaly: may I ask why the need for a struct here? How about `<algo>:<hex>`?

agreed on indexing.

api/v1alpha1/graphrevision_types.go
a-hilaly: agreed. removed.

api/v1alpha1/graphrevision_types.go
a-hilaly: Looking back at @n3wscott's  [design](https://github.com/kubernetes-sigs/kro/blob/main/docs/design/proposals/status-conditions.md), his intent was actually to move the project to `metav1.Condition`; the custom `Condition` type predates that work. What landed was the local condition manager pattern, but not the full type migration, so GR currently follows the existing project wide condition shape

api/v1alpha1/resourcegraphdefinition_types.go
a-hilaly: agree, changed to GR. Can you say a bit more about what you mean by diff here?

api/v1alpha1/resourcegraphdefinition_types.go
a-hilaly: i was overthinking it. i'm keeping only `LastIssuedRevision`. keeping this one because it's used as the high water mark, we can always introduce more informational fields in the future.

api/v1alpha1/graphrevision_types.go
a-hilaly: good point, removed.

cmd/controller/main.go
a-hilaly: This is handled by `--rgd-max-graph-revisions` We already prune against the current retention limit during RGD reconcile

api/v1alpha1/graphrevision_types.go
a-hilaly: This field is mainly informational for users, not something I expect to be sufficient to reconstruct previous concrete state

it surfaces the inferred dependency topology from the compiled graph, not the exact realized object set

i'm happy to drop it for now and come up with something better in the future

docs/design/proposals/graph-revisions.md
a-hilaly: mainly for low collision risk, it's a boring choice that works. not opinionated here. what were you thinking, fnvs?

docs/design/proposals/graph-revisions.md
a-hilaly: correct they are surfaced as `GraphVerified` and the synthesized Ready conditions instead

docs/design/proposals/graph-revisions.md
a-hilaly: This is desired because I want the delete and recreate of an RGD with the same name to just work as continuity, not start a brand new revision lineage because the object UID changed

docs/design/proposals/graph-revisions.md
a-hilaly: Good catch - what are you suggesting? IMO we do need the RGD name to be part of the GV name.

docs/design/proposals/graph-revisions.md
a-hilaly: agreed

pkg/controller/graphrevision/controller.go
a-hilaly: good point

pkg/controller/graphrevision/controller.go
a-hilaly: I kept Active as the runtime term for consistency with the RGD naming - as in active is set up and is usable. that said i agree they are close today

pkg/controller/graphrevision/controller.go
a-hilaly: good point, fixed

pkg/controller/instance/controller.go
a-hilaly: since we're using one queue i'm woried about high churn

pkg/controller/instance/controller.go
a-hilaly: afaik requeue.Needed() is not a controller-runtime error retry with rate limited exponential backoffm i think it can instantly retry

pkg/controller/instance/controller.go
a-hilaly: I’d prefer not to parameterize this, yet. Latest is the only defined selection policy today - `Active`/Previous start to encode rollout and fallback semantics that we haven’t actually specified

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: I think we should surface it, but as a separate informational condition rather than something that feeds Ready/Active

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: agreed, it's some sort of sorcery

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: yep that's fair

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: yep two RGDs won't be reconciling at the same time thanks to queue dedups

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: profile driven development :D

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: Yeah, it’s for restart correctness. We need the full revision set before proceeding so we recover the watermark correctly and don’t make issuance/serving decisions against a partially warmed registry. replay order is not guaranteed, so we can’t assume the newest revision shows up first

I could be wrong tho, i'll keep thinking about it

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: it's float only... TIL `max` is a thing in 1.25

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: agree

pkg/graph/hash/hash.go
a-hilaly: this was mainly for `runtime.RawExtension` which come in as yaml but could be really anything.. i agree CS would be a clear final canonicalization step after that than relying json.Marshall convensions

pkg/graph/hash/hash.go
a-hilaly: that's what I thought you'd suggest :)

api/internal.kro.run/v1alpha1/graphrevision_types.go
jakobmoellerdev: string is fine for me as well :+1

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: TBH cannot even remember my self, just ignore until my LLM mind stops hallucinating

api/v1alpha1/graphrevision_types.go
jakobmoellerdev: not saying we drop it, more that we might need additional info in controller to accomodate this state fully. totally fine for now

docs/design/proposals/graph-revisions.md
jakobmoellerdev: fnv is the way!

docs/design/proposals/graph-revisions.md
jakobmoellerdev: I think we might want the hash of the full name as part of the key that we use for resolution (we have the label change coming in soon to accomodate this). this way we can do a unique lookup without relying on name. WDYT?

pkg/controller/instance/controller.go
jakobmoellerdev: could you point me at the code for that? maybe my fundamentals are too rusty...

a-hilaly: https://github.com/kubernetes-sigs/kro/blob/main/pkg/dynamiccontroller/dynamic_controller.go#L264-L267

a-hilaly: Maybe it's my fundamentals that are too rusty.. exponential backoff comes from the queue IIRC

helm/values.yaml
```diff
@@ -136,6 +138,8 @@ config:
     maxCollectionSize: 1000
     # Maximum size of collection dimension per forEach defined in ResourceGraphDefinitions
     maxCollectionDimensionSize: 10
+    # Maximum number of GraphRevisions to retain per ResourceGraphDefinition
+    maxGraphRevisions: 20
```
jakobmoellerdev: lets go for 5 or something smaller for now, impact can be very large

api/internal.kro.run/v1alpha1/graphrevision_types.go
jakobmoellerdev: since the spec is anyhow immutable, should we move revision hash calculation to status and let it be computed by revision controller?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
jakobmoellerdev: could you elaborate here why not to choose a spec field selector? we already have it in spec anyhow right?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -207,3 +305,184 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+		Revision: revision,
+		SpecHash: specHash,
+		State:    revisions.RevisionStatePending,
+	})
+	return graphRevision, nil
+}
+
+func graphRevisionName(rgdName string, revision int64) string {
+	hash := graphRevisionNameHash(rgdName)
+	suffix := fmt.Sprintf("-%s-r%06d", hash, revision)
```
jakobmoellerdev: I think we should not prefix the revision number with leading zeros, since the hash conflict potential rises

if you really want to trim that you should change the algorithm to have better conflict potential with less characters imho

pkg/graph/hash/hash.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+	"hash/fnv"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/yaml"
+
+	"github.com/kubernetes-sigs/kro/api/v1alpha1"
+)
+
+var jsonMarshal = json.Marshal
```
jakobmoellerdev, line 33: what is happening

pkg/graph/hash/hash.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Spec computes a deterministic FNV-128a hash for an RGD spec.
+//
+// FNV is chosen over a cryptographic hash because this is a content-identity
+// fingerprint, not a security boundary. FNV-128a is fast, well-distributed,
+// and sufficient for collision resistance at the scale of revision lineages.
+//
+// The hash is based on a canonicalized form of the spec:
```
jakobmoellerdev: where do you actually do canonicalization?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -207,3 +305,184 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+	maxPrefixLen := 253 - len(suffix)
+	if len(rgdName) > maxPrefixLen {
+		rgdName = rgdName[:maxPrefixLen]
+	}
+	return rgdName + suffix
+}
+
+func graphRevisionNameHash(rgdName string) string {
+	sum := sha256.Sum256([]byte(rgdName))
```
jakobmoellerdev: probably fnv again

pkg/metadata/labels.go
```diff
@@ -216,6 +229,22 @@ func safeVersion(version string) string {
...
+
+	if raw, err := hex.DecodeString(specHash); err == nil {
+		encoded := strings.ToLower(base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(raw))
+		if validation.IsValidLabelValue(encoded) == nil {
+			return encoded
+		}
+	}
+
+	sum := sha256.Sum256([]byte(specHash))
+	return strings.ToLower(base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(sum[:]))
```
jakobmoellerdev: maybe we should do the same for rgd name encoding?

pkg/controller/graphrevision/controller.go
```diff
@@ -0,0 +1,204 @@
...
+func (r *GraphRevisionReconciler) Reconcile(ctx context.Context, obj *internalv1alpha1.GraphRevision) (ctrl.Result, error) {
+	if !obj.DeletionTimestamp.IsZero() {
+		if err := r.setUnmanaged(ctx, obj); err != nil {
+			return ctrl.Result{}, err
+		}
+		// Evict only after finalizer removal succeeds. If patching fails, the GR is
+		// still present in the API and must remain visible in cache for warmup paths.
+		r.registry.Delete(obj.Spec.ResourceGraphDefinitionName, obj.Spec.Revision)
+		return ctrl.Result{}, nil
+	}
```
michaelhtm, line 114: can the user delete the GraphRevision? If so, should we block it? or trigger a recreate when deleted?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,499 @@
...
+(compilation error). These states are internal scheduling signals between
+controllers. The external GraphRevision API surfaces them through conditions
+instead of a separate state enum.
+
+### API Changes
+
+#### New CRD: GraphRevision
+
+```yaml
+apiVersion: kro.run/v1alpha1
```
michaelhtm:
```suggestion
apiVersion: internal.kro.run/v1alpha1
```

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,499 @@
...
+4. On success: Put Active (with compiled graph), set GraphVerified=True.
+5. On failure: Put Failed, set GraphVerified=False.
+6. Update status (topological order, resources, conditions).
+7. On deletion: evict from registry, remove finalizer.
+
+### Naming Convention
+
+GraphRevision names follow the pattern `{rgdName}-rNNNNNN` for normal-size names.
+For very long RGD names that must be truncated to stay under the 253-byte limit,
+the name is `{truncatedRGDName}-{shortHash}-rNNNNNN`, where `shortHash` is a
```
michaelhtm: how does the revision increase/decrease?
do we want it to go back to 0 at some point again? Or could we see it reaching n digits worst case? reaching the max limit for resource name

helm/values.yaml
a-hilaly: done. separately I want to add logic to keep revisions alive if instances are still pinned to them - need to think about how to do that cleanly though, it's not trivial..

api/internal.kro.run/v1alpha1/graphrevision_types.go
a-hilaly: Hash needs to stay in spec - it's an identity field, not an observation. The RGD controller uses it to check if a revision already exists before creating one. If it's in status, tha lookup doesn't work at creation time. maybe we can do some verification yes.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: A field indexer on `spec.resourceGraphDefinitionName` would work too but requires registering a custom indexer to get the same results we get for free with labels

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: Agreed on the hash length, bumping from 8 to 12 hex chars. Dropping the zero padding on the revision number too. the monotonic counter can grow over 6 digits anyways.

Also swapping the order to `{rgdName}-r{revision}-{hash}` - revision number is what you actually scan for, hash is just a collision guard on truncated RGD names. Names will look like `my-webapp-r3-2a97516c354b`

pkg/graph/hash/hash.go
a-hilaly: lets tests inject marshal failures to cover error paths. Used in `withJSONMarshalStub` - really went for 100% test coverage here

pkg/graph/hash/hash.go
a-hilaly: Canonicalization is in normalizeRawExtension -> `json.Unmarshal` into any -> `json.Marshal` back out (sorts map keys). Same content of same bytes produce same hash

I verified verified against a live cluster that the API server already sorts keys in `RawExtension` at every nesting depth, so the `round trip` is technically redundant. Keeping it as a safety net since it's not a documented guarantee. Also switched from` yaml.Unmarshal` to `json.Unmarshal`

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: You were right - switched to a field indexer on `spec.specHash/rgdName` since are immutable. Keeping the RGD name label for `kubectl` and informational purposes but lookups now use the field selector. Also dropped the UID label from `GraphRevisions` since lineage is name-based.

pkg/graph/hash/hash.go
jakobmoellerdev: you should probably add a comment on this. that said will have to check the stub check

pkg/graph/hash/hash.go
jakobmoellerdev: I thought about this some more and I think this is not okay to merge without switching to proper JCS because switching later will break older canonicalized RGD hashes. I am afraid of a world in which we have a bug here and are never able to switch away from it again.

api/internal.kro.run/v1alpha1/graphrevision_types.go
jakobmoellerdev: why is hash an identity field? currently afaik the only identity field we have is the namespace+name. Hash is not authoritative (even if we want it to be, the current model doesnt do that). And since spec can never change change, hash wont as well.

a-hilaly: "Identity" is maybe too strong a word but it's not identity in the namespace/name sense. it does serve as the dedup key: when the RGD controller it compares its current spec hash against `latestRevision.Spec.snapshot.SpecHash` to decide whether a new revision issuance is needed

a-hilaly: If we move it to status, the RGD controller would have to create the GR without a hash, wait for the GR controller to reconcile and populate `.status.specHash` and then come back and read it, it's just an extra round trip for something we already know at creation time. My thinking is that the hash is a computed property of the snapshot that's frozen at creation, same as generation and spec itself

a-hilaly: Happy to hear if there's a concrete problem this causes though if it's purely about convention I'd rather keep the simpler flow

docs/design/proposals/graph-revisions.md
a-hilaly: [Discussed of line about kubectl output ordering]

pkg/graph/hash/hash.go
a-hilaly: are you worried about floats in `resources[].template` ? jsonMarshal https://pkg.go.dev/encoding/json#Marshal and its float formatting is stable, but it's not formally specified like JCS

That said - if we ever need to switch canonicalization, the cost is one re issuance per RGD. Not broken state, just a revision bump. We're not really locked in?

docs/design/proposals/graph-revisions.md
a-hilaly: Monotonically increasing, never resets. High-water mark persisted in `status.lastIssuedRevision` so we never reuse a number. that said, if you delete all revisions and recreate the RGD it will re-start from 0

pkg/controller/graphrevision/controller.go
a-hilaly: Yes, users can delete them. The finalizer ensures we cleanly evict from the registry before the object is removed. On next reconcile, the RGD controller sees the latest is gone (if it was the one deleted) and re issues if needed. Self-healing - no need to block deletion

api/v1alpha1/graphrevision_types.go
a-hilaly: Good point, the create will fail, an error that the RGD will expose in it's conditions

pkg/graph/hash/hash.go
a-hilaly: (I'm still thinking about this)

jakobmoellerdev: I think it would immediately result in broken state because the resulting hash would be differing, which would then lead to an algorithm and value mismatch when comparing it to the active revision. since the revision is immutable its impossible to change without issuing a new revision, but Im not sure if we should reissue revisions if we have a hash mismatch

api/internal.kro.run/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,112 @@
...
+// ResourceGraphDefinitionSnapshot captures the identity and spec of a
+// ResourceGraphDefinition at a point in time.
+type ResourceGraphDefinitionSnapshot struct {
+	// Name identifies the source ResourceGraphDefinition by name.
+	// This is the authoritative identity for lineage/adoption decisions.
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// UID is the UID of the source ResourceGraphDefinition when this revision
+	// was issued. Informational only — lineage is name-based, not UID-based.
```
jakobmoellerdev: I think we should drop uid since we dont use it atm. we can always add back later

api/internal.kro.run/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,112 @@
...
+	// This includes user-authored schema and resource templates.
+	//
+	// +kubebuilder:validation:Required
+	Spec krov1alpha1.ResourceGraphDefinitionSpec `json:"spec"`
+}
+
+// GraphRevisionStatus defines the observed state of GraphRevision.
+type GraphRevisionStatus struct {
+	// TopologicalOrder is the ordered list of resource IDs based on dependencies.
+	TopologicalOrder []string `json:"topologicalOrder,omitempty"`
```
jakobmoellerdev, line 61: the active topological order is not definitive and it can change based on includeWhen conditions for each instance, so im not sure how useful this order here really is. WDYT? (is actually also true already for current RGDs)

api/v1alpha1/conditions.go
jakobmoellerdev: If its Compiled and Validated => why is it Verified?

Shouldnt it be `CompiledAndValidated` ? or just `Validated` ?

Alternatively maybe change the comment

pkg/controller/graphrevision/controller.go
```diff
@@ -0,0 +1,214 @@
...
+		maxConcurrentReconciles: maxConcurrentReconciles,
+	}
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *GraphRevisionReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.Client = mgr.GetClient()
+
+	logConstructor := func(req *reconcile.Request) logr.Logger {
+		log := mgr.GetLogger().WithName("graph-revision-controller").WithValues(
```
jakobmoellerdev: why do we need such a complex log constructor?

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -91,6 +100,25 @@ func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) e
 	r.clientSet.SetRESTMapper(mgr.GetRESTMapper())
 	r.instanceLogger = mgr.GetLogger()
 
+	// Index GraphRevisions by RGD name so listGraphRevisions can use a field
+	// selector on the immutable spec field instead of a label selector. Labels
+	// are kept for kubectl/informational purposes but are not trusted for
+	// functional lookups since they can be modified externally.
+	if err := mgr.GetFieldIndexer().IndexField(
```
jakobmoellerdev: isnt this a client side index? i was expecting an index on the CRD itself? https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#crd-selectable-fields

I dont see any selectable fields on the CRD

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -16,77 +16,123 @@ package resourcegraphdefinition
...
 )
 
-// reconcileResourceGraphDefinition orchestrates the reconciliation of a ResourceGraphDefinition by:
-// 1. Processing the resource graph
-// 2. Ensuring CRDs are present
-// 3. Setting up and starting the microcontroller
+const (
+	// defaultRequeueDelay is the delay before requeuing when waiting for
+	// asynchronous state to converge (e.g. cache warmup, pending compilation).
+	defaultRequeueDelay = 3 * time.Second
```
jakobmoellerdev: this probably needs to be parameterized

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -16,77 +16,123 @@ package resourcegraphdefinition
...
+		}
+		return ctrl.Result{}, rgd.Status.TopologicalOrder, rgd.Status.Resources, err
 	}
-	if crd, err = r.crdManager.Get(ctx, crd.Name); err != nil {
-		mark.KindUnready(err.Error())
-	} else {
-		mark.KindReady(crd.Status.AcceptedNames.Kind)
+
+	latestRevision := latestRevisionView.Revision
+	if latestRevision == nil || latestRevisionView.RuntimeEntry == nil || latestRevisionView.RuntimeEntry.SpecHash != currentSpecHash {
```
jakobmoellerdev: these conditions deserve their own log reasoning on why the build happens i think

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -16,77 +16,123 @@ package resourcegraphdefinition
...
+		if _, _, err := r.buildResourceGraphDefinition(ctx, rgd); err != nil {
+			mark.InvalidResourceGraph(err.Error())
+			return ctrl.Result{}, nil, nil, err
+		}
+		mark.ResourceGraphValid()
+
+		revision := max(rgd.Status.LastIssuedRevision, latestRevisionView.RevisionNumber) + 1
+		createdGR, createErr := r.createGraphRevision(ctx, rgd, revision, currentSpecHash)
+		if createErr != nil {
+			mark.RevisionLineageFailed(createErr.Error())
```
jakobmoellerdev: for me this would be a createGraphRevision failed.

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -16,77 +16,123 @@ package resourcegraphdefinition
...
+	topologicalOrder, resourcesInfo, err := r.ensureServingState(
+		ctx,
+		rgd,
+		latestRevisionView.RuntimeEntry.CompiledGraph,
+		mark,
+	)
+
+	// Best-effort retention GC: runs inline after serving state is established.
+	// Failures are logged and recorded as a metric but do not block the reconcile.
+	if gcErr := r.garbageCollectGraphRevisions(ctx, rgd); gcErr != nil {
```
jakobmoellerdev, line 159: arguably this should then be a separate control loop no?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -106,18 +152,69 @@ func (r *ResourceGraphDefinitionReconciler) setupMicroController(
...
 		r.metadataLabeler,
 		r.dynamicController.Coordinator(),
 	)
 }
 
-// reconcileResourceGraphDefinitionGraph processes the resource graph definition to build a dependency graph
-// and extract resource information. It skips the expensive build pipeline when the
-// RGD's .metadata.generation has not changed since the last successful build.
-func (r *ResourceGraphDefinitionReconciler) reconcileResourceGraphDefinitionGraph(_ context.Context, rgd *v1alpha1.ResourceGraphDefinition) (*graph.Graph, []v1alpha1.ResourceInformation, error) {
+func (r *ResourceGraphDefinitionReconciler) reconcileRevisionLineage(
```
jakobmoellerdev: his state machine needs explanation and why things are moving the way they are

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -207,3 +345,209 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+	RuntimeEntry   *revisions.Entry
+}
+
+// getLatestGraphRevisionView returns the latest revision view and whether the
+// latest revision is present in the in-memory registry.
+//
+// Only the latest revision needs to be warmed because the reconcile loop only
+// ever serves the latest compiled graph. Older revisions may still be compiling
+// in the GR controller — that's fine, they don't block serving or issuance.
+func (r *ResourceGraphDefinitionReconciler) getLatestGraphRevisionView(
```
jakobmoellerdev, line 458: this needs a todo that we need to have n-1 at the very least right?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -207,3 +345,209 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+		rgdName = rgdName[:maxPrefixLen]
+	}
+	// Trim trailing dots/hyphens so the joined name stays DNS-1123 valid.
+	// RGD names can contain dots (DNS-1123 subdomains allow them), so
+	// truncation can land on a dot, producing "name.-r1-hash" which is invalid.
+	rgdName = strings.TrimRight(rgdName, ".-")
+	return rgdName + suffix
+}
+
+// graphRevisionNameHash computes a short FNV-128a hash of the RGD name for use
```
jakobmoellerdev: I think we agreed that fnv32 is sufficient collision here

pkg/controller/resourcegraphdefinition/controller_reconcile.go
```diff
@@ -207,3 +345,209 @@ func (e *microControllerError) Unwrap() error { return e.err }
...
+		}
+		if err := r.Delete(ctx, &revision); err != nil {
+			ignoreNotFound := client.IgnoreNotFound(err)
+			if ignoreNotFound != nil {
+				return fmt.Errorf("deleting graph revision %q: %w", revision.Name, err)
+			}
+		}
+	}
+
+	r.revisionsRegistry.DeleteRevisionsBefore(rgd.Name, minRevisionToKeep)
```
jakobmoellerdev: dont we need to wait until the revision is actually gone before killing it from the registry?

pkg/controller/resourcegraphdefinition/controller_status.go
```diff
@@ -81,6 +97,13 @@ func (r *ResourceGraphDefinitionReconciler) updateStatus(
 	})
 }
 
+func isConditionTrueForObservedGeneration(cond *v1alpha1.Condition, generation int64) bool {
+	if cond == nil {
+		return false
+	}
+	return cond.IsTrue() && cond.ObservedGeneration == generation
```
jakobmoellerdev: nit: this is same as returning cond != nil && cond.IsTrue() && cond.ObservedGeneration == generation which can be inlined

pkg/controller/resourcegraphdefinition/controller_status.go
```diff
@@ -112,21 +135,29 @@ func (r *ResourceGraphDefinitionReconciler) setUnmanaged(ctx context.Context, rg
...
 
 // NewConditionsMarkerFor creates a marker to manage conditions and sub-conditions for ResourceGraphDefinitions.
 //
 // ```
 // Ready
-//	├─ ResourceGraphAccepted - This controller has accepted the spec.schema and spec.resources.
 //	├─ KindReady - The CRD status created on behalf of this RGD.
 //	└─ ControllerReady - The status of the controller thread reconciling this resource.
+//
+// ResourceGraphAccepted and RevisionLineageResolved are informational conditions
```
jakobmoellerdev: IMHO this should not be surfaced then. I think it breaks with our existing behavior in that all conditions we add currently are contributing to Ready. I dont see a large benefit in an extra condition for that reason.

pkg/graph/hash/hash.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return "", fmt.Errorf("marshal normalized spec: %w", err)
+	}
+
+	h := fnv.New128a()
+	h.Write(data)
+	return hex.EncodeToString(h.Sum(nil)), nil
+}
+
+func normalizeSpec(spec v1alpha1.ResourceGraphDefinitionSpec) (v1alpha1.ResourceGraphDefinitionSpec, error) {
```
jakobmoellerdev, line 65: since this is a custom normalization, we need to version it off and encode it as part of our hash, otherwise it will later on not be possible to transition to a new normalization.

jakobmoellerdev: I think we agreed that this should be a hash all scenario and we dont mind reordering for now so we could land on JCS.

api/internal.kro.run/v1alpha1/graphrevision_types.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	krov1alpha1 "github.com/kubernetes-sigs/kro/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// GraphRevisionSpec defines the desired state of GraphRevision.
+// It captures an immutable snapshot of the source ResourceGraphDefinition.
+type GraphRevisionSpec struct {
```
jakobmoellerdev, line 23: Im gonna point out the elephant in the Room. Why do we call "ResourceGraphDefinition" that, but we dont call this here "ResourceGraphRevision"?

pkg/metadata/labels.go
```diff
@@ -219,6 +232,22 @@ func safeVersion(version string) string {
...
+	}
+
+	if raw, err := hex.DecodeString(specHash); err == nil {
+		encoded := strings.ToLower(base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(raw))
+		if validation.IsValidLabelValue(encoded) == nil {
+			return encoded
+		}
+	}
+
+	sum := sha256.Sum256([]byte(specHash))
```
jakobmoellerdev: this is still a sha256, when it should probably be fnv right?

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: Inline is intentional here, not a shortcut. GC runs specifically after `ensureServingState` succeeds - we only prune old revisions once we've confirmed the latest one is compiled. A separate loop would lose that sequencing guarantee and risk deleting old revisions before the new one is proven healthy


Happy to revisit post release

pkg/controller/resourcegraphdefinition/controller_reconcile.go
a-hilaly: Delegated to GR controller

pkg/controller/resourcegraphdefinition/controller_status.go
a-hilaly: Ready means "CRD established, controller running, instances can be served" ResourceGraphAccepted and RevisionLineageResolved are diagnostic signals about the current reconcile, not serving state. Making them block Ready would flip the rgd to Inactive on every spec change while the new revision compiles - even though the old revision is still serving just fine. That's a lie about system health

pkg/graph/hash/hash.go
a-hilaly: the hash is internal - it's not persisted on GraphRevision specs or exposed as an API contract anymore. It only lives in the in-memory registry and as a label for kubectl convenience, I added a doc comment on `Spec()`.

a-hilaly: I thought about this, but I think it's better that we handle reordering - especially for resources. since this is internal we can always switch if the need arises, and controller restarts with new algorithms would work just fine

api/internal.kro.run/v1alpha1/graphrevision_types.go
a-hilaly: Haaa, i personally think the `Resource` in RGD was mistake. 2/ i think GraphRevision aligns more with long term vision

api/internal.kro.run/v1alpha1/graphrevision_types.go
a-hilaly: This is not for instances it's for GRs, and it includes the node ids not the real resources. Just keeping consistent with RGD. I don't mind nuking it along with ResourceInfo but i think both are actually useful for debugging

api/internal.kro.run/v1alpha1/graphrevision_types.go
a-hilaly: droppped

pkg/controller/resourcegraphdefinition/controller_status.go
a-hilaly: Anyways non-Ready conditions are not API contract (as stated in our docs) so we can experiment with them right now.

pkg/controller/resourcegraphdefinition/controller.go
a-hilaly: i thought actually we don't need that a hit in informer cache is enough?

jakobmoellerdev: Informer cache can go stale and then you rely on sync interval, what's the issue with a server side field?

a-hilaly: Added crd selector field. 

However my thinking was that this is the same informer backing GR controller ... if it's stale it's stale on both controllers and maybe it's a good thing that they are both in sync? (using client selectors?)

pkg/controller/graphrevision/controller.go
michaelhtm: Should we trigger a create if the user deleted the last revision?

PR #1090, opened 2026-03-05
fix: return a requeue error when waiting for readiness

fabianburth:
While fixing a regression that prevented errors from being surfaced properly (see https://github.com/kubernetes-sigs/kro/pull/1078), we also noticed that we currently do not requeue in case of `ErrWaitingForReadiness` (see comment https://github.com/kubernetes-sigs/kro/pull/1078#discussion_r2882244738).

This PR now causes the return of a requeue error in case of `ErrWaitingForReadiness`.


pkg/controller/instance/resources.go
```diff
@@ -595,13 +595,8 @@ func (c *Controller) processApplyResults(
...
-	for _, state := range rcx.StateManager.NodeStates {
-		if state.Err != nil && !errors.Is(state.Err, runtime.ErrWaitingForReadiness) {
-			errs = append(errs, state.Err)
-		}
-	}
-	if err := errors.Join(errs...); err != nil {
+	// Aggregate all node errors
+	if err := rcx.StateManager.NodeErrors(); err != nil {
 		return fmt.Errorf("apply results contain errors: %w", err)
 	}
```
fabianburth, line 600: This leads to a delayed requeue for `ErrWaitingForReadiness` now, with a default of (I think) 3 seconds. Not sure if this might lead too unfortunate load characteristics.

pkg/controller/instance/controller.go
```diff
@@ -202,9 +202,6 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 		} else {
 			rcx.Mark.ResourcesNotReady("resource reconciliation error")
 		}
-	case InstanceStateInProgress:
-		err := rcx.StateManager.NodeErrors()
-		rcx.Mark.ResourcesNotReady("awaiting resource readiness: %v", err)
```
a-hilaly, line 218: Removing this case looks risky. `IN_PROGRESS` is still a valid state outside the readiness wait path, for example when reconciliation is suspended. With this branch removed, those flows now fall into "unknown instance state" instead of preserving the current non-error behavior. I think we should either keep the IN_PROGRESS handling here or explicitly special-case suspended reconciliation.

fabianburth: I removed it because it was me, who introduced it in the first place, in the previous [PR](https://github.com/kubernetes-sigs/kro/pull/1078). Since I introduced it particularly for the waiting for readiness case, I assumed it would not be used anymore catching the waiting for readiness in the code mentioned above. So, I primarily meant to clean up after myself.

Do you still want to keep it?

a-hilaly: I think the concrete issue here is suspended reconciliation rather than `IN_PROGRESS` in general. 

So I think we should either keep the `InstanceStateInProgress` handling here, or explicitly special case suspended reconciliation with a `instanceStateSuspended` thing.

happy to have the suspended state modeling addressed in a separate PR

Issue #542, opened 2025-05-08
Support for handling resource life-cycle modifiers


barney-s:
### Feature Description

**Problem Statement**:

We want to support different life-cycle modifiers such as:
1. Create  : allowed or not , force re-create if it exists
2. Delete : allowed or not 
3. Update : allowed or not 

Some example use cases:
1. Annote a service account. May be create. But dont delete it.
2. Create a DB object, but dont delete it.
3. Create a Job but dont delete it.

**Proposed Solution**:

```
resources:
  - id: something
     template: ...
     lifecycle:
          create: always* | once | never
          delete: always* | abandon
          update: always* | never
```

**Alternatives Considered**:

For KCC (https://cloud.google.com/config-connector/docs/overview), the individual objects support its own lifecycle annotation to deal with the underlying GCP objects. We can use it for these resources. 

Use annotations on the resource being created.
```
resources:
   id: something
   template:
      ...
      metadata:
         annotation:
            kro.lifecycle.delete: abandon
```

Advantage is this allows user to modify object lifecycle without modifying RGD for break-glass.

**Additional Context**:

#537

Resource `annotateDefaultServiceAccount` is the default SA created by Kubernetes for a namespace. We want to annotate it for Workload identity but not delete it. 

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

a-hilaly:
related https://github.com/kro-run/kro/issues/183

ellistarn:
I was discussing something similar with @a-hilaly in DMs:

---

[] <-- **Watcher** (just read)
[Create] <-- (create and read, but don't update? Footgun?)
[Create, Update] <-- **Donor** (create and update it, but never delete? This is Retain)
[Create, Update, Delete] <-- **Owner** (this is normal management)
[Update] <-- (dont create it, but update it, and never delete. Footgun?)
[Update, Delete] <-- **Adopter** (don't create it, but you own this now)
[Delete] <--(read it, never change it, but kill it? Footgun?)

4 real use cases that aren't footguns: **Owner**, **Watcher**, **Donor**, **Adopter**.

We were debating whether or not to decompose each of the actions or to roll the combinatorics into some first class concepts.

---
There are a few things in your post that aren't capture here, like create once, though I'm not sure I understand the use case there.

barney-s:
[...]

PR #1091, opened 2026-03-05
KREP-014 Resource lifecycles

NicholasBlaskey:
Addresses #542

Add design document for resource lifecycles.

### DIFF
--- a/docs/design/proposals/resource-lifecycles.md
+++ b/docs/design/proposals/resource-lifecycles.md
@@ -73,9 +73,9 @@ that it is no longer an owner of replicas and then report a terminal
 conflict. Users would be expected to see this conflict and resolve it
 on a case by case basis.
 
-On delete, if there are no other owners we just delete the object as is. If
-there are other owners we do an empty SSA releasing ownership of all of our
-fields.
+On instance delete or pruning if the rgd no longer requires the resource, if there 
+are no other owners we just delete the object as is. If there are other owners we do
+an empty SSA releasing ownership of all of our fields.
 
 These semantics allow partial objects. With this you can define a Kro resource
 that just changes a single field or a few fields.
--- a/docs/design/proposals/resource-lifecycles.md
+++ b/docs/design/proposals/resource-lifecycles.md
@@ -2,161 +2,193 @@
 
 ## Problem statement
 
-Resource lifecycles and ownership have a wide potential scope. Here are some user stories grouped by broad use case.
+Resource lifecycles and ownership have a wide potential scope. A further discussion on other potential items is listed in the not-in-scope section.
 
-CRUD policies
-- I want to run resources on deletion because I may want to snapshot a database before deleting it.
-- I want to be able to update immutable fields in Kro because I want to change a job spec.
-- I want to create a resource and never update it because some resources like PVC might be expensive or difficult to update.
-- I want to only read a resource because I don't want to use external refs and want to reference the resource.
+This document addresses the following user stories:
 - I want to avoid deleting resources because some resources like databases are too risky to be deleted through Kro.
-- I want to control deletion order because I may have implicit dependencies or a more complicated teardown workflow needed for resource cleanup.
-
-Migration into Kro and between Kro resources
 - I want to safely do migrations into Kro from existing resources because I want to manage my cluster in Kro and want to avoid downtime and risk of updating resources poorly.
 - I want to safely do migrations between Kro RGDs or instances because I want to be able to refactor and evolve my Kro definitions over time and avoid downtime.
 
-Singletons
-- I want to define applications with resources that are defined only once across all Kro instances because resources like PrivateEndpoint are one per cluster.
-
-Resource contribution
-- I want to create a controller that updates a subset of fields on a resource because it is a useful pattern to control a small number of fields in Kubernetes.
-
 ## Proposal
 
-The proposal is to add a new mode of operating resources
-through an annotation called `kro.run/ownership=shared`.
+The proposal is to let adoption be handled by the current behaviour.
+The current behaviour applies over other resources as long as they
+are not being managed by a Kro instance.
 
-Shared mode will be a new primitive that solves migrating
-between Kro resources, Singletons, and resource contribution.
+To avoid deleting resources we will add a new CRD field.
+```
+apiVersion: kro.run/v1alpha1
+  kind: ResourceGraphDefinition
+  spec:
+    resources:
+      - id: database
+        lifecycle:
+          deletePolicy: Retain
+```
+
+The `deletePolicy` field accepts two values:
+- `Delete` (default): Remove the resource from the cluster
+- `Retain`: Remove Kro labels indicating it is Kro-managed instead of deleting it
+
+For migration or refactoring use cases, you would add the delete policy field,
+then delete or update the RGD to orphan the resources, and finally adopt them
+into another RGD or instance. 
 
 #### Overview
 
-##### Exclusive
+##### Adoption
 
-The default mode is exclusive. A Kro resource applied with exclusive
-will cause errors for any other Kro instance trying to use it. Otherwise,
-exclusive will work the same as it does today. Exclusive resources will
-adopt resources from other systems but raise errors if they have Kro
-applyset labels.
+While adopting resources and migrating them safely is a major concern, 
+this is a subset of the more general problem of trying to understand
+what updates to Kro resources will update your cluster state. We will 
+recommend using future CLI commands 
+like `kro preview` to help customers understand the actions that will be taken on their cluster.
 
-##### Shared
+The general workflow will be to design your RGD and run the CLI command `kro preview` to
+go through an interactive diff of the resources the RGD will expect to create.
+The user will be responsible for understanding and confirming this diff is correct.
 
-On create shared mode will error if the resource is exclusively owned. If it is not,
-we will apply and start sharing ownership of the fields. In shared mode we will always
-SSA with force=false. If we do not conflict we will just add ourselves as managing the field.
+There will be some RGDs that will be non-deterministic (a job that fails if rand() > 0.5), 
+but these will be the minority of RGDs, especially in migration use cases. There is also 
+the potential that you run a dry-run and the cluster state changes when you go to apply the resource.
+This is not a unique problem to Kro adoption. In system migrations it is typical to avoid making
+updates to resources that are in the process of undergoing a migration.
 
-If we do conflict we will check if this is a terminal conflict that needs to be
-surfaced to users. We consider conflicts terminal if we do not own the field and
-another Kro field manager does. Otherwise, we consider these conflicts overridable.
+The goal here is to leverage an already planned feature of the CLI 
+and enable safer adoption in the majority of cases. 
 
-| We own field | Another Kro owns field | Conflict Type |
-|--------------|------------------------|---------------|
-| true | false | No conflict |
-| true | true | Overridable |
-| false | false | Overridable |
-| false | true | **Terminal** |
+We will update documentation to clarify this as a stated feature of Kro.
 
-Terminal conflicts will be surfaced in the instance status condition until the conflict is resolved.
+##### Orphan
 
-Our mechanism for overriding conflict will be a merge-patch editing the field
-managers directly with a precondition checking that the resource version of
-the object has not changed to avoid race conditions.
+For the purpose of orphaning, we consider both deleting an RGD or instance and 
+pruning (where a resource is deleted because the instance or RGD definition no longer requires it) 
+to apply the same mechanism. 
 
-This mechanism ensures we can make updates to fields without leading into
-deadlock. For example, if we have instance A and B apply the same object
-with replicas=3, any attempt to update either A or B will run into a conflict.
-If we just forced over the conflict then we would run into the two objects
-fighting each other. When we update A with replicas=5, we will kick B out of
-field ownership. When B tries to apply again it will see a conflict, realize
-that it is no longer an owner of replicas and then report a terminal
-conflict. Users would be expected to see this conflict and resolve it
-on a case by case basis.
+When orphaning, we will check if the lifecycle.deletePolicy field is set to Retain and remove Kro-applied labels
+instead of deleting the object.
 
-On instance delete or pruning if the rgd no longer requires the resource, if there 
-are no other owners we just delete the object as is. If there are other owners we do
-an empty SSA releasing ownership of all of our fields.
+We will leave the Kro field manager for simplicity. Parsing and patching out the managed field
+is extra complexity. Leaving the field manager also gives clarity about what Kro contributed to
+the object. It will require other systems to run with force=true but this is pretty common, Kro 
+does it.
 
-These semantics allow partial objects. With this you can define a Kro resource
-that just changes a single field or a few fields.
+CEL expressions that reference the schema will be allowed. This will allow creating 
+non-prod vs production RGDs that may have different requirements on retaining resources.
+We will avoid allowing dependencies to other resources for simplicity initially.
 
-These also allow singletons and shared dependencies between Kro installations.
-Using these shared dependencies can accomplish migrating a resource safely
-between Kro instances and RGDs.
+Example with CEL expression:
+```yaml
+resources:
+  - id: database
+    lifecycle:
+      deletePolicy: '${schema.spec.environment == "production" ? "Retain" : "Delete"}'
+    template:
+      apiVersion: v1
+      kind: PersistentVolumeClaim
+      # ...
+``` 
 
-#### Design details
+##### Migration
 
-There are two main complexities with this design
+An example migration use case is refactoring your RGDs. You may have 
+initially bundled a webserver and a backend API into one RGD but now want to organize
+them separately so different teams can own their own infrastructure.
 
-##### Label updates
+You may have more complicated migrations. You may have an application that was an instance
+per namespace and want to refactor it to be a single instance utilizing collections. 
 
-We have a large amount of labels that only support a single Kro instance managing
-the resource like `applyset.kubernetes.io/part-of={applySetID}`.
+Example migration steps
+```
+  # Step 1: Add retain to old RGD
+  apiVersion: kro.run/v1alpha1
+    kind: ResourceGraphDefinition
+    spec:
+      resources:
+        - id: database
+          lifecycle:
+            deletePolicy: Retain # Set deletePolicy to Retain
 
-We need to keep the old labels around for compatibility but start adding additional
-labels to be of the form `applyset-{applySetID}` to support multiple owners.
+  # Step 2: Delete old RGD (resources get orphaned)
+  kubectl delete rgd monolith
 
-We will track an annotation called `internal.kro.run/migrated-to-new-labels` that we will look
-at to know if we should use the new or old labels in label selectors.
+  # Step 3: Create new RGDs that reference the same resources
+  # Kro will adopt them since they're no longer managed
+```
 
-##### Field manager
+## Other solutions considered
 
-The actual parsing and updating of field managers is non-trivial. We will need to be very careful to
-correctly parse the error message and be able to successfully update field managers.
+### Adoption safety policies
+
+There are various levels of safety we could build into Kro for adoption. We could add policies like `AdoptOrCreate`,
+`AdoptOnly`, `CreateOnly` to give finer control over adopt behaviour. 
+
+The issue with these policies is that while they do provide some safety, they are drastically less safe than 
+a preview command like `kro preview`. The preview command would let you understand not just if resources will be
+created but also if any updates are potentially disruptive. If you successfully adopt a database resource, but during 
+that adoption set memory to 10 bytes, you just caused an outage.
+
+### Adoption matching
+
+We could guarantee that adopted resources will not change initially. We could do a Kubernetes server side 
+dry run for resources with an `Adopt` policy and try to validate that there will be no differences.
+
+Performing the diff will be non-trivial since we will intend to make some updates like adding labels.
+This approach is also not perfect, admission webhooks with side effects could be skipped. We would also run into
+an issue if any updates happen between our dry run and applying the object.
+
+This idea has merit and could be a potentially valuable feature. However, this would be 
+prematurely solving the problem of adoption safety. We should see gaps from the CLI preview
+command before adding this complexity.
+
+### CRD field vs annotation
+
+We could choose an annotation `alpha.kro.run/delete-policy: retain` instead of a CRD field.
+
+Choosing an annotation over a field has the following benefits:
+1. A CRD field is harder to make breaking changes to. An alpha annotation will allow us to evolve the project quicker and easier than committing to an RGD CRD field now.
+2. Annotations are more easily understood by the Kubernetes ecosystem. It would be very easy for a platform team to automatically add annotations or require all database CRs have this annotation present.
+3. Easier to inspect and have confidence a resource will not be deleted.
+4. Allows templating mechanisms to transfer over.
+
+The reason to use a CRD field instead of an annotation is type safety. It would be too easy for a customer to accidentally 
+typo the annotation then delete their production database. We could partially type check the annotation
+but nothing could stop a user from accidentally specifying something like `rko.run`. 
+
+### Top level
+
+A potential enhancement would be to orphan all resources in an RGD. We could add a top level policy field then allow overriding
+at the lower level. This is reasonable but not initially necessary. It is very possible the idea of defaulting CRD fields
+per RGD will be solved more generally (users already want to default a list of labels to apply to everything).
 
 ## Scoping
 
 #### What is in scope for this proposal?
 
-This proposal addresses resource contribution, singletons, and migrating resources between Kro installations safely.
+The scope of this proposal is minimal; all this document proposes is adding the lifecycle.deletePolicy field.
 
 #### What is not in scope?
 
-CRUD policies are not in scope. One could imagine modeling `delete: retain` as setting a dummy
-owner like `applyset-retain: true`. This would not work however since our semantics for delete
-include doing an empty SSA. This empty SSA would fail if there is a dummy owner that does not
-own any fields.
+##### kro CLI
 
-Migrating into Kro is not in scope. To handle migration into Kro, a robust dry run would be ideal
-to preview what would happen.
+While the kro CLI preview is referenced, exact details will be left to the CLI KREP.
 
-## Other solutions considered
+##### Other CRUD policies
 
-The most controversial part of the solution will be handling conflicts.
+- I want to run resources on deletion because I may want to snapshot a database before deleting it.
+- I want to be able to update immutable fields in Kro because I want to change a job spec.
+- I want to create a resource and never update it because some resources like PVC might be expensive or difficult to update.
+- I want to only read a resource because I don't want to use external refs and want to reference the resource.
+- I want to control deletion order because I may have implicit dependencies or a more complicated teardown workflow needed for resource cleanup.
 
-There are a couple of alternatives that we could take to avoid fiddling with field managers manually.
+These use cases will need to be tackled separately. 
 
-The fundamental issue that makes this hard is SSA operates on a binary force=true or force=false. We really
-want to force=true for fields we can overwrite and force=false for fields we don't own but another Kro instance owns.
+##### Singletons and resource contribution
 
-### Accept some race conditions
+- I want to define applications with resources that are defined only once across all Kro instances because resources like PrivateEndpoint are one per cluster.
+- I want to create a controller that updates a subset of fields on a resource because it is a useful pattern to control a small number of fields in Kubernetes.
 
-Simplest idea could be
-
-1. apply with force=false
-2. if we run into conflicts check if they are all on fields we own, then apply with force=true
-
-Issue is if object state is
-```
-replicas (owned by A, owned by B): 4
-```
-
-1. A attempts to apply replicas=5, image=aNewImage, fails with conflict over replicas
-2. B successfully applies image=bNewImage
-3. A only sees replicas conflict, then thinks it's good to force apply
-
-With this we just overwrote another Kro's image field that we don't own. We only thought it was safe to
-force because we saw the single conflict but there was another one that we would not have decided to override.
-We don't thrash from here, A would own both fields and B would report a conflict
-but, we don't have any strong guarantees in the system
-
-### Apply twice
-
-We could try and split the patch into overridable fields and apply with force=true while we
-apply non overridable fields with force=false.
-
-This is mainly listed to be exhaustive for possibilities but this solution feels like it adds complexity
-with no strong guarantees.
+A separate KREP will address these. 
 
 ## Testing strategy
 
@@ -169,6 +201,6 @@ Testing will follow existing patterns using chainsaw for e2e tests. No special i
 Unit tests will be added to cover changed code.
 
 End-to-end tests will validate
-- resource contribution example
-- shared ownership with conflict cases
-- shared ownership failing to use exclusive resource 
+- orphaning resources with deletePolicy: Retain
+- validating we can adopt a resource that has been orphaned
+- pruning behavior respects deletePolicy field
--- a/docs/design/proposals/resource-lifecycles.md
+++ b/docs/design/proposals/resource-lifecycles.md
@@ -2,164 +2,161 @@
 
 ## Problem statement
 
-The problem we are trying to solve are resource adoption and resource orphaning. 
+Resource lifecycles and ownership have a wide potential scope. Here are some user stories grouped by broad use case.
 
-The goal of resource adoption is to enable customers to safely migrate resources created outside Kro to be managed by Kro. For example, you may have run `kubectl apply -f ...` on a database CR but now want to have Kro manage this resource without causing downtime or impact to existing services. The customer in this case would want the instance to error if it is not found instead of creating a duplicate resource. Customers may want to manage resources under Kro to enable things like RGD configuration updates to propagate to the resource.
+CRUD policies
+- I want to run resources on deletion because I may want to snapshot a database before deleting it.
+- I want to be able to update immutable fields in Kro because I want to change a job spec.
+- I want to create a resource and never update it because some resources like PVC might be expensive or difficult to update.
+- I want to only read a resource because I don't want to use external refs and want to reference the resource.
+- I want to avoid deleting resources because some resources like databases are too risky to be deleted through Kro.
+- I want to control deletion order because I may have implicit dependencies or a more complicated teardown workflow needed for resource cleanup.
 
-The goal of resource orphaning is to allow customers to leave resources behind. An example could be you may want to switch from a Kro controller per namespace to a single Kro installation per cluster. You would want to Orphan resources to allow the other Kro installation to manage the resources.
+Migration into Kro and between Kro resources
+- I want to safely do migrations into Kro from existing resources because I want to manage my cluster in Kro and want to avoid downtime and risk of updating resources poorly.
+- I want to safely do migrations between Kro RGDs or instances because I want to be able to refactor and evolve my Kro definitions over time and avoid downtime.
+
+Singletons
+- I want to define applications with resources that are defined only once across all Kro instances because resources like PrivateEndpoint are one per cluster.
+
+Resource contribution
+- I want to create a controller that updates a subset of fields on a resource because it is a useful pattern to control a small number of fields in Kubernetes.
 
 ## Proposal
 
-The proposal is to add a `resourcePolicy` string enum option to the resource configuration with various configuration options.
-```
-resourcePolicy: Owner | Adopt | Orphan 
-```
+The proposal is to add a new mode of operating resources
+through an annotation called `kro.run/ownership=shared`.
+
+Shared mode will be a new primitive that solves migrating
+between Kro resources, Singletons, and resource contribution.
 
 #### Overview
 
-The default resource policy will be `Owner`. This matches the current behaviour today. 
+##### Exclusive
 
-`Adopt` will avoid creating resources and will take ownership of existing resources and manage them as if kro had created them. `Adopt` will throw an error if resources do not exist.
+The default mode is exclusive. A Kro resource applied with exclusive
+will cause errors for any other Kro instance trying to use it. Otherwise,
+exclusive will work the same as it does today. Exclusive resources will
+adopt resources from other systems but raise errors if they have Kro
+applyset labels.
 
-`Orphan` allows the instance to be deleted without deleting certain resources. Kro removes the ApplySet ownership labels, leaving the resources orphaned and unmanaged.
+##### Shared
 
-We will support configuring these with cel expressions. For example, a schema option can be used on the RGD to determine if the instance should be `Adopt` or `Owner`.
-```
-resources:
-  - id: database
-    kind: RDSInstance
-    resourcePolicy: ${has(schema.spec.adoptResource) ? "Adopt" : "Owner"}
-```
+On create shared mode will error if the resource is exclusively owned. If it is not,
+we will apply and start sharing ownership of the fields. In shared mode we will always
+SSA with force=false. If we do not conflict we will just add ourselves as managing the field.
 
-We would only initially support referencing schema fields in the cel expressions analogous to how `includeWhen` works today.
+If we do conflict we will check if this is a terminal conflict that needs to be
+surfaced to users. We consider conflicts terminal if we do not own the field and
+another Kro field manager does. Otherwise, we consider these conflicts overridable.
+
+| We own field | Another Kro owns field | Conflict Type |
+|--------------|------------------------|---------------|
+| true | false | No conflict |
+| true | true | Overridable |
+| false | false | Overridable |
+| false | true | **Terminal** |
+
+Terminal conflicts will be surfaced in the instance status condition until the conflict is resolved.
+
+Our mechanism for overriding conflict will be a merge-patch editing the field
+managers directly with a precondition checking that the resource version of
+the object has not changed to avoid race conditions.
+
+This mechanism ensures we can make updates to fields without leading into
+deadlock. For example, if we have instance A and B apply the same object
+with replicas=3, any attempt to update either A or B will run into a conflict.
+If we just forced over the conflict then we would run into the two objects
+fighting each other. When we update A with replicas=5, we will kick B out of
+field ownership. When B tries to apply again it will see a conflict, realize
+that it is no longer an owner of replicas and then report a terminal
+conflict. Users would be expected to see this conflict and resolve it
+on a case by case basis.
+
+On delete, if there are no other owners we just delete the object as is. If
+there are other owners we do an empty SSA releasing ownership of all of our
+fields.
+
+These semantics allow partial objects. With this you can define a Kro resource
+that just changes a single field or a few fields.
+
+These also allow singletons and shared dependencies between Kro installations.
+Using these shared dependencies can accomplish migrating a resource safely
+between Kro instances and RGDs.
 
 #### Design details
 
-We would need to update the Resource config option to take in `resourcePolicy`. 
+There are two main complexities with this design
 
-##### Adopt 
+##### Label updates
 
-Currently, Kro handles most of the adoption use case. If you have an instance point to a preexisting Kubernetes object, Kro will take ownership of it. The current behaviour is essentially `AdoptOrCreate`.
+We have a large amount of labels that only support a single Kro instance managing
+the resource like `applyset.kubernetes.io/part-of={applySetID}`.
 
-To get the `Adopt` semantics, we will need to update the reconciliation loop to throw an error if the resource does not exist and has `resourcePolicy: Adopt`.
+We need to keep the old labels around for compatibility but start adding additional
+labels to be of the form `applyset-{applySetID}` to support multiple owners.
 
-The one exception to this resource adoption is if another Kro instance is managing the resource. Then a `resource belongs to a different ApplySet` will be thrown. This is not planned to be changed to prevent Kro installations fighting over a resource.
+We will track an annotation called `internal.kro.run/migrated-to-new-labels` that we will look
+at to know if we should use the new or old labels in label selectors.
 
-#### Orphan
+##### Field manager
 
-The only difference for resources with `Orphan` is deletion will remove ApplySet labels with an update call instead of deleting the Kubernetes resource.
-
-When an instance with `resourcePolicy: Orphan` is deleted, Kro will remove the `applyset.kubernetes.io/part-of` label from the resource and skip issuing the delete call. The resource remains in the cluster without any ownership labels.
-
-Removing the ownership labels allows other Kro instances to take ownership of this.
+The actual parsing and updating of field managers is non-trivial. We will need to be very careful to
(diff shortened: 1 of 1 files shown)


docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+## Proposal
+
+The proposal is to add a `resourcePolicy` string enum option to the resource configuration with various configuration options.
+```
+resourcePolicy: Owner | Adopt | Orphan 
+```
+
+#### Overview
+
+The default resource policy will be `Owner`. This matches the current behaviour today.
```
cheeseandcereal: Compared to 'adopt/oprhan', is the word 'owner' a little confusing?

I mean, once a resources is adopted, wouldn't the behavior of 'owner' and 'adopt' be the same?

And for 'orphan', wouldn't this just be something that needs to be updated only before a deletion to control deletion behavior? It almost feels like that's a bit unrelated from creation controls that owner/adopt provide.

I wonder if 'deletion policy', and 'creation policy' would make more sense as dimensions than trying to combine these into a single thing.

If I want to do something like 'adopt' on creation, and 'orphan' on deletion, it will require a resource update after creation if these are combined as-is.

NicholasBlaskey: > I mean, once a resources is adopted, wouldn't the behavior of 'owner' and 'adopt' be the same?

Yeah correct

> I wonder if 'deletion policy', and 'creation policy' would make more sense as dimensions than trying to combine these into a single thing.

This is the basically the alternative proposed. This was called out as a foot gun in a previous discussion because it might be weird to think about Kro only managing the update lifecycle of a resource and not the create or deletion.
https://github.com/kubernetes-sigs/kro/issues/542#issuecomment-3202236827

If this is a common usecase we could also make a new policy called `AdoptAndOrphan`.

I'm also not opposed to doing the other solution of having multiple dimensions to configure.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
+# Resource lifecycles
+
+## Problem statement
+
+The problem we are trying to solve are resource adoption and resource orphaning. 
+
+The goal of resource adoption is to enable customers to safely migrate resources created outside Kro to be managed by Kro. For example, you may have run `kubectl apply -f ...` on a database CR but now want to have Kro manage this resource without causing downtime or impact to existing services. The customer in this case would want the instance to error if it is not found instead of creating a duplicate resource. Customers may want to manage resources under Kro to enable things like RGD configuration updates to propagate to the resource.
```
a-hilaly: FYI wrapping markdown lines can still render as normal paragraph but make reviews easier and "targeted" - https://martin-ueding.de/posts/hard-vs-soft-line-wrap/

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+The goal of resource orphaning is to allow customers to leave resources behind. An example could be you may want to switch from a Kro controller per namespace to a single Kro installation per cluster. You would want to Orphan resources to allow the other Kro installation to manage the resources.
+
+## Proposal
+
+The proposal is to add a `resourcePolicy` string enum option to the resource configuration with various configuration options.
+```
+resourcePolicy: Owner | Adopt | Orphan 
+```
```
a-hilaly: are there any similar concepts in terraform/cloudformation/pulumi we can get inspired from?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Overview
+
+The default resource policy will be `Owner`. This matches the current behaviour today. 
+
+`Adopt` will avoid creating resources and will take ownership of existing resources and manage them as if kro had created them. `Adopt` will throw an error if resources do not exist.
```
a-hilaly: what if the resource was owned by another kro instance/rgd? what's be behaviour?

docs/design/proposals/resource-lifecycles.md
ellistarn: I think this might overlap with resource contribution.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+`Orphan` allows the instance to be deleted without deleting certain resources. Kro removes the ApplySet ownership labels, leaving the resources orphaned and unmanaged.
+
+We will support configuring these with cel expressions. For example, a schema option can be used on the RGD to determine if the instance should be `Adopt` or `Owner`.
+```
+resources:
+  - id: database
+    kind: RDSInstance
+    resourcePolicy: ${has(schema.spec.adoptResource) ? "Adopt" : "Owner"}
```
a-hilaly: how about just `policy` or `policies`, i'm specifically thing that a resource can have multiple policies. e.g `Owner` and `Orphan`

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+The goal of resource adoption is to enable customers to safely migrate resources created outside Kro to be managed by Kro. For example, you may have run `kubectl apply -f ...` on a database CR but now want to have Kro manage this resource without causing downtime or impact to existing services. The customer in this case would want the instance to error if it is not found instead of creating a duplicate resource. Customers may want to manage resources under Kro to enable things like RGD configuration updates to propagate to the resource.
+
+The goal of resource orphaning is to allow customers to leave resources behind. An example could be you may want to switch from a Kro controller per namespace to a single Kro installation per cluster. You would want to Orphan resources to allow the other Kro installation to manage the resources.
+
+## Proposal
+
+The proposal is to add a `resourcePolicy` string enum option to the resource configuration with various configuration options.
+```
+resourcePolicy: Owner | Adopt | Orphan
```
ellistarn: the parent field is resources, so should this be `resources[i].policy` instead of `resources[i].resourcePolicy`?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+We will support configuring these with cel expressions. For example, a schema option can be used on the RGD to determine if the instance should be `Adopt` or `Owner`.
+```
+resources:
+  - id: database
+    kind: RDSInstance
+    resourcePolicy: ${has(schema.spec.adoptResource) ? "Adopt" : "Owner"}
+```
+
+We would only initially support referencing schema fields in the cel expressions analogous to how `includeWhen` works today.
```
a-hilaly: `includeWhen` supports referencing other resources now. I'm not sure if the policy should allow that as well. in fact i'm not super sure we want to support CEL expressions in `resourcePolicy`

michaelhtm: nit: this functionality has now changed. https://github.com/kubernetes-sigs/kro/pull/1104

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Design details
```
a-hilaly: I think this section is missing an `Owner` implementation detail, especially for situation where a resource is owned by another RGD/instance

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: conflict due to applyset labels

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+##### Additional resource policies
+
+This design does allow for the expansion of additional resource policies.
+
+One idea discussed before is `SharedOwnership`. This would effectively be a singleton. You may have resources that should be created at most once per cluster. `PrivateEndpoint` is an example of this. Multiple instances would point to the same object and only create it if the object does not exist. They would only delete it when they are the last instance to manage it. This feature should be implemented eventually but carries enough complexity that decoupling this from this design is needed.
+
+Another option could be `watcher`. This has significant overlap with externalRefs so is avoided for now. In the future we could consider deprecating externalRef and making `watcher` solve this use case.
+
+## Other solutions considered
```
ellistarn: What about an approach that uses metadata on the underlying objects? i.e. imagine KRO keeps track of all references. If KRO points at a resource that already exists, but doesn't have a reference, it adopts. If KRO tries to remove a resource that doesn't already exist, it abandons. This also covers the multi-owner migration approach. I'm not sure it's right, but we should explore. 

We'd need to think through how this relates to `Contribute` as well.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+We would need to update the Resource config option to take in `resourcePolicy`. 
+
+##### Adopt 
+
+Currently, Kro handles most of the adoption use case. If you have an instance point to a preexisting Kubernetes object, Kro will take ownership of it. The current behaviour is essentially `AdoptOrCreate`.
+
+To get the `Adopt` semantics, we will need to update the reconciliation loop to throw an error if the resource does not exist and has `resourcePolicy: Adopt`.
+
+The one exception to this resource adoption is if another Kro instance is managing the resource. Then a `resource belongs to a different ApplySet` will be thrown. This is not planned to be changed to prevent Kro installations fighting over a resource.
```
a-hilaly: This reminds me of reference counting, maybe we need a section on it.

michaelhtm: could this happen if another controller that uses applyset owns the resource?
Also, should we support two kro instances to co-own a resource?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Orphan
+
+The only difference for resources with `Orphan` is deletion will remove ApplySet labels with an update call instead of deleting the Kubernetes resource.
```
a-hilaly: I think it should remove all the kro labels

docs/design/proposals/resource-lifecycles.md
pnz1990: adoption is a creation decision and orphaning is a deletion decision, and they should be independent imo?

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: have some other examples in notes section

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+##### adoption checking ownership with other systems
+
+One situation that may come up is you adopt a pod that is being managed by a replicaset. This may lead to both the replicaset and Kro making updates to this resource.
+
+This will be left up to the user to validate anything they adopt is not going to be managed by other resources in unexpected ways.
+
+##### adoption doesn't cause updates
+
+One problem could be you may want to have a migration of resources without any changes to those resources. To do this you may want a feature where you ensure a resource can be adopted without making any updates to the field.
```
a-hilaly: Isn't this just external references?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+To get the `Adopt` semantics, we will need to update the reconciliation loop to throw an error if the resource does not exist and has `resourcePolicy: Adopt`.
+
+The one exception to this resource adoption is if another Kro instance is managing the resource. Then a `resource belongs to a different ApplySet` will be thrown. This is not planned to be changed to prevent Kro installations fighting over a resource.
+
+#### Orphan
+
+The only difference for resources with `Orphan` is deletion will remove ApplySet labels with an update call instead of deleting the Kubernetes resource.
+
+When an instance with `resourcePolicy: Orphan` is deleted, Kro will remove the `applyset.kubernetes.io/part-of` label from the resource and skip issuing the delete call. The resource remains in the cluster without any ownership labels.
```
pnz1990: What's the behavior for Orphan resources during updates, before the instance is ever deleted? If I patch the instance spec, does kro still reconcile changes onto the resource? The doc only covers the deletion path. If kro does keep updating it, the resource isn't really orphaned until deletion — which is fine, just needs to be said.

pnz1990: If kro skips updates too, that's a meaningful behavior difference that needs to be explicit both here and in the implementation.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+Removing the ownership labels allows other Kro instances to take ownership of this.
+
+## Scoping
+
+#### What is in scope for this proposal?
+
+The scope for this proposal is just the `Orphan` and `Adopt` use case.
```
michaelhtm: would `Owner` fail if the resource exists and is not managed by the kro instance?

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: One disadvantage to this is we might get unclear semantics 

Say we introduced a sharedowner in the future, does sharedowner and adopt overlap? Can you add sharedowner and owner? Might be not clear to try and reason about the overlap for users

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+This is out of scope to avoid adding extra complexity into this feature.
+
+##### updatePolicy
+
+Some resources like job spec are immutable and can not be updated by a typical patch command. Instead of issuing an update, a user may want to have the option of configuring a resource to be deleted and recreated.
+
+A potential design could have considered this as a part of the resource policy.
+
+For the given proposal `updatePolicy` does not fit in clearly and would end up being a separate field and topic.
```
a-hilaly: Just thinking outloud what if we did something like:
```
resources:
- id: r
  policies:
  - AdoptOrCreate
  - Orphan
  - UpdateOnImmutableErrors??
```

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Design details
+
+We would need to update the Resource config option to take in `resourcePolicy`. 
+
+##### Adopt 
+
+Currently, Kro handles most of the adoption use case. If you have an instance point to a preexisting Kubernetes object, Kro will take ownership of it. The current behaviour is essentially `AdoptOrCreate`.
+
+To get the `Adopt` semantics, we will need to update the reconciliation loop to throw an error if the resource does not exist and has `resourcePolicy: Adopt`.
```
pnz1990: Does a missing Adopt resource block the whole instance from becoming Ready, or just that one resource?

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: agreed should not be ```resources.resourcePolicy``` -- though policy does seem somewhat vague

docs/design/proposals/resource-lifecycles.md
pnz1990: also what does status.conditions look like

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: yeah can be a two way door to add more cel expression support later

I do think that setting this per instance could have a valid usecase otherwise your RGDs may not be flexible for different use cases

docs/design/proposals/resource-lifecycles.md
michaelhtm: should we rename the field `createPolicy`? if we consider to add `updatePolicy`

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: Would you expect ```Owner``` to be any different than current behavior today?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
...
+
+End to end tests will validate
+- failing to adopt a resource because it is missing
+- successfully adopting a resource
+- validating the adopted resource can be updated
+- orphaning a resource
+- validating that another instance can adopt this resource
+- failing to adopt a resource owned by another instance
+
+## Discussion and notes
```
JamesGAWS: Can we add a future directions calling out other resourcePolicy values we can add in the future?

docs/design/proposals/resource-lifecycles.md
pnz1990: If someone flips schema.spec.adoptResource on an already-running instance, does kro re-evaluate and do something?

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: included ```sharedOwner``` in not in scope, is that what you had in mind?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,210 @@
+# Resource lifecycles
+
+## Problem statement
+
+The problem we are trying to solve are resource adoption and resource orphaning.
```
JamesGAWS: Can we start this krep by enumerating all possible use cases? We can pick and choose what we want to implement first but an overview of the options will help support a consistent UX.

Possible use cases not addressed in this krep are order changes on delete, create and create and run resources for a period before finishing delete (for backup or cleanup jobs for example).

vrabbi:
I think an approach like in crossplane of a management policies field which recieves an array of policies allowed is nice but confusing for end users sometimes.
I think having a policies field with then sub fields creation, update, delete and for each of those allow a different set of relevant values for example 
policies.delete=orphan 
policies.create=adopt
policies.update=ReplaceOnImmutableChange

This seems more explicit and easy for a user to understand.

I do think an ObserveOnly mode like in crossplane would be really nice to have as well though which may be something like
Policies.create=Never
Policies.update=Never
Policies.delete=Orphan

It may also need a policies.read=Yes

Naming and value names is not something im good at, but i think the more explicit the types of policies the easier it will be for users to reason about

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,174 @@
...
+SSA with force=false. If we do not conflict we will just add ourselves as managing the field.
+
+If we do conflict we will check if this is a terminal conflict that needs to be
+surfaced to users. We consider conflicts terminal if we do not own the field and
+another Kro field manager does. Otherwise, we consider these conflicts overridable.
+
+| We own field | Another Kro owns field | Conflict Type |
+|--------------|------------------------|---------------|
+| true | false | No conflict |
+| true | true | Overridable |
```
michaelhtm: if two "kro"s own the field, wouldn't they start fighting?
What are the cases where we want to override this?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,174 @@
...
+
+This mechanism ensures we can make updates to fields without leading into
+deadlock. For example, if we have instance A and B apply the same object
+with replicas=3, any attempt to update either A or B will run into a conflict.
+If we just forced over the conflict then we would run into the two objects
+fighting each other. When we update A with replicas=5, we will kick B out of
+field ownership. When B tries to apply again it will see a conflict, realize
+that it is no longer an owner of replicas and then report a terminal
+conflict. Users would be expected to see this conflict and resolve it
+on a case by case basis.
```
michaelhtm: does this only apply to two kro instances? Or could we see HPA and kro co-owning a field?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,174 @@
...
+
+On instance delete or pruning if the rgd no longer requires the resource, if there 
+are no other owners we just delete the object as is. If there are other owners we do
+an empty SSA releasing ownership of all of our fields.
+
```
michaelhtm: would this also set all the fields kro owns to `null`?

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: No trashing should ever occur, if there is no conflict there is no need to make constant updates

If there is a conflict one of the Kros will lose and see a conflict error

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: Yes we could, however we would see it happen in the same way we do today 

I this the responsibility will end up being on the user to avoid external systems fighting with kro

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: I believe SSA semantics would release the field ownership and set it to the value of other owners

If we were the only owner of a field then it would get set to null

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,174 @@
...
+the object has not changed to avoid race conditions.
+
+This mechanism ensures we can make updates to fields without leading into
+deadlock. For example, if we have instance A and B apply the same object
+with replicas=3, any attempt to update either A or B will run into a conflict.
+If we just forced over the conflict then we would run into the two objects
+fighting each other. When we update A with replicas=5, we will kick B out of
+field ownership. When B tries to apply again it will see a conflict, realize
+that it is no longer an owner of replicas and then report a terminal
+conflict. Users would be expected to see this conflict and resolve it
```
spattk: What would be the UX in this case?
Is that in the scope or out of scope?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,174 @@
...
+between Kro resources, Singletons, and resource contribution.
+
+#### Overview
+
+##### Exclusive
+
+The default mode is exclusive. A Kro resource applied with exclusive
+will cause errors for any other Kro instance trying to use it. Otherwise,
+exclusive will work the same as it does today. Exclusive resources will
+adopt resources from other systems but raise errors if they have Kro
```
cheeseandcereal: I wonder if we should re-evaluate any of this behavior with this change. It feels a little arbitrary to adopt resources from other systems, but not with kro specifically.

"Exclusive" doesn't really imply this behavior difference imo.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: They see in their status: failed to apply due to field conflicts ..., along with some information about the conflicts

They then would have to update one or more of their rgds to no longer have the conflict

docs/design/proposals/resource-lifecycles.md
cheeseandcereal: > an empty SSA releasing ownership of all of our fields

This isn't exactly how this works. If you are the sole owner of a field, doing an SSA without that field will tell kubernetes to attempt to delete that field. It will not simply "release ownership" without modifying the field unless there is a joint owner.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: Yeah this has come up in some discussions. Right now Kro is safe by default only if there is another Kro instance managing the resource. Kro is unsafe by default if it is not Kro.

I don't think this would be an easy behavior change to make currently 

FWIW exclusive isn't even being defined explicitly to users (the ownership annotation is either shared or it is not)

docs/design/proposals/resource-lifecycles.md
spattk: > If there is a conflict one of the Kros will lose and see a conflict error

Just curious, is there a possibility of picking up a bad/later update and the cluster gets stuck in a wedged state until the user take actions which would have been prevented had we taken the other update?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,174 @@
...
+
+##### Shared
+
+On create shared mode will error if the resource is exclusively owned. If it is not,
+we will apply and start sharing ownership of the fields. In shared mode we will always
+SSA with force=false. If we do not conflict we will just add ourselves as managing the field.
+
+If we do conflict we will check if this is a terminal conflict that needs to be
+surfaced to users. We consider conflicts terminal if we do not own the field and
+another Kro field manager does. Otherwise, we consider these conflicts overridable.
```
cheeseandcereal: > another Kro field manager

Does this imply we need to change the field manager to some deterministic string based on the RGD?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,174 @@
+# Resource lifecycles
+
+## Problem statement
+
+Resource lifecycles and ownership have a wide potential scope. Here are some user stories grouped by broad use case.
+
+CRUD policies
+- I want to run resources on deletion because I may want to snapshot a database before deleting it.
```
cheeseandcereal: What does this have to do with shared/exclusive mode proposed below? I'm not sure I connect the dots.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: would be applyset id which is generated from the gvk / namespace / name of the instance

docs/design/proposals/resource-lifecycles.md
cheeseandcereal: > If there are other owners we do an empty SSA

Changing delete behavior based on field ownership is a dangerous game. It can cause really un-intuitive side effects. For example, running HPA against a deployment now means that kro's deletion behavior totally changes. Or someone manually `kubectl edit` a resource once could cause a similar permanent future behavior change as long as that object exists (but on delete/recreate it will have different behavior).

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: Yes, we essentially have first write wins semantics. 

Its very possible your first write is wrong and then your manual updates to other instances won't be able to fix it.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
+# Resource lifecycles
+
+## Problem statement
+
+Resource lifecycles and ownership have a wide potential scope. A further discussion on other potential items is listed in the not-in-scope section.
+
+This document addresses the following user stories:
+- I want to avoid deleting resources because some resources like databases are too risky to be deleted through Kro.
+- I want to safely do migrations into Kro from existing resources because I want to manage my cluster in Kro and want to avoid downtime and risk of updating resources poorly.
+- I want to safely do migrations between Kro RGDs or instances because I want to be able to refactor and evolve my Kro definitions over time and avoid downtime.
```
jakobmoellerdev, line 10: Hmmm, Not sure I do not want to handle KRO Cross-RGD migration with resource lifecycles because it would mean that we would have to have better orphan behavior in here for existing RGDs.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+apiVersion: kro.run/v1alpha1
+kind: ResourceGraphDefinition
+spec:
+  resources:
+    - id: database
+      lifecycle: "${policy().withRetain()}"
+```
+
+The `policy()` function provides methods to configure lifecycle behavior:
+- `policy().withRetain()` - Retain resource when instance is deleted
```
jakobmoellerdev, line 30: Is this referring to the resource or the actual instance?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+- `policy().withDelete()` - Delete resource (default)
+
+The CEL expression evaluates to a map. Static maps are also supported but not recommended to avoid future breakages:
```
jakobmoellerdev, line 33: For me this here is equivalent to a Policy field and I dont see value in having this in a Cel Sentinel. Why do we want to offer something on top of the deletePolicy?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+The general workflow will be to design your RGD and run the CLI command `kro preview` to
+go through an interactive diff of the resources the RGD will expect to create.
+The user will be responsible for understanding and confirming this diff is correct.
+
+There will be some RGDs that will be non-deterministic (a job that fails if rand() > 0.5), 
+but these will be the minority of RGDs, especially in migration use cases. There is also 
+the potential that you run a dry-run and the cluster state changes when you go to apply the resource.
+This is not a unique problem to Kro adoption. In system migrations it is typical to avoid making
+updates to resources that are in the process of undergoing a migration.
```
jakobmoellerdev, line 61: I dont understand how this impacts the proposal above? to me it feels like 2 problems described separately

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+The goal here is to leverage an already planned feature of the CLI
```
jakobmoellerdev, line 63: I dont understand which feature and how you plan to do this

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+##### Orphan
+
+For the purpose of orphaning, we consider both deleting an RGD or instance and 
+pruning (where a resource is deleted because the instance or RGD definition no longer requires it) 
+to apply the same mechanism. 
+
+When orphaning, we will evaluate the lifecycle expression and check if deletePolicy is set to retain, 
+then remove Kro-applied labels instead of deleting the object.
+
+We will leave the Kro field manager for simplicity. Parsing and patching out the managed field
```
jakobmoellerdev, line 77: this is not good IMHO. we need to remove the field manager once we orphan because otherwise it will look like kro is still managing the object

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+is extra complexity. Leaving the field manager also gives clarity about what Kro contributed to
+the object. It will require other systems to run with force=true but this is pretty common, Kro 
+does it.
```
jakobmoellerdev, line 80: This is not the correct behavior long term. KRO needs to become smarter and not force everytime. Also other tools dont force by default if they are not configured to do so.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+CEL expressions that reference the schema will be allowed. This will allow creating 
+non-prod vs production RGDs that may have different requirements on retaining resources.
+We will avoid allowing dependencies to other resources for simplicity initially.
+
+Example with conditional CEL expression:
+```yaml
+resources:
+  - id: database
+    lifecycle: '${schema.spec.environment == "production" ? policy().withRetain() : policy()}'
```
jakobmoellerdev, line 90: this cel expression does not read cleanly. This could be implemented with

```
lifecycle:
  deletePolicy:  '${schema.spec.environment == "production" ? "Retain" : "Delete" }'
```

without the expression sentinels and is much easier to read.

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Other solutions considered
+
+### Adoption safety policies
+
+There are various levels of safety we could build into Kro for adoption. We could add policies like `AdoptOrCreate`,
+`AdoptOnly`, `CreateOnly` to give finer control over adopt behaviour. 
+
+The issue with these policies is that while they do provide some safety, they are drastically less safe than 
+a preview command like `kro preview`. The preview command would let you understand not just if resources will be
```
jakobmoellerdev, line 131: a preview command is by definition a preview and thus it cannot be compared to an actual behavior change in the instance. Even if `kro preview` was safe, executing the actual change is still just as dangerous and would need to be covered with safety checks

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+created but also if any updates are potentially disruptive. If you successfully adopt a database resource, but during 
+that adoption set memory to 10 bytes, you just caused an outage.
```
jakobmoellerdev, line 133: I dont understand how this is safeguarded or relevant to the proposal, I believe it needs some more explanation

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Adoption matching
+
+We could guarantee that adopted resources will not change initially. We could do a Kubernetes server side
```
jakobmoellerdev, line 137: what does "not change initially" mean here?

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+We could guarantee that adopted resources will not change initially. We could do a Kubernetes server side 
+dry run for resources with an `Adopt` policy and try to validate that there will be no differences.
+
+Performing the diff will be non-trivial since we will intend to make some updates like adding labels.
+This approach is also not perfect, admission webhooks with side effects could be skipped. We would also run into
+an issue if any updates happen between our dry run and applying the object.
+
+This idea has merit and could be a potentially valuable feature. However, this would be 
+prematurely solving the problem of adoption safety. We should see gaps from the CLI preview
```
jakobmoellerdev, line 145: I feel like the options youre outlining are alternatives to the preview command only, and not to your actual resource lifecycle proposal

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+The reason to use a CRD field instead of an annotation is type safety. It would be too easy for a customer to accidentally 
+typo the annotation then delete their production database. We could partially type check the annotation
+but nothing could stop a user from accidentally specifying something like `rko.run`. 
+
+### Static CRD fields vs CEL library
+
+We could have defined lifecycle as a static CRD schema field instead of a CEL expression evaluated by the `policy()` library.
+
+Using CEL allows adding new policy behaviors by extending the library without requiring CRD schema migrations.
```
jakobmoellerdev, line 166: This is moot, because you can also add policies without any schema migration, its non-breaking

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+
+#### What is in scope for this proposal?
+
+The scope of this proposal is minimal; all this document proposes is adding the lifecycle field with policy() CEL function support.
+
+#### What is not in scope?
+
+##### kro CLI
+
+While the kro CLI preview is referenced, exact details will be left to the CLI KREP.
```
jakobmoellerdev, line 184: Correct Choice IMHO, but feels like the proposal is not talking enough about the policies and edge cases and too much about semantics when used together with CLI

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+- I want to control deletion order because I may have implicit dependencies or a more complicated teardown workflow needed for resource cleanup.
+
+These use cases will need to be tackled separately. 
+
+##### Singletons and resource contribution
+
+- I want to define applications with resources that are defined only once across all Kro instances because resources like PrivateEndpoint are one per cluster.
+- I want to create a controller that updates a subset of fields on a resource because it is a useful pattern to control a small number of fields in Kubernetes.
+
+A separate KREP will address these.
```
jakobmoellerdev, line 201: Im not sure about this. If you add a policy field now, how would this ever be correctly addressable? To me it feels like partial ownership needs to be solved conceptially before we agree on this lifecycle policy field

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: resource

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: My original design had many field options like 
```
lifecycle:
   deletePolicy: ...
   updatePolicy: ...
```

The feedback I got then was it would be too many fields and we would be committing to that structure while if we went the cel route we would have more flexibility to evolve over the time.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: An earlier revision of this design addressed partial ownership
https://github.com/kubernetes-sigs/kro/pull/1091/changes/6824e0b2acbdccbdda53e2f120007948e1edd8a6

The feedback I got was that partial ownership ends up being very complicated and the use cases were not that important currently so we should defer it.

At least how I was thinking about partial ownership, retain vs delete did not fit in clearly in partial ownership.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: If we went with 

```
policy:
   deletePolicy: ...
```

while we could easily migrate to 
```
policy:
   deletePolicy: ...
   updatePolicy: ...
```

we could not easily migrate to something like 
```
policies: 
   - retainInsteadOfDelete
   - createAndDeleteInsteadOfUpdate
```

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: A part of the proposal here is that we should not add any specific policies for adoption, this is trying to give alternatives and describe ways we could model adoption as a policy.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: The thinking is that when we patch a resource to add Kro labels to indicate we are managed it by Kro that patch does not change the functionality of the Kubernetes resource. 

We could then change it if you made an update to an RGD or instance but at the very least that initial cut over from non Kro managing to Kro managing would not cause functional changes

docs/design/proposals/resource-lifecycles.md
jakobmoellerdev: While I agree, this is completely different from what the sentence describes, because its referencing CEL expressions here and the `policy` library.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: A well implemented preview command would give confidence but not give guarantees.

I don't believe that we can have any strict guarantee of safety. Someone could have a mutating webhook marked with side effects and that could make any action taken by Kro unsafe with absolutely no potential for us to foresee this.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: If we do expect policies to be correlated the builder might read easier 

For example if we added like a withBackup policy that like backed up pvcs (I'm not saying this makes any sense but just as an example)

```
    lifecycle: '${schema.spec.environment == "production" ? policy().withRetain().withBackup() : policy()}'
```

is cleaner than 
```
lifecycle:
  deletePolicy:  '${schema.spec.environment == "production" ? "Retain" : "Delete" }'
  backupPolicy: '${schema.spec.environment == "production" ? "Backup" : "None }'
```


Though a policy like ```updatePolicy: recreateOnUpdate` probably would be uncorrelated and cleaner as not as a cel expression

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: Not opposed to doing this, just gets kinda messy trying to manually mess with field managers

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: The suggestion would be to have them update the existing RGDs with orphan

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: Yeah I agree they are seperate, but to give some background the two main use cases that were called out for resource lifecycles were adoption and orphaning. I wanted to try and touch on how users could do adoption with this current proposal despite not suggesting any changes. 

I have seen feedback about Kro that it doesn't support adoption, but this isn't true today.

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: The rough idea would be to essentially run Kro reconciles loop in the CLI (not apply any resources just dry run / diff them again a cluster). 

This wouldn't be easy and would have a ton of complications like how to handle resources that depend on other resource statuses / fields but for the use case of all my resources already exist and I want to bring them into Kro it should work pretty well

We couldn't preview every RGD / update but I do think we could give a lot of value with this

docs/design/proposals/resource-lifecycles.md
NicholasBlaskey: Its relevant to show that just adding a policy like ```createPolicy: adoptOnyErrorOnResourceNotExisting``` does not provide much value on its own

The safe guard is gaining some confidence using the preview command

docs/design/proposals/resource-lifecycles.md
```diff
@@ -0,0 +1,216 @@
...
+- I want to safely do migrations into Kro from existing resources because I want to manage my cluster in Kro and want to avoid downtime and risk of updating resources poorly.
+- I want to safely do migrations between Kro RGDs or instances because I want to be able to refactor and evolve my Kro definitions over time and avoid downtime.
+
+## Proposal
+
+The proposal is to let adoption be handled by the current behaviour.
+The current behaviour applies over other resources as long as they
+are not being managed by a Kro instance.
+
+To avoid deleting resources we will add a new `lifecycle` field that accepts a CEL expression:
```
illrill, line 18: I think this be framed a bit more generally than trying to model scenario-specific lifecycles. What if we instead express a small, **stackable** set of **management actions** an RGD instance is allowed to take on a managed resource - taking inspiration from Crossplane's [management policies](https://docs.crossplane.io/v1.20/concepts/managed-resources/#managementpolicies).

Thinking about it, there are in fact two related but distinct "categories" of actions:
- **Reconcile** - what the instance is allowed to do to ensure a resource exists and matches desired state.
- **Ownership** - how the instance is allowed to take, transfer, or drop ownership of a resource.

In KRO this distinction is particularly relevant because ownership is explicit (via KRO metadata / ApplySet tracking). So beyond just reconciling resources, we also need to be clear about who is allowed to own a resource and how that ownership can change over time.

---

Consider the following set of management actions, placed in e.g. a `managementActions` array field:

| Category     | Action     | Description |
|--------------|------------|------------|
| Reconcile    | `Manage`   | Ensure the resource exists and matches desired state |
| Reconcile    | `Delete`   | Allow removing the resource on prune / teardown |
| Ownership    | `Claim`    | Take ownership of an unowned resource |
| Ownership    | `Release`  | Drop ownership without deleting |
| Ownership    | `Takeover` | Replace another instance as owner |

With a model like this, different lifecycle scenarios can be expressed as **combinations of actions**:

### `[Manage, Delete]`  
Full ownership (good default). Use when resource lifecycle is fully tied to the instance.
- Ensure resource exists and is up to date  
- Instance owns it  
- Delete on prune/teardown  

### `[Manage]`  
Retain/orphan. Useful when resource should outlive the instance.
- Ensure resource exists and is up to date  
- Instance owns it while active  
- Retain on prune/teardown  

### `[Claim, Manage]`  
Adopt existing or create, then manage. Useful when adopting existing resources or implementing singletons.
- If exists and unowned, claim it  
- If absent, ensure it exists  
- Once owned, keep it up to date  
- Fail if owned by another instance unless `Takeover`  

Bonus: When multiple RGD instances resolve the **same resource identity** (`apiVersion/kind/name/namespace`), this naturally gives **singleton-style** behaviour:
- Resolve by identity  
- Ensure it exists  
- Claim if unowned  
- Fail if already owned (unless `Takeover`)  

### `[Claim]`  
Claim-only. Useful for staged migration or ownership without mutation.
- Claim existing unowned resource  
- No reconciliation or deletion  

### `[Claim, Manage, Delete]`  
Adopt or create, then fully manage. Useful when "adopting" resources into KRO management.
- Claim if exists and unowned  
- Ensure it exists if absent  
- Keep it up to date  
- Delete on prune/teardown  

### `[Manage, Release]`  
Manage, then release. Useful for handoff or retaining resources without ownership.
- Ensure resource exists and is up to date  
- Instance owns it while active  
- Release ownership on prune/teardown  

### `[Takeover, Manage]`  
Forced migration. Useful for controlled ownership migration.
- Take ownership from another instance  
- Once owned, ensure it exists and is up to date  

### Invalid combinations

- `[Delete]` - no reconciliation or ownership; unclear what is being deleted  
- `[Release]` - no ownership established; nothing to release  
- `[Takeover]` - ownership changes but no subsequent reconciliation  
- `[Delete, Release]` - conflicting teardown behavior (delete vs retain)  
- `[Claim, Takeover]` - ambiguous ownership acquisition (unowned vs already-owned)  
- `[Takeover, Delete]` - takes ownership only to delete, without managing the resource  

These could be rejected at admission time with clear validation errors.

jakobmoellerdev: honestly a good take imho. I like this view.

NicholasBlaskey: In a previous discussion there was a comment with a similar user interface and the proposed idea was to present higher level preset supported options like "Orphaner" that would represent "Manage, Release"
https://github.com/kubernetes-sigs/kro/issues/542#issuecomment-3202236827

Personally I feel like there is some complexity in presenting this as a list. We need to work through the combination of options. Its not clear to me why a user shouldn't be allowed to both ```claim``` and ```takeover``` at the same time to adopt from another system regardless if it is Kro or not. 

Crossplanes policies here `managementPolicies` seem to be strict subsets of actions `  managementPolicies: ["Create", "Delete", "Observe"]`. There is no overlap in any of the policy options. The proposed ```Delete, Release``` have overlap and are contradictory which put us into this complexity of trying to understand the combinations situation. 

Though it seems they also expose a ```deletePolicy``` on top of their management policies. I'm not sure if the lesson there is to avoid designing something that has this overlap in it, or that having this overlap is useful and we may want to decompose this problem more.

jakobmoellerdev: Agreed on Overlap, but I feel like it is easily possible to define allowed combinations and behavior with this in mind though.

NicholasBlaskey: I think we can avoid the overlap. Suppose we define a list of actions

```
# Can allow taking over non Kro resources
Adopt               

# Can allow taking over Kro resources. Without this error if resource exists and owned by another Kro.
Takeover

# Allows creating resources when they exist. Without this error if object does not exist.
Create

# Allows updating resources. Without this resources are immutable.
Update

# Allows deleting resources. Without this retain.
Delete
```

Then to model release/retain we would just drop the ```delete``` and have a policies like ```Create, Update```

For modifying a job spec that is immutable, we would either need to add another `UpdatePolicy` unrelated field or introduce another policy like `RecreateOnImmutableError`

illrill: Glad you're open to the idea! It puts control over lifecycle scenarios in the hands of RGD author, rather than trying to anticipate, model (and name) each one as a preset. I'm guessing this is also a bit more straightforward to implement, and that we wouldn't (couldn't?) need to leave any scenarios out of scope.

Good call to split `Manage` into `Create` + `Update`, I didn't consider the "immutable scenario". 👍

Also agree that `Release` is unnecessary, it can be implicit behavior if `Delete` is not present.

illrill: > Though it seems they also expose a `deletePolicy` on top of their management policies. I'm not sure if the lesson there is to avoid designing something that has this overlap in it, or that having this overlap is useful and we may want to decompose this problem more.

For reference, Crossplane started out with `deletionPolicy` but that's now deprecated and replaced by the more flexible `managementPolicies`.

PR #1097, opened 2026-03-06
Fix ResourceGraphDefinition delete predicate handling

a-hilaly:
Allow the primary ResourceGraphDefinition watch to reconcile when the
deletion timestamp changes, not just when generation changes. This fixes
finalizer-driven cleanup paths that were skipped on delete.

Also add table-driven predicate tests covering generation changes,
deletion timestamp transitions, and ignored status-only updates.

Closes https://github.com/kubernetes-sigs/kro/pull/1037

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller.go
+++ b/pkg/controller/resourcegraphdefinition/controller.go
@@ -104,8 +104,7 @@ func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) e
 
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("ResourceGraphDefinition").
-		For(&v1alpha1.ResourceGraphDefinition{}).
-		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		For(&v1alpha1.ResourceGraphDefinition{}, builder.WithPredicates(resourceGraphDefinitionPrimaryWatchPredicate())).
 		WithOptions(
 			ctrlrtcontroller.Options{
 				LogConstructor:          logConstructor,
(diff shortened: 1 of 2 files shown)


pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -126,6 +125,23 @@ func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) e
 		Complete(reconcile.AsReconciler[*v1alpha1.ResourceGraphDefinition](mgr.GetClient(), r))
 }
 
+func resourceGraphDefinitionPrimaryPredicate() predicate.Predicate {
```
michaelhtm: nit: should we add a comment describing primaryPredicate?
also why the name primaryPredicate?

pkg/controller/resourcegraphdefinition/controller_test.go
```diff
@@ -0,0 +1,99 @@
...
+			},
+			want: true,
+		},
+		{
+			name: "keeps delete events enabled",
+			run: func(pred predicate.Predicate) bool {
+				return pred.Delete(event.DeleteEvent{Object: newPredicateTestRGD(1, nil)})
+			},
+			want: true,
+		},
```
michaelhtm, line 882: what's the difference between a delete event and the event of a deletionTimestamp being added?
Why do we want to reconcile on both?

michaelhtm (commented):
🚀🚀

pkg/controller/resourcegraphdefinition/controller.go
a-hilaly: Renamed this to `resourceGraphDefinitionPrimaryWatchPredicate` and added a doc comment. `primary` here means the predicate attached to the primary `For(&ResourceGraphDefinition{})` watch, as opposed to the secondary CRD metadata watch in the same controller setup

pkg/controller/resourcegraphdefinition/controller_test.go
a-hilaly: The important signal for this fix is the update where `meteadata.deletionTimestamp` is added. The terminal delete event is too late for that path here.. I’ve updated the predicate to ignore primary delete events and added a test to make that explicit

jakobmoellerdev (commented):
/lgtm
/unhold

Issue #1088, opened 2026-03-05
CRD not updated when schema default values change


mvgijssel:
## Bug Description                                                                            
                                                                                                
  When updating a ResourceGraphDefinition with a changed default value (e.g., `timeout: "1m30s"`
   → `timeout: "2m"`), the generated CRD is not updated because `pkg/graph/crd/compat/schema.go`
   does not compare `Default` field values.                                                     
                                                                                                
  The `DefaultChanged` constant exists in `pkg/graph/crd/compat/changes.go` but is never used in
   the comparison logic in `schema.go`.                                                         
                                                                                                
  ## Steps to Reproduce                                                                         
                                                                                                
  1. Create an RGD with a field that has a default value:                                       
```yaml                                                                                    
     spec:                                                                                      
       schema:                                                                                  
         apiVersion: v1alpha1                                                                   
         kind: MyResource                                                                       
         spec:                                                                                  
           timeout: string | default="1m30s"   
```                                                 
                                                                                                
  2. Apply the RGD - CRD is created with default "1m30s"                                        
  3. Update the RGD to change the default:                                                      
```yaml
  spec:                                                                                         
    schema:                                                                                     
      apiVersion: v1alpha1                                                                      
      kind: MyResource                                                                          
      spec:                                                                                     
        timeout: string | default="2m"                                                          
```
  4. Apply the updated RGD                                                                      
  5. Check the CRD:        
```bash                                                                     
  kubectl get crd myresources.kro.run -o jsonpath='{.spec.versions[0].schema.openAPIV3Schema.pro
  perties.spec.properties.timeout.default}'                                                     
```
                                                                                                
  Expected Behavior                                                                             
                                                                                                
  CRD should be updated with the new default value "2m".                                        
                                                                                                
  Actual Behavior                                                                               
                                                                                                
  CRD retains the old default value "1m30s". The controller logs show "CRD is up-to-date"       
  because report.HasChanges() returns false.                                                    
                                                                                                
  Root Cause Analysis                                                                           
                                                                                                
  In pkg/graph/crd/compat/schema.go, the compare() function checks for:                         
  - Description changes                                                                         
  - Type changes                                                                                
  - Pattern changes                                                                             
  - Property additions/removals                                                                 
  - Required field changes                                                                      
  - Enum value changes                                                                          
  - Array items changes                                                                         
                                                                                                
  But it does not check for Default value changes.                                              
                                                                                                
  The fix would add something like:                                                             

[...]

PR #1098, opened 2026-03-06
feat(compat): detect default value changes as non-breaking in schema comparison

michaelhtm:
Issue [#1088](https://github.com/kubernetes-sigs/kro/issues/1088)

The CRD schema compatibility checker did not account for changes to
default values, causing them to be silently ignored during RGD updates.
Add default value comparison logic that classifies additions, removals,
and modifications of defaults as non-breaking changes.

### DIFF
--- a/pkg/graph/crd/compat/schema.go
+++ b/pkg/graph/crd/compat/schema.go
@@ -14,6 +14,8 @@
 package compat
 
 import (
+	"bytes"
+
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
@@ -50,8 +52,18 @@ func compare(path string, oldSchema, newSchema *v1.JSONSchemaProps) *Report {
 		result.AddNonBreakingChange(
 			path+".description",
 			DescriptionChanged,
-			"",
-			"",
+			oldSchema.Description,
+			newSchema.Description,
+		)
+	}
+
+	// default value changes are non-breaking
+	if !defaultsEqual(oldSchema.Default, newSchema.Default) {
+		result.AddNonBreakingChange(
+			path+".default",
+			DefaultChanged,
+			getDefaultValue(oldSchema.Default),
+			getDefaultValue(newSchema.Default),
 		)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/graph/crd/compat/schema.go
```diff
@@ -55,6 +57,19 @@ func compare(path string, oldSchema, newSchema *v1.JSONSchemaProps) *Report {
 		)
 	}
 
+	// default value changes are non-breaking
+	//
+	// TODO: Should removing a default from a required field be a breaking
```
jakobmoellerdev: IMHO: I think it should be. Leaving final vote to @a-hilaly 

Nit: Arguably if we are very strict even adding a default can break behavior but im not dying on that hill.

spattk: I might be wrong, but I thought in KRO there was no way for a field to be `required` and have a `default`. Isn't it?

pkg/graph/crd/compat/schema.go
```diff
@@ -55,6 +57,19 @@ func compare(path string, oldSchema, newSchema *v1.JSONSchemaProps) *Report {
...
+	// change?
+	if !defaultsEqual(oldSchema.Default, newSchema.Default) {
+		result.AddNonBreakingChange(
+			path+".default",
+			DefaultChanged,
+			"",
```
spattk: nit: I realized, there are a bunch of places in the changeDesc code that we don't pass the `oldValue` and the `newValue` even if they are available. Is there a specific reason or that doesn't serve much purpose for now?

I believe, it would be helpful here if we have the old and new value just to have a log for debugging purpose? https://github.com/kubernetes-sigs/kro/blob/main/pkg/graph/crd/compat/changes.go#L161-L162

pkg/graph/crd/compat/schema.go
michaelhtm: here's an example of how you can define required/default fields in kro: https://github.com/kubernetes-sigs/kro/blob/main/examples/kubernetes/defaults/rg.yaml

pkg/graph/crd/compat/schema.go
michaelhtm: I was mainly following the pattern of how we detect non breaking changes in kro, since they're not really logged anywhere. This may be useful to someone who's importing the library tho. We can add them

pkg/graph/crd/compat/schema.go
michaelhtm: Added REQUIRED_DEFAULT_REMOVED

pkg/graph/crd/compat/schema.go
michaelhtm: added

PR #1101, opened 2026-03-07
docs: KREP-016 Resource Contribution

ellistarn:
Introduces a design proposal for resource contribution: a new `patch` node type that allows KRO to contribute fields to resources it does not own via server-side apply.

- Motivates the design via the general Kubernetes multi-controller cooperation pattern
- Proposes `patch` as a distinct node type alongside `template` and `externalRef`
- Documents two canonical examples: sidecar injector and a task state machine with separate RGDs per phase transition
- Opens discussion on naming and whether `externalRef` should be renamed to `watch`

### DIFF
--- a/docs/design/proposals/resource-contribution.md
+++ b/docs/design/proposals/resource-contribution.md
@@ -1,0 +1,290 @@
+# KREP-016 Resource Contribution
+
+## Summary
+
+Today, KRO can observe a resource it doesn't own via `externalRef`, but it can never write to it.
+The only way KRO can mutate a resource is to own it -- to have created it via a resource template.
+This KREP introduces a new node type, `patch`, which points to a resource KRO does not own and
+contributes fields to it via server-side apply. `externalRef` remains read-only. `patch` is the
+write counterpart.
+
+## Motivation
+
+It is common in Kubernetes for multiple independent control loops to cooperate on a single object's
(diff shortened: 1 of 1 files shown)


docs/design/proposals/resource-contribution.md
```diff
@@ -0,0 +1,290 @@
+# KREP-016 Resource Contribution
```
jakobmoellerdev, line 1: I think the idea of this KREP is very good but I would prefer to have the patch behavior be part of our normal resource handling. For example, shouldn't it be perfectly possible to apply a partial status / field spec in our current apply? I feel like a top level key for this might be too much.

For me it should already possible to achieve this behavior by allowing better configurability of the existing resource pattern, for example with policy specifications on create / update. Maybe this should even be part of the lifecycle modifiers mentioned here: https://github.com/kubernetes-sigs/kro/pull/1091/changes

ellistarn: It could potentially roll up into a lifecycle modifier. I am super open to this, actually. 

The main reason I separated it is because I wanted to separate "I am creating this resource" from "I am just updating it". 

I think lifecycle is the wrong name, and we need to get the granularity right, but agree this may be better achieved there. Let me explore another rev.

ellistarn:
folded into https://github.com/kubernetes-sigs/kro/pull/1091

PR #1103, opened 2026-03-08
test(instance): harden instance reconciler coverage

a-hilaly:
Add broad unit coverage for reconcile success, requeue, suspend, deletion,
collection, external reference, apply-result, and status/state branches.

```
go test -cover ./pkg/controller/instance -count=1               
ok      github.com/kubernetes-sigs/kro/pkg/controller/instance  0.597s  coverage: 88.1% of statements
```


jakobmoellerdev (commented):
/lgtm

PR #933, opened 2026-01-02
KREP-008: docs: add proposal for resource-backed includeWhen evaluation

shivansh-source:
proposal for #845 

Add a design proposal for allowing includeWhen to reference upstream
resources and participate in dependency graph construction. The proposal
defines the intended validation and runtime semantics, clarifies how
resource-backed includeWhen should interact with dependency readiness and
contagious exclusion, and uses ACK-based examples to illustrate the
expected behavior.

PR #1104, opened 2026-03-08
allow `includeWhen` to reference other resources

a-hilaly:
Implementing KREP-8 https://github.com/kubernetes-sigs/kro/pull/933

Teach the graph builder to treat includeWhen references as real
dependencies, so resource-backed conditions participate in DAG
construction and cycle detection instead of being limited to schema-only
checks.

At runtime, evaluate resource-backed includeWhen expressions against
observed upstream state and return data-pending while referenced
dependencies are not ready, rather than failing reconciliation or
skipping too early. Propagate that pending state through instance
processing so apply and prune wait for a definitive decision.

Update the API/docs to describe the new behavior and add builder,
runtime, and integration coverage for dependency inference, contagious
exclusion, cycle handling, and pruning when resource-backed conditions
flip.

### DIFF
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -199,7 +199,7 @@ type Resource struct {
 	// IncludeWhen is a list of CEL expressions that determine whether this resource should be created.
 	// All expressions must evaluate to true for the resource to be included.
 	// If not specified, the resource is always included.
-	// Expressions may reference schema fields and upstream resources. They are
+	// Expressions may reference schema fields and/or fields in other resources in the RGD. They are
 	// re-evaluated during reconciliation, so resources may be created later or
 	// pruned later as conditions change.
 	// Example: ["schema.spec.enableMonitoring == true", "network.status.ready == true"]
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -537,7 +537,7 @@ func (b *Builder) buildDependencyGraph(
 			return nil, err
 		}
 
-		includeWhenDeps, err := extractConditionDependencies(inspector, node.IncludeWhen, nil)
+		includeWhenDeps, err := extractConditionDependencies(inspector, node.IncludeWhen)
 		if err != nil {
 			return nil, fmt.Errorf("failed to extract dependencies from includeWhen: %w", err)
 		}
@@ -818,7 +818,7 @@ func buildStatusSchema(
 
 // inspectExpressionRestricted uses the shared inspector to parse an expression,
 // then validates that only the allowed identifiers are referenced.
-// This is used for restricted contexts like forEach (only known resources).
+// This is used for restricted contexts like includeWhen (only schema) or readyWhen (only self).
 func inspectExpressionRestricted(inspector *ast.Inspector, expr string, allowedIdentifiers []string) (ast.ExpressionInspection, error) {
 	result, err := inspector.Inspect(expr)
 	if err != nil {
@@ -905,18 +905,16 @@ func extractDependencies(inspector *ast.Inspector, expr *krocel.Expression, iter
 }
 
 // extractConditionDependencies extracts resource dependencies from condition
-// expressions such as includeWhen and readyWhen. It also populates
-// expr.References for later validation. iteratorVars can be used for
-// collection-scoped bindings like "each".
+// expressions such as includeWhen. It also populates expr.References for later
+// validation.
 func extractConditionDependencies(
 	inspector *ast.Inspector,
 	expressions []*krocel.Expression,
-	iteratorVars []string,
 ) ([]string, error) {
 	var allDeps []string
 
 	for _, expression := range expressions {
-		nodeDeps, _, err := extractDependencies(inspector, expression, iteratorVars)
+		nodeDeps, _, err := extractDependencies(inspector, expression, nil)
 		if err != nil {
 			return nil, err
 		}
@@ -1119,17 +1117,14 @@ func validateAndCompileNode(builderCache *celcache.BuilderCache, sessionCache *c
 		// readyWhen expressions can ONLY reference the node itself (or 'each' for collections).
 		// At runtime, IsResourceReady/IsCollectionReady only has the resource in scope.
 		allowedVar := node.Meta.ID
-		var iteratorVars []string
 		if node.Meta.Type == NodeTypeCollection {
 			allowedVar = EachVarName
-			iteratorVars = []string{EachVarName}
 		}
 
-		if _, err := extractConditionDependencies(inspector, node.ReadyWhen, iteratorVars); err != nil {
-			return fmt.Errorf("resource %q readyWhen: %w", node.Meta.ID, err)
-		}
-		if err := validateConditionReferences(node.ReadyWhen, []string{allowedVar}); err != nil {
-			return fmt.Errorf("resource %q readyWhen: %w", node.Meta.ID, err)
+		for _, expression := range node.ReadyWhen {
+			if _, err := inspectExpressionRestricted(inspector, expression.Original, []string{allowedVar}); err != nil {
+				return fmt.Errorf("resource %q readyWhen: %w", node.Meta.ID, err)
+			}
 		}
 
 		// For readyWhen on collections, we need "each" variable which isn't in the shared env.
--- a/pkg/runtime/node.go
+++ b/pkg/runtime/node.go
@@ -112,10 +112,11 @@ func (n *Node) IsIgnored() (bool, error) {
 		if !ok {
 			return false, fmt.Errorf("includeWhen dependency %q not wired into runtime", depID)
 		}
-		if err := dep.checkObservedReadiness(); err != nil {
-			if errors.Is(err, ErrWaitingForReadiness) {
-				return false, fmt.Errorf("includeWhen dependency %q not ready: %s (%w)", depID, err.Error(), ErrDataPending)
-			}
+		err := dep.checkObservedReadiness()
+		if errors.Is(err, ErrWaitingForReadiness) {
+			return false, fmt.Errorf("includeWhen dependency %q not ready: %s (%w)", depID, err.Error(), ErrDataPending)
+		}
+		if err != nil {
 			return false, fmt.Errorf("includeWhen dependency %q: %w", depID, err)
 		}
 	}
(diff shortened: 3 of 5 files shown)


api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -190,14 +190,19 @@ type Resource struct {
 	// ReadyWhen is a list of CEL expressions that determine when this resource is considered ready.
 	// All expressions must evaluate to true for the resource to be ready.
 	// If not specified, the resource is considered ready when it exists.
-	// Example: ["self.status.phase == 'Running'", "self.status.readyReplicas > 0"]
+	// Examples:
```
jakobmoellerdev, line 193: i think the self specification should still stay valid right? or do we drop this completely

pkg/controller/instance/resources.go
```diff
@@ -213,6 +213,9 @@ func (c *Controller) processNode(
 
 	ignored, err := node.IsIgnored()
 	if err != nil {
+		if runtime.IsDataPending(err) {
```
jakobmoellerdev, line 243: I disagree with this change I believe. Why would a data pending error not be an error state?

pkg/graph/builder.go
```diff
@@ -899,6 +905,46 @@ func extractDependencies(inspector *ast.Inspector, expr *krocel.Expression, iter
 	return resourceDeps, iteratorRefs, nil
 }
 
+// extractConditionDependencies extracts resource dependencies from condition
+// expressions such as includeWhen and readyWhen. It also populates
+// expr.References for later validation. iteratorVars can be used for
+// collection-scoped bindings like "each".
+func extractConditionDependencies(
```
jakobmoellerdev, line 910: I reaally really dont like that we have to pass iterator variables specifically. shouldnt this be part of the inspector knowledge?

pkg/runtime/node.go
```diff
@@ -92,12 +92,49 @@ func (n *Node) IsIgnored() (bool, error) {
...
+			needed[ref] = struct{}{}
+			if ref != graph.InstanceNodeID {
+				resourceRefs[ref] = struct{}{}
+			}
+		}
+	}
+
+	// Resource-backed includeWhen conditions evaluate against observed upstream
+	// state, so they must wait until those dependencies are ready. The caller
+	// already checked contagious ignore propagation above, so only inspect the
```
jakobmoellerdev, line 108: what would be a contagious ignore propagation?

pkg/runtime/node.go
```diff
@@ -92,12 +92,49 @@ func (n *Node) IsIgnored() (bool, error) {
...
+	// dependency's observed readiness here.
+	for depID := range resourceRefs {
+		dep, ok := n.deps[depID]
+		if !ok {
+			return false, fmt.Errorf("includeWhen dependency %q not wired into runtime", depID)
+		}
+		if err := dep.checkObservedReadiness(); err != nil {
+			if errors.Is(err, ErrWaitingForReadiness) {
```
jakobmoellerdev: nit move errors.Is out of the err != nil check (this is handled inside the errors.Is anyway)

pkg/runtime/node_test.go
```diff
@@ -691,6 +691,68 @@ func TestNode_IsIgnored_WithCEL(t *testing.T) {
...
+					withObserved(map[string]any{"data": map[string]any{"enabled": "false"}}).build()
+				middle := newTestNode("middle", graph.NodeTypeResource).
+					withDep(source).
+					withIncludeWhenRefs("source.data.enabled == 'true'", "source").build()
+				return newTestNode("child", graph.NodeTypeResource).
+					withDep(middle).
+					withTemplateExpr("middle.data.value", variable.ResourceVariableKindDynamic).
+					build()
+			},
+			wantIgnored: true,
```
jakobmoellerdev, line 754: Why is ignored a state, but data pending is an error? arguably also ignore should be an error indicating why it is ignored. WDYT?

website/docs/docs/concepts/rgd/02-resource-definitions/02-conditional-creation.md
```diff
@@ -38,20 +38,22 @@ spec:
     enabled: true  # Ingress will be created
 ```
 
-If `ingress.enabled` is `false` or not provided, the Ingress resource is skipped entirely.
+If `ingress.enabled` is `false`, the Ingress resource is skipped entirely. If the field may be omitted, give it a default or guard the expression so a missing field does not raise an evaluation error.
```
jakobmoellerdev, line 41: this now needs an example

website/docs/docs/concepts/rgd/02-resource-definitions/02-conditional-creation.md
```diff
@@ -38,20 +38,22 @@ spec:
...
 
 ## How includeWhen Works
 
 `includeWhen` is a list of CEL expressions that control whether a resource is created:
 
 - If **all** expressions evaluate to `true`, the resource is included
 - If **any** expression evaluates to `false`, the resource is skipped
+- Conditions are re-evaluated on later reconciliations; if a previously included resource starts evaluating to `false`, kro prunes it
```
jakobmoellerdev, line 49: I think this is good behavior! is pruning still happening in order?

As in => `A->B(includeWhen=>true)->C->D` moves to `A->B(includeWhen=>false)` then pruning order should do `D->C` right

api/v1alpha1/resourcegraphdefinition_types.go
michaelhtm: i believe this was an outdated comment. Right now, the resource should reference itself by id

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -190,14 +190,19 @@ type Resource struct {
...
+	// - Single resource: ["database.status.phase == 'Running'", "database.status.readyReplicas > 0"]
+	// - Collection: ["each.status.phase == 'Running'"]
 	//
 	// +kubebuilder:validation:Optional
 	ReadyWhen []string `json:"readyWhen,omitempty"`
 	// IncludeWhen is a list of CEL expressions that determine whether this resource should be created.
 	// All expressions must evaluate to true for the resource to be included.
 	// If not specified, the resource is always included.
-	// Example: ["schema.spec.enableMonitoring == true"]
+	// Expressions may reference schema fields and upstream resources. They are
```
michaelhtm: nit: 
```suggestion
	// Expressions may reference schema fields and/or fields in the other resources of the RGD. They are
```

pkg/graph/builder.go
```diff
@@ -813,7 +819,7 @@ func buildStatusSchema(
 
 // inspectExpressionRestricted uses the shared inspector to parse an expression,
 // then validates that only the allowed identifiers are referenced.
-// This is used for restricted contexts like includeWhen (only schema) or readyWhen (only self).
+// This is used for restricted contexts like forEach (only known resources).
```
michaelhtm: nit: why are we updating the comment? i believe this function is used by ReadyWhen and Status expressions

pkg/graph/builder.go
```diff
@@ -1076,14 +1122,17 @@ func validateAndCompileNode(node *Node, inspector *ast.Inspector, env *cel.Env,
...
 		if node.Meta.Type == NodeTypeCollection {
 			allowedVar = EachVarName
+			iteratorVars = []string{EachVarName}
 		}
 
-		for _, expression := range node.ReadyWhen {
-			if _, err := inspectExpressionRestricted(inspector, expression.Original, []string{allowedVar}); err != nil {
-				return fmt.Errorf("resource %q readyWhen: %w", node.Meta.ID, err)
-			}
+		if _, err := extractConditionDependencies(inspector, node.ReadyWhen, iteratorVars); err != nil {
```
michaelhtm: are we changing readyWhen to be dependent on other resources as well?

pkg/runtime/node.go
michaelhtm: i believe it's
parent is ignored, all dependents are ignored

pkg/runtime/node_test.go
michaelhtm: Error for data pending triggers a requeue, while for ignored, we just skip the dependent and move on to the next node. I do see the necessity for a Data pending state tho, just to give users visibility of what's happening

api/v1alpha1/resourcegraphdefinition_types.go
a-hilaly: `self` was wrong, it's either resource id or `each` for collections

pkg/controller/instance/resources.go
a-hilaly: `data pending` here is not a reconcile failure. It means the `includeWhen` expression cannot be evaluated yet because the referenced upstream data is not available, for example when a status field has not been populated yet. In that case we should neither treat the condition as `false` nor mark the node as errored we should wait, requeue, and reevaluate once the upstream resource has the data. Nothing special here, just standard kro behaviour (pretty similar to template resolution behaviour)

pkg/runtime/node.go
a-hilaly: the reason why i created https://github.com/kubernetes-sigs/kro/pull/1125

pkg/runtime/node.go
a-hilaly: Makes sense, updated.

pkg/runtime/node_test.go
a-hilaly: there 4 different outcomes:
- true
- false
- we don't have enough data
- expression evaluation truly errored (0 division for example)

website/docs/docs/concepts/rgd/02-resource-definitions/02-conditional-creation.md
a-hilaly: AFAIK there is no reverse topology guarantee today. oprhans are deleted by applyset in a flag concurrent pass. Good catch - should be addressed in a later PR

a-hilaly: At the same time, i don't know how simple this will be. When do we do concurrent deletes? when do we reverse delete? step by step pruning and requeues? i think this is an applyset issue and out of scope for this PR

a-hilaly: https://github.com/kubernetes-sigs/kro/issues/1135

pkg/graph/builder.go
a-hilaly: No, this is not changing `readyWhen` to depend on other resources

pkg/graph/builder.go
jakobmoellerdev: Did you manage to check this one already? https://github.com/kubernetes-sigs/kro/pull/1104#discussion_r2917866614

jakobmoellerdev (commented):
apart from the builder extension this is good to go for me

pkg/graph/builder.go
a-hilaly: I remove the iterator variables from the function signature, anything else missing?

PR #1105, opened 2026-03-08
dynamiccontroller: retain parent informers for shared GVR watches

a-hilaly:
Parent watches and instance level child/externalRef watches share a single
`WatchManager` informer per `GVR`. The watch coordinator tears down orphaned
child watches by GVR when the last instance request disappears. In the
chained RGD case, that allowed externalRef watch cleanup to stop an
informer that was still serving a registered parent GVR, leaving the
parent controller without a live watch.

Introduce explicit parent retention in `WatchManager` and make parent
registration/deregistration acquire and release that retention. Normal
`StopWatch` calls now refuse to stop a retained informer, while informer
shutdown still happens once the final remaining user is gone: either after
the last parent deregisters, or after the last child/external watch request
is removed.

Also add regression coverage for the shared-GVR lifecycle:
- unit coverage for parent-retained watches in WatchManager
- dynamic-controller tests for parent/child informer coexistence
- an integration test covering RGD chaining with externalRef watches

### DIFF
--- a/pkg/dynamiccontroller/coordinator.go
+++ b/pkg/dynamiccontroller/coordinator.go
@@ -352,6 +352,14 @@ func (c *WatchCoordinator) WatchRequestCount() (scalar, collection int) {
 	return
 }
 
+// HasRequestsForGVR reports whether any child or external watch requests still
+// exist for the given GVR.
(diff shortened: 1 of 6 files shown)


jakobmoellerdev:
/lgtm

for an alternative, do you mind cross checking with https://github.com/kubernetes-sigs/kro/pull/1115 ?

PR #1107, opened 2026-03-08
feat(simpleschema): generate CRD printer columns from printColumn markers

a-hilaly:
Teach SimpleSchema to parse `printColumn="TITLE"` and carry printer-column
intent through schema transformation into CRD `additionalPrinterColumns`.

Generated columns are inferred from scalar spec fields, expanded through
custom types, merged with explicit CRD columns, and deduped by JSONPath.
Invalid marker usage now fails fast (empty title, non-scalar target type,
or unsupported path segment for Kubernetes JSONPath).

Update unit and integration tests, docs, examples, and website highlighting.

Example UX:
```yaml
schema:
  spec:
    image: string | default="nginx" printColumn="IMAGE"
    ingress:
      enabled: boolean | default=false printColumn="ENABLED"
```

### DIFF
--- a/examples/kubernetes/additionalPrinterColumns/rg.yaml
+++ b/examples/kubernetes/additionalPrinterColumns/rg.yaml
@@ -8,9 +8,9 @@ spec:
     kind: Application
     spec:
       name: string
-      image: string | default="nginx"
+      image: string | default="nginx" printColumn="IMAGE"
       ingress:
-        enabled: boolean | default=false
+        enabled: boolean | default=false printColumn="ENABLED"
     status:
       deploymentConditions: ${deployment.status.conditions}
       availableReplicas: ${deployment.status.availableReplicas}
@@ -18,12 +18,6 @@ spec:
       - jsonPath: .status.availableReplicas
         name: Available replicas
         type: integer
-      - jsonPath: .spec.image
-        name: Image
-        type: string
-      - jsonPath: .spec.ingress.enabled
-        name: Enabled Ingress
-        type: boolean
 
   resources:
     - id: deployment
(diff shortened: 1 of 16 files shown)


examples/kubernetes/additionalPrinterColumns/rg.yaml
```diff
@@ -8,22 +8,16 @@ spec:
     kind: Application
     spec:
       name: string
-      image: string | default="nginx"
+      image: string | default="nginx" kubectlPrint="IMAGE"
       ingress:
-        enabled: boolean | default=false
+        enabled: boolean | default=false kubectlPrint="ENABLED"
```
shabbskagalwala: Maybe instead of `kubectlPrint`, we can follow kubebuilder’s naming and use something like `printColumn` ?

Kubebuilder does it using annotations like this 
```
// +kubebuilder:printcolumn:name="ENABLED"
```

a-hilaly: Good point! i did hesitate a lot before calling it `kubectlPrint`.. i just thought that it's more explicit to make it about "kubectl" - if you haven't been building controllers before it's hard to know what this marker will do.

I think our options are:
- `print` (aligns with simple schema philosophy)
- `printColumn` (aligns with kubebuilder naming)
- `additionalPrinterColumn` (too long)
- `kubectl`
- `kubectlPrint`

would like to hear your thoughts @jakobmoellerdev @michaelhtm @ellistarn @matthchr

jakobmoellerdev: I think we should align with kubebuilder here as its so familiar to most. That said i think this requires a few more approvers on here

jakobmoellerdev:

/close

PR #1108, opened 2026-03-08
feat(cel): add deterministic seeded integer generation function

a-hilaly:
The random CEL library only supported generating deterministic strings
via `random.seededString`. This adds `random.seededInt(min, max, seed)`
which generates a deterministic integer in `[min, max)` from a SHA-256
hash of the seed string. This is useful for cases like generating stable
NodePort values from a resource UID.

The implementation reads the first 8 bytes of the hash as a uint64 and
maps it into the requested range. Input validation rejects cases where
min >= max. The doc comment on `Random()` is also updated to document
both library functions in a consistent format.

### DIFF
--- a/pkg/cel/library/random.go
+++ b/pkg/cel/library/random.go
@@ -27,23 +27,24 @@ const (
 	alphanumericChars = "0123456789abcdefghijklmnopqrstuvwxyz"
 )
 
-// Random returns a CEL library that provides functions to generate random text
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (commented):
/lgtm
/hold

generally okay for me, just looking for a second approver on this due to new function scope

michaelhtm (approved):
/lgtm

PR #1109, opened 2026-03-08
fix(cel/ast): derive known functions from CEL environment

a-hilaly:
The inspector maintained a hardcoded `knownFunctions` list that had to
be updated every time a new CEL function or library was added. This was
fragile and easy to get out of sync.

This change replaces the static list with a call to `env.Functions()`,
which returns all functions registered in the CEL environment at
construction time. The inspector now automatically picks up any custom
function without code changes. A nil-env guard keeps the zero-value
constructor safe.

Also adds thorough test coverage for the new behavior, verifying that
kro's custom CEL functions (e.g. `random.seededString`, `json.marshal`,
`base64.encode`) are correctly classified as functions rather than
resource dependencies or unknown identifiers, including edge cases
where resource names collide with function namespace prefixes.

### DIFF
--- a/pkg/cel/ast/inspector.go
+++ b/pkg/cel/ast/inspector.go
@@ -106,20 +106,11 @@ type Inspector struct {
 	loopVars map[string]struct{}
 }
 
-// knownFunctions contains the list of all CEL functions that are supported
-//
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (commented):
/lgtm

PR #1110, opened 2026-03-08
website: overhaul overview docs with interactive resource graph diagrams

a-hilaly:
Overhaul the website overview page to better explain kro through clearer
messaging and interactive diagrams.

Replace the old overview doc with a rewritten "What is kro?" page, including
updated intro copy, a stronger "In practice" section, aligned RGD examples,
and clearer "Get started" guidance.

preview: https://a-hilaly.github.io/kro/next/docs/overview


website/src/components/IntroFlow/index.tsx
jakobmoellerdev, line 1: something is off on the intro flow, the turning cog is not centered

a-hilaly: do you have a screenshot? browser?

a-hilaly: ohh actually no, keep staring at it, the gear is rotating :), like a sun

a-hilaly: fixed

website/docs/docs/overview.mdx
```diff
@@ -0,0 +1,165 @@
...
+# What is kro?
+
+**kro** (Kube Resource Orchestrator) lets you turn a set of Kubernetes resources
+into a reusable API. You define the API schema, describe the resources behind
+it in YAML, and connect them with CEL expressions. kro turns that definition
+into a CRD, watches for instances of that API, and reconciles the underlying
+resources for each one.
+
+A **ResourceGraphDefinition (RGD)** is the controller definition: it describes the
+input API users work with and the resources each instance should produce. kro
```
jakobmoellerdev:
```suggestion

A **ResourceGraphDefinition (RGD)** defines your input API: it describes the interface that users work with and the resources each instance should produce. kro
```

website/docs/docs/overview.mdx
```diff
@@ -0,0 +1,165 @@
...
+runtime failures.
+
+<ControllerFlow />
+
+## How it works
+
+A ResourceGraphDefinition has two parts: a **schema** that defines your API
+surface (the fields users fill in), and **resource templates** that reference
+those fields with CEL expressions. kro parses the expressions, infers the
+dependency graph, generates a CRD, and stands up a controller - all at runtime.
```
jakobmoellerdev: nit: link to the various sections in our docs

website/docs/docs/overview.mdx
```diff
@@ -0,0 +1,165 @@
...
+service/my-app   ClusterIP   80/TCP
+```
+
+Users apply one `WebApp`. kro creates the `ConfigMap`, `Bucket`, `Deployment`,
+and `Service`, and writes useful outputs like `bucketArn` back onto the same
+object.
+
+### The definition behind it
+
+A **ResourceGraphDefinition** defines the API under `spec.schema` and the
```
jakobmoellerdev: reference simple schema here

website/docs/docs/overview.mdx
a-hilaly: Is it just the input API? i believe it's more than that but I agree the "controller definition" is too broad

a-hilaly: changed to blueprint

website/docs/docs/overview.mdx
```diff
@@ -0,0 +1,165 @@
...
+      template:
+        apiVersion: v1
+        kind: Service
+        metadata:
+          name: ${deployment.metadata.name}
+        spec:
+          selector: ${deployment.spec.selector.matchLabels}
+```
+
+`bucket.status.arn` does not exist when you apply this definition. kro waits for
```
jlbutler, line 151: might be clunky but think it's worthwhile to clarify here that any resources in the cluster can be managed. in this case, s3, but could be another resource type on other clouds, etc?

i think it's pretty subtle and not the main thrust of this section, but good to clarify that this is just an example.

jlbutler: options to consider:

* could use the optional ingress resource to make it a bit more abstract.
* could use S3 if on AWS, Blob store if on Azure, etc? 

or, i could be over-rotating. just want to keep it clear

PR #1111, opened 2026-03-08
docs(website): remove pre-v0.7.0 entries from versions.json

a-hilaly:
stop publishing docs versions older than v0.7.0


jakobmoellerdev (commented):
/lgtm

a-hilaly:
i thought maybe i could keep them for sometime, until we have better automation. but happy to nuke them in this PR

jakobmoellerdev:
I think since they would be dead code right now otherwise we should drop them. theyre still available in commit history if we really need them

PR #1112, opened 2026-03-08
docs: add "Building Abstractions with kro" guides

a-hilaly:
Introduces guides for single-resource RGDs, multi-resource RGDs, and RGD chaining under a new top-level "Building Abstractions with kro" section. Covers when and why to use each pattern, with worked examples using native Kubernetes resources and ACK CRDs

### DIFF
--- a/website/docs/docs/building-abstractions/10-single-resource-rgd.md
+++ b/website/docs/docs/building-abstractions/10-single-resource-rgd.md
@@ -261,6 +261,25 @@ Wrapping is usually the wrong fit when:
 In those cases, letting users work with the native resource directly is often
 cleaner than publishing another API surface that adds little real value.
 
+:::warning Limitations and Future Work
+kro currently supports [breaking change detection](/docs/concepts/rgd/overview#breaking-changes)
+for schema fields (field removal, type changes, new required fields, enum
+restrictions, pattern changes) and blocks them by default. However, several
+areas are still evolving:
+
+- **Versioning** — multi-version CRD support and migration paths are not yet
+  available. See [KREP-009](https://github.com/kubernetes-sigs/kro/pull/935).
+- **Resource lifecycle** — fine-grained control over create/update/delete
+  behavior per resource is planned. See
+  [KREP-014](https://github.com/kubernetes-sigs/kro/pull/1091).
+- **Deletion policy** — configurable owner-reference and deletion semantics are
+  under proposal. See
+  [KREP-004](https://github.com/kubernetes-sigs/kro/pull/763).
+
+Keep these gaps in mind when deciding whether to publish a wrapper — you will
+own its API surface and any future migrations.
+:::
+
 ## Single Resource RGD vs. Grouping
 
 Single-resource abstractions and grouping are related, but they solve different


website/docs/docs/advanced/02-1-resource-wrapping.md
```diff
@@ -0,0 +1,273 @@
+---
+sidebar_position: 2
+---
+
+# Resource Wrapping
```
jakobmoellerdev: Why do we call this Wrapping? feels like this would stick around for long so just wanna make sure we deliberately name

website/docs/docs/building-abstractions/20-multi-resource-rgd.md
```diff
@@ -0,0 +1,212 @@
...
+
+Use grouping when the value comes from packaging related resources that should
+always travel together. Use wrapping when the value comes from redefining the
+API for one resource. Use chaining when the building blocks should be reusable
+as independent APIs.
+
+For those related patterns, see [Resource Wrapping](./02-1-resource-wrapping.md)
+and [RGD Chaining](./02-rgd-chaining.md).
+
+## Operational Notes
```
jakobmoellerdev, line 219: We should add a guide on when NOT to wrap Imho. Also api upgrade necessity and other caveats apply here that are not mentioned.

website/docs/docs/advanced/02-1-resource-wrapping.md
a-hilaly: Just a terminology i've always used. not married to the name tho. How about `Single resource abstraction` or `Resource adapter` ? cc @jlbutler thoughts?

jakobmoellerdev: we could call it `Building Abstractions with KRO` and keep it more open. No one is forcing us to adopt a specific terminus for this pattern. this reads more like a guide anyhow. WDYT?

a-hilaly: We do already have https://kro.run/docs/advanced/rgd-chaining. Should we put these 3 under the same guide "Building abstractions with KRO" ? 

cc @jlbutler

a-hilaly: moved it all to a "Building abstractions with KRO" and rename wrapping/grouping to single resource RGD/multi resource RGD

website/docs/docs/building-abstractions/10-single-resource-rgd.md
```diff
@@ -0,0 +1,297 @@
...
+          "managed-by": "kro",
+          "tenant": schema.metadata.?labels["tenant"].orValue("shared"),
+          "namespace": schema.metadata.namespace
+        })}
+```
+
+Putting the platform tags in the second argument to `merge()` makes them win on
+key conflicts, so the wrapper always injects the tags you require.
+
+## When Not To Wrap
```
jakobmoellerdev, line 247: do we have a known issues caveat link for our schema evolution KREP / issue?

jakobmoellerdev (commented):
mostly LGTM. great read

michaelhtm (approved):
This are really interesting use cases @a-hilaly

PR #1114, opened 2026-03-08
Add a Grafana dashboard for KRO controller metrics.

a-hilaly:
Store the dashboard at `config/grafana/dashboard.json` and link to the raw
GitHub URL from the metrics docs for direct import.

screenshot:

Screenshot 2026-03-08 at 12-37-49 KRO Controller Metrics - Dashboards - Grafana

### DIFF
--- a/config/grafana/dashboard-scale-testing.json
+++ b/config/grafana/dashboard-scale-testing.json
@@ -1,0 +1,9643 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "graphTooltip": 1,
+  "panels": [
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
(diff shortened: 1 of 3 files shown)


jakobmoellerdev (commented):
Nice addition for observability. A few things to address:

**Duplicate UID — will cause overwrites**
Both `dashboard.json` and `dashboard-scale-testing.json` share `"uid": "kro-controller-metrics"` and `"title": "kro"`. Importing both into the same Grafana instance will silently overwrite the first. Give the scale testing dashboard a distinct UID and title.

**Scale testing dashboard has hardcoded test resource names**
The "Total Managed Children" / "Managed Child Resources" panels enumerate ~50 specific custom resource metrics (`graphheavy5001` through `graphheavy5050`). This is clearly from a specific scale test run and won't work for anyone else. Either generalize with a regex pattern or document that this dashboard is environment-specific and needs customization.

**Hardcoded `namespace="kro"` everywhere**
Consider adding a `$namespace` template variable (like the existing `$datasource`) so users deploying to a different namespace don't have to edit every query.

**Inconsistent rate intervals**
`dashboard.json` mixes `[5m]`, `[2m]`, and `[$__rate_interval]`. The scale testing dashboard uses `[2m]` throughout. Prefer `$__rate_interval` consistently — it's the Grafana best practice and adapts to the scrape interval.

**Minor**
- Missing trailing newline on both JSON files
- No docs/README for the scale testing dashboard — worth a sentence explaining when/why to use it vs the main dashboard
- `dashboard.json` hardcodes `service="kro-metrics"` and `job="apiserver-direct"` — same portability concern as namespace

jakobmoellerdev:

/close

PR #1116, opened 2026-03-08
perf(graph): eliminate duplicate CEL type building during RGD construction

jakobmoellerdev:
Three targeted optimizations to avoid redundant SchemaDeclTypeWithMetadata calls:

1. Add TypedEnvironmentWithProvider() that returns the DeclTypeProvider already created internally by DefaultEnvironment, replacing separate CreateDeclTypeProvider + TypedEnvironment call pairs.

2. Change getCelTypeFromSchema to use O(1) provider.FindDeclType lookup instead of re-converting schemas per field descriptor.

3. Reuse templatesEnv for readyWhen validation instead of creating a per-resource TypedEnvironment.


a-hilaly (approved):
👍 
/lgtm

PR #1121, opened 2026-03-08
KREP-017 template field omission with `omit()`

a-hilaly:
Add a design proposal for explicit field omission in resource templates via
`${expr.omit(value)}`.

The KREP frames the problem as a template-semantics limitation rather than
a CEL-evaluation limitation: once a field is present in a template, kro
always includes it in the apply payload, even when omission is the only
valid outcome for the target API.

The proposal defines omit as an explicit, template-only mechanism for
omitting fields without changing the meaning of plain null, and documents
its scope, compiler/runtime behavior, and relationship to prior discussion
around implicit null omission.

### DIFF
--- a/docs/design/proposals/omit-when.md
+++ b/docs/design/proposals/omit-when.md
@@ -1,0 +1,379 @@
+# KREP-017: Template Field Omission with `omit()`
+
+## Summary
+
+Today, once a resource template defines a field, that field is present in the
+object kro sends in its apply request. CEL can decide what value is written to
+that field, but it cannot express that the field should be omitted entirely.
+That becomes a practical limitation for Kubernetes APIs that distinguish
+between leaving a field out and sending that field with an empty or placeholder
+value. For many APIs, omission is valid, while `null`, `""`, or similar
+substitutes are not.
+
+This proposal introduces `omit()`, a zero-argument CEL function that returns a
+sentinel value. When the resolver encounters this sentinel as a field's
+resolved value, it removes the containing field from the rendered object before
+sending the apply request. Users express omission conditions using standard CEL
+control flow:
+
+```yaml
+policy: ${schema.spec.policy != "" ? schema.spec.policy : omit()}
+```
+
+KRO could instead choose a global default and omit certain resolved values
+automatically, for example by omitting all resolved `null` values. But no such
+default is compatible with all CRDs and all fields. Some fields need omission
+to remain valid; other fields are nullable and need to preserve explicit
+`null`. This proposal therefore keeps omission explicit and field-local.
+
+The proposal is intentionally narrow. It does not change CEL evaluation, it
+does not reinterpret plain `null`, and it does not apply to `includeWhen`,
+`readyWhen`, `forEach`, or other non-template expression contexts. The goal is
+to add one explicit mechanism for template-driven field omission without
+changing the meaning of existing templates.
+
+## Motivation
+
+GitHub issue [#928](https://github.com/kubernetes-sigs/kro/issues/928) asks for
+field-level conditional inclusion in templates. The motivating example is an
+ACK S3 `Bucket` where `policy: ""` is rejected, but omitting `policy`
+entirely is valid.
+
+This is not specific to ACK, and it is not primarily about CEL expressiveness.
+KRO already uses CEL extensively. The missing capability is that template
+structure is fixed once authored: if a field exists in the template, kro will
+include that field in the apply payload. Users can vary the field's value, but
+they cannot express that the field itself should be absent.
+
+That matters because, under Kubernetes API validation, "field absent" and
+"field present with `null`" are distinct states. For many APIs, especially
+third-party CRDs, that distinction is the difference between a valid request
+and a rejected one.
+
+In practice, that leaves users with workarounds that either do not preserve
+intent or do not scale:
+
+- return `null` and hope the target schema accepts it
+- return `""` or `{}` and hope the target API treats that like absence
+- split one logical resource into multiple `includeWhen` variants just to
+  control one optional field
+- move omission logic outside KRO entirely
+
+PR [#1003](https://github.com/kubernetes-sigs/kro/pull/1003) explored implicit
+omission of resolved `null` values. The discussion there established the right
+boundary for this proposal. The problem is real, but changing the meaning of
+plain `null` globally would be a breaking semantic shift. Omission should
+remain an explicit authoring decision rather than a side effect of ordinary
+expression evaluation.
+
+This matters especially for CRDs because `nullable` is opt-in in CRD schemas. A
+field only accepts explicit `null` when the schema marks it as nullable. In the
+common Kubebuilder/controller-tools path, fields are non-nullable unless API
+authors opt in with `+nullable`, so many third-party CRDs reject explicit
+`null` while still accepting the field being absent.
+
+The design goal, then, is not to make `null` behave differently. It is to give
+template authors one explicit way to say "omit this field."
+
+## Proposal
+
+### Overview
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/omit-when.md
```diff
@@ -0,0 +1,354 @@
...
+Today, once a resource template defines a field, that field is present in the
+object kro sends in its apply request. CEL can decide what value is written to
+that field, but it cannot express that the field should be omitted entirely.
+That becomes a practical limitation for Kubernetes APIs that distinguish
+between leaving a field out and sending that field with an empty or placeholder
+value. For many APIs, omission is valid, while `null`, `""`, or similar
+substitutes are not.
+
+This proposal introduces an explicit, template-only omission operator:
+`${expr.omitWhen(value)}`. kro evaluates `expr` and `value` using existing CEL
```
jakobmoellerdev: omitWhen is different from includeWhen but has similar syntax. I think we are walking a very dangerous path here. this would be the first syntax element in KRO that infers runtime behavior on the LHS of an expression based on the value of the RHS.

Thinking about
`myfield: ${myvalue.omitWhen(null)}`

in this case the expression on the right side now controls field presence for the left side (myfield)

jakobmoellerdev: I think we could at least adopt another option (for completeness of consideration) that has something like

`${myfield.omitWhen(null)}: ${myvalue}`

(Im not opinionated towards either, just came through reading this)

docs/design/proposals/omit-when.md
```diff
@@ -0,0 +1,354 @@
...
+how kro interprets the result of a template field expression. If evaluation
+produces an ordinary value, kro writes it as usual. If it produces the custom
+wrapper with its omit bit set, kro removes the containing field instead. If the
+omit bit is not set, kro unwraps the contained value and writes it normally.
+This also composes naturally with branching: in an expression such as
+`${condition ? left.omitWhen(x) : right}`, the selected branch evaluates to
+either an ordinary value or the omission wrapper, and the resolver determines
+the final field outcome from that single result.
+
+## In Scope
```
jakobmoellerdev, line 248: this proposal lacks an implementation approach to be considered naturally feasible

docs/design/proposals/omit-when.md
```diff
@@ -0,0 +1,354 @@
...
+selected branch resolves to omission. An unselected branch must have no effect
+on the final field result.
+
+The initial proposal omits only the containing field. It does not recursively
+prune parent objects that become empty after omission.
+
+### Implementation Shape
+
+The implementation should treat `omitWhen(...)` as a template-resolution
+wrapper, not as a change to ordinary CEL evaluation. The function returns a
```
jakobmoellerdev: arguably that means it should maybe not be part of the cel expression itself, it could also be arbitrarily defined in our language

docs/design/proposals/omit-when.md
```diff
@@ -0,0 +1,354 @@
...
+`${expr.omitWhen(value)}`. kro evaluates `expr` and `value` using existing CEL
+semantics. After evaluation, if the resolved value of `expr` equals `value`,
+kro omits the containing field from the rendered object before sending the
+apply request; otherwise, it writes the resolved value as usual.
+
+KRO could instead choose a global default and omit certain resolved values
+automatically, for example by omitting all resolved `null` values. But no such
+default is compatible with all CRDs and all fields. Some fields need omission
+to remain valid; other fields are nullable and need to preserve explicit
+`null`. This proposal therefore keeps omission explicit and field-local.
```
michaelhtm, line 27: should we mention here the issues introduced by applyset?

docs/design/proposals/omit-when.md
```diff
@@ -0,0 +1,354 @@
...
+The proposed user-facing API is a member-style form on standalone template
+field expressions:
+
+```yaml
+spec:
+  minReadySeconds: ${schema.spec.minReadySeconds.omitWhen(null)}
+  ingressClassName: ${schema.spec.ingress.className.omitWhen("")}
+  policy: ${schema.spec.policy.omitWhen("")}
+  replicas: ${(true ? 10 : 5).omitWhen(7)}
+  maybePolicy: ${schema.spec.enabled ? schema.spec.policy.omitWhen("") : "default"}
```
michaelhtm: this feels confusing..what if the user wants to specify policy that will be used somewhere else. 
They would need to define it twice so they can make omitWhen work.

Maybe we can see if there's something else we could use

docs/design/proposals/omit-when.md
michaelhtm: i agree..
can we also just introduce a function that just omits the field?
something like 
`${omitField()}`
It can be used in conditionals, or ternary operators.

docs/design/proposals/omit-when.md
cheeseandcereal: +1. Was going to ask if we've considered this alternative.

I wonder what a solution would look like where some sort of new sentinel value which indicates 'omit' as opposed to a new required expression evaluation for this control.

That may also be simpler to implement. And then what is proposed here would still be achievable by template authors, just writing a cel expression which conditionally resolves to this special sentinel.

docs/design/proposals/omit-when.md
a-hilaly: getting feedback from the comments, i'm revising the KREP to use `omit()` instead of `.omitWhen(value)`. Updated KREP incoming

a-hilaly: > `${myfield.omitWhen(null)}: ${myvalue}`

I'm planning on opening up https://github.com/kubernetes-sigs/kro/pull/1127 for reviews soon, i believe it's pretty much a seperate, larger effort.

docs/design/proposals/omit-when.md
a-hilaly: Updating

docs/design/proposals/omit-when.md
a-hilaly: I like the sentinel approach - it's the simplest primitive that solves the problem.

@jakobmoellerdev If omission isn't part of the CEL expression, the only other options are instructions outside the template (something like `omitFields` at the resource level) or new inline template syntax (like `policy?: ${...}` or sibling `_omitWhen` keys). Both force you to express the omission condition separately from the value, which means duplication and more CEL expressions/moving parts to keep in sync.

@cheeseandcereal exactly right @michaelhtm implementation in #1139 does exactly the sentinel approach you described

a-hilaly:
@michaelhtm @jakobmoellerdev i've updated the doc, let's get this merged?

docs/design/proposals/omit-when.md
```diff
@@ -0,0 +1,379 @@
...
+
+Because `omit()` is a plain CEL value that composes with standard control flow,
+users can express arbitrarily complex omission conditions without needing
+special syntax.
+
+### Compiler Behavior
+
+At compile time, the compiler should:
+
+- register `omit()` as a zero-argument function in the CEL environment
```
jakobmoellerdev, line 175: how does the omit zero value interact with other meta types? (i.e. with optionals)?

a-hilaly: Not sure I understand the question, can you share what scenario's you're thinking of specifically?

jakobmoellerdev: Im thinking of what the value is if you run something like a new optional of an omit value via constructor. `optional.of(omit())` would be an example.

a-hilaly: I do believe it will be unwrapped and falls back into a sentinel value https://github.com/michaelhtm/kro/blob/7b43c4c0547f431c0245214650048f1d88d4d8fc/pkg/cel/conversion/conversions.go#L66

PR #1123, opened 2026-03-09
test: verify externalref integration cases reconcile on watched updates

a-hilaly:
Extend the existing externalref integration suite to verify that referenced
resource changes trigger reconciliation, instead of only validating initial
resolution.


jakobmoellerdev (commented):
/lgtm

PR #1124, opened 2026-03-09
resourcegraphdefinition: wait for CRD deletion to complete

a-hilaly:
This patch makes deletion follow the same behavioral contract that creation
already relies on. On create, the RGD controller blocks until the generated
CRD is actually ready before marking the kind ready and continuing
reconciliation. On delete, cleanup was weaker: it treated a successful
DELETE request as if the CRD were already gone.

This is intentionally a narrow behavioral correction, not a larger
lifecycle redesign. Longer term, the options we can consider include moving
this kind of waiting out of reconcile and into a requeue / event-driven
model, or revisiting whether more of this lifecycle should be expressed via
ownerReferences. If we want to go further than that, and especially if we
want explicit transitional states such as Activating and Deactivating, that
should go through KREP-level design work rather than being folded into this
fix.

That mismatch showed up in the integration suite. The test
"CRD CRD Creation [It] should delete CRD when ResourceGraphDefinition is
deleted" removes an RGD and then polls the generated CRD until it returns
NotFound. In CI we observed the controller log "Deleting CRD", proceed with
cleanup, and remove the finalizer, while the API server continued to serve
the CRD. The test eventually timed out with "Expected an error, got nil".
This change makes deletion honor the stronger contract by polling for
NotFound after CRD deletion is requested and only reporting cleanup success
once the CRD has actually disappeared. The change is covered by an RGD
controller unit test, while the existing integration spec remains the
end-to-end regression.

### DIFF
--- a/pkg/client/crd.go
+++ b/pkg/client/crd.go
@@ -218,7 +218,29 @@ func (w *CRDWrapper) Delete(ctx context.Context, name string) error {
 	if err != nil && !apierrors.IsNotFound(err) {
 		return fmt.Errorf("failed to delete CRD: %w", err)
 	}
-	return nil
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	return w.waitForDelete(ctx, name)
+}
+
+// waitForDelete waits for a CRD to be fully removed from the API server.
+func (w *CRDWrapper) waitForDelete(ctx context.Context, name string) error {
+	log := logr.FromContext(ctx)
+	log.Info("Waiting for CRD to be deleted", "name", name)
+
+	return wait.PollUntilContextTimeout(ctx, w.pollInterval, w.timeout, true,
+		func(ctx context.Context) (bool, error) {
+			_, err := w.Get(ctx, name)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return true, nil
+				}
+				return false, err
+			}
+
+			return false, nil
+		})
 }
 
 // waitForReady waits for a CRD to become ready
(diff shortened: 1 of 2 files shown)


pkg/client/crd.go
```diff
@@ -218,7 +218,29 @@ func (w *CRDWrapper) Delete(ctx context.Context, name string) error {
...
 	}
-	return nil
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	return w.waitForDelete(ctx, name)
+}
+
+// waitForDelete waits for a CRD to be fully removed from the API server.
+func (w *CRDWrapper) waitForDelete(ctx context.Context, name string) error {
```
michaelhtm, line 228: would this block a worker from reconciling other RGDs until the CRD is deleted?
should we consider requeuing instead?

pkg/client/crd.go
michaelhtm, line 1: can we ensure we delete the CRD before shutting down the controller here? https://github.com/kubernetes-sigs/kro/blob/main/pkg/controller/resourcegraphdefinition/controller_cleanup.go#L34-L50

If not, if there are any instances that have not been cleaned up, they will no longer be reconciled

pkg/client/crd.go
michaelhtm: nvm i see it's addressed in PR description

a-hilaly: Yeah it's just temporary hack, and i'm the first to dislike it.

pkg/client/crd.go
a-hilaly: Good call. We should, perhaps in a separate PR with proper integration tests etc...

michaelhtm: picking it up after this PR is merged

michaelhtm (approved):
/lgtm

jakobmoellerdev (commented):
/lgtm

please lets follow up on this with a proper fix as soon as possible.

jakobmoellerdev:

/close

PR #1125, opened 2026-03-09
KREP-018: Partial Dependencies in Branching Expressions

a-hilaly:
kro currently infers dependencies from CEL expressions by collecting every
referenced resource and treating all of them as hard dependencies. That model is
too coarse for branching expressions, where only one branch is active at
runtime. In practice, this makes dependency handling overly strict and can lead
to contagious exclusion, where a dependency on an unselected branch still causes
the resource to be excluded.

KREP-018 proposes making partial dependencies explicit. The compiler preserves
branch structure when analyzing the checked CEL AST and emits branch-aware
dependency information alongside the original CEL program. The runtime then uses
that information to require only the dependencies on the selected path while
keeping final value evaluation unchanged.

### DIFF
--- a/docs/design/proposals/partial-dependencies.md
+++ b/docs/design/proposals/partial-dependencies.md
@@ -1,0 +1,255 @@
+# KREP-018: Partial Dependencies in Branching Expressions
+
+## Summary
+
+KRO currently infers dependencies by analyzing the CEL AST for a template
+expression, collecting every resource reference it finds, and turning each one
+into a hard dependency. That works for straight-line expressions, but it is too
+coarse for branching expressions such as ternaries, and that creates a
+correctness problem we refer to here as contagious exclusion. In an expression
+like `${schema.spec.useCache ? cache.status.endpoint :
+database.status.endpoint}`, the runtime needs either `cache` or `database`
+depending on the branch taken. But KRO infers both as required, so excluding
+one referenced resource can prematurely exclude the dependent resource even
+when the selected branch could have been satisfied.
+
+This proposal makes the compiler and runtime aware of partial dependencies in
+branching expressions. The compiler should preserve the branch structure of the
+checked CEL AST and lower it into a branch-aware dependency/control-flow IR,
+and the runtime should use that IR while preserving the original CEL program
+for final value evaluation.
+
+For ternaries, including nested ternaries, dependency activation should follow
+CEL selection semantics: conditions are evaluated in order, and the first
+branch whose predicate evaluates to true wins. Only the dependencies on the
+selected path should be considered required for the node to proceed.
+
+## Motivation
+
+The core problem is not that the compiler sees too many references; it is that
+the current runtime model treats every reference in a branching expression as
+unconditionally required.
+
+The canonical example is:
+
+```yaml
+spec:
+  endpoint: ${schema.spec.useCache
+      ? cache.status.endpoint
+      : database.status.endpoint}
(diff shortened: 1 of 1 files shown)


docs/design/proposals/partial-dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+one referenced resource can prematurely exclude the dependent resource even
+when the selected branch could have been satisfied.
+
+This proposal makes the compiler and runtime aware of partial dependencies in
+branching expressions. The compiler should preserve the branch structure of the
+checked CEL AST and lower it into a branch-aware dependency/control-flow IR,
+and the runtime should use that IR while preserving the original CEL program
+for final value evaluation.
+
+For ternaries, including nested ternaries, dependency activation should follow
```
jakobmoellerdev, line 22: This produces a problem. The CEL AST cannot know which branch is chosen. As such we would have to defer node rendering and graph rendering to node level. Maybe this is something we want but it will significantly complicate the current arch if we are not careful

docs/design/proposals/partial-dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+## Summary
+
+KRO currently infers dependencies by analyzing the CEL AST for a template
+expression, collecting every resource reference it finds, and turning each one
+into a hard dependency. That works for straight-line expressions, but it is too
+coarse for branching expressions such as ternaries, and that creates a
+correctness problem we refer to here as contagious exclusion. In an expression
+like `${schema.spec.useCache ? cache.status.endpoint :
+database.status.endpoint}`, the runtime needs either `cache` or `database`
+depending on the branch taken. But KRO infers both as required, so excluding
```
spattk, line 12: the same problem lo.ternary has. :-|

docs/design/proposals/partial-dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+one referenced resource can prematurely exclude the dependent resource even
+when the selected branch could have been satisfied.
+
+This proposal makes the compiler and runtime aware of partial dependencies in
+branching expressions. The compiler should preserve the branch structure of the
+checked CEL AST and lower it into a branch-aware dependency/control-flow IR,
```
spattk, line 18: What's IR?
(edit)nvm, I think it's Intermediate Representation

docs/design/proposals/partial-dependencies.md
```diff
@@ -0,0 +1,255 @@
...
+
+This proposal does not include:
+
+- new CEL syntax
+- rewriting user-authored CEL into new user-visible CEL expressions
+- soft dependencies, soft edges, or cycle breaking
+- solving every form of partial resolution
+- changing non-branching expression semantics
+
+## Other Solutions Considered
```
NicholasBlaskey, line 215: can we evaluate the expression partially for compile time values like schema? 

`schema.spec.useCache ? cache.status.endpoint : database.status.endpoint` can we not replace the expression internally with `cache.status.endpoint` or `database.status.endpoint`?

This won't work for all runtime values but could be a lot of the problem

docs/design/proposals/partial-dependencies.md
NicholasBlaskey: this could apply to like functions too right? 

If I had a function that was implemented as ```returnFirst(a, b)``` are we going to be able to realize that we don't need to resolve `b`

PR #1126, opened 2026-03-09
KREP-019: Deferred Fields and Soft Dependencies

a-hilaly:
This KREP introduces `defer(expr)` for standalone template fields to model soft
dependencies explicitly in KRO. A deferred expression does not block initial
resource creation when referenced data is still pending; instead it evaluates to
an effective `null` for that reconcile, or can be composed with `omitWhen(null)`
from KREP-018 to omit the field while the value is still unavailable.

The proposal defines `defer()` as a compiler-recognized wrapper around an
ordinary CEL expression, validates cycles on the hard-dependency graph, and
describes runtime behavior in terms of normal graph walks. Deferred expressions
are re-evaluated on each reconcile while the graph is being walked, but
`defer()` does not create its own infinite requeue loop, which makes readiness
an important part of using deferred fields correctly.

### DIFF
--- a/docs/design/proposals/defer.md
+++ b/docs/design/proposals/defer.md
@@ -1,0 +1,378 @@
+# KREP-019: Deferred Fields and Soft Dependencies
+
+## Problem statement
+
+Today KRO treats data dependencies as hard edges. If resource `A` references
+resource `B`, then `A` must wait for `B`, and dependency cycles are rejected.
+That is the right default, but it rules out graphs where both resources can be
+created first and only need each other's data later.
+
+The missing capability is an explicit way to say "this field may be resolved
+after initial creation." Without that, ordinary resource templates fail with
+`ErrDataPending` as soon as a referenced value is absent, so a graph that could
+converge across reconciles is rejected as cyclic up front.
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/defer.md
```diff
@@ -0,0 +1,378 @@
...
+resource `B`, then `A` must wait for `B`, and dependency cycles are rejected.
+That is the right default, but it rules out graphs where both resources can be
+created first and only need each other's data later.
+
+The missing capability is an explicit way to say "this field may be resolved
+after initial creation." Without that, ordinary resource templates fail with
+`ErrDataPending` as soon as a referenced value is absent, so a graph that could
+converge across reconciles is rejected as cyclic up front.
+
+KRO already has a narrow precedent for this behavior: instance status can be
```
jakobmoellerdev, line 15: I believe I disagree with this problem statement. for me this is what optional values are designed for. IMHO it would be enough to promote our optional checks to also have impact on the graph. Also when you define a defer, you must automatically declare a default value. This is all part of optionals API and I dont see why we now need to introduce defer.

Am I missing a case? Maybe differentiate it concretely from optionals

PR #1129, opened 2026-03-09
perf(cel): add caching for CEL compilation artifacts to optimize performance

jakobmoellerdev:
- Introduce caching for CEL compilation artifacts (DeclTypes, environments, programs) in `compilationCache`, scoped per `Builder` instance, as well as for the RGD build itself
- Optimize `TypedEnvironmentWithProvider`, `ParseCheckAndCompile`, and related methods to reuse cached results.
- Add robust unit tests for caching behavior and concurrent access.
- Enhance RGD processing by reducing redundant CEL operations.

Depending on load characteristics this can decrease heap pressure by up to 50%. We are working on adding consistent Scale tests.


pkg/controller/resourcegraphdefinition/build_cache.go
a-hilaly, line 1: I'd prefer we drop the `buildCache` from this PR. The generation based key assumes the build output only depends on the RGD spec, but the builder also resolves external CRD schemas via it's resolver. If a referenced CRDs schema changes in the cluster we need to rebuild the graph even though the RGD generation hasn't changed~ (The GraphRevisions work is trying to introducing caching at the `GraphRevision` level with support for recompilation when needed)

I think the CEL compilation caching is the real win in this PR and stands well on its own.

pkg/cel/types.go
```diff
@@ -27,14 +28,23 @@ import (
 	apiservercel "k8s.io/apiserver/pkg/cel"
 )
 
+// fieldTypeMapCache caches FieldTypeMap results keyed by *apiservercel.DeclType pointer.
+// Safe because DeclType is immutable after MaybeAssignTypeName, and FieldTypeMap
+// is a pure function of its input type.
+var fieldTypeMapCache sync.Map // key: *apiservercel.DeclType, value: map[string]*apiservercel.DeclType
+
```
a-hilaly: We don't have any cache eviction here, do we wanna have one? if an RGD is deleted there is no need to keep DeclType for unused types.

a-hilaly: Second thought i believe we need to do this caching per rgdBuild process, not cross RGD builds.

pkg/cel/cache.go
a-hilaly, line 1: deserves to be it's own package IMO

pkg/graph/builder.go
```diff
@@ -94,6 +94,9 @@ type Builder struct {
 	// schemaResolver is used to resolve the OpenAPI schema for the resources.
 	schemaResolver resolver.SchemaResolver
 	restMapper     meta.RESTMapper
+	// celCache holds cached CEL compilation artifacts (DeclTypes, environments,
+	// programs) scoped to this Builder instance.
+	celCache krocel.CompilationCache
```
a-hilaly: I was torn on this one for a bit and now i understand why. The `CompilationCache` bundles two kinds of entries with very different reuse profiles:
1/ `DeclTypes` and typed environments benefit from cross RGD caching - multiple RGDs referencing the same GVKs share identical schemas from the resolver
2/ Programs, checked ASTs, and extended envs are RGD-specific - expressions and parent envs differ per RGD, so cross RGD cache hits are near zero. These just accumulate objects from deleted RGDs on the long lived build sessions and prevent the GC from doing it's job.

Worth splitting into builder cache and build session cache?

a-hilaly (commented):
Overall solid optimizations - my main concern is around cache scoping - not all cached artifacts have the same reuse profile, and some would be better as per build throwaways than long lived on the Builder singleton/session. I left inline comments for specifics

pkg/controller/resourcegraphdefinition/build_cache.go
jakobmoellerdev: removed build cache for now

pkg/graph/builder.go
jakobmoellerdev: agreed and split

a-hilaly (approved):
LGTM

/cc @michaelhtm @NicholasBlaskey

michaelhtm (approved):
Great work @jakobmoellerdev 
/lgtm

pkg/cel/cache/keys.go
```diff
@@ -0,0 +1,80 @@
...
+
+	var b strings.Builder
+	for i, name := range names {
+		if i > 0 {
+			b.WriteByte(';')
+		}
+		b.WriteString(name)
+		b.WriteByte(':')
+		b.WriteString("0x")
+		b.WriteString(strconv.FormatUint(uint64(uintptr(unsafe.Pointer(schemas[name]))), 16))
```
a-hilaly, line 77: Missed this, but i'd prefer avoiding using unsafe in non-testing code @jakobmoellerdev

PR #1130, opened 2026-03-10
KREP-021: CEL write-back — stateful transitions in ResourceGraphDefinitions

pnz1990:
## Summary

kro's CEL expressions are read-only today: they evaluate against a snapshot of the parent instance CR and produce derived values in `status` fields and child resource templates, but cannot persist a computed value back into the CR across reconcile cycles.

This blocks stateful transition patterns (certificate rotation timestamps, progressive delivery phase advancement, retry counters, autoscaling cooldown enforcement, quota accumulation) that today each require a bespoke external controller — the exact reconcile-patch loop kro already performs internally.

## What this proposal adds

A new virtual resource type in the RGD `resources` list — a **write-back node** — that evaluates CEL expressions and persists the results back into the instance CR rather than creating a Kubernetes resource.

Two alternatives are presented, differing only in where the result is written:

- **Alt A: `specPatch`** — writes to instance `spec` using a dedicated SSA field manager (`kro.run/specpatch`). Simpler implementation; requires `ignoreDifferences` or SSA mode in GitOps tooling.
- **Alt B: `stateWrite`** — writes to a new `status.state` sub-object. GitOps-compatible by construction (GitOps tools never touch `status`); slightly more implementation complexity (must relax `withStatusOmitted` for the `status.state` sub-path).

Both integrate with the existing DAG, `includeWhen`, and idempotency machinery. Both are ~300 lines of Go across 4 files.

The proposal explicitly enumerates what write-back nodes **cannot** do (per-request entropy, array element mutation by index, time-triggered transitions, cross-instance mutations) and does not overclaim.

## Origin

This proposal emerged from building [Krombat](https://github.com/pnz1990/krombat), a Kubernetes-native turn-based RPG using kro RGDs as the game state machine. The external-controller workaround is operational there today; this proposal aims to eliminate the need for it in the simple cases.

## Questions for maintainers

1. Which alternative aligns better with kro's long-term API philosophy?
2. Is the `withStatusOmitted` restriction in `buildContext` intentional as a general rule?
3. Should write-back nodes use a `type:` discriminator on resource entries, or a top-level `transitions:` stanza?

Happy to revise based on feedback or implement a proof-of-concept for whichever alternative the maintainers prefer.

### DIFF
--- a/docs/design/proposals/cel-writeback/proposal.md
+++ b/docs/design/proposals/cel-writeback/proposal.md
@@ -1,4 +1,4 @@
-# CEL Write-Back: Stateful Transitions in ResourceGraphDefinitions
+# KREP-021: CEL Write-Back — Stateful Transitions in ResourceGraphDefinitions
 
 **Author:** Rafael Panazzo (https://github.com/pnz1990)
 **Status:** Draft / Open Question (see §Design Details and §Other Solutions Considered)


docs/design/proposals/cel-writeback/proposal.md
```diff
@@ -0,0 +1,514 @@
...
+game state machine. The write-back limitation manifests concretely there — DoT
+ticks (`heroHP -= 5`, `poisonTurns -= 1`) must survive across reconcile cycles,
+but today the Go backend handles all HP mutations and patches the Dungeon CR spec
+directly. That external controller pattern is exactly what this proposal aims to
+make unnecessary for the simple cases.
+
+The Krombat use cases that *cannot* be addressed by this proposal — combat dice
+rolls using Attack CR UIDs as random seeds, per-monster HP arrays, and room
+transition logic with multi-field atomic mutations — remain in the Go backend
+permanently. The proposal is honest about this boundary.
```
shabbskagalwala, line 514: Apologies but would this KREP be similar https://github.com/kubernetes-sigs/kro/pull/1035 ?

pnz1990:
Thanks for flagging — these are worth engaging carefully.

**On KREP-011 (RGD Variables / `data` nodes)**

KREP-011 and this proposal are solving adjacent but distinct problems, and they would compose naturally. A `data` node computes an intermediate value once per reconcile cycle from current inputs — it is a pure function with no side effects and no memory beyond that cycle. That is exactly the right tool for reusable expressions, computed container lists, and DRY naming prefixes.

The problem this proposal addresses is the *absence of memory across reconcile cycles*. A `data` node computing `${schema.spec.poisonTurns - 1}` produces the correct next value — but there is nowhere to store it so the following reconcile sees `poisonTurns = 2` instead of `3` again. The gap is not in computation; it is in persistence.

Framed simply: KREP-011 gives you a scratchpad within a single reconcile. This proposal asks for a notebook that survives between reconciles. Both are needed for a complete declarative state machine, and neither replaces the other.

If both were implemented, a natural composition pattern emerges: use a `data` node to compute intermediate values cleanly, then feed them into a write-back node to persist the results. The separation of concerns feels right.

**On KREP-016 (Resource Contribution / `patch` nodes)**

This is the more interesting question, and the overlap is real — but we think it is narrower than it first appears.

KREP-016's motivating problem is *cross-RGD cooperation on shared external objects*: a `task-scheduler` RGD and a `task-completion` RGD that neither created the `Task` CR but both need to write `status.phase` to it. The `patch` node is designed around SSA field-manager ownership, with the manager keyed to the RGD name so multiple independent RGDs can each claim a non-overlapping slice of the same object.

This proposal's motivating problem is *intra-RGD temporal state*: a single RGD that needs to read what it wrote in the previous reconcile cycle and act on it. The target is always the instance CR the RGD is currently reconciling — not an external object, not a sibling RGD's managed resource.

Where the two genuinely intersect: Alt A (`specPatch`) and KREP-016's `patch` node share the same underlying mechanism — SSA with a dedicated field manager. If KREP-016 were extended to permit targeting `self` (the current instance CR by its own `apiVersion/kind/name`), then Alt A becomes a degenerate case of KREP-016's `patch` node. We think that is a reasonable unification and we would be glad to drop Alt A entirely and express the self-patching case through KREP-016's `patch` node if the maintainers prefer to consolidate there.

That said, we see two areas where KREP-016 as currently written doesn't yet address what this proposal needs, and wanted to raise them explicitly rather than assume:

**1. Convergence gating for non-idempotent transitions.**

KREP-016's examples are all idempotent contributions — "this RGD always wants `status.conditions[PlatformManaged] = True` on every Deployment." The patch fires every reconcile, and that is correct because the value is stateless and deterministic from fixed inputs.

A write-back counter — `retryCount + 1`, `poisonTurns - 1` — is different: the expression needs to fire *once per triggering event*, then stop. Without a convergence guard, `retryCount` increments without bound. The `includeWhen` mechanism (e.g., `${schema.spec.retryCount < schema.spec.maxRetries && step1Job.status.failed > 0}`) is what makes the transition safe. KREP-016 doesn't discuss `includeWhen` on `patch` nodes — its model appears to assume patch expressions are inherently idempotent. We aren't sure whether `includeWhen` is already assumed to apply to `patch` nodes (since it applies to `template` and `externalRef` nodes generally), or whether there's a specific reason it wasn't mentioned. Either way, this is the question we'd want settled before relying on KREP-016's `patch` to cover this case.

**2. Alt B (`stateWrite` → `status.kstate`) doesn't overlap with KREP-016 at all.**

KREP-016 targets external resources via SSA on the main object body. Alt B writes to the `/status` subresource of the current instance CR — a different API path that KREP-016 neither uses nor is designed for. The separation of `spec` (user-owned) from `status` (controller-owned) is what makes Alt B GitOps-compatible without any `ignoreDifferences` configuration. This alternative stands entirely outside KREP-016's scope regardless of how that proposal evolves.

**How we would adapt if both 011 and 016 land first**

If KREP-016 is accepted and extended to permit self-targeting with `includeWhen` support on `patch` nodes, we would revise this proposal accordingly:

- **Drop Alt A.** A self-targeting `patch` node with `includeWhen` is `specPatch` — no reason to maintain a separate node type for the same mechanism.
- **Retain Alt B unchanged.** `stateWrite` → `status.kstate` is not expressible through a `patch` node and addresses a different ownership concern.
- **Note** that KREP-011's `data` nodes and write-back nodes compose cleanly for complex transition logic.

If KREP-016 lands without self-targeting or without `includeWhen` applying to `patch` nodes, this proposal remains necessary in full.

The core ask of this proposal — a mechanism for *triggered, convergent, self-directed state transitions* — isn't addressed by either KREP-011 or KREP-016 as currently written. We're very open to restructuring or consolidating if the maintainers see a cleaner path, and appreciate the pointers to both proposals.

jakobmoellerdev:
I will close this KREP. 
it has not been aligned with anyone of the maintainer team and it is very obvious heavy AI generation is in place without due review.

/close

PR #1131, opened 2026-03-10
Make string templates compile to concentation 

NicholasBlaskey:
Previously the codebase treated string templates as a special list of expressions. Replace this with building string template expressions as concatenation.

This has several advantages in simplying the code
- doesn't require separate code paths for templates vs expressions
- only required to consider single expression case
- can leave type checking to cel

One callout is the error format for type mismatch has changed, new error looks like
```
    Message:               failed to validate resource "deployment": failed to compile template expression "good-${json.marshal({\"name\": schema.spec.name})}-${4}" at path "spec.template.spec.containers[0].env[5].value": ERROR: <input>:1:58: found no matching overload for '_+_' applied to '(string, int)'
```

### DIFF
--- a/pkg/cel/expression.go
+++ b/pkg/cel/expression.go
@@ -75,6 +75,14 @@ func NewUncompiledSlice(exprs ...string) []*Expression {
 	return result
 }
 
+// UserExpression returns the user-facing expression string for error messages.
+func (e *Expression) UserExpression() string {
+	if e.OriginalTemplate != "" {
+		return e.OriginalTemplate
+	}
+	return e.Original
+}
+
 // Eval evaluates the compiled expression and returns the result.
 func (e *Expression) Eval(ctx map[string]any) (any, error) {
 	startTime := time.Now()
@@ -85,12 +93,12 @@ func (e *Expression) Eval(ctx map[string]any) (any, error) {
 
 	out, _, err := e.Program.Eval(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("eval %q: %w", e.Original, err)
+		return nil, fmt.Errorf("eval %q: %w", e.UserExpression(), err)
 	}
 
 	native, err := conversion.GoNativeType(out)
 	if err != nil {
-		return nil, fmt.Errorf("convert %q: %w", e.Original, err)
+		return nil, fmt.Errorf("convert %q: %w", e.UserExpression(), err)
 	}
 
 	return native, nil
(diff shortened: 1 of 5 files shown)


jakobmoellerdev:
/ok-to-test

this is awesome, thanks a lot!

pkg/graph/parser/parser.go
```diff
@@ -281,17 +279,34 @@ func parseString(field string, path string, expectedTypes []string) ([]variable.
...
+			Path:       path,
 		}}, nil
 	}
 	return nil, nil
 }
 
+// buildStringTemplate builds a CEL concatenation expression from a string
+// template with embedded expressions.
+// e.g. "prefix-${expr1}-${expr2}" → `"prefix-" + expr1 + "-" + expr2`
+func buildStringTemplate(original string, matches []exprMatch) string {
```
a-hilaly, line 294: I think this is where the secret sauce needs one more ingredient: we probably need to wrap each interpolated CEL fragement in parenthesis when lowering the templates in concatenations. Right now `prefix-${cond ? a: b}-sufix` becomes `"prefix-"+ cond ? a: b + "-suffix}` which changes precedence in CEL. The safe form is `"prefix-" + (cond ? a: b) + "suffix`. so we should probably emit `(...)` around each inner expression

pkg/runtime/resolver/resolver.go
```diff
@@ -92,48 +91,24 @@ func (r *Resolver) resolveField(field variable.FieldDescriptor) ResolutionResult
...
-
-		// setValueAtPath cannot fail here: if getValueFromPath succeeded,
-		// the path is valid and traversable.
-		_ = r.setValueAtPath(field.Path, replaced)
-		result.Resolved = true
-		result.Replaced = replaced
+	expr := field.Expression.Original
+	resolvedValue, ok := r.data[expr]
+	if !ok {
+		result.Error = fmt.Errorf("no data provided for expression: %s", expr)
```
a-hilaly: Do we want to use the OriginalTemplate here?

pkg/cel/expression.go
a-hilaly, line 102: ditto let's use OriginalTemplate here?

a-hilaly (commented):
Really nice work on this @NicholasBlaskey !

pkg/graph/parser/parser.go
NicholasBlaskey: good catch

added parenthesis to building the expression

pkg/runtime/resolver/resolver.go
NicholasBlaskey: went through and updated all error messages that use field.Expression

michaelhtm (approved):
Great PR @NicholasBlaskey

PR #1132, opened 2026-03-11
feat(instance): block graph propagation on terminating managed resources

a-hilaly:
When a managed resource is manually deleted, Kubernetes can keep the
object visible during finalization with a `deletionTimestamp` set. In
that window, kro was still treating the resource as present and could
continue reconciliation as if nothing was wrong.

That makes the behavior misleading for users and unsafe for dependency
ordering. A terminating managed resource cannot be recreated yet, and
downstream resources should not keep progressing while one of their
managed inputs is actively being deleted.

Detect terminating managed single resources and managed collection
items during normal reconcile, mark the instance as
`ResourcesReady=False` with reason `ResourceDeleting`, and requeue until
deletion completes instead of continuing apply/prune work.

### DIFF
--- a/pkg/controller/instance/controller.go
+++ b/pkg/controller/instance/controller.go
@@ -16,6 +16,7 @@ package instance
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"maps"
 	"strings"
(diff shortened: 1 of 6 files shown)


pkg/controller/instance/controller.go
```diff
@@ -194,6 +195,12 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 	if !ok || !strings.EqualFold(reconcileState, "disabled") {
 		rcx.Mark.ReconciliationActive()
 		if err := c.reconcileNodes(rcx); err != nil {
+			var deletingErr *resourceDeletingError
+			if errors.As(err, &deletingErr) {
+				rcx.Mark.ResourcesDeleting("%v", deletingErr)
+				_ = c.updateStatus(rcx)
```
jakobmoellerdev, line 201: shouldnt a status error propagate as well?

a-hilaly: it looks like we don't do that often in this file. shall we revisit that?

Issue #1031, opened 2026-02-09
Feature: Add allow breaking changes to controller argument in addition to the RGD annotation


shabbskagalwala:
### Feature Description

**Problem Statement**:
Kro recently started recognizing breaking changes https://kro.run/docs/concepts/rgd/overview#annotations
And allows RGD authors to propagate these changes via an annotation. This is a much needed change however with teams owning multiple RGDs, this can be cumbersome when developing RGDs in sandbox environments, where each RGD needs to be annotated to allow new field additions or making fields required. 

**Proposed Solution**:
Add a flag on the controller arg to allow breaking changes like 

```
--allow-breaking-changes
```

Which can by default be set to `false` 

> Note: I do understand this can have a blast radius where breaking changes recognized by Kro will be allowed across all RGDs in the cluster

**Alternatives Considered**:
- Adding the annotation `kro.run/allow-breaking-changes` default to all RGDs - knowing the implications 
- This is however similar in nature to the controller argument 

**Additional Context**:

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue
* If you are interested in working on this feature, please leave a comment

jakobmoellerdev:
My personal take here is that we should at the moment not introduce additional behavior flags on top of this annotation until we get some feedback. maybe lets park this idea and see how users respond. Im not against a flag idea there in general.

shabbskagalwala:
Thanks @jakobmoellerdev, sounds good - we can park this for now - this is just a nice to have, make life easy kind of thing when multiple teams are iterating constantly on RGDs. 

There is a workaround, using the annotation, which we are adding across all RGDs in non-prod, but not having to do that and have control via a flag on the controller would be a great thing :) 

We can revisit later if the maintainers think this is a good to add feature

a-hilaly:
I'm running into a similar situation where we're managing multiple RGDs across non prod environments and having to annotate each one individually gets tedious quickly. A controller level flag would be a nice quality of life improvement.

I'm thinking that now we have [a helm chart rgd config](https://github.com/kubernetes-sigs/kro/blob/main/helm/values.yaml#L133), this could be exposed as something like `config.rgd.allowBreakingChangesDefault` ?

a-hilaly:
[...]

PR #1138, opened 2026-03-12
fix: reconcile on annotation changes

NicholasBlaskey:
Fix a bug of Kro not reconciling on annotation changes

A user could update an RGD with `kro.run/allow-breaking-changes` but this update may not be considered by kro until another reconcile is triggered.

Mentioned in this issue https://github.com/kubernetes-sigs/kro/issues/1031#issuecomment-3960682798


NicholasBlaskey:
closing since this is duplicated

PR #1139, opened 2026-03-12
feat(cel): add omit() for conditional field omission in templates

michaelhtm:
Implements KREP-017 #1121
When a CEL expression resolves to the `omit()` sentinel, the resolver
removes the containing field from the rendered object instead of writing
a value. This lets users author optional fields that are absent from the apply
payload when a condition is met, which is required by many CRDs that distinguish
between field absence and an explicit null/empty value.

Since kro uses ServerSideApply, this field will no longer be managed by
kro. Any drifts that may occur to the omitted field will not be
detected.

Example usage:
```yaml
policy: ${schema.spec.policy != "" ? schema.spec.policy : omit()}
```

### DIFF
--- a/pkg/cel/conversion/conversions.go
+++ b/pkg/cel/conversion/conversions.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/cel-go/common/types"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/common/types/traits"
+	"github.com/kubernetes-sigs/kro/pkg/cel/sentinels"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
@@ -68,6 +69,10 @@ func GoNativeType(v ref.Val) (interface{}, error) {
 	case types.NullType:
 		return nil, nil
 	default:
+		// Check if this is the omit sentinel before falling through to error.
+		if _, ok := v.Value().(sentinels.Omit); ok {
+			return sentinels.Omit{}, nil
+		}
 		// For types we can't convert, return as is with an error
 		return v.Value(), fmt.Errorf("%w: %v", ErrUnsupportedType, v.Type())
 	}
--- a/pkg/cel/environment.go
+++ b/pkg/cel/environment.go
@@ -119,6 +119,11 @@ func BaseDeclarations() []cel.EnvOption {
 			library.Maps(),
 			library.JSON(),
 			library.Lists(),
+			// Omit() is registered globally so CEL can parse and type-check it
+			// everywhere. The graph builder rejects it in restricted contexts
+			// (includeWhen, readyWhen, forEach) via inspectExpressionRestricted
+			// and validateAndCompileForEach.
+			library.Omit(),
 		}
 	})
 	return cachedBaseDeclarations
--- a/pkg/cel/library/omit.go
+++ b/pkg/cel/library/omit.go
@@ -1,0 +1,103 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package library
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/kubernetes-sigs/kro/pkg/cel/sentinels"
+)
+
+// omitVal is the CEL ref.Val wrapper for the omit sentinel.
+type omitVal struct{}
+
+// singleton — there's only one omit sentinel.
+var omitInstance = &omitVal{}
+
+func (v *omitVal) ConvertToNative(typeDesc reflect.Type) (any, error) {
+	if typeDesc == reflect.TypeOf(sentinels.Omit{}) {
+		return sentinels.Omit{}, nil
+	}
+	if typeDesc.Kind() == reflect.Interface {
+		return sentinels.Omit{}, nil
+	}
+	return nil, fmt.Errorf("unsupported native conversion from omit to %v", typeDesc)
+}
+
+func (v *omitVal) ConvertToType(typeVal ref.Type) ref.Val {
+	if typeVal == types.TypeType {
+		return types.NewObjectType("kro.omit")
+	}
+	return types.NewErr("unsupported conversion from omit to %v", typeVal)
+}
+
+func (v *omitVal) Equal(other ref.Val) ref.Val {
+	return types.NewErr("omit() is a field-removal sentinel and cannot be compared")
+}
+
+func (v *omitVal) Type() ref.Type {
+	return types.NewObjectType("kro.omit")
+}
+
+func (v *omitVal) Value() any {
+	return sentinels.Omit{}
+}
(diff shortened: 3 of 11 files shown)


pkg/cel/conversion/conversions.go
a-hilaly, line 1: i think your changes deserve a `sentinels.go` file where we want to declare this sentinel and future sentinels.

pkg/runtime/resolver/resolver.go
```diff
@@ -113,43 +124,75 @@ func (r *Resolver) resolveField(field variable.FieldDescriptor) ResolutionResult
...
+
+	// Remove the target field from its parent.
+	last := segments[len(segments)-1]
+	if last.Index >= 0 {
+		// Array element removal is not yet supported. Splicing an element out
+		// shifts all subsequent indices, so when multiple elements in the same
+		// array are omitted the resolver must process them from the highest
+		// index to the lowest to keep the remaining indices valid. The current
+		// Resolve loop does not guarantee that ordering.
+		return fmt.Errorf("omit does not support removing array elements (path: %s)", path)
```
a-hilaly: Let's support array element omission. Since `omit()` returns a value that gets written into the resource like anything else, we can handle arrays with a "simple post resolution pass" - walk the resource, find any array that contains an omit marker, rebuild it without those elements. `[a, b, omit(), d]` comes `[a, b, d]`. No index tracking needed.

test/integration/suites/core/omit_test.go
a-hilaly, line 1: Can we get a test for array item omission please?

pkg/cel/environment.go
```diff
@@ -117,6 +117,7 @@ func BaseDeclarations() []cel.EnvOption {
 			library.Random(),
 			library.Maps(),
 			library.JSON(),
+			library.Omit(),
```
a-hilaly, line 126: Add a comment noting that restricted context rejection is handled by the compiler

pkg/graph/builder.go
```diff
@@ -837,9 +837,31 @@ func inspectExpressionRestricted(inspector *ast.Inspector, expr string, allowedI
...
+	for _, fc := range result.FunctionCalls {
+		if fc.Name == "omit" {
+			return ast.ExpressionInspection{}, fmt.Errorf("omit() can only be used in resource template expressions")
+		}
+	}
 	return result, nil
 }
 
+// expressionUsesOmit reports whether the given expression contains a call to omit().
+func expressionUsesOmit(inspector *ast.Inspector, expr string) bool {
```
a-hilaly: This calls `inspector.Inspect()` on expressions already inspected elsewhere in the validation flow. how about just looking at the result instead of reinspecting?

pkg/graph/builder.go
```diff
@@ -837,9 +837,31 @@ func inspectExpressionRestricted(inspector *ast.Inspector, expr string, allowedI
...
 		return ast.ExpressionInspection{}, fmt.Errorf("uses unknown functions: %v", result.UnknownFunctions)
 	}
+	// omit() is registered globally but is only valid in resource template
+	// expressions. Reject it in restricted contexts (includeWhen, readyWhen,
+	// instance status).
+	for _, fc := range result.FunctionCalls {
+		if fc.Name == "omit" {
+			return ast.ExpressionInspection{}, fmt.Errorf("omit() can only be used in resource template expressions")
+		}
+	}
```
a-hilaly: ^ same to what you're doing here

pkg/graph/builder.go
```diff
@@ -1115,7 +1142,10 @@ func validateAndCompileTemplates(
 	typeProvider *krocel.DeclTypeProvider,
 	iteratorTypes map[string]*cel.Type,
 ) error {
-	// If we have iterator types (from forEach), extend the environment with those declarations
+	// TODO: Should we also prevent omit() on required resource fields like
+	// metadata.name? Currently nothing stops a user from
+	// writing e.g. `metadata.name: ${omit()}`, which would produce an
+	// invalid resource at apply time.
```
a-hilaly: happy to have this as future work. i think in general the compiler can do a pass to make sure required fields are all set to non null values and non null-resolvable CEL expressions

pkg/graph/builder_test.go
```diff
@@ -3486,6 +3487,16 @@ func newUnitInspector(t *testing.T, ids ...string) *ast.Inspector {
 	return ast.NewInspectorWithEnv(env, ids)
 }
 
+func newUnitInspectorWithOmit(t *testing.T, ids ...string) *ast.Inspector {
+	t.Helper()
+	env, err := krocel.DefaultEnvironment(
+		krocel.WithResourceIDs(ids),
+		krocel.WithCustomDeclarations([]cel.EnvOption{library.Omit()}),
+	)
```
a-hilaly: default environement already include the Omit library

pkg/cel/library/omit_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	out, _, err := prg.Eval(map[string]interface{}{})
+	require.NoError(t, err)
+
+	native, err := conversion.GoNativeType(out)
+	require.NoError(t, err)
+	assert.True(t, conversion.IsOmitSentinel(native))
+}
+
+func TestOmitInTernary(t *testing.T) {
```
a-hilaly, line 47: wanna test weird things like `omit() + 1` (should error) `omit() == omit()` (should be error too?)

pkg/cel/library/omit.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return types.NewErr("unsupported conversion from omit to %v", typeVal)
+}
+
+func (v *omitVal) Equal(other ref.Val) ref.Val {
+	if _, ok := other.(*omitVal); ok {
+		return types.True
+	}
+	return types.False
+}
```
a-hilaly, line 52: why do we want to allow things like `omit() == omit()` ?

test/integration/suites/core/omit_test.go
a-hilaly, line 1: need a case that updates `instance1` to set `optional: "now-present"` and verify the field appears on next reconcile.

a-hilaly (commented):
Overall good stuff, thanks Michael! please rebase. i left a few comments inline

pkg/graph/builder_test.go
michaelhtm: this was an older change, removing

pkg/cel/conversion/conversions.go
michaelhtm: that makes sense. Addressing this

pkg/runtime/resolver/resolver.go
michaelhtm: For arrays, one other issue i noticed was, what do we want to do if we have the following expression:
```yaml
a:
  b:
  - ${omit()}
```
do we want to apply b as empty, or should omit propagate the removal?

pkg/cel/library/omit.go
michaelhtm: you're right! i assumed this was required for type checking, removing

pkg/graph/builder.go
```diff
@@ -1159,7 +1165,14 @@ func validateAndCompileTemplates(
 	typeProvider *krocel.DeclTypeProvider,
 	iteratorTypes map[string]*cel.Type,
 ) error {
-	// If we have iterator types (from forEach), extend the environment with those declarations
+	// NOTE: omit() is allowed in template expressions. Restricted-context
+	// rejection (includeWhen, readyWhen, forEach) is handled by the compiler
+	// in inspectExpressionRestricted and validateAndCompileForEach.
+	//
+	// TODO: Add validation that required resource fields (e.g. metadata.name)
```
jakobmoellerdev, line 1172: I think this should be added in this PR.

michaelhtm: I believe this should be a separate PR that ensures required fields are not null, since currently kro does not catch them in compile time. 
Eg. something like this currently compiles
```yaml
  - id: bucket
    template:
      apiVersion: s3.services.k8s.aws/v1alpha1
      kind: Bucket
      metadata:
        name: null
```

a-hilaly: agreed there is a bigger picture here, happy to take a seperate PR.

a-hilaly (approved):

PR #1140, opened 2026-03-13
Use RGD metadata.name as the label for ResourceGraphDefinition controller metrics.

a-hilaly:
Replace the rgd_kind label with the idiomatic `name` label keyed on
  metadata.name. The RGD name is the natural resource identifier, whereas
  kind is a derived schema artifact.


michaelhtm:
Should we keep some informational details of the RGD as a label? It may be useful when debugging which RGD has issues.

a-hilaly:
@michaelhtm brought it back, just using rgd metadata name now

PR #1143, opened 2026-03-13
Add client-go and reflector metrics registration

a-hilaly:
controller-runtime v0.16+ stopped registering client-go latency, size,                                                                                      
  and retry histograms, and never registers reflector metrics. Add a                                                                                          
  `pkg/metrics` package that fills both gaps by assigning adapters directly                                                                                   
  to client-go globals and calling `cache.SetReflectorMetricsProvider`.

  New metrics:
  - `rest_client_request_duration_seconds` — histogram by verb
  - `rest_client_rate_limiter_duration_seconds` — histogram by verb
  - `rest_client_request_size_bytes` — histogram by verb
  - `rest_client_response_size_bytes` — histogram by verb
  - `rest_client_request_retries_total` — counter by code/method

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -35,6 +35,7 @@ import (
 	resourcegraphdefinitionctrl "github.com/kubernetes-sigs/kro/pkg/controller/resourcegraphdefinition"
 	"github.com/kubernetes-sigs/kro/pkg/dynamiccontroller"
 	"github.com/kubernetes-sigs/kro/pkg/graph"
+	"github.com/kubernetes-sigs/kro/pkg/metrics"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -52,6 +53,8 @@ func init() {
 }
 
 func main() {
+	metrics.Register()
+
 	var (
 		metricsAddr                                 string
 		enableLeaderElection                        bool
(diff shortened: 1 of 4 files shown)


pkg/metrics/doc.go
a-hilaly, line 1: @jakobmoellerdev what do you think of moving all metric declarations here and getting rid of the init functions all over the place?

cmd/controller/main.go
```diff
@@ -52,6 +53,8 @@ func init() {
 }
 
 func main() {
+	metrics.Register()
```
jakobmoellerdev, line 56: now instead of calling register without params, what about passing it the registry so people can inject their own metrics registry (i.e. when running outside of the controller runtime registries)?

a-hilaly: not sure if it's the right scope for this PR, we have many other metrics that could use the same pattern. I don't want us to have divergent APIs. happy to get all metrics moved here and use the same register func

jakobmoellerdev: The main issue i see is that its impossible to use your own registry right now so it is very hard for reuse.

a-hilaly: I'm happy to iterate on all the metrics stuff and make it more flexible, in an appropriate PR, just low prio for 0.9-rc1 imo

PR #1145, opened 2026-03-13
feat(cel): add cel.bind() support to default environment and AST inspector

### LINKED ISSUE #1144: feat(cel): add cel.bind() support to the default CEL environment and AST inspector
### Feature Description

**Problem Statement**:

`cel.bind()` is a macro from the `cel-go` extensions library (`ext.Bindings()`) that lets you assign an intermediate value to a local variable and reuse it in the same expression:

```
cel.bind(total, base + bonus, total * multiplier)
```

Without this, the only option is to repeat sub-expressions, which makes complex RGD CEL expressions verbose and hard to follow. For example, computing a value from several schema fields and then using that value in multiple branches requires duplicating the computation every time it appears.

There are actually two separate gaps today:

1. `ext.Bindings()` is not registered in `BaseDeclarations()`, so any expression using `cel.bind()` fails at RGD validation time with a compile error.
2. The AST inspector in `pkg/cel/ast/inspector.go` has no awareness of `cel.bind(varName, init, body)` scoping. When walking the expression tree to collect resource dependencies, the bound variable name gets reported as an unknown identifier, which causes spurious dependency errors on well-formed expressions.

**Proposed Solution**:

Two small changes:

1. Add `ext.Bindings()` to `BaseDeclarations()` in `pkg/cel/environment.go` alongside the existing `ext.Lists()`, `ext.Strings()`, etc.

2. In `inspectCall()` in `pkg/cel/ast/inspector.go`, detect the `bind` function name, inspect the init expression normally to collect its dependencies, then introduce the bound variable into the local scope before inspecting the body, and clean it up afterward. This is the same scoping pattern already used for `forEach` iterator variables.

I have a working implementation of both and am happy to send a PR.

**Alternatives Considered**:

Users can work around the missing `ext.Bindings()` by nesting expressions directly, but this quickly becomes unreadable for any non-trivial CEL expression. There is no workaround for the AST inspector issue — it produces false errors on valid expressions.

**Additional Context**:

`cel.bind()` is part of the standard `cel-go` extensions library that kro already depends on, so this adds no new dependency.

pnz1990:
Fixes #1144

## What

Two small changes to wire in `cel.bind()` support:

**1. `pkg/cel/environment.go`** — add `ext.Bindings()` to `BaseDeclarations()`

```go
ext.Bindings(),
```

Without this, any RGD expression using `cel.bind()` fails at validation time:
```
ERROR: found no matching overload for '@not_strictly_needed' applied to '()'
```

**2. `pkg/cel/ast/inspector.go`** — teach `inspectCall` to scope bound variables

The dependency-analysis walker had no awareness of `cel.bind(varName, init, body)` semantics. When walking the body, `varName` was treated as an undeclared identifier and reported as an unknown resource dependency, causing spurious errors on well-formed expressions.

The fix detects the `bind` function name, inspects the init expression normally, then introduces the bound variable into `loopVars` before inspecting the body — the same scoping pattern already used for `forEach` iterator variables.

## Tests

- `pkg/cel/environment_test.go` — `TestDefaultEnvironment_CelBind`: compile + eval of basic, reuse, nested, and string bind expressions through `DefaultEnvironment()`
- `pkg/cel/ast/inspector_test.go` — two cases in `TestNewInspectorWithEnv_CustomFunctionsNotResources`: single bind and nested bind, verifying the bound variable is not reported as an unknown identifier and that resource dependencies in the init expression are still collected correctly

### DIFF
--- a/pkg/cel/ast/inspector.go
+++ b/pkg/cel/ast/inspector.go
@@ -250,35 +250,6 @@ func (a *Inspector) inspectCall(ast *celast.AST, call celast.CallExpr, path stri
 
 	fn := call.FunctionName()
 
-	// Handle cel.bind(varName, initExpr, bodyExpr) — the variable name in
-	// the first argument must be added to loopVars before inspecting the body
-	// (third argument) so it isn't reported as an unknown identifier.
-	if fn == "bind" && len(call.Args()) == 3 {
-		varArg := call.Args()[0]
-		initArg := call.Args()[1]
-		bodyArg := call.Args()[2]
-
-		// Extract the variable name from the first argument (should be an Ident)
-		if varArg.Kind() == celast.IdentKind {
-			varName := varArg.AsIdent()
-			// Inspect the init expression (no special scoping)
-			out.merge(a.inspectExpr(ast, initArg, ""))
-			// Add bound variable to loop vars for the body
-			_, wasSet := a.loopVars[varName]
-			a.loopVars[varName] = struct{}{}
-			out.merge(a.inspectExpr(ast, bodyArg, ""))
-			if !wasSet {
-				delete(a.loopVars, varName)
-			}
-		} else {
-			// Fallback: inspect all args normally
-			for _, arg := range call.Args() {
-				out.merge(a.inspectExpr(ast, arg, ""))
-			}
-		}
-		return out
-	}
-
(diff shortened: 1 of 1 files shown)


a-hilaly (approved):
Thanks Rafael!
/lgtm

PR #1146, opened 2026-03-13
fix missing dynamic controller registration, and wire up missing metrics

jakobmoellerdev:
- **Unify informer ownership into a named owner-set model** in WatchManager, replacing the split `parentRefs` refcount + 
  `coordinator.HasRequestsForGVR()` check with a single `owners map[GVR]map[string]struct{}`. Informers now auto-stop when
  their last owner calls `ReleaseWatch`, eliminating the need for callers to manually check and stop.
  - **Store and explicitly remove the WatchManager's internal event handler** (`handlerReg`) in `stopWatchLocked`, ensuring
   deterministic cleanup instead of relying on GC after informer shutdown.
  - **Simplify `Register`/`Deregister`** in DynamicController: collapse 3 separate rollback paths into a single deferred
  cleanup, extract `parentEventHandlerFor()` and `enqueueExistingInstances()` helpers, and use `RunWithContext` instead of
  `Run(stopCh)`.
  - **Wire up 5 existing but uninstrumented Prometheus metrics**: `watchCount`, `instanceWatchCount`, `watchRequestCount`,
  `routeTotal`, `routeMatchTotal` — plus instrument `informerSyncDuration` in both the success and timeout paths of
  `ensureWatch`

### DIFF
--- a/pkg/dynamiccontroller/coordinator.go
+++ b/pkg/dynamiccontroller/coordinator.go
@@ -316,7 +316,9 @@ func (c *WatchCoordinator) RemoveParentGVR(parentGVR schema.GroupVersionResource
 		}
 		delete(c.instances, key)
 	}
-	c.decInstanceWatchCount(parentGVR)
+	// All instances for this parent are gone — delete the gauge outright
+	// rather than decrementing once (which would undercount for N > 1).
+	instanceWatchCount.DeleteLabelValues(parentGVR.String())
 
 	orphanedGVRs := c.findOrphanedGVRsLocked(affectedGVRs)
 	c.mu.Unlock()
@@ -460,15 +462,15 @@ func (c *WatchCoordinator) removeScalarIndexLocked(key instanceKey, req *WatchRe
 		byName[nn] = filtered
 	}
 	gvrStr := req.GVR.String()
+	if removed > 0 {
+		watchRequestCount.WithLabelValues(gvrStr, "scalar").Sub(float64(removed))
+	}
 	if len(byName) == 0 {
 		delete(c.scalarIndex, req.GVR)
-		// If this GVR is also gone from collectionIndex, delete its gauge labels.
 		if len(c.collectionIndex[req.GVR]) == 0 {
 			watchRequestCount.DeleteLabelValues(gvrStr, "scalar")
 			watchRequestCount.DeleteLabelValues(gvrStr, "collection")
 		}
-	} else if removed > 0 {
-		watchRequestCount.WithLabelValues(gvrStr, "scalar").Sub(float64(removed))
 	}
 }
 
@@ -510,18 +512,17 @@ func (c *WatchCoordinator) removeCollectionIndexLocked(key instanceKey, req *Wat
 		filtered = append(filtered, e)
 	}
 	gvrStr := req.GVR.String()
+	if removed > 0 {
+		watchRequestCount.WithLabelValues(gvrStr, "collection").Sub(float64(removed))
+	}
 	if len(filtered) == 0 {
 		delete(c.collectionIndex, req.GVR)
-		// If this GVR is also gone from scalarIndex, delete its gauge labels.
 		if len(c.scalarIndex[req.GVR]) == 0 {
 			watchRequestCount.DeleteLabelValues(gvrStr, "scalar")
 			watchRequestCount.DeleteLabelValues(gvrStr, "collection")
 		}
 	} else {
 		c.collectionIndex[req.GVR] = filtered
-		if removed > 0 {
-			watchRequestCount.WithLabelValues(gvrStr, "collection").Sub(float64(removed))
-		}
 	}
 }
 
--- a/go.mod
+++ b/go.mod
@@ -58,6 +58,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/magiconair/properties v1.8.7 // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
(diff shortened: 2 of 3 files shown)


pkg/dynamiccontroller/watch_manager.go
```diff
@@ -60,62 +57,67 @@ type WatchManager struct {
...
 	m.mu.Unlock()
 
-	if err := m.EnsureWatch(gvr); err != nil {
-		m.ReleaseParentWatch(gvr)
+	if err := m.ensureWatch(gvr); err != nil {
+		m.ReleaseWatch(gvr, ownerID)
 		return err
 	}
 	return nil
 }
```
a-hilaly: This owner add + ensure sequence is racy when `EnsureWatch` and `ReleaseWatch` run concurrently. If goroutine A adds `ownerID` here and unlocks, goroutine B can call `ReleaseWatch`, remove the last owner, and return before any informer exists. A then continues into `ensureWatch()` and creates a live informer with zero owners, so the watch can stay running indefinitely after the last release. Since register/deregister churn can happen in parallel here, ownership changes and informer creation need to be atomic

I'm running some churn tests on this branch to validate this.

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -368,74 +368,75 @@ func (dc *DynamicController) Register(
...
 
-		gvrCount.Inc()
-		handlerAttachTotal.WithLabelValues("parent").Inc()
-		handlerCount.WithLabelValues("parent").Inc()
-		dc.log.V(1).Info("Attached parent watch", "gvr", parent)
+	// Register event handler directly on the parent informer.
+	reg, err := dc.watches.GetInformer(parent).AddEventHandler(dc.parentEventHandlerFor(parent))
+	if err != nil {
+		return fmt.Errorf("add parent handler %s: %w", parent, err)
 	}
```
a-hilaly, line 381: the old code had nil checks on `GetInformer`+rollback,  i think we should bring that back

pkg/dynamiccontroller/coordinator.go
```diff
@@ -539,6 +540,36 @@ func (w *instanceWatcher) Done(commit bool) {
 	w.coordinator.doneInstance(key)
 }
 
+// updateWatchRequestMetrics recalculates the watch request gauge metrics
+// from the current index state. Must be called with c.mu held (at least RLock).
+func (c *WatchCoordinator) updateWatchRequestMetrics() {
```
a-hilaly: This only sets gauges for `GVRs` currently in the indexes. When a `GVR` is fully removed from both `scalarIndex` and `collectionIndex` its watchRequestCount labels are never zeroed - the gauge retains its last non-zero value permanently. Same issue with `instanceWatchCount` below: removed parent GVRs leave stale gauge values.

we need a 
```go
watchRequestCount.DeleteLabelValues(gvr.String(), "scalar")
watchRequestCount.DeleteLabelValues(gvr.String(), "collection")
```

in the `stopWatches` or `RemoveInstance`/`RemoveParentGVR`

pkg/dynamiccontroller/dynamic_controller.go
jakobmoellerdev: I dont understand when this would ever return nil.

jakobmoellerdev: I reverted this and added a test

pkg/dynamiccontroller/watch_manager.go
jakobmoellerdev: I have reworked the watch registration to make sure this doesnt happen

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: I have adjusted the metrics registration so that we dont need this helper method.

pkg/dynamiccontroller/coordinator.go
```diff
@@ -474,33 +484,58 @@ func (c *WatchCoordinator) findOrphanedGVRsLocked(gvrs []schema.GroupVersionReso
...
+// decInstanceWatchCount decrements the instanceWatchCount gauge for the given
+// parent GVR, deleting the label set entirely when no instances remain.
+// Must be called with c.mu held, after the instance has been deleted.
+func (c *WatchCoordinator) decInstanceWatchCount(parentGVR schema.GroupVersionResource) {
+	for key := range c.instances {
+		if key.parentGVR == parentGVR {
+			instanceWatchCount.WithLabelValues(parentGVR.String()).Dec()
+			return
+		}
 	}
```
a-hilaly, line 538: `RemoveParentGVR` deletes all instances for a `parentGVR` in a loop, then calls `decInstanceWatchCount` once. If there were N instances, only 1 Dec() or DeleteLabelValues happens. The gauge will be wrong for any N > 1

pkg/dynamiccontroller/coordinator.go
```diff
@@ -457,8 +459,16 @@ func (c *WatchCoordinator) removeScalarIndexLocked(key instanceKey, req *WatchRe
...
 		byName[nn] = filtered
 	}
+	gvrStr := req.GVR.String()
 	if len(byName) == 0 {
 		delete(c.scalarIndex, req.GVR)
+		// If this GVR is also gone from collectionIndex, delete its gauge labels.
+		if len(c.collectionIndex[req.GVR]) == 0 {
+			watchRequestCount.DeleteLabelValues(gvrStr, "scalar")
+			watchRequestCount.DeleteLabelValues(gvrStr, "collection")
+		}
```
a-hilaly, line 473: we need to handle if len(removed) > 0 here

a-hilaly: how about instead
```go
  if removed > 0 {                                                                                                                                                                                               
      watchRequestCount.WithLabelValues(gvrStr, "scalar").Sub(float64(removed))
  }                                                                                                                                                                                                              
  if len(byName) == 0 {
      delete(c.scalarIndex, req.GVR)                                                                                                                                                                             
      if len(c.collectionIndex[req.GVR]) == 0 {
          watchRequestCount.DeleteLabelValues(gvrStr, "scalar")
          watchRequestCount.DeleteLabelValues(gvrStr, "collection")                                                                                                                                              
      }                                                                                                                                                                                                          
  }
```

jakobmoellerdev: agreed, good callout!

pkg/dynamiccontroller/coordinator.go
jakobmoellerdev: I have now replaced the call for parents with instanceWatchCount.DeleteLabelValues(parentGVR.String()) to avoid this

Issue #367, opened 2025-03-09
Create DevContainer for KRO


yashisrani:
- DevContainer makes developer life easier. 
- Set up a development container (DevContainer) configuration for the KRO project to streamline the development environment setup. 
- The DevContainer should include all necessary tools, dependencies, and configurations to ensure consistency across development environments.

a-hilaly:
Generally we try to avoid putting dev specific configuration in the kro repository, it is up to users to choose their setup. Thank you @yashisrani 

/close

PR #1148, opened 2026-03-13
feat(cel): add index-mutation functions to the lists library

### LINKED ISSUE #1147: feat(cel): add index-mutation functions to the lists library (setIndex, insertAt, removeAt)
### Feature Description

**Problem Statement**:

CEL's built-in list operations are all read-only. Functions like `filter()`, `map()`, `slice()`, `reverse()`, and `distinct()` return new lists derived from an existing one, but there is no way to return a new list with a single element replaced, inserted, or removed at a specific index.

This gap is particularly noticeable when writing RGD expressions that manage ordered arrays in instance specs — for example, a `[]int` field tracking per-replica state, or a `[]string` field holding a slot allocation table. The only workaround today is to reconstruct the entire list with a `map()` comprehension and a conditional, which is verbose and hard to read:

```cel
# Replacing element at index 2 — current workaround
schema.spec.slots.map(i, i == 2 ? newValue : schema.spec.slots[i])
```

The intent is buried. The index variable `i` is reused as both a loop counter and a value, which only works because the list is `[]int`. For `[]string` or `[]dyn` it breaks entirely — `i` is the index, not the element, so you'd need:

```cel
lists.range(size(schema.spec.tags)).map(i, i == 2 ? "new-tag" : schema.spec.tags[i])
```

This is workable but noisy, and it requires knowing about `lists.range` just to perform what is conceptually a single-element update.

**Proposed Solution**:

Three new functions in `pkg/cel/library/lists.go`, alongside the existing `maps.merge` pattern:

```
lists.setIndex(list(dyn), int, dyn)   -> list(dyn)   // replace element at index
lists.insertAt(list(dyn), int, dyn)   -> list(dyn)   // insert element before index
lists.removeAt(list(dyn), int)        -> list(dyn)   // remove element at index
```

The same expressions from above become:

```cel
lists.setIndex(schema.spec.slots, 2, newValue)
lists.insertAt(schema.spec.tags, 0, "prepended")
lists.removeAt(schema.spec.items, idx)
```

All three functions are pure — they return a new list and do not mutate the input. Bounds are checked and an error is returned for out-of-range indices. `insertAt` with `index == size(list)` appends (equivalent to `list + [value]`).

**Why this belongs in kro rather than cel-go**:

`cel-go/ext` lists covers general-purpose functional operations (sort, distinct, flatten, slice). Index-mutation is a different category — it's about producing modified copies of a list at a known position, which is an application-layer concern driven by how Kubernetes resource specs are structured (ordered arrays of scalars or objects). The existing `maps.merge` in `pkg/cel/library/maps.go` is the right precedent: kro maintains its own library for operations that are specific to the resource-management domain.

**Alternatives Considered**:

- `map()` + conditional: works for simple cases but requires `lists.range` to get a counter, is verbose, and the intent is not clear to someone reading the RGD for the first time.
- `cel.bind()` + `map()`: reduces repetition but doesn't change the fundamental verbosity.
- No workaround exists for `insertAt` or `removeAt` without reconstructing the entire list manually.

**Additional Context**:

`slice()` is already available via `ext.Lists()` and covers the read-only slicing case, so it is deliberately excluded from this proposal.

I have a working implementation and am happy to send a PR if the approach looks good.

pnz1990:
## What

Three new functions in `pkg/cel/library/lists.go`, alongside the existing `maps.merge` pattern. All are pure — they return a new list and do not modify the input.

**1. `pkg/cel/library/lists.go`** — new functions:

```
lists.setIndex(list(dyn), int, dyn) -> list(dyn)   // replace element at index
lists.insertAt(list(dyn), int, dyn) -> list(dyn)   // insert before index; index == size appends
lists.removeAt(list(dyn), int)      -> list(dyn)   // remove element at index
```

Bounds are checked on all three. Out-of-range index returns an error. Implemented with `types.NewRefValList` so mixed-type lists round-trip correctly.

**2. `pkg/cel/environment.go`** — register `library.Lists()` in `BaseDeclarations()`

**No overlap with existing libraries** — `cel-go/ext.Lists()` and `k8s.io/apiserver` lists library both cover read-only operations only (`slice`, `sort`, `distinct`, `flatten`, `isSorted`, `sum`, `min`, `max`, `indexOf`). None of them have mutation functions.

## Why

CEL has no way to return a new list with a single element changed at a known index. The only workaround is to reconstruct the entire list through a `map()` comprehension with a conditional — which requires knowing `lists.range` just to get a counter, and breaks for non-int element types:

```cel
# before: intent buried, breaks for list(string)
lists.range(size(schema.spec.tags)).map(i, i == 2 ? "new-tag" : schema.spec.tags[i])

# after
lists.setIndex(schema.spec.tags, 2, "new-tag")
```

There is no workaround at all for `insertAt` or `removeAt` without rebuilding the list manually element by element.

## Tests

- `pkg/cel/library/lists_test.go` — `TestListsSetIndex`, `TestListsInsertAt`, `TestListsRemoveAt`: int and string elements, bounds errors, edge cases (empty list, single-element list, append via `insertAt`)
- `TestListsComposition`: chains of operations to verify purity (input not mutated, round-trips)

Closes #1147

### DIFF
--- a/pkg/cel/library/lists.go
+++ b/pkg/cel/library/lists.go
@@ -26,43 +26,43 @@ import (
 // Lists returns a CEL library that provides index-mutation functions for lists.
 // All functions are pure — they return a new list and do not modify the input.
 //
-// # SetIndex
+// # SetAtIndex
 //
 // Returns a new list with the element at index replaced by value.
 // Index must be in [0, size(list)).
 //
-//	lists.setIndex(list(T), int, T) -> list(T)
+//	lists.setAtIndex(list(T), int, T) -> list(T)
 //
 // Examples:
 //
-//	lists.setIndex([1, 2, 3], 1, 99)          // [1, 99, 3]
-//	lists.setIndex(["a", "b", "c"], 0, "z")   // ["z", "b", "c"]
+//	lists.setAtIndex([1, 2, 3], 1, 99)          // [1, 99, 3]
+//	lists.setAtIndex(["a", "b", "c"], 0, "z")   // ["z", "b", "c"]
 //
-// # InsertAt
+// # InsertAtIndex
 //
 // Returns a new list with value inserted before the element at index.
 // Index must be in [0, size(list)]. An index equal to size(list) appends.
 //
-//	lists.insertAt(list(T), int, T) -> list(T)
+//	lists.insertAtIndex(list(T), int, T) -> list(T)
 //
 // Examples:
 //
-//	lists.insertAt([1, 2, 3], 1, 99)   // [1, 99, 2, 3]
-//	lists.insertAt([1, 2, 3], 0, 99)   // [99, 1, 2, 3]
-//	lists.insertAt([1, 2, 3], 3, 99)   // [1, 2, 3, 99]
+//	lists.insertAtIndex([1, 2, 3], 1, 99)   // [1, 99, 2, 3]
+//	lists.insertAtIndex([1, 2, 3], 0, 99)   // [99, 1, 2, 3]
+//	lists.insertAtIndex([1, 2, 3], 3, 99)   // [1, 2, 3, 99]
 //
-// # RemoveAt
+// # RemoveAtIndex
 //
 // Returns a new list with the element at index removed.
 // Index must be in [0, size(list)).
 //
-//	lists.removeAt(list(T), int) -> list(T)
+//	lists.removeAtIndex(list(T), int) -> list(T)
 //
 // Examples:
 //
-//	lists.removeAt([1, 2, 3], 1)   // [1, 3]
-//	lists.removeAt([1, 2, 3], 0)   // [2, 3]
-//	lists.removeAt([1, 2, 3], 2)   // [1, 2]
+//	lists.removeAtIndex([1, 2, 3], 1)   // [1, 3]
+//	lists.removeAtIndex([1, 2, 3], 0)   // [2, 3]
+//	lists.removeAtIndex([1, 2, 3], 2)   // [1, 2]
 func Lists(options ...ListsOption) cel.EnvOption {
 	l := &listsLibrary{version: math.MaxUint32}
 	for _, o := range options {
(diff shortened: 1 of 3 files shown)


pkg/cel/library/lists.go
```diff
@@ -0,0 +1,232 @@
...
+		// lists.setIndex(arr list(dyn), index int, value dyn) -> list(dyn)
+		cel.Function("lists.setIndex",
+			cel.Overload("lists.setIndex_list_int_dyn",
+				[]*cel.Type{listDyn, cel.IntType, cel.DynType},
+				listDyn,
+				cel.FunctionBinding(listsSetIndex),
+			),
+		),
+
+		// lists.insertAt(arr list(dyn), index int, value dyn) -> list(dyn)
```
jakobmoellerdev: This likely needs adjust for Typed Parameters instead of using Dyn hardcoding. Can we code them not against dyn but `cel.TypeParamType(name)`?

pnz1990: updated to use cel.TypeParamType("T"). also added a TestListsTypeInference test .

pkg/cel/library/lists.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func (l *listsLibrary) ProgramOptions() []cel.ProgramOption {
+	return nil
+}
+
+// listsSetLegacy is the original lists.set implementation, typed list(int) only.
+// Kept for backwards compatibility with existing RGDs using lists.set.
+func listsSetLegacy(args ...ref.Val) ref.Val {
+	if len(args) != 3 {
```
a-hilaly: This is legacy now, shall we remove it?

pkg/cel/library/lists.go
```diff
@@ -0,0 +1,232 @@
...
+	return []cel.EnvOption{
+		// lists.set is kept for backwards compatibility with existing RGDs.
+		// It is typed list(int) only. New code should use lists.setIndex.
+		cel.Function("lists.set",
+			cel.Overload("lists.set_list_int_int",
+				[]*cel.Type{cel.ListType(cel.IntType), cel.IntType, cel.IntType},
+				cel.ListType(cel.IntType),
+				cel.FunctionBinding(listsSetLegacy),
+			),
+		),
```
a-hilaly: ditto, we don't need to registry legacy, as it's not legacy for kro

pkg/cel/library/lists.go
```diff
@@ -0,0 +1,232 @@
...
+// Examples:
+//
+//	lists.removeAt([1, 2, 3], 1)   // [1, 3]
+//	lists.removeAt([1, 2, 3], 0)   // [2, 3]
+//	lists.removeAt([1, 2, 3], 2)   // [1, 2]
+func Lists() cel.EnvOption {
+	return cel.Lib(&listsLibrary{})
+}
+
+type listsLibrary struct{}
```
a-hilaly: Can you add the version option pattern (`ListsVersion`, `ListsOption`) for consistency with `maps.go` and `json.go`?

pnz1990:
**Downstream rebase note (for maintainers and author)**

Once this PR merges, the following rebase chain is required:

1. **Rebase `pnz1990/kro` branch `cel-writeback-d`** onto the new upstream `main` — this branch carries additional krombat-private patches (specPatch, stateWrite, CSV library, seededInt) on top of the lists functions introduced here
2. **Update `github.com/pnz1990/krombat` backend `go.mod`** — the `replace` directive for `github.com/kubernetes-sigs/kro` is pinned to a `cel-writeback-d` commit SHA; after the rebase it must be updated to the new SHA so the backend CEL playground (issue #367) picks up the merged lists functions

Tracking issue on the consuming project: https://github.com/pnz1990/krombat/issues/367

pkg/cel/library/lists.go
```diff
@@ -0,0 +1,217 @@
...
+
+// Lists returns a CEL library that provides index-mutation functions for lists.
+// All functions are pure — they return a new list and do not modify the input.
+//
+// # SetIndex
+//
+// Returns a new list with the element at index replaced by value.
+// Index must be in [0, size(list)).
+//
+//	lists.setIndex(list(T), int, T) -> list(T)
```
jakobmoellerdev: shouldnt this be `setAtIndex()` since youre not setting the actual index? Below functions are also not named ideally I think.

pkg/cel/library/lists.go
jakobmoellerdev, line 1: The new functions should be documented at https://kro.run/next/docs/concepts/rgd/cel-expressions#available-cel-libraries I think, I am noticing now that were probably lacking docs for existing owned functions by kro as well. @a-hilaly

pkg/cel/library/lists.go
pnz1990: ty. renamed all functions for consistency

pkg/cel/library/lists.go
pnz1990: added the kro-owned index mutation library to the Available CEL Libraries

PR #1149, opened 2026-03-13
feat(telemetry): Add events, metrics and logs for instance status condition transitions

spattk:
#### What
Adds observability for instance status condition transitions — Prometheus metrics, Kubernetes events, and structured logs. The whole thing is __feature-gated__ behind `--enable-instance-telemetry` (default: `false`), so zero impact unless you opt in.

#### Why

Today we have no visibility into how instance conditions change over time. When debugging stuck or flapping instances, you're flying blind. This gives us:

- __`kro_instance_condition_current_status_seconds`__ (gauge) — how long a condition has been in its current state. Useful for catching stuck instances.
- __`kro_instance_condition_transitions_total`__ (counter) — tracks every status/reason flip. Good for alerting on flapping.
- __K8s Events__ — `Eventf` on the instance for each transition, so `kubectl describe` shows the history.
- __Structured logs__ — `condition transitioned` log lines with old/new status, reason, etc.

#### Performance note

This adds per-instance Prometheus label cardinality (gvr × namespace × name × condition_type × status × reason), which can grow with cluster size. The feature gate keeps this off by default. When enabled, the resync period is shortened to 60s for level-based stuck detection — the `DeepEqual` guard in `updateStatus` prevents unnecessary API writes so this is safe. 

If folks running 10,000s of instances hit performance bottlenecks with this enabled, we'll optimize in a follow-up — e.g., by reading instances from the informer cache instead of issuing a live GET per reconcile, and tuning the cardinality model to reduce memory footprint.

#### Testing
- Validated the tests passed in controller_test.go
- Ran it against a kubernetes clusters to see the metrics and events for the instance object.

```
Events:
  Type    Reason          Age   From                     Message
  ----    ------          ----  ----                     -------
  Normal  ResourcesReady  10s   kro-instance-controller  Status condition transitioned, Type: ResourcesReady, Status: False -> True, Reason: AllResourcesReady
  Normal  Ready           10s   kro-instance-controller  Status condition transitioned, Type: Ready, Status: False -> True, Reason: Ready
```

```
curl -s localhost:8078/metrics | grep kro_instance_condition

kro_instance_condition_current_status_seconds{condition_status="False",condition_type="ReconciliationSuspended",gvr="kro.run/v1alpha1/applications",name="my-app-instance",namespace="default",reason="Active"} 288.984465523
```

### DIFF
--- a/pkg/controller/instance/controller.go
+++ b/pkg/controller/instance/controller.go
@@ -138,7 +138,7 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 	}()
 
 	//--------------------------------------------------------------
-	// 1. Load instance; if gone, nothing to do
+	// 1. Load instance; snapshot conditions for telemetry diff
 	//--------------------------------------------------------------
 	inst, err := c.client.Dynamic().
 		Resource(c.gvr).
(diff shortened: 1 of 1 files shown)


spattk:
Will split the PR into 
- events: https://github.com/kubernetes-sigs/kro/pull/1151
- and open up a follow up PR for metrics

PR #1151, opened 2026-03-13
feat(events): Add events for instance status condition transition

spattk:
### What

Emits Kubernetes Events on instance status condition transitions. Every time a condition's status changes (e.g., `ResourcesReady` goes from `False` → `True`, or `InstanceManaged` appears for the first time), the controller fires an `Eventf` on the instance object. This is always on — no feature gate, no opt-in required.

### Why

Today, when an instance gets stuck or flaps between states, there's no native Kubernetes breadcrumb trail to show what happened. You have to dig through controller logs and correlate timestamps manually.

With condition events:

- __`kubectl describe`__ on any instance now shows a history of condition transitions — what changed, from/to status, and reason — right in the Events section.
- __Event watchers__ (monitoring tools, dashboards, alert pipelines) can subscribe to these transitions without parsing logs.

This lays the foundation for richer telemetry (metrics, structured logs) in a follow-up, but stands on its own as immediately useful observability that every user gets out of the box.

#### With no feature gate passed or when `InstanceConditionEvents=false`
```
go run ./cmd/controller --zap-log-level=info --allow-crd-deletion
Events:                    <none>
```

#### With `InstanceConditionEvents=true`
```
go run ./cmd/controller --zap-log-level=info --allow-crd-deletion --feature-gates=InstanceConditionEvents=true

Events:
  Type     Reason                   Age   From                                 Message
  ----     ------                   ----  ----                                 -------
  Normal   InstanceManaged          7s    kro/applications.kro.run-controller  none -> True: Managed
  Normal   GraphResolved            7s    kro/applications.kro.run-controller  none -> True: Resolved
  Warning  ReconciliationSuspended  7s    kro/applications.kro.run-controller  none -> False: Active
  Warning  ResourcesReady           7s    kro/applications.kro.run-controller  none -> False: NotReady
  Warning  Ready                    7s    kro/applications.kro.run-controller  none -> False: NotReady
  Normal   ResourcesReady           6s    kro/applications.kro.run-controller  False -> True: AllResourcesReady
  Normal   Ready                    6s    kro/applications.kro.run-controller  False -> True: Ready
```

### DIFF
--- a/pkg/controller/instance/controller.go
+++ b/pkg/controller/instance/controller.go
@@ -27,11 +27,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
 	kroclient "github.com/kubernetes-sigs/kro/pkg/client"
 	"github.com/kubernetes-sigs/kro/pkg/dynamiccontroller"
+	"github.com/kubernetes-sigs/kro/pkg/features"
 	"github.com/kubernetes-sigs/kro/pkg/graph"
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
 	"github.com/kubernetes-sigs/kro/pkg/requeue"
@@ -90,6 +92,9 @@ type Controller struct {
 	childResourceLabeler metadata.Labeler
 	reconcileConfig      ReconcileConfig
 	coordinator          *dynamiccontroller.WatchCoordinator
+
+	// eventRecorder emits K8s Events on condition transitions.
+	eventRecorder record.EventRecorder
 }
 
 // NewController constructs a new controller with static RGD.
@@ -102,6 +107,7 @@ func NewController(
 	instanceLabeler metadata.Labeler,
 	childResourceLabeler metadata.Labeler,
 	coord *dynamiccontroller.WatchCoordinator,
+	eventRecorder record.EventRecorder,
 ) *Controller {
 	return &Controller{
 		log:                  log,
@@ -112,6 +118,7 @@ func NewController(
 		childResourceLabeler: childResourceLabeler,
 		reconcileConfig:      reconcileConfig,
 		coordinator:          coord,
+		eventRecorder:        eventRecorder,
 	}
 }
 
@@ -135,7 +142,7 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 	}()
 
 	//--------------------------------------------------------------
-	// 1. Load instance; if gone, nothing to do
+	// 1. Load instance; snapshot conditions for event diff
 	//--------------------------------------------------------------
 	ri := c.client.Dynamic().Resource(c.gvr)
 	var inst *unstructured.Unstructured
@@ -153,6 +160,20 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 		return err
 	}
 
+	// Emit condition events on every return path (behind feature gate).
+	var rcx *ReconcileContext
+	if features.FeatureGate.Enabled(features.InstanceConditionEvents) {
+		initialConditions := conditionsFromInstance(inst)
+		defer func() {
+			obj := inst
+			if rcx != nil {
+				obj = rcx.Instance
+			}
+			finalConditions := conditionsFromInstance(obj)
+			emitConditionEvents(c.eventRecorder, obj, initialConditions, finalConditions)
+		}()
+	}
+
 	//--------------------------------------------------------------
 	// 2. Create a fresh runtime for this reconciliation
 	//--------------------------------------------------------------
@@ -165,7 +186,7 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 	//--------------------------------------------------------------
 	// 3. Build reconciliation context (clients, mapper, labeler, runtime)
 	//--------------------------------------------------------------
-	rcx := NewReconcileContext(
+	rcx = NewReconcileContext(
 		ctx, log, c.gvr,
 		c.client.Dynamic(),
 		c.client.RESTMapper(),
--- a/pkg/controller/instance/events.go
+++ b/pkg/controller/instance/events.go
@@ -1,0 +1,85 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package instance
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/kubernetes-sigs/kro/api/v1alpha1"
+)
+
+// emitConditionEvents fires a K8s Event for every status condition that
+// transitioned between the initial and final snapshots.
+func emitConditionEvents(
+	recorder record.EventRecorder,
+	inst *unstructured.Unstructured,
+	initial, final []v1alpha1.Condition,
+) {
+	initialByType := indexConditionsByType(initial)
+
+	for _, cond := range final {
+		old, existed := initialByType[cond.Type]
+		if existed && old.Status == cond.Status {
+			continue
+		}
+
+		// default shown in transition events when a condition appears for the first time (e.g. "none -> True")
+		oldStatus := "none"
+		if existed {
+			oldStatus = string(old.Status)
+		}
+
+		reason := ""
+		if cond.Reason != nil {
+			reason = *cond.Reason
+		}
+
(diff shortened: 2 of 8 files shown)


pkg/controller/instance/events.go
```diff
@@ -0,0 +1,75 @@
...
+// emitConditionEvents fires a K8s Event for every status condition that
+// transitioned between the initial and final snapshots.
+func emitConditionEvents(
+	recorder record.EventRecorder,
+	inst *unstructured.Unstructured,
+	initial, final []v1alpha1.Condition,
+) {
+	if recorder == nil {
+		return
+	}
```
a-hilaly: can we make sure that recover is never nil to avoid this code path?

pkg/controller/instance/events.go
```diff
@@ -0,0 +1,75 @@
...
+			cond.Type, oldStatus, cond.Status, reason,
+		)
+	}
+}
+
+// conditionsFromInstance extracts status conditions from an unstructured object.
+func conditionsFromInstance(inst *unstructured.Unstructured) []v1alpha1.Condition {
+	if inst == nil {
+		return nil
+	}
```
a-hilaly: instance should and will never be nil

pkg/controller/instance/test_helpers_test.go
```diff
@@ -228,6 +228,7 @@ func newControllerUnderTest(t *testing.T, raw *dynamicfake.FakeDynamicClient, g
 		metadata.NewKROMetaLabeler(),
 		metadata.NewKROMetaLabeler(),
 		newControllerTestCoordinator(t),
+		nil, // eventRecorder
```
a-hilaly: any fake recorders we can use?

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -90,6 +94,7 @@ func (r *ResourceGraphDefinitionReconciler) SetupWithManager(mgr ctrl.Manager) e
 	r.Client = mgr.GetClient()
 	r.clientSet.SetRESTMapper(mgr.GetRESTMapper())
 	r.instanceLogger = mgr.GetLogger()
+	r.eventRecorder = mgr.GetEventRecorderFor("kro-instance-controller")
```
a-hilaly: i think it's too early to create recorder here, this means that all RGDs share the same source name. ideally we want to create a recorder per RGD/managed-GVK, that way it's easier to distinguish events across instances.

pkg/controller/resourcegraphdefinition/controller_test.go
```diff
@@ -158,6 +159,10 @@ func (m *stubManager) GetCache() cache.Cache {
 	return m.cache
 }
 
+func (m *stubManager) GetEventRecorderFor(_ string) record.EventRecorder {
+	return record.NewFakeRecorder(100)
```
a-hilaly: what's the 100 for?

pkg/controller/instance/events.go
```diff
@@ -0,0 +1,75 @@
...
+
+		reason := ""
+		if cond.Reason != nil {
+			reason = *cond.Reason
+		}
+
+		recorder.Eventf(inst, "Normal", string(cond.Type),
+			"Condition %s: %s -> %s, Reason: %s",
+			cond.Type, oldStatus, cond.Status, reason,
+		)
```
a-hilaly: What about `not-Normal` transitions? The event type should be `corev1.EventTypeWarning` when the condition moves to a worse state (True -> false)

pkg/controller/instance/events.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/kubernetes-sigs/kro/api/v1alpha1"
+)
+
+// emitConditionEvents fires a K8s Event for every status condition that
+// transitioned between the initial and final snapshots.
+func emitConditionEvents(
```
a-hilaly, line 28: i think this deserves some unit tests

pkg/controller/instance/events.go
```diff
@@ -0,0 +1,75 @@
...
+		)
+	}
+}
+
+// conditionsFromInstance extracts status conditions from an unstructured object.
+func conditionsFromInstance(inst *unstructured.Unstructured) []v1alpha1.Condition {
+	if inst == nil {
+		return nil
+	}
+	return (&unstructuredWrapper{inst}).GetConditions()
```
a-hilaly, line 75: not a blocker but we're doing a JSON marshal/unmarshal roud-trip for free here. For better perf we might want to keep raw Conditions at rsc level

pkg/controller/instance/events.go
a-hilaly: Condition type is already in the Reason column - no need to repeat it in the message. how about
```go
  recorder.Eventf(inst, "Normal", string(cond.Type),
      "%s -> %s (%s)", // or status changed: %s -> %s (%s)
      oldStatus, cond.Status, reason,
  )
```

which will produce something like:
```
  Type    Reason          Message
  Normal  ResourcesReady  False -> True (AllResourcesReady)
```

pkg/controller/instance/controller.go
```diff
@@ -241,9 +249,16 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
...
+	// 8. Persist status/conditions and emit events
 	//--------------------------------------------------------------
-	return c.updateStatus(rcx)
+	statusErr := c.updateStatus(rcx)
+
+	if statusErr == nil {
+		finalConditions := conditionsFromInstance(rcx.Instance)
+		emitConditionEvents(c.eventRecorder, rcx.Instance, initialConditions, finalConditions)
+	}
+
```
a-hilaly: how about a `defer` instead? there are serveral return branchs (deletion path/ ensure managed failure/reconcile nodes failure etc...)

pkg/controller/instance/events.go
spattk: done, once I made a fakeRecorder this became a no-op.

pkg/controller/instance/test_helpers_test.go
spattk: done.

pkg/controller/resourcegraphdefinition/controller_test.go
spattk: it's a default bucket size I have kept for now. We can revisit this later if required.

pkg/controller/instance/events.go
spattk: good callout. added.

pkg/controller/instance/events.go
spattk: yes makes sense. made changes

pkg/controller/instance/events.go
spattk: yeah, will do it in the next rev. Added a todo(perf) for now.

pkg/controller/instance/controller.go
spattk: yeah, I think I planned that initially but somehow fell through the cracks during removing the code for metrics. Added the defer back

pkg/controller/instance/events.go
spattk: done

pkg/controller/resourcegraphdefinition/controller.go
spattk: yes makes sense. moved the code inside to `setupMicroController`

jakobmoellerdev:
because such state changes can be quite taxing on the API server Im thinking of gating this feature. WDYT @a-hilaly ?

a-hilaly:
>because such state changes can be quite taxing on the API server Im thinking of gating this feature. WDYT @a-hilaly ?

It's very hard to tell without having events for normal/default instance evenst (like node created/ready/deleted)

a-hilaly:
@jakobmoellerdev probably you're right

jakobmoellerdev:
we have introduced a gate for this to address concerns, this should account for this gate and then we can get it merged.

spattk:
Nice, thanks Jakob!

spattk:
Updated the description to include both cases of valdiation that feature-gates are disabled by default and instance transition events are only emitted when the feature flag is enabled.

pkg/controller/instance/controller.go
```diff
@@ -153,6 +160,22 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
...
 
+	initialConditions := conditionsFromInstance(inst)
+
+	// Emit condition events on every return path (behind feature gate)
+	var rcx *ReconcileContext
+	defer func() {
+		if !features.FeatureGate.Enabled(features.InstanceConditionEvents) {
+			return
+		}
+		obj := inst
```
a-hilaly: should we wrap the full defer function?

pkg/controller/instance/controller.go
```diff
@@ -153,6 +160,22 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 		return err
 	}
 
+	initialConditions := conditionsFromInstance(inst)
```
a-hilaly: how about instanciating initial conditions as nil, and only call `conditionsFromInstance` before the defer if the feature gate is enabled

pkg/controller/instance/events.go
```diff
@@ -0,0 +1,85 @@
...
+	return corev1.EventTypeWarning
+}
+
+// conditionsFromInstance extracts status conditions from an unstructured object.
+//
+// TODO(perf): This performs a JSON marshal/unmarshal round-trip via
+// unstructuredWrapper.GetConditions(). For better performance we could
+// keep raw []v1alpha1.Condition at the resource level to avoid the
+// conversion on every reconciliation.
+func conditionsFromInstance(inst *unstructured.Unstructured) []v1alpha1.Condition {
```
a-hilaly, line 74: can we have a follow up PR to address this? will very likely show up in scale tests :)

pkg/controller/instance/controller.go
spattk: done

pkg/controller/instance/controller.go
spattk: yeah, that can be cleaner. done

pkg/controller/instance/events.go
spattk: Yeah, for sure, I had some brainstorming on the followup PR. Coming up soon.

NicholasBlaskey (approved):

PR #1155, opened 2026-03-16
chore: implement annotation change predicate to allow reconciles on a…

shabbskagalwala:
…nnotation changes on the rgd controller

This PR is based on the issue comment here https://github.com/kubernetes-sigs/kro/issues/1031#issuecomment-3960682798

Added unit tests to the existing tests to test behavior

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller.go
+++ b/pkg/controller/resourcegraphdefinition/controller.go
@@ -172,15 +172,16 @@ func resourceGraphDefinitionPrimaryWatchPredicate() predicate.Predicate {
 	}
 }
 
-// annotationChangedPredicate a implements an update predicate function only on the
-// the allow-breaking-changes annotation, while suppressing delete events.
-// The default `predicate.AnnotationChangedPredicate` will not skip delete events
-// using it along with predicate.Or will not suppress delete events
+// annotationChangedPredicate triggers a reconcile only when the
+// allow-breaking-changes annotation is changed between "true" and "false"
+// Adding the annotation as "false" from no annotation or vice versa, is the same
+// and we do not trigger a reconcile.
+// Delete events are suppressed to avoid redundant reconciles
 func annotationChangedPredicate() predicate.Predicate {
 	return predicate.TypedFuncs[client.Object]{
 		UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
-			oldVal := e.ObjectOld.GetAnnotations()[v1alpha1.AllowBreakingChangesAnnotation]
-			newVal := e.ObjectNew.GetAnnotations()[v1alpha1.AllowBreakingChangesAnnotation]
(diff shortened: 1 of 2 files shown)


a-hilaly (approved):
👍

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -164,6 +172,19 @@ func resourceGraphDefinitionPrimaryWatchPredicate() predicate.Predicate {
...
+func annotationChangedPredicate() predicate.Predicate {
+	return predicate.TypedAnnotationChangedPredicate[client.Object]{
+		TypedFuncs: predicate.TypedFuncs[client.Object]{
+			DeleteFunc: func(event.TypedDeleteEvent[client.Object]) bool {
+				return false
+			},
+		},
+	}
+}
+
```
michaelhtm: nit: should we change the update predicate to reconcile only on changes to annotations we manage? (eg. AllowBreakingChanges)

Not a blocking comment, but I was wondering if we can avoid unnecessary reconciles

shabbskagalwala: Makes sense- else when any applyset annotations `applyset.kubernetes.io/*` are added these will trigger a reconcile. Another thing is with ArgoCD when it adds the `argocd.argoproj.io/tracking-id`, which would also trigger a reconcile.

Question is should we only reconcile on annotation `allow-breaking-changes` or all kro managed annotations including applyset annotations?

pkg/controller/resourcegraphdefinition/controller.go
```diff
@@ -164,6 +172,23 @@ func resourceGraphDefinitionPrimaryWatchPredicate() predicate.Predicate {
...
 
+// annotationChangedPredicate a implements an update predicate function only on the
+// the allow-breaking-changes annotation, while suppressing delete events.
+// The default `predicate.AnnotationChangedPredicate` will not skip delete events
+// using it along with predicate.Or will not suppress delete events
+func annotationChangedPredicate() predicate.Predicate {
+	return predicate.TypedFuncs[client.Object]{
+		UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
+			oldVal := e.ObjectOld.GetAnnotations()[v1alpha1.AllowBreakingChangesAnnotation]
+			newVal := e.ObjectNew.GetAnnotations()[v1alpha1.AllowBreakingChangesAnnotation]
```
a-hilaly: IMO, going from no annotation to `kro.run/allow-breaking-changes: "false"` and vice versa, should not trigger a reconciliation.

pkg/controller/resourcegraphdefinition/controller.go
shabbskagalwala: Updated to only reconcile on allow-breaking-changes annotation

pkg/controller/resourcegraphdefinition/controller.go
shabbskagalwala: updated :)

PR #1159, opened 2026-03-17
register watches during deletion to accelerate convergence at scale

a-hilaly:
During deletion the controller relied solely on delayed requeue to detect
when child resources were removed. At large scale this adds unnecessary                                                                                                                          
latency - each deletion waits for the requeue timer instead of reacting
to the actual delete event. Watches are only registered for resources
confirmed to exist (after GET/LIST) not for already-gone resources.

This is the first of two changes: a follow up will make the requeue
interval configurable and allow disabling it entirely when watches
provide sufficient signal.

### DIFF
--- a/pkg/controller/instance/deletion.go
+++ b/pkg/controller/instance/deletion.go
@@ -122,6 +122,10 @@ func (c *Controller) planNodesForDeletion(
 				state.SetDeleted()
 				continue
 			}
+			// Watch collection items so we get notified when they're deleted.
+			for _, item := range items {
+				requestWatch(rcx, rid, nodeMeta.GVR, item.GetName(), item.GetNamespace())
+			}
 			node.SetObserved(items)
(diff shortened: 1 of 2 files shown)


pkg/controller/instance/deletion.go
```diff
@@ -140,6 +144,8 @@ func (c *Controller) planNodesForDeletion(
 				state.SetError(err)
 				return nil, err
 			}
+			// Watch so we get notified when the resource is deleted.
+			requestWatch(rcx, rid, nodeMeta.GVR, obj.GetName(), obj.GetNamespace())
```
jakobmoellerdev, line 148: we might not want to do that because it could trigger a relist. ideally this would be an ephemeral watch without list

jakobmoellerdev:
/close 
closing for staleness, feel free to reopen if needed still

PR #1160, opened 2026-03-17
fix: show root cause error

fabianburth:
Since we currently only surface the `lastUnresolvedErr`, we frequently suppress the root cause error containing the cel expression that failed to evaluate. This makes debugging quite hard.

Instead, show the `firstUnresolvedErr` as we assume that this is a root cause.

### DIFF
--- a/pkg/controller/instance/metrics.go
+++ b/pkg/controller/instance/metrics.go
@@ -41,7 +41,7 @@ var (
 		prometheus.HistogramOpts{
 			Name:    "instance_reconcile_duration_seconds",
 			Help:    "Duration of instance reconciliation in seconds per GVR",
-			Buckets: prometheus.DefBuckets,
+			Buckets: []float64{.01, .05, .1, .25, .5, 1, 2.5, 5, 10, 15, 20, 25, 30, 45, 60, 120},
 		},
 		[]string{"gvr"},
 	)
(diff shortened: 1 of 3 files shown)


pkg/controller/instance/resources.go
```diff
@@ -165,19 +165,22 @@ func (c *Controller) processNodes(
...
+			// surface the root error
+			if firstUnresolvedErr == nil {
+				firstUnresolvedErr = err
+			}
 		}
 		resources = append(resources, resourcesToAdd...)
 	}
 
-	return resources, lastUnresolvedErr
+	return resources, firstUnresolvedErr
```
michaelhtm, line 183: Can we return a concatenation of all the errors instead? maybe sorted first to last

jakobmoellerdev: I love how this is going full circle :D I believe @a-hilaly you mentioned that you didnt want aggregated errors in a previous run right?

a-hilaly: correct, we i don't think we aggregate

jakobmoellerdev: @a-hilaly lets please revisit this. currently not ready conditions are not showing the actual blocking expression. the user experience is not good. 

on top of this: We have to revisit this error model. In a graph with 3 nodes, where we have 

```
A -> B
  -> C
```

then B,C can error out independently. we should show both errors here. this would not be happening

PR #1161, opened 2026-03-17
KREP-022: `graph` top level field with node data in Instance 

jakobmoellerdev:
- Add `graph` to CRDs generated by ResourceGraphDefinitions (RGDs) that embeds graph information from kro to the instance without interfering with user defined status.
- Populate `graph` during reconciliation to provide resource-level insight, aiding debugging, auditing, and tooling. Also provides baseline for revision migration
- Support truncation of large resource collections by using only identity selectors


docs/design/proposals/managed-resources-status/managed-resources-status.md
```diff
@@ -0,0 +1,277 @@
...
+| `nodeID` | string | The graph node identifier from the RGD `resources[].id` |
+| `apiVersion` | string | API group and version (e.g. `apps/v1`, `v1` for core) |
+| `kind` | string | Resource kind |
+| `nodeType` | string | One of: Resource, Collection, External, ExternalCollection |
+| `state` | string | Node reconciliation state: SYNCED, IN_PROGRESS, ERROR, SKIPPED, DELETING, DELETED, WAITING_FOR_READINESS |
+| `total` | integer | Total observed resource count for this node (omitted when not truncated) |
+| `truncated` | boolean | Whether the `resources` list was capped (omitted when false) |
+| `resources` | array | List of concrete resource references (may be truncated) |
+| `resources[].name` | string | Resource name |
+| `resources[].namespace` | string | Resource namespace (empty for cluster-scoped) |
```
a-hilaly: Do we want the graph revision here as well? nodes and their templates might belong to different revisions

docs/design/proposals/managed-resources-status/managed-resources-status.md
```diff
@@ -0,0 +1,277 @@
...
+```yaml
+status:
+  state: ACTIVE
+  conditions: [...]
+  managedResources:
+    - nodeID: "deployment"
+      apiVersion: "apps/v1"
+      kind: "Deployment"
+      nodeType: "Resource"
+      state: "SYNCED"
```
a-hilaly: How about `Ready` instead of `Synced`? i think we might need more statuses in here reflecting the state of readyWhen/includeWhen

docs/design/proposals/managed-resources-status/managed-resources-status.md
```diff
@@ -0,0 +1,277 @@
...
+        - name: "myapp-worker-2"
+          namespace: "default"
+    - nodeID: "external-config"
+      apiVersion: "v1"
+      kind: "ConfigMap"
+      nodeType: "External"
+      state: "SYNCED"
+      resources:
+        - name: "shared-config"
+          namespace: "default"
```
a-hilaly: Not sure about external refs being included, it's not a managed resource.

docs/design/proposals/managed-resources-status/managed-resources-status.md
```diff
@@ -0,0 +1,277 @@
...
+  state: "SYNCED"
+  total: 500
+  truncated: true
+  resources:
+    - name: "myapp-worker-0"
+      namespace: "default"
+    # ... up to maxManagedResourcesPerNode entries
+```
+
+The limit is configured via `RGDConfig.MaxManagedResourcesPerNode`, which
```
a-hilaly: Isn't this just `rgd.maxCollectionItems`?

jakobmoellerdev: this would be purely for the status to limit the collection size that is reported. I think otherwise we can have collections that spiral out of control

docs/design/proposals/managed-resources-status/managed-resources-status.md
jakobmoellerdev: agreed, but revision would be not per node but on the entire instance right? If we are in between revisions we could introduce a node state that tracks which node is on which revision, but this could be confusing

docs/design/proposals/managed-resources-status/managed-resources-status.md
jakobmoellerdev: The existing codebase uses NodeStateSynced internally (api/v1alpha1/instance_state.go). The proposal's status field can map NodeStateSynced → READY in the buildManagedResources() helper without changing the internal state machine.
 Alternatively, this KREP could propose renaming the internal constant too — but that's a larger change. We should keep the mapping approach for now and note this as a future consideration.

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+to the number and size of its managed resources. For graphs with many nodes, or
+with collection expansions, this quickly becomes an etcd size problem. The only
+safe data to store is a compact identity reference, not full resource content.
+
+## Proposal
+
+Introduce a dedicated `graph` field at the top level of every kro-generated
+Instance CRD, sitting alongside `spec` and `status`. The `graph` field is
+written exclusively by kro using its own field manager, keeping it
+operationally separate from user-controlled `status`. It contains only the
```
Skarlso: Did you mean user facing here...? Because the status is not 'controlled' by the user, right?

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+track its own view of the resource graph: which managed resources exist, their
+topological ordering, dependency relationships, and the current reconciliation
+state. Today this data is mixed into the user-facing `status` block, which
+creates two problems.
+
+**API surface pollution.** The `status` subresource is the contract between
+kro and the user. Fields such as `topologicalOrder` and
+`resources []ResourceInformation` are kro implementation details based on its graph projection that should
+not be part of that contract. Every time kro's internal bookkeeping changes,
+the user-visible API changes with it.
```
Skarlso: Is this true in the sense that the API changes...? Isn't the API constant, it's just the populated _data_ that changes? 🤔

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+The approach taken in this proposal is to emulate subresource semantics using
+[server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/)
+field ownership:
+
+- `graph` is declared as a top-level field in the CRD OpenAPI schema, alongside
+  `spec` and `status`.
+- kro applies the `graph` field using a dedicated field manager
+  (`kro-graph-manager`). Server-side apply records ownership of every field
+  under `graph` to that manager.
+- No other actor — user, admission webhook, or GitOps controller — should claim
```
Skarlso, line 70: I'm not sure on this. ArgoCD claims ownership on everything that it pushes, doesn't it?
If I do `kubectl get myapp -o yaml > my-app.yaml` and then commit that, that will include the graph field, and now ArgoCD is pushing the graph field as well, right?

Skarlso: Ah yes. Okay, so you can do `managedFieldsManagers` in argocd and ignore kro. I think we should be okay, if we just document this the right way.

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+kro and the user. Fields such as `topologicalOrder` and
+`resources []ResourceInformation` are kro implementation details based on its graph projection that should
+not be part of that contract. Every time kro's internal bookkeeping changes,
+the user-visible API changes with it.
+
+**Scalability.** If managed resource objects (or even rich per-node detail) were
+stored inside the instance status, the instance object would grow proportionally
+to the number and size of its managed resources. For graphs with many nodes, or
+with collection expansions, this quickly becomes an etcd size problem. The only
+safe data to store is a compact identity reference, not full resource content.
```
Skarlso: This isn't happening today though, right? Instance state doesn't store anything. So this is rather a consideration for this proposal's current suggested changes. Would this be better suited in the "Alternative Considerations" section? Or am I misunderstanding this? :)

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
+# Managed Resources Status via Graph Field
+
+## Problem statement
+
+When kro reconciles a ResourceGraphDefinition (RGD) instance, it needs to
+track its own view of the resource graph: which managed resources exist, their
+topological ordering, dependency relationships, and the current reconciliation
+state. Today this data is mixed into the user-facing `status` block, which
+creates two problems.
```
Skarlso: I would add here, "status block _of the RDG_.". Because this change significantly would like to shift that into the respective `Instance`. Right?

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+}
+```
+
+##### Instance reconciler changes
+
+`pkg/controller/instance/status.go`:
+
+- The existing `updateStatus` path continues to write `status` (conditions,
+  state, user-projected fields).
+- Add a new `updateGraph` function that writes only the `graph` field, using
```
Skarlso, line 271: Would this mean that we add an extra update call? Just thinking, this would now effectively `double` the number of calls done / reconcile. Thinking out loud about the implications this would be a significant one. But not very much of a concern I believe. Although with a very chatty graph, this _could_ be significant. Not sure about the math here. :)

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+  server-side apply with field manager `kro-graph-manager`. Because `graph` is
+  a distinct top-level field, this update never conflicts with user writes to
+  `status`.
+- `graph.state` mirrors `status.state` so that consumers needing kro's
+  reconciliation state do not have to cross-reference two fields.
```
Skarlso: Yeah, because of this there are now +1 calls for everything, right?

Skarlso: Actually, can the graph be written in the same PATCH as the status? 🤔

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+    # collection node — identified by label selector
+    - id: worker
+      group: apps
+      version: v1
+      kind: Deployment
+      revision: 3
+      selector:
+        matchLabels:
+          kro.run/instance-id: "e5f6g7h8-..."
+          kro.run/node-id: worker
```
Skarlso, line 456: This means that a consumer, in order to know which is which, would have to first check if there is a selector or not. You already have nodeType. How about adding that, so users can actually do a switch or infer it from that instead of having to do a `if selector != nil { I have a collection`?

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+  x-kubernetes-preserve-unknown-fields: false
+  properties:
+    state:
+      type: string
+      description: >
+        kro's view of the reconciliation state for this instance.
+        Mirrors status.state so that consumers of the graph field do not
+        need to cross-reference the status block.
+    resources:
+      type: array
```
Skarlso, line 134: I only know this because of my crd parser project. :D Only declaring this as an array will mean that however edits the array owns it FULLY. I think there are markers in open api spec that you can use for partial editing. Let me look that up...

https://kubernetes.io/docs/reference/using-api/server-side-apply/ -> search for `x-kubernetes-list-type`

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+// collection (forEach) nodes it carries a label selector instead.
+type ManagedGraphResourceRef struct {
+	// ID is the resource identifier as defined in the RGD spec.
+	ID string `json:"id"`
+	// Group is the API group of the managed resource. Empty for core group.
+	Group string `json:"group,omitempty"`
+	// Version is the API version of the managed resource.
+	Version string `json:"version"`
+	// Kind is the kind of the managed resource.
+	Kind string `json:"kind"`
```
Skarlso: Why not just `apiVersion + kind`? 🤔 Wouldn't that be more standard?

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+  `graph.resources` matches the RGD's dependency order together with the revision.
+- **Field manager tests**: Verify that a user patching `status` does not
+  overwrite `graph`, and a kro patch to `graph` does not overwrite user
+  `status` fields.
+- **E2E tests** (`test/e2e/`): Smoke-test that existing chainsaw tests still
+  pass with the new schema; no `topologicalOrder` field in `status`.
+
+## Discussion and notes
+
+- **Why SSA and not a strategic merge patch?** SSA tracks per-field ownership,
```
Skarlso: Correct me please if I'm wrong the SMP is only supported for built-in types, right? For CRDs you can only do JSON merge patch or whatever the hell it's called. :)

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,505 @@
...
+  making it impossible for another manager to silently overwrite `graph` fields.
```
Skarlso: `impossible` is an overstatement, right? With the array as it is right now, anyone writing the entire thing will own is as the per the server-side-apply rules.

docs/design/proposals/managed-resources-status.md
Skarlso: SSA populated `metadata.managedFields` right...? Does that mean that we will grow that together with these new fields and potentially increase its size significantly?

Skarlso (commented):
I have a couple comments here and there. I didn't touch on the technical side of Kro, but more of the implications on the Kubernetes side. :) On the technical side, I still lack the necessary knowledge really. :)

Skarlso:
/assign

I'll take this over. :) 🙇‍♀️

Skarlso:
I think this proposal will be a bit different/will have another angle once this lands: https://github.com/kubernetes-sigs/kro/pull/1318

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,592 @@
...
+
+**Scalability.** Whatever replaces this must not grow with the number or size of
+the managed resources — for graphs with many nodes, or with collection
+expansions, that becomes an etcd size problem. The only safe data to store is a
+compact identity reference, not full resource content.
+
+## Proposal
+
+Introduce a dedicated `graph` field at the top level of every kro-generated
+Instance CRD, sitting alongside `spec` and `status`. The `graph` field is
```
tomasaschan, line 26: IIUC the example below correctly, this shows up on the _Instance CR_, not on the CRD. (The CRD is the _definition_ of the new resource type, an `apiextensions.k8s.io/v1/CustomResourceDefinition` instance, while the CR is the `example.com/v1/MyApp` instance). Right?

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,592 @@
...
+require infrastructure beyond what Kro already uses.
+
+**GitOps.** Because `graph` is in the object body, `kubectl get -o yaml`
+displays it, so an instance checked into git has a `graph` block and the
+GitOps controller either reports permanent drift or fights Kro for the field.
+Argo CD's fix is `ignoreDifferences.managedFieldsManagers:
+[kro-graph-manager]`; Flux's is `spec.driftDetection.ignore` with
+`fromFieldPath: graph`. Both must be documented user-facing, not only here. Kro
+never treats a foreign write to `graph` as an error, so a misconfigured setup
+degrades to churn rather than a broken instance.
```
tomasaschan, line 94: This means this change is effectively breaking to anyone who uses these tools - and anyone using any other tool with a different diffing mechanism might end up being unable to use kro at all, if their tooling does not have support for ignoring diffs like this. (Not sure what the likelihood of that is in practice, but it's one data point against this approach.)

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,592 @@
...
+
+## Other solutions considered
+
+**Keep everything in `status`.** Rejected. `status` is the user-facing contract
+and should not carry Kro implementation details. Mixing the two makes the API
+harder to understand and harder to evolve independently.
+
+**Companion `GraphState` CRD per instance.** A separate namespaced CRD
+(e.g. `InstanceGraph`) owned by each instance via OwnerReference would give
+true API separation, but makes synchronization brittle.
```
tomasaschan, line 488: > but makes synchronization brittle

How is it more brittle than the current approach, given that using a dedicated field manager will require using separate API requests anyway?

docs/design/proposals/managed-resources-status.md
```diff
@@ -0,0 +1,592 @@
...
+Kubernetes printer columns use JSONPath expressions and JSONPath has no
+`length()` or `count()` function.
+The `graph` data remains fully accessible via `-o yaml` and `-o jsonpath` for
+users and tooling that need it.
+
+## Other solutions considered
+
+**Keep everything in `status`.** Rejected. `status` is the user-facing contract
+and should not carry Kro implementation details. Mixing the two makes the API
+harder to understand and harder to evolve independently.
```
tomasaschan, line 484: I would argue that _the entire generated CRD is the user-facing contract_, which makes this reasoning apply to the proposed approach too.

tomasaschan (commented):
I think the idea to store a graph is nice, but the approach is probably the wrong one, for two main reasons:

Firstly, I think the proposed workarounds for gitops tools are indicators that this is an approach that goes against the grain of the Kubernetes echosystem.

Secondly - and more importantly - I think this pollutes the user-facing contract in ways that are really undesirable. As a platform engineer, I want to be able to expose a clean and minimal API for my users to declare their intent. I also want that API to present to my users some clear and actionable information about the status of their stuff. A real example from our org is wrapping usage of Google Cloud Bigtable (including config for IAM, backups, autoscaling, app profiles for cross-region connectivity, etc); they specify <20 lines of YAML, we generate dozens of resources that would take hundreds of lines to configure. With the proposal here, every time my user does `kubectl get keyvaluestore foo -oyaml` they'll get their own 20 lines, another 20-ish lines of status, and multiple pages of graph metadata that they do not care about at all, because it's a platform concern, not a user concern.

The proposal mentions an alternative approach with a companion resource rather than storing the info on the CR directly; I think that is _a lot_ more attractive from an API design point of view, and worth exploring deeper before rejecting. (The comment about brittleness seems unsubstantiated at best, maybe unfounded...)

docs/design/proposals/managed-resources-status.md
Skarlso: I think for things like argo and flux this is common behavior. Like, there are many fields being ignored already by people who use these tools because of other generated values. 🤔 Adding one more seems not to big of a concern.

jakobmoellerdev:
> The comment about brittleness seems unsubstantiated at best, maybe unfounded...

Saying something is unfounded or unsubstantiated without having tried or tested it yourself is very problematic in my view, especially when the comment has not yet been answered.

That being said, I believe this entire proposal is probably moot as we could replace the companion resource idea with https://github.com/kubernetes-sigs/kro/pull/1302 since every resource can now be backed by a graph that is completely independently tracked. This is effectively a very similar API design surface and also addresses your concern of not having to view the entire graph structure when you interact with the actual contract the user defines. Im gonna close for that reason.

/close

vast-raymond-page:
> I think the idea to store a graph is nice, but the approach is probably the wrong one, for two main reasons:
> 
> Firstly, I think the proposed workarounds for gitops tools are indicators that this is an approach that goes against the grain of the Kubernetes echosystem.
> 
> Secondly - and more importantly - I think this pollutes the user-facing contract in ways that are really undesirable. As a platform engineer, I want to be able to expose a clean and minimal API for my users to declare their intent. I also want that API to present to my users some clear and actionable information about the status of their stuff. A real example from our org is wrapping usage of Google Cloud Bigtable (including config for IAM, backups, autoscaling, app profiles for cross-region connectivity, etc); they specify <20 lines of YAML, we generate dozens of resources that would take hundreds of lines to configure. With the proposal here, every time my user does `kubectl get keyvaluestore foo -oyaml` they'll get their own 20 lines, another 20-ish lines of status, and multiple pages of graph metadata that they do not care about at all, because it's a platform concern, not a user concern.
> 
> The proposal mentions an alternative approach with a companion resource rather than storing the info on the CR directly; I think that is _a lot_ more attractive from an API design point of view, and worth exploring deeper before rejecting. (The comment about brittleness seems unsubstantiated at best, maybe unfounded...)

I find this really interesting. I as a platform engineer want my RGI to have an inspectable inventory of everything it manages. I think of the RGI as the owner of everything it creates. So like helm records state in secrets, FluxCD stores an inventory under the status, and ArgoCD I'm sure does something similar, I think of KRO as a resource orchestrator and owner. So my expectation is KRO should expose in its status an inventory of all the resources it owns/manages.

I'd almost suggest that my view is the more kubernetes-centric and that KRO is not creatingg a lightweight CRD to view a made up resource. I think a crossplane controller or the yoke project for writing a minimal app is a better fit for fitting a CRD that explicitly doesn't expose what its managing.

I've been really struggling to get KRO to expose a useful inventory because using CEL to loop over every defined object identifier with a for_each gets unwieldly, and I would prefer if KRO exposed everything under .status.inventory as a user of the product.

PR #1162, opened 2026-03-17
docs: document cluster-scoped instance support

a-hilaly:
Document the `scope` field introduced in KREP-10 that allows generating
cluster-scoped instance CRDs. The schema docs now explain the field, its
values, and the validation requiring explicit namespaces on namespaced
child resources. Also updated the instances docs to reflect that not all
instances are namespaced.

### DIFF
--- a/website/docs/docs/concepts/15-instances.md
+++ b/website/docs/docs/concepts/15-instances.md
@@ -108,7 +108,7 @@ Resources created by kro (Deployments, Services, ConfigMaps, etc.) receive label
 | `kro.run/kro-version` | Version of kro managing the resource |
 | `kro.run/instance-id` | UID of the instance that created this resource |
 | `kro.run/instance-name` | Name of the instance |
-| `kro.run/instance-namespace` | Namespace of the instance |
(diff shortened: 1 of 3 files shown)


website/docs/docs/concepts/15-instances.md
```diff
@@ -108,7 +108,7 @@ Resources created by kro (Deployments, Services, ConfigMaps, etc.) receive label
 | `kro.run/kro-version` | Version of kro managing the resource |
 | `kro.run/instance-id` | UID of the instance that created this resource |
 | `kro.run/instance-name` | Name of the instance |
-| `kro.run/instance-namespace` | Namespace of the instance |
+| `kro.run/instance-namespace` | Namespace of the instance (empty for cluster-scoped instances) |
```
jakobmoellerdev: am I the only one who thinks we should just not set the label here?

a-hilaly: agreed we don't need to stamp it

a-hilaly: https://github.com/kubernetes-sigs/kro/pull/1175

PR #1164, opened 2026-03-18
refactor(dynamic): refactor dynamic controller to be truly atomic

jakobmoellerdev:
Replace the eager index-update design (current/previous maps, nodeID-based
identity, 4 near-duplicate lifecycle methods) with a fundamentally simpler
model where Watch() is a pure local append and Done(true) commits atomically
by diffing against the previously committed set.

Key changes:
- Watch() no longer locks the coordinator or touches shared state
- Done(false) is a true no-op (pending discarded, committed stays)
- Watches identified by content {GVR,Name,Namespace,Selector} not nodeID
- Single committed slice per instance replaces current+previous maps
- gvrRefCount enables O(1) orphan detection

builds on top of https://github.com/kubernetes-sigs/kro/pull/1146

### DIFF
--- a/pkg/dynamiccontroller/coordinator.go
+++ b/pkg/dynamiccontroller/coordinator.go
@@ -153,6 +153,7 @@ func (c *WatchCoordinator) addWatch(key instanceKey, req WatchRequest) error {
 			previous: make(map[string]*WatchRequest),
 		}
 		c.instances[key] = state
+		instanceWatchCount.WithLabelValues(key.parentGVR.String()).Inc()
 	}
 
 	// Fix reverse index orphaning on nodeID reuse: if an existing entry
@@ -178,7 +179,6 @@ func (c *WatchCoordinator) addWatch(key instanceKey, req WatchRequest) error {
 		}
 	}
 
-	c.updateWatchRequestMetrics()
 	gvr := req.GVR
 	c.mu.Unlock()
 
@@ -213,7 +213,6 @@ func (c *WatchCoordinator) abortInstance(key instanceKey) {
 
 	state.current = make(map[string]*WatchRequest)
 
-	c.updateWatchRequestMetrics()
 	orphanedGVRs := c.findOrphanedGVRsLocked(affectedGVRs)
 	c.mu.Unlock()
 
@@ -246,7 +245,6 @@ func (c *WatchCoordinator) doneInstance(key instanceKey) {
 	state.previous = state.current
 	state.current = make(map[string]*WatchRequest)
 
-	c.updateWatchRequestMetrics()
 	orphanedGVRs := c.findOrphanedGVRsLocked(affectedGVRs)
 	c.mu.Unlock()
 
@@ -279,8 +277,8 @@ func (c *WatchCoordinator) RemoveInstance(parentGVR schema.GroupVersionResource,
 	}
 
 	delete(c.instances, key)
+	c.decInstanceWatchCount(parentGVR)
 
-	c.updateWatchRequestMetrics()
 	orphanedGVRs := c.findOrphanedGVRsLocked(affectedGVRs)
 	c.mu.Unlock()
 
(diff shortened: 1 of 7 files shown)


pkg/dynamiccontroller/coordinator.go
```diff
@@ -102,151 +109,115 @@ type collectionEntry struct {
...
-	c.mu.Lock()
-
-	state, ok := c.instances[key]
-	if !ok {
-		c.mu.Unlock()
-		return
+	// Add routes that are new.
+	ensureGVRs := make(map[schema.GroupVersionResource]struct{})
+	for id, req := range newSet {
+		if _, ok := oldSet[id]; !ok {
```
a-hilaly, line 195: This drops the old retry path for transient watch startup failures. If `EnsureWatch` times out once, the request is still recorded in `committed` and the next reconcile with the same identity skips this block entirely because `oldSet == newSet`

On the main branch every `Watch()` retried `EnsureWatch` so the watch could recover on the next reconcile

how about ensuring after building the new set

```go
ensureGVRs := make(map[schema.GroupVersionResource]struct{}, len(newSet))
for _, req := range newSet {
    ensureGVRs[req.GVR] = struct{}{}
}

c.mu.Unlock()

for gvr := range ensureGVRs {
    if err := c.watches.EnsureWatch(gvr, "coordinator"); err != nil {
        c.log.Error(err, "Failed to ensure watch", "gvr", gvr)
    }
}
```

pkg/dynamiccontroller/coordinator.go
```diff
@@ -511,46 +465,32 @@ func (NoopInstanceWatcher) Watch(_ WatchRequest) error { return nil }
...
 
-// Watch registers a watch request for this instance.
+// Watch appends a watch request to the local pending set.
+// No locks are acquired and the call never fails.
 func (w *instanceWatcher) Watch(req WatchRequest) error {
-	return w.coordinator.addWatch(instanceKey{
-		parentGVR: w.parentGVR,
-		instance:  w.instance,
-	}, req)
+	w.pending = append(w.pending, req)
```
a-hilaly, line 483: This changes watch registration from immediate to deferred-until-Done(true). The instance controller calls `Done(..)` in a defer at the end of reconcile, so child/external events that happen between `Watch()` and reconcile return can be missed. The previous implementation installed the route from `Watch()` specifically to avoid that event gap?

a-hilaly: reframing:

This defers route installation until `Done(true)`. For GVRs that are already being watched by a shared informer, that creates an event gap: if a child/external event lands after `Watch()` but before deferred `Done(true)` the informer observes it, but this instance has no route yet, so `queue.Add` is never called and the follow up reconcile can be lost

pkg/dynamiccontroller/coordinator.go
```diff
@@ -102,151 +109,115 @@ type collectionEntry struct {
...
-	state.current = make(map[string]*WatchRequest)
-
-	orphanedGVRs := c.findOrphanedGVRsLocked(affectedGVRs)
+	orphanedGVRs := c.findOrphanedGVRsLocked(removedGVRs)
 	c.mu.Unlock()
 
+	// Ensure informers for new GVRs (outside lock).
+	for gvr := range ensureGVRs {
+		if err := c.watches.EnsureWatch(gvr, "coordinator"); err != nil {
+			c.log.Error(err, "Failed to ensure watch", "gvr", gvr)
```
NicholasBlaskey, line 218: what is the consequence here of a failure?

Do we retry this?

pkg/dynamiccontroller/watch_manager.go
```diff
@@ -31,20 +31,17 @@ import (
...
-	// TODO: If watch ownership gets more complex, replace the split
-	// parent-retention + coordinator-index model with a unified owner set in
-	// WatchManager so informer lifetime is derived from one source of truth.
-	parentRefs map[schema.GroupVersionResource]int
-	client     metadata.Interface
-	resync     time.Duration
-	log        logr.Logger
+	// owners tracks who retains each GVR. An informer is eligible for
+	// shutdown only when its owner set is empty.
+	owners map[schema.GroupVersionResource]map[string]struct{}
```
NicholasBlaskey, line 41: does this address https://github.com/kubernetes-sigs/kro/issues/1169? 

we could defiantly solve that in other ways with avoiding sending shutdown signals but feels somewhat adjacent to this

PR #1165, opened 2026-03-18
refactor(runtime,instance): split large files and simplify state management

jakobmoellerdev:
Reduce cognitive load in the instance controller (~7K LOC) and node runtime (~3.1K LOC) through six incremental refactors:

1. Split resources.go by node type — external ref/collection processing moved to resources_external.go, collection processing moved to resources_collection.go. Zero behavioral change.

2. Split node.go by concern — context building to node_context.go, readiness checks to node_readiness.go, resolution logic to node_resolve.go. Zero behavioral change.

3. Consolidate expression caching — add EvalCached() on expressionEvaluationState, making evalBoolExpr/evalListExpr thin type-assertion wrappers and eliminating duplicated cache-check-and-set across 4 locations.

4. Make resolution mode explicit — add resolveMode enum and unify GetDesired/GetDesiredIdentity into a shared resolve(mode) method that encodes which vars to resolve, whether to check dep readiness, and whether to cache results.

5. Return NodeState as values — process*Node methods return ([]Resource, NodeState, error) instead of mutating shared *NodeState pointers. processNode becomes the single state registration point via SetNodeState.

6. Extract reconcileNodes pipeline — replace 90-line method with reconcileResult struct flowing through buildApplyInputs, applyResources, pruneIfSafe, and finalizeState phases.

All changes are internal (no public API changes).

### DIFF
--- a/pkg/controller/instance/resources.go
+++ b/pkg/controller/instance/resources.go
@@ -22,7 +22,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
-	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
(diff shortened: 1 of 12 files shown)


pkg/runtime/node_test.go
a-hilaly, line 1: non-blocking: ideally the tests need to move along with the new changes

pkg/runtime/state.go
```diff
@@ -45,3 +45,19 @@ type expressionEvaluationState struct {
 	// ResolvedValue holds the cached result. Nil until Resolved=true.
 	ResolvedValue any
 }
+
+// EvalCached evaluates the expression with caching. If the expression was
+// already evaluated, the cached value is returned. Otherwise the expression
+// is evaluated against the filtered context, and the result is cached.
+func (e *expressionEvaluationState) EvalCached(ctx map[string]any) (any, error) {
```
a-hilaly, line 52: I believe before this refactor `evalBoolExpr` / `evalListExpr` did not mark a `null` result as resolved, so the expression could be reevaluated later if the node’s observed state changed within the same reconcile. With `EvalCached()`, `nil` is now cached as the final value

I agree this only affects expressions that legitimately evaluate to `null` (not the more common missing-field / pending case), so the impact may be narrow

jakobmoellerdev: I will think about this some more. I think that we might want to explicitly opt out of the cache for certain expressions then, if we know that the expression uses input from the cluster. WDYT?

a-hilaly: ~~yeah, i'm happy with nuking the cache for now, the scale tests showed very low eval duration so i'm not concerned about it~~

pkg/controller/instance/resources.go
```diff
@@ -262,55 +286,47 @@ func (c *Controller) processNode(
...
 
 // processRegularNode builds applyset inputs for a single-resource node.
+// Returns the resources to apply and the resulting node state.
 func (c *Controller) processRegularNode(
 	rcx *ReconcileContext,
 	node *runtime.Node,
-	state *NodeState,
 	desiredList []*unstructured.Unstructured,
-) ([]applyset.Resource, error) {
+) ([]applyset.Resource, NodeState, error) {
```
NicholasBlaskey, line 327: much prefer this returning instead of relying on each path to know they need to set the state

PR #1166, opened 2026-03-18
compile constant CEL expressions at graph eval time

NicholasBlaskey:
Compile constant CEL expressions per RGD instead of per instance

This causes two main advantages 
1. we can catch expression eval errors in constant expressions sooner
2. we only do the work of eval constant expressions per rgd instead of per instance

Test with an rgd with constant expression of `${string(5/0)}`
```
Status:
  Conditions:
    Last Transition Time:  2026-03-18T19:35:45Z
    Message:               condition "KindReady" is awaiting reconciliation
    Observed Generation:   1
    Reason:                AwaitingReconciliation
    Status:                Unknown
    Type:                  KindReady
    Last Transition Time:  2026-03-18T19:35:45Z
    Message:               condition "ControllerReady" is awaiting reconciliation
    Observed Generation:   1
    Reason:                AwaitingReconciliation
    Status:                Unknown
    Type:                  ControllerReady
    Last Transition Time:  2026-03-18T19:35:46Z
    Message:               failed to validate resource "deployment": constant at "spec.template.metadata.labels.app": division by zero
    Observed Generation:   1
    Reason:                InvalidResourceGraph
    Status:                False
    Type:                  ResourceGraphAccepted
    Last Transition Time:  2026-03-18T19:35:46Z
    Message:               failed to validate resource "deployment": constant at "spec.template.metadata.labels.app": division by zero
    Observed Generation:   1
    Reason:                InvalidResourceGraph
    Status:                False
    Type:                  Ready
  State:                   Inactive
Events:                    <none>

```

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -46,6 +46,7 @@ import (
 	schemaresolver "github.com/kubernetes-sigs/kro/pkg/graph/schema/resolver"
 	"github.com/kubernetes-sigs/kro/pkg/graph/variable"
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
+	runtimeresolver "github.com/kubernetes-sigs/kro/pkg/runtime/resolver"
 	"github.com/kubernetes-sigs/kro/pkg/simpleschema"
 )
 
@@ -615,7 +616,7 @@ func extractTemplateDependencies(
 
 	for _, templateVariable := range node.Variables {
 		expression := templateVariable.Expression
-		nodeDeps, iteratorRefs, err := extractDependencies(inspector, expression, iteratorNames)
+		nodeDeps, iteratorRefs, callsAtLeastOneFunction, err := extractDependencies(inspector, expression, iteratorNames)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to extract dependencies: %w", err)
 		}
@@ -628,6 +629,8 @@ func extractTemplateDependencies(
 			templateVariable.Kind = variable.ResourceVariableKindIteration
 		} else if len(nodeDeps) > 0 && templateVariable.Kind == variable.ResourceVariableKindStatic {
 			templateVariable.Kind = variable.ResourceVariableKindDynamic
+		} else if len(expression.References) == 0 && !callsAtLeastOneFunction {
+			templateVariable.Kind = variable.ResourceVariableKindConstant
 		}
 
 		// Dependencies are tracked in Expression.References
@@ -670,7 +673,7 @@ func extractForEachDependencies(
 	for _, iter := range node.ForEach {
 		// Only pass iteratorNames - we want to detect iterator cross-references.
 		// schema references in forEach are valid (e.g schema.spec.regions).
-		nodeDeps, iteratorRefs, err := extractDependencies(inspector, iter.Expression, iteratorNames)
+		nodeDeps, iteratorRefs, _, err := extractDependencies(inspector, iter.Expression, iteratorNames)
 		if err != nil {
 			return nil, fmt.Errorf("failed to extract dependencies from forEach iterator %q: %w", iter.Name, err)
 		}
@@ -708,7 +711,7 @@ func buildInstanceNode(
 		statusVariable.Path = path
 
 		// Extract dependencies from the expression
-		deps, _, err := extractDependencies(inspector, statusVariable.Expression, nil)
+		deps, _, _, err := extractDependencies(inspector, statusVariable.Expression, nil)
 		if err != nil {
 			return nil, fmt.Errorf("failed to extract dependencies from expression %q: %w", statusVariable.Expression, err)
 		}
@@ -893,15 +896,16 @@ func inspectExpressionRestricted(inspector *ast.Inspector, expr string, allowedI
 func extractDependencies(inspector *ast.Inspector, expr *krocel.Expression, iteratorVars []string) (
 	resourceDeps []string,
 	iteratorRefs []string,
+	callsAtLeastOneFunction bool,
 	err error,
 ) {
 	inspectionResult, err := inspector.Inspect(expr.Original)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to inspect expression: %w", err)
+		return nil, nil, false, fmt.Errorf("failed to inspect expression: %w", err)
 	}
 
 	if !features.FeatureGate.Enabled(features.CELOmitFunction) && inspectionResult.UsesOmit() {
-		return nil, nil, fmt.Errorf("omit() requires the CELOmitFunction feature gate to be enabled")
+		return nil, nil, false, fmt.Errorf("omit() requires the CELOmitFunction feature gate to be enabled")
 	}
 
 	// Populate expression references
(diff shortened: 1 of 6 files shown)


pkg/graph/variable/variable.go
```diff
@@ -113,6 +113,19 @@ const (
...
+	// The evaluated value replaces the expression in the template, so runtime
+	// never sees them.
+	//
+	// For example:
+	//   data:
+	//     computed: ${1 + 2}  <- constant, becomes "3" in template
+	//     encoded: ${base64.encode("hello")}  <- constant function call
+	//
+	// We rely on an invariant that all kro CEL functions must be pure.
+	// This allows constant folding of function calls with literal arguments.
```
michaelhtm: should we explicitly define the CEL functions we consider are invariant? This way, if in the future we decide to add support for functions like `time.Now()`, we don't accidentally miss this?

NicholasBlaskey: Yeah I thought about an explicit list here

I think one assumption I have is just adding time.Now would break a ton of existing assumptions in the system. I could be wrong but I think a lot else breaks before this breaks if we allow non deterministic cel functions

jakobmoellerdev:
IMO The assumption that expressions are constant is moot. an expression alone is never constant as it must be reevaluated. It is impossible to know if an expression is constant value or not without knowing its input and function set behavior.

Thinking about any expression ${myfunc}, you must know that myfunc never returns a different value when called at a different time.

Thus I don't think this is acceptable.

Happy to hear thoughts or what I am missing

a-hilaly:
@jakobmoellerdev an expression is concidered constant if `else if len(expression.References) == 0 {` (inferred by the builder), hence no inputs hence always the same output (unless if we introduce things like `time.now()` as @michaelhtm stated)

Now i'm thinking we should exclude functions that call random.X

jakobmoellerdev:
> @jakobmoellerdev an expression is concidered constant if `else if len(expression.References) == 0 {` (inferred by the builder), hence no inputs hence always the same output (unless if we introduce things like `time.now()` as @michaelhtm stated)
> 
> Now i'm thinking we should exclude functions that call random.X

It is not guaranteed that a function will return the same result at any given time. To differentiate constant / non-constant we would have to categorize every single function we offer. random / time.Now are just samples for me but dont solve the bigger issue: CEL does not expose this characteristic of a function

NicholasBlaskey:
>  It is impossible to know if an expression is constant value or not without knowing its input and function set behavior.

My understanding is Kro is depending on this assumption that function set is deterministic now. A lot of the other parts of the system appear to break if cel functions are non deterministic. Suppose we do allow ```time.Now``` or another non deterministic function. That environment variable would get updated only whenever kro is reconciled. That may be often or it may be hours between that variable getting updated. It would be very hard to reason about if not an outright bug.

I would argue that caching the result of an expression per input would be easier to reason about than allowing the same input to change the graph.

> Now i'm thinking we should exclude functions that call random.X

Our cel random library is deterministic. I do agree broadly I don't want to be in the business of picking certain cel functions to cache over others.

Another option I'm not opposed to would be to cache constant cel expressions only if they have no functions.

jakobmoellerdev:
> > It is impossible to know if an expression is constant value or not without knowing its input and function set behavior.
> 
> My understanding is Kro is depending on this assumption that function set is deterministic now. A lot of the other parts of the system appear to break if cel functions are non deterministic. Suppose we do allow `time.Now` or another non deterministic function. That environment variable would get updated only whenever kro is reconciled. That may be often or it may be hours between that variable getting updated. It would be very hard to reason about if not an outright bug.
> 
> I would argue that caching the result of an expression per input would be easier to reason about than allowing the same input to change the graph.
> 
> > Now i'm thinking we should exclude functions that call random.X
> 
> Our cel random library is deterministic. I do agree broadly I don't want to be in the business of picking certain cel functions to cache over others.
> 
> Another option I'm not opposed to would be to cache constant cel expressions only if they have no functions.

I think we have never explicitly established deterministic functions, and in general, our entire resource evaluation system is non-deterministic, so im unsure what you mean with a lot of the other parts of the system break. I think the fact that we only recompute once reconciliation happens is expected no?

NicholasBlaskey:
> I think we have never explicitly established deterministic functions, and in general, our entire resource evaluation system is non-deterministic, so im unsure what you mean with a lot of the other parts of the system break. I think the fact that we only recompute once reconciliation happens is expected no?

Would this not be a massive footgun? It would be certainly incorrect if we allowed `time.Now` and someone tried to use it as part of their `includeWhen` to have a certain resource exist during working hours. The unexpected part is when this reconciliation time happens. Kro is no longer reacting to the state on the cluster but changing due to unrelated reconciles and arbitrary time intervals. 

I'm open to excluding caching constant cel expressions if they have functions if we want to leave open non deterministic functions possibility the line.

NicholasBlaskey:
> I'm open to excluding caching constant cel expressions if they have functions if we want to leave open non deterministic functions possibility the line.

@jakobmoellerdev updated the PR to do this

Could potentially allow all standard library functions, but for now left it at caching no functions for now

pkg/graph/builder.go
```diff
@@ -935,14 +940,27 @@ func extractDependencies(inspector *ast.Inspector, expr *krocel.Expression, iter
...
-		return nil, nil, fmt.Errorf("uses unknown functions: %v", inspectionResult.UnknownFunctions)
+		return nil, nil, false, fmt.Errorf("uses unknown functions: %v", inspectionResult.UnknownFunctions)
+	}
+
+	// Check if expression calls any non-operator functions.
+	// CEL represents operators (like +, -, /, etc.) as function calls,
+	// but they're pure and safe to evaluate at build time.
+	// Use CEL's official operators package to distinguish operators from functions.
+	hasNonOperatorCall := false
+	for _, fc := range inspectionResult.FunctionCalls {
```
jakobmoellerdev: I believe this could probably be part of the inspection TBH instead of being checked here. WDYT?

pkg/graph/variable/variable.go
```diff
@@ -113,6 +113,11 @@ const (
 	//
 	// Expressions not referencing iterator variables remain static or dynamic.
 	ResourceVariableKindIteration ResourceVariableKind = "iteration"
+	// ResourceVariableKindConstant represents a constant expression.
+	// These have no variable references and no function calls (operators like
+	// +, -, *, / are allowed). They are evaluated once at build time, and the
+	// evaluated value replaces the expression in the template.
+	ResourceVariableKindConstant ResourceVariableKind = "constant"
```
jakobmoellerdev, line 120: I think the word constant is a bit misleading. We already have VariableKind Static. I think both of these now need a revisit.

Another question: What about iterations with pure operators. Can they ever be constant?

Also dont we now have the ability to calculate certain includeWhen and readyWhen conditions containing pure operators ahead of time so we dont have to do it at runtime as well? Since these expressions never change, they could appear anywhere

jakobmoellerdev (commented):
I think this is going into a good direction but the variable kind mapping strikes me as off.

pkg/graph/builder.go
NicholasBlaskey: moved logic to inspection package

pkg/graph/variable/variable.go
NicholasBlaskey: >I think the word constant is a bit misleading. We already have VariableKind Static. I think both of these now need a revisit.

Do you think this is a naming problem here? Or do you think this is a conceptual problem too?  

For potential names something like `BuildTimeConstant` / `RuntimeStatic`

> Another question: What about iterations with pure operators. Can they ever be constant?

Yes we could. I do think this would be a larger change since we could then apply further optimizations to the iterator variables if we knew they were constant. We could apply something akin to loop unrolling.

>Also dont we now have the ability to calculate certain includeWhen and readyWhen conditions containing pure operators ahead of time so we dont have to do it at runtime as well?

Yes we could. 

We could also potential optimize out `includeWhen` conditions too if they are constant. Either removing the resource or the consideration of the resource depending on if we should include or exclude the resource. 

Could apply the same idea to `readyWhen`

jakobmoellerdev:

/close

PR #1167, opened 2026-03-18
feat(graph): reject omit() on required resource metadata fields

michaelhtm:
Reject omit() on metadata.name (always) and metadata.namespace (when
the instance CRD is cluster-scoped and the resource is namespaced).

Adds validateNoOmitOnRequiredFields as a dedicated validation pass
that runs after dependency extraction, where node metadata and the
instance scope are available.

### DIFF
--- a/pkg/graph/validation.go
+++ b/pkg/graph/validation.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
+	"github.com/kubernetes-sigs/kro/pkg/cel/ast"
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/graph/validation.go
```diff
@@ -301,6 +302,37 @@ func validateTemplateConstraints(
...
 
+// validateNoOmitOnRequiredFields checks that omit() is not used on fields that
+// must resolve to a concrete value. Currently enforced fields:
+//   - metadata.name: always required for any Kubernetes resource
+//   - metadata.namespace: required when the instance is cluster-scoped and the
+//     resource itself is namespaced (no instance namespace to inherit)
+func validateNoOmitOnRequiredFields(nodes map[string]*Node, inspector *ast.Inspector, isInstanceNamespaced bool) error {
+	for _, node := range nodes {
+		for _, v := range node.Variables {
+			switch v.Path {
```
NicholasBlaskey: question: should we check if the field definition in the schema of the object itself is required?

Is this already an issue? Can I define a resource and miss fields and then have it fail to apply?

michaelhtm: The reason we may not want to do this are in scenarios where kro might want to create a resource, and later omit the required field, as it would be managed by another controller (eg. Horizontal Autoscaler with deployment.replicas)

a-hilaly: qq: do we allow resources to reference them selfs in kro?

michaelhtm: I don't think we do, except for ReadyWhen
https://github.com/kubernetes-sigs/kro/blob/main/pkg/graph/dag/dag.go#L152-L155

PR #1168, opened 2026-03-18
test: integration test case, omit() removes field

michaelhtm:
Verify that when an instance field transitions from present back to
empty, omit() drops the field from the SSA apply payload and the
previously-managed field is removed from the ConfigMap.


jakobmoellerdev (commented):
/lgtm

PR #1174, opened 2026-03-19
KREP-013: Graph Revisions

a-hilaly:
This proposal introduces `GraphRevision`, a cluster-scoped immutable snapshot of
a compiled RGD spec. On each accepted RGD spec change, the RGD controller issues
a new GraphRevision. A dedicated GraphRevision controller compiles the snapshot
and stores the result in a shared in-memory registry. Instance controllers
resolve their compiled graph from the registry instead of receiving it directly.
This KREP covers snapshot issuance, compilation, registry, and instance
resolution. Revision pinning, leveled topologies, and propagation control are
follow-up work that builds on this foundation.

### DIFF
--- a/docs/design/proposals/graph-revisions.md
+++ b/docs/design/proposals/graph-revisions.md
@@ -1,0 +1,730 @@
+# KREP-013: Graph Revisions
+
+## Summary
+
+This proposal introduces `GraphRevision`, a cluster-scoped immutable snapshot of
+a compiled RGD spec. On each accepted RGD spec change, the RGD controller issues
+a new GraphRevision. A dedicated GraphRevision controller compiles the snapshot
+and stores the result in a shared in-memory registry. Instance controllers
+resolve their compiled graph from the registry instead of receiving it directly.
+This KREP covers snapshot issuance, compilation, registry, and instance
+resolution. Revision pinning, leveled topologies, and propagation control are
+follow-up work that builds on this foundation.
+
+## Motivation
+
+The current model has no concept of graph history. When an RGD spec changes, the
+old compiled graph is gone. This creates three categories of problems:
+
+**No rollback or diffing.** Operators cannot see what changed between RGD edits.
+When an update breaks instances, there is no way to inspect the previous graph,
+diff it against the current one, or roll back to a known-good state. Debugging
+requires reading git history (if it exists) rather than cluster state.
+
+**No controlled rollout.** All instances immediately pick up the new graph on
+their next reconcile. There is no way to roll out changes in levels, test a
+new revision against a canary set, or hold some instances on the old graph while
+validating the new one. For platform teams managing hundreds of instances, an
+RGD change is all-or-nothing.
+
+**No foundation for propagation control.** Features like leveled applies (apply
+changes to a subset of instances, observe, then proceed - see
+[KREP-005](https://github.com/kubernetes-sigs/kro/pull/859) and
+[KREP-006](https://github.com/kubernetes-sigs/kro/pull/861)) require a stable
+revision identity that instances can be pinned to. Without immutable snapshots,
+there is nothing to pin to. The instance controller receives its graph through a
+closure - there is no addressable revision object.
+
+### Concrete example
+
+A platform team runs an RGD that manages a `WebApp` custom resource. 200
+instances exist across the cluster. The team updates the RGD to add a sidecar
+container to the deployment template.
+
+Without GraphRevisions: all 200 instances pick up the new graph on their next
+reconcile. If the sidecar has a bug, all 200 instances are affected. The team
+cannot see what the previous graph looked like, cannot roll back without
+reverting the RGD spec (and hoping they remember the exact previous state), and
+cannot test the change on a subset first.
+
+With GraphRevisions: the RGD controller creates revision 4. The team can diff
+revision 3 vs 4 to see exactly what changed. In future work, they can pin 5
+canary instances to revision 4, verify the sidecar works, then either pause
+behind a manual review gate or promote the remaining 195 in levels. If
+something goes wrong, revision 3 still exists as an immutable snapshot they can
+roll back to.
+
+## Proposal
+
+### Overview
+
+RGDs are cluster-scoped (`+kubebuilder:resource:scope=Cluster`), so
+GraphRevisions are cluster-scoped as well - they follow their parent's scope.
+
+A new cluster-scoped CRD, `GraphRevision`, stores an immutable snapshot of an
+RGD spec. A new controller reconciles GraphRevision objects by compiling them
+and storing the compiled graph in a shared in-memory registry. The RGD
+controller becomes the issuer of revisions after pre-flight validation, but it
+remains the owner of the generated CRD and the dynamic controller. The instance
+controller becomes a consumer of the registry.
+
+```
+ ┌────────────────┐
+ │ RGD Controller │
+ └───────┬────────┘
+         │ creates
+         ▼
+ ┌───────────────────┐       ┌──────────────────┐
+ │  GraphRevision    │──────▶│  GR Controller   │
+ │  (API object)     │       │  (compiles graph) │
+ └───────────────────┘       └────────┬─────────┘
+                                      │ stores
+                                      ▼
+                             ┌────────────────────┐
+                             │  In-Memory Registry │
+                             │  (compiled graphs)  │
+                             └────────┬────────────┘
+                                      │ reads
+                                      ▼
+                             ┌────────────────────┐
+                             │ Instance Controller │
+                             └─────────────────────┘
+```
+
+### Core Concepts
+
+**Immutable snapshots.** A GraphRevision's spec is immutable (enforced by a CEL
+validation rule: `self == oldSelf`). Once created, it never changes. A new RGD
+spec change produces a new GraphRevision, never an update to an existing one.
+
+**Pre-flight issuance gate.** Not every attempted RGD edit becomes a
+GraphRevision. The RGD controller first compiles the current spec. If that
+fails, the edit is surfaced as invalid on the RGD and no new GraphRevision is
+created. Revision history records accepted graph snapshots, not arbitrary failed
+edits.
+
+**Monotonic revision numbers.** Each GraphRevision has a `revision` field,
+monotonically increasing per RGD lineage. The lineage key is the source RGD
+name, so a recreated RGD with the same name continues from the existing lineage.
+The RGD persists its high-water mark in `status.lastIssuedRevision` so revision
+numbers never go backwards within a lineage, even across GC cycles.
+
+On delete+recreate, the new RGD's `status.lastIssuedRevision` starts at 0, but
+the RGD controller lists existing GraphRevisions by name (field index on
+`spec.snapshot.name`) and computes the next revision as
+`max(status.lastIssuedRevision, highestExistingRevision) + 1`. This means:
+
+- If GRs from the previous incarnation survive (GC still in progress or
+  retention keeps them), the lineage continues from the highest existing
+  revision number.
+- If all previous GRs have been garbage collected, the revision resets to 1 -
+  which is safe because there are no surviving objects to collide with.
+
+**Internal hashing for change detection.** A canonical FNV-64a hash of the
+normalized RGD spec determines whether a new revision is needed (see Hashing
+section). The hash is computed in-process and stored only in the in-memory
+registry - never persisted on the GraphRevision API object. This makes the hash
+algorithm an internal implementation detail that can change between binary
+versions without migration. A label (`kro.run/graph-revision-hash`) is written
+to the GraphRevision for kubectl discoverability but is never read back.
+Semantically identical edits - whitespace changes, YAML field reordering,
+comment differences that normalize to the same canonical JSON - do not produce
+new revisions.
+
+**Three-state registry.** The in-memory registry tracks each revision as Pending
+(created, not yet compiled), Active (compiled, graph available), or Failed
+(compilation error). These states are internal scheduling signals between
+controllers. The external GraphRevision API surfaces them through conditions
+instead of a separate state enum.
+
+### API Changes
+
+#### New CRD: GraphRevision
+
+The `internal.kro.run` API group signals that this API is subject to breaking
+changes between releases. Operators can inspect GraphRevision objects via
+`kubectl` for debugging and observability, but external automation should use
+RGD status fields (conditions, `lastIssuedRevision`) rather than scraping
+GraphRevision objects directly. The API group may move to `kro.run` if/when the
+schema stabilizes.
+
+```yaml
+apiVersion: internal.kro.run/v1alpha1
+kind: GraphRevision
+metadata:
+  name: my-webapp-r00003
+  labels:
+    kro.run/resource-graph-definition-name: my-webapp
+    kro.run/graph-revision-hash: "a1b2c3d4e5f6" # informational, not authoritative
+  ownerReferences:
+    - apiVersion: kro.run/v1alpha1
+      kind: ResourceGraphDefinition
+      name: my-webapp
+spec:
+  revision: 3
+  snapshot:
+    name: my-webapp
+    generation: 2
+    spec: # immutable copy of rgd.spec
+      schema:
+        kind: WebApp
+        apiVersion: v1alpha1
+      resources:
+        - id: deployment
+          template: ...
+status:
+  topologicalOrder: ["configmap", "deployment", "service"]
+  conditions:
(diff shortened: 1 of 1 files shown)


docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
+# KREP-013: Graph Revisions
+
+## Summary
+
+This proposal introduces `GraphRevision`, a cluster-scoped immutable snapshot of
+a compiled RGD spec. On each accepted RGD spec change, the RGD controller issues
```
NicholasBlaskey, line 6: what is the definition of accepted here? Is it defined by a status?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+diff it against the current one, or roll back to a known-good state. Debugging
+requires reading git history (if it exists) rather than cluster state.
+
+**No controlled rollout.** All instances immediately pick up the new graph on
+their next reconcile. There is no way to roll out changes in levels, test a
+new revision against a canary set, or hold some instances on the old graph while
+validating the new one. For platform teams managing hundreds of instances, an
+RGD change is all-or-nothing.
+
+**No foundation for propagation control.** Features like leveled applies (apply
```
cheeseandcereal, line 30: Isn't this paragraph basically repeating the above one about controlled rollout, but more general?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+Without GraphRevisions: all 200 instances pick up the new graph on their next
+reconcile. If the sidecar has a bug, all 200 instances are affected. The team
+cannot see what the previous graph looked like, cannot roll back without
+reverting the RGD spec (and hoping they remember the exact previous state), and
+cannot test the change on a subset first.
+
+With GraphRevisions: the RGD controller creates revision 4. The team can diff
+revision 3 vs 4 to see exactly what changed. In future work, they can pin 5
+canary instances to revision 4, verify the sidecar works, then either pause
+behind a manual review gate or promote the remaining 195 in levels. If
```
cheeseandcereal, line 53: To be clear, this is defining future possible implementation of another future krep with propogation rollout, not this one, right?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+## Proposal
+
+### Overview
+
+RGDs are cluster-scoped (`+kubebuilder:resource:scope=Cluster`), so
+GraphRevisions are cluster-scoped as well - they follow their parent's scope.
+
+A new cluster-scoped CRD, `GraphRevision`, stores an immutable snapshot of an
+RGD spec. A new controller reconciles GraphRevision objects by compiling them
+and storing the compiled graph in a shared in-memory registry. The RGD
```
spattk, line 66: is there an upper limit to the number of revisions we are storing in the memory registry?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+roll back to.
+
+## Proposal
+
+### Overview
+
+RGDs are cluster-scoped (`+kubebuilder:resource:scope=Cluster`), so
+GraphRevisions are cluster-scoped as well - they follow their parent's scope.
+
+A new cluster-scoped CRD, `GraphRevision`, stores an immutable snapshot of an
```
NicholasBlaskey, line 64: if ```GraphRevision``` are immutable did we think about if we even needed an ```GraphRevision``` controller? 

Could we have just used them as a data store?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+On delete+recreate, the new RGD's `status.lastIssuedRevision` starts at 0, but
+the RGD controller lists existing GraphRevisions by name (field index on
+`spec.snapshot.name`) and computes the next revision as
+`max(status.lastIssuedRevision, highestExistingRevision) + 1`. This means:
+
+- If GRs from the previous incarnation survive (GC still in progress or
+  retention keeps them), the lineage continues from the highest existing
+  revision number.
+- If all previous GRs have been garbage collected, the revision resets to 1 -
+  which is safe because there are no surviving objects to collide with.
```
cheeseandcereal, line 121: Will this divergence of behavior be confusing? It seems weird to me that a brand new creation could have different behavior like this, but maybe that doesn't matter. Thoughts?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
+# KREP-013: Graph Revisions
+
+## Summary
+
+This proposal introduces `GraphRevision`, a cluster-scoped immutable snapshot of
+a compiled RGD spec. On each accepted RGD spec change, the RGD controller issues
+a new GraphRevision. A dedicated GraphRevision controller compiles the snapshot
+and stores the result in a shared in-memory registry. Instance controllers
```
JamesGAWS, line 8: Why in memory and not a separate CR?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+| --------------------- | --------------------------- | ----------------------------------------------------------- |
+| `revision`            | int64                       | Monotonic revision number per RGD lineage.                  |
+| `snapshot.name`       | string                      | Source RGD name. Authoritative identity for lineage.        |
+| `snapshot.uid`        | types.UID                   | Source RGD UID at issuance time. Informational only.        |
+| `snapshot.generation` | int64                       | Source RGD generation at issuance time. Informational only. |
+| `snapshot.spec`       | ResourceGraphDefinitionSpec | Immutable copy of the source RGD spec.                      |
+
+Status conditions follow the established kro pattern (KREP status-conditions):
+`Ready` as top-level, `GraphVerified` as sub-condition. The mapping is:
+`Pending => Ready=Unknown`, `Active => Ready=True`, `Failed => Ready=False`.
```
michaelhtm: Would failed ever happen? If we only create GraphRevisions on valid RGDs

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+- GraphRevision CRD and controller
+- In-memory spec hashing for change detection (algorithm is an implementation
+  detail)
+- In-memory revision registry with Pending/Active/Failed states
+- RGD controller as revision issuer (create, GC)
+- Instance controller consuming compiled graphs from registry
+- OwnerReference-based GC on RGD deletion
+- Retention-based GC (configurable max revisions per RGD)
+- Finalizer on GraphRevision for clean registry eviction on delete
+- Startup replay: GR controller rebuilds the in-memory registry from persisted
```
cheeseandcereal, line 327: Will this add "cold-start"/scaling constraints we may be concerned about? Or do we think this will be performant enough.

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+**OwnerReference cascade.** Each GraphRevision has an OwnerReference pointing to
+its source RGD. When the RGD is deleted, Kubernetes garbage collection deletes
+all its GraphRevisions. The GR controller's finalizer ensures registry entries
+are evicted before the API object disappears. The RGD controller does not evict
+registry entries itself during cleanup - each GR's finalizer handles its own
+eviction as the cascade proceeds. On RGD deletion, the microcontroller is shut
+down before any cleanup begins to prevent data races between the instance
+controller (reading from registry) and the cascade (evicting from registry).
+
+**Retention floor.** A separate periodic goroutine runs retention GC on a
```
NicholasBlaskey: was a time based retention floor considered?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+configurable interval (default 60s). For each RGD, if the number of live
+(non-terminating) GraphRevisions exceeds `--rgd-max-graph-revisions` (default
+10), it sorts revision numbers, computes a retention floor (the
```
NicholasBlaskey: does this mean we could delete revisions currently being used?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+- GraphRevision: `GraphVerified=False`, `Ready=False`
+
+The previous Active revision remains in the registry and is not evicted. This is
+the foundation for automatic rollback (future work): the data to fall back to
+already exists, what's missing is the policy for when to do so. When automatic
+rollback lands, the behavior changes from "hard block, operator fixes" to
+"degrade to previous revision, surface the failure as a condition, operator
+decides whether to fix or roll forward."
+
+Older retained revisions that fail recompilation on restart do not block the
```
spattk, line 309: Isn't this a problem if we need to rollback to an older version. How can we be sure if it's safe to rollback if need arises in such situation?

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+### Garbage Collection
+
+Two GC mechanisms:
+
+**OwnerReference cascade.** Each GraphRevision has an OwnerReference pointing to
+its source RGD. When the RGD is deleted, Kubernetes garbage collection deletes
+all its GraphRevisions. The GR controller's finalizer ensures registry entries
+are evicted before the API object disappears. The RGD controller does not evict
+registry entries itself during cleanup - each GR's finalizer handles its own
+eviction as the cascade proceeds. On RGD deletion, the microcontroller is shut
```
cheeseandcereal, line 514: [Not an issue] Are we really canonically calling these "microcontroller"s ? Lol. I had to do a double-take on that word.

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+down before any cleanup begins to prevent data races between the instance
+controller (reading from registry) and the cascade (evicting from registry).
+
+**Retention floor.** A separate periodic goroutine runs retention GC on a
+configurable interval (default 60s). For each RGD, if the number of live
+(non-terminating) GraphRevisions exceeds `--rgd-max-graph-revisions` (default
+10), it sorts revision numbers, computes a retention floor (the
+Nth-from-latest), and deletes all revisions below that floor. The registry is
```
cheeseandcereal: Is this just being done to save resources? (I assume yes, but confirming)

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+                             └─────────────────────┘
+```
+
+### Core Concepts
+
+**Immutable snapshots.** A GraphRevision's spec is immutable (enforced by a CEL
+validation rule: `self == oldSelf`). Once created, it never changes. A new RGD
+spec change produces a new GraphRevision, never an update to an existing one.
+
+**Pre-flight issuance gate.** Not every attempted RGD edit becomes a
```
k1ranpk, line 100: Are there cases, other than failed compile, where an RGD revision is not considered as a new version?

docs/design/proposals/graph-revisions.md
a-hilaly: It's both, CR for persistence, in-memory for instances

docs/design/proposals/graph-revisions.md
a-hilaly: An update to the api server

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+
+RGDs are cluster-scoped (`+kubebuilder:resource:scope=Cluster`), so
+GraphRevisions are cluster-scoped as well - they follow their parent's scope.
+
+A new cluster-scoped CRD, `GraphRevision`, stores an immutable snapshot of an
+RGD spec. A new controller reconciles GraphRevision objects by compiling them
+and storing the compiled graph in a shared in-memory registry. The RGD
+controller becomes the issuer of revisions after pre-flight validation, but it
+remains the owner of the generated CRD and the dynamic controller. The instance
+controller becomes a consumer of the registry.
```
k1ranpk, line 69: How do the operators ensure that the instance count is pinned to a certain graph revision?

docs/design/proposals/graph-revisions.md
a-hilaly: There is an upper limit for GraphRevision CRs in general, which limits how many you have in memory.

docs/design/proposals/graph-revisions.md
a-hilaly: Seperation of concerns makes it easier to reason about these systems. RGD Controller is already complex enough.

docs/design/proposals/graph-revisions.md
a-hilaly: Correct or that could also be part of propagation rollout if we want to increase it's scope

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+effect (the new GR stays Pending, the RGD controller requeues with
+`WaitingForCompilation`). However, existing instances continue serving on their
+current Active revision - the registry retains previously compiled entries in
+memory. The blast radius is "new RGD changes don't propagate," not "existing
+instances stop working." Operators can diagnose a stuck `WaitingForCompilation`
+state by checking GR controller pod health and logs, or via the
+`kro_graph_revision_compilations_total` metric (a flatline when activity is
+expected signals a stuck controller).
+
+**Latest revision failure is a hard block.** If the latest GraphRevision fails
```
JamesGAWS, line 287: Is there a metric that RGD authors can alarm on?

docs/design/proposals/graph-revisions.md
a-hilaly: Yes just a paragraph mentioning the other kreps

docs/design/proposals/graph-revisions.md
```diff
@@ -0,0 +1,716 @@
...
+compilation, the RGD controller sets `ResourceGraphAccepted=False` with reason
+`InvalidResourceGraph` and returns an error. The instance controller sees
+`Failed` from `GetLatestRevision()` and returns a terminal error - no requeue.
+There is no automatic fallback to the previous Active revision.
+
+This is a deliberate choice for this KREP. The alternative - silently falling
```
JamesGAWS, line 293: Can we make this configurable? If users are able to scrape a metric and alarm when a broken RGD is created, then these failures are loud but don't break the service. Best of both worlds!

docs/design/proposals/graph-revisions.md
a-hilaly: That's a good question, very likely a seperate krep for instance counting and watch state reporting in general

docs/design/proposals/graph-revisions.md
a-hilaly: Yes, added to the doc

docs/design/proposals/graph-revisions.md
a-hilaly: beside network issues, there isn't a case

docs/design/proposals/graph-revisions.md
a-hilaly: correct, graphs are expensive in memory

docs/design/proposals/graph-revisions.md
a-hilaly: yes :)

docs/design/proposals/graph-revisions.md
a-hilaly: correct. this can be addressed in future kreps, but holding RGD/graph deletion when instances are still consuming them

docs/design/proposals/graph-revisions.md
a-hilaly: The major reason why a graph would fail to recompile is if the schemas have changed (and the graph is no longer valid), which is safer to consider them failing and block rollbacks

docs/design/proposals/graph-revisions.md
a-hilaly: changed, no longer a goroutine but part of the RGD reconciliation process

docs/design/proposals/graph-revisions.md
a-hilaly: scale tests showed that this is quiet fast. nothing to be concerned about

docs/design/proposals/graph-revisions.md
a-hilaly: If there are no graph revisions and no water mark there is no way for us to know from where to start. it's just mechanically correct.

docs/design/proposals/graph-revisions.md
a-hilaly: We can make it configurable in future versions

PR #1176, opened 2026-03-19
feat(resolver): add CRD informer-backed schema resolver with lazy extraction

a-hilaly:
Replace the TTL-only schema resolution with a three-layer chained resolver:
core (compiled-in) -> CRD (informer-backed) -> fallback (TTL-cached discovery)

The CRD resolver uses a custom GVK indexer on the CRD informer for O(1)
lookups and lazily extracts `OpenAPI` schemas on first access, caching them
until the CRD is updated or deleted. This gives event driven invalidation
without paying the memory cost of converting schemas for every CRD in the
cluster -- only CRDs that kro actually references are materialized.

Builder is a pure consumer that takes a `SchemaResolver` interface with no
lifecycle concerns. Resolver wiring, ordering, and lifecycle are owned by
`cmd/controller/main.go`. The CRD resolver takes a
`CustomResourceDefinitionInformer` directly rather than a `clientset`, and the
informer factory is registered with the controller manager for lifecycle
management.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -239,9 +239,8 @@ func main() {
 		os.Exit(1)
 	}
 	crdInformerFactory := apiextensionsinformers.NewSharedInformerFactory(apiextensionsClientset, 0)
-	crdResolver := schemaresolver.NewCRDSchemaResolver(
-		crdInformerFactory.Apiextensions().V1().CustomResourceDefinitions(),
-	)
+	crdInformer := crdInformerFactory.Apiextensions().V1().CustomResourceDefinitions()
+	crdResolver := schemaresolver.NewCRDSchemaResolver(crdInformer)
 	if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
 		crdInformerFactory.Start(ctx.Done())
 		crdInformerFactory.WaitForCacheSync(ctx.Done())
@@ -270,6 +269,7 @@ func main() {
 		dc,
 		resourceGraphDefinitionGraphBuilder,
 		instanceRequeueInterval,
+		crdInformer.Informer(),
 		resourceGraphDefinitionConcurrentReconciles,
 		graph.RGDConfig{
 			MaxCollectionSize:          rgdMaxCollectionSize,
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -27,10 +27,10 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
 	ctrlconfig "sigs.k8s.io/controller-runtime/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
@@ -230,9 +230,7 @@ func main() {
 		BurstLimit:      burstLimit,
 	}, set.Metadata(), set.RESTMapper())
 
-	// Schema resolution: core → CRD informer → fallback discovery.
-	coreResolver := schemaresolver.DefaultCoreResolver()
-
+	// Shared CRD informer for schema resolution and RGD controller.
 	apiextensionsClientset, err := apiextensionsclient.NewForConfigAndClient(restConfig, set.HTTPClient())
 	if err != nil {
 		setupLog.Error(err, "unable to create apiextensions clientset")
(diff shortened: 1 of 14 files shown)


cmd/controller/main.go
```diff
@@ -210,17 +216,47 @@ func main() {
...
+	// Shared CRD informer for schema resolution and RGD controller.
+	apiextensionsClientset, err := apiextensionsclient.NewForConfigAndClient(restConfig, set.HTTPClient())
 	if err != nil {
-		setupLog.Error(err, "unable to create resource graph definition graph builder")
+		setupLog.Error(err, "unable to create apiextensions clientset")
+		os.Exit(1)
+	}
+	crdInformerFactory := apiextensionsinformers.NewSharedInformerFactory(apiextensionsClientset, 0)
+	crdInformer := crdInformerFactory.Apiextensions().V1().CustomResourceDefinitions()
+	if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
```
jakobmoellerdev: this is a complete custom implementation of basically a normal controller for CRDs, I think this should be unified to controller runtime

jakobmoellerdev: Additionally I think we should restrict the informer to only watch for kro based CRDs to limit reach of the events

jakobmoellerdev (commented):
While I like the general idea of a CRD informer to give us events on the schemas, I don't really like how its implemented here in a fully custom index informer. I would much prefer the normal controller runtime informer setup that is present in the other controllers. What was the main reasoning for choosing this way here?

pkg/graph/schema/resolver/crd_resolver.go
```diff
@@ -0,0 +1,322 @@
...
+// the CRD is updated or deleted. A controller watches CRD changes and evicts
+// cached schemas so the next resolve re-extracts from the updated CRD.
+type CRDSchemaResolver struct {
+	mu      sync.RWMutex
+	schemas map[schema.GroupVersionKind]*spec.Schema
+	// crdGVKs is a reverse index: CRD name → GVKs tracked for that CRD.
+	// Used to evict schemas when a CRD is deleted (the object is already
+	// gone from the cache by the time the reconciler runs).
+	crdGVKs map[string][]schema.GroupVersionKind
+	cache   cache.Cache
```
jakobmoellerdev, line 64: i think you can just pass the controller runtime client here

pkg/graph/schema/resolver/crd_resolver.go
```diff
@@ -0,0 +1,322 @@
...
+		cache:   c,
+	}
+}
+
+// SetupWithManager registers a GVK field index on the cache and a controller
+// that watches CRD updates/deletes to evict stale cached schemas. Must be
+// called before the manager is started.
+func (r *CRDSchemaResolver) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetCache().IndexField(
+		context.Background(),
```
jakobmoellerdev, line 84: nit, pass cmd context

pkg/graph/schema/resolver/crd_resolver.go
```diff
@@ -0,0 +1,322 @@
...
+		).
+		Complete(r)
+}
+
+// Reconcile evicts cached schemas when a CRD is updated or deleted.
+func (r *CRDSchemaResolver) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	if err := r.cache.Get(ctx, client.ObjectKey{Name: req.Name}, crd); err != nil {
+		if apierrors.IsNotFound(err) {
+			// CRD deleted — evict using the reverse index.
```
jakobmoellerdev, line 117: nit: move cache functions out so you dont have confusing lock paths

pkg/graph/schema/resolver/crd_resolver.go
```diff
@@ -0,0 +1,322 @@
...
+		r.mu.RLock()
+		if s, ok := r.schemas[gvk]; ok {
+			r.mu.RUnlock()
+			return s, nil
+		}
+		r.mu.RUnlock()
+
+		// Look up the CRD from the cache's GVK field index.
+		var crdList apiextensionsv1.CustomResourceDefinitionList
+		if err := r.cache.List(context.Background(), &crdList,
```
jakobmoellerdev, line 190: nit, needs different context

pkg/graph/schema/resolver/crd_resolver.go
```diff
@@ -0,0 +1,322 @@
...
+			Kind:    crd.Spec.Names.Kind,
+		})
+	}
+	return gvks
+}
+
+// injectKubeEnvelope adds the standard Kubernetes envelope fields (metadata,
+// apiVersion, kind) to a CRD schema. CRD specs only contain spec/status —
+// the API server injects the rest when serving aggregated OpenAPI.
+func injectKubeEnvelope(s *spec.Schema, namespaced bool) {
```
jakobmoellerdev, line 271: nit, should probably move to kroschema

pkg/graph/schema/resolver/crd_resolver.go
```diff
@@ -0,0 +1,322 @@
...
+		if v.Schema == nil || v.Schema.OpenAPIV3Schema == nil {
+			return nil, nil
+		}
+		return convertCRDSchema(v.Schema.OpenAPIV3Schema)
+	}
+	return nil, nil
+}
+
+// convertCRDSchema converts a CRD JSONSchemaProps to a kube-openapi spec.Schema.
+func convertCRDSchema(jsonSchema *apiextensionsv1.JSONSchemaProps) (*spec.Schema, error) {
```
jakobmoellerdev, line 307: nit: dont we already have a schema converter for this somewhere?

jakobmoellerdev (commented):
looks much better already!

jakobmoellerdev:
/close

will be fixed by new graph engine

PR #1179, opened 2026-03-20
fix(cel): handle `oneOf` string|number schemas in `UnstructuredToVal`

a-hilaly:
`resource.Quantity` in OpenAPI v3 is represented as
`oneOf: [{type: string}, {type: number}]` with no top level type and no
`x-kubernetes-int-or-string` extension. This caused `UnstructuredToVal` to
error when evaluating CEL expressions against resources with `Quantity`
fields (e.g `ResourceQuota.spec.hard`)

Detect the `oneOf` string|number pattern and delegate to `NativeToValue`
Add `ResourceQuota` with `Quantity` schema to the fake resolver.

### DIFF
--- a/pkg/cel/unstructured/valunstructured.go
+++ b/pkg/cel/unstructured/valunstructured.go
@@ -193,11 +193,30 @@ func UnstructuredToVal(unstructured interface{}, schema common.Schema) ref.Val {
 	}
 
 	if schema.IsXPreserveUnknownFields() {
-		// Adjusted so KRO gives access to unknown fields best effort
+		// kro divergence: gives access to unknown fields best effort
+		return types.DefaultTypeAdapter.NativeToValue(unstructured)
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/cel/unstructured/valunstructured.go
```diff
@@ -193,13 +193,32 @@ func UnstructuredToVal(unstructured interface{}, schema common.Schema) ref.Val {
 	}
 
 	if schema.IsXPreserveUnknownFields() {
-		// Adjusted so KRO gives access to unknown fields best effort
+		// kro divergence: gives access to unknown fields best effort
+		return types.DefaultTypeAdapter.NativeToValue(unstructured)
+	}
+
+	// kro divergence: resource.Quantity in OpenAPI v3 has no top-level type,
```
jakobmoellerdev, line 200: good catch, this might even be worth it to push into upstream tbh

PR #1180, opened 2026-03-20
fix(externalref): allow external collections to list across all namespaces

a-hilaly:
External collection nodes use an empty namespace to signal "list across                                                                                                                           all namespaces." The existing `normalizeNamespaces` call was injecting                                                                                                                            the instance namespace into external collection resources, which broke                                                                                                                            cross-namespace listing.                                                                                                                                                                         
                                                                                                                                                                                                  
This patch adds a `NodeTypeExternalCollection` guard alongside the                                                                                                                                existing `NodeTypeInstance` check so that neither node type gets its                                                                                                                              namespace overwritten. Tests cover the full matrix of node types                                                                                                                                  (regular, instance, external collection) combined with namespace-scoped                                                                                                                           and cluster-scoped variants

### DIFF
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -129,7 +129,8 @@ type ExternalRefMetadata struct {
 	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name,omitempty"`
 	// Namespace is the namespace of the external resource.
-	// If empty, the instance's namespace will be used.
+	// For single external refs (Name), defaults to the instance's namespace.
+	// For external collections (Selector), empty means list across all namespaces.
 	//
 	// +kubebuilder:validation:Optional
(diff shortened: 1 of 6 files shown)


api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -129,7 +129,8 @@ type ExternalRefMetadata struct {
 	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name,omitempty"`
 	// Namespace is the namespace of the external resource.
-	// If empty, the instance's namespace will be used.
+	// For single external refs (Name), defaults to the instance's namespace.
+	// For external collections (Selector), empty means list across all namespaces.
```
jakobmoellerdev, line 133: Im not sure this is a good idea. I really think there should be a dedicated alias for this in the namespace like "*". Listing across all namespaces is very expensive for serious clusters.

PR #1184, opened 2026-03-20
make instance requeue interval configurable

a-hilaly:
The delayed requeue duration used when the instance reconciler waits
for cluster state to settle was hardcoded to 3s. Extract it into an
`--instance-requeue-interval` flag (default unchanged) and thread it
from main through the RGD controller into `instance.ReconcileConfig`

Setting the interval to 0 disables delayed requeues entirely -
delayedRequeue returns `requeue.None` instead of `requeue.NeededAfter`
Helm values use `config.instance.requeueInterval` rather than a flat
key to avoid a breaking rename when more instance knobs are added

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -73,6 +73,7 @@ func main() {
 		rateLimit     int
 		burstLimit    int
 		// reconciler parameters
+		instanceRequeueInterval time.Duration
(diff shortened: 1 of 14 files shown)


jakobmoellerdev (commented):
/lgtm

PR #1187, opened 2026-03-20
feat(helm): expose feature gates in chart values

a-hilaly:
Adds `config.featureGates` as a map in `values.yaml` rendered into a
single `--feature-gates flag` on the controller args. Empty by default,
so existing deployments are unaffected.


jakobmoellerdev (commented):
/lgtm

PR #1189, opened 2026-03-20
harden `BuilderCache` stores and cache iterator environments

a-hilaly:
`BuilderCache` methods used plain `Store` on miss, so concurrent callers
could overwrite an entry another goroutine just stored - producing
duplicate objects for the same key. Switch all four methods to
`LoadOrStore` so the first writer wins.

`forEach` iterator environments were built fresh via `env.Extend()` on
every build. Add `ExtendWithVariables` to `SessionCache` keyed by
`(parent, varsKey)` where `varsKey` is a canonical encoding of iterator
variable names and types. This deduplicates within a single RGD build
when multiple nodes share the same forEach signature.

### DIFF
--- a/pkg/cel/cache/builder_cache.go
+++ b/pkg/cel/cache/builder_cache.go
@@ -53,7 +53,9 @@ func (c *BuilderCache) SchemaDeclType(schema *spec.Schema, create func(*spec.Sch
 	builderCacheMissesTotal.WithLabelValues("decl_type").Inc()
 	declType := create(schema)
 	if declType != nil {
-		c.declTypes.Store(schema, declType)
+		if actual, loaded := c.declTypes.LoadOrStore(schema, declType); loaded {
+			return actual.(*apiservercel.DeclType)
+		}
 		builderCacheSize.WithLabelValues("decl_type").Inc()
 	}
 	return declType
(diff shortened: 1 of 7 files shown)


jakobmoellerdev (commented):
why did you ask me to move this into separate caches first? I already had this impl before and now were moving back to it again? https://github.com/kubernetes-sigs/kro/pull/1129#discussion_r2916347093

a-hilaly:
@jakobmoellerdev took me a few rounds of scale testing to understand the tradeoffs - i thought cross RGD program caching wasn't really useful, untill i ran into it..

This is actually part 1 of 2 changes needed to stabilize the cache, next PR incoming

a-hilaly:
@jakobmoellerdev  https://github.com/kubernetes-sigs/kro/pull/1190

jakobmoellerdev:
I dunno about that. Please check my previous commit on the PR i had before. I dont think we should cache on property level on schemas (https://github.com/kubernetes-sigs/kro/pull/1129/changes/08fdce1cc46b07138daf0c885f3dd957fc317bd1) but rather only on decl types. the current pr set makes the code hard to maintain imho

a-hilaly:
@jakobmoellerdev went back and forth on this one honestly. had some scale tests hammering similar RGDs that made the cross-RGD hits look useful, but the instance CRD schema makes the env unique per RGD so real hit rate is low

after GRs land we should probably revisit some of this caching work, the sharing boundaries will look different

PR #1190, opened 2026-03-20
add `schema.Cache` for pointer stable OpenAPI field lookups

a-hilaly:
`spec.Schema.Properties` is a value-typed map, so every index copies
the struct and `&prop` gives you a fresh pointer each time. Both
`lookupSchemaAtField` and `parser.getFieldSchema` did this, which
meant the pointer-keyed `DeclType` cache on `BuilderCache` never hit - 
every lookup recomputed and stored a duplicate entry.

`schema.Cache` returns a stable `*spec.Schema` for a given (parent,
field) pair via `sync.Map.LoadOrStore`. The builder owns it and shares
it with the parser, which is now a struct taking a `SchemaLookup`
interface instead of bare functions doing `&prop`. Pointer keys are
deliberate — new schema fetches produce new parents, so stale entries
are never hit without needing explicit eviction.

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -63,11 +63,13 @@ func NewBuilder(clientConfig *rest.Config, httpClient *http.Client) (*Builder, e
 		return nil, fmt.Errorf("failed to create dynamic REST mapper: %w", err)
 	}
 
+	schemaCache := schema.NewCache()
 	rgBuilder := &Builder{
 		schemaResolver: schemaResolver,
 		restMapper:     rm,
 		celCache:       celcache.NewBuilderCache(),
-		schemaCache:    schema.NewCache(),
+		schemaCache:    schemaCache,
+		parser:         parser.New(schemaCache),
 	}
 	return rgBuilder, nil
 }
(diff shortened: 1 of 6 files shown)


jakobmoellerdev:
Im wondering if we achieve the same by just using the type name in the key: https://github.com/kubernetes-sigs/kro/commit/4bda43f5f0f741cc6657f173de922603aedb9b96 do you think this would solve that? mainly wondering because im unsure if i wanna keep this low level field correctness in maintenance

a-hilaly:
@jakobmoellerdev Both points addressed in the second commit.

On string keys & https://github.com/kubernetes-sigs/kro/commit/4bda43f5f0f741cc6657f173de922603aedb9b96, type names use the user-defined resource ID, not the GVK so two RGDs with id "deployment" pointing at different types would silently collide. Pointer keys also invalidate naturally on schema changes without needing eviction logic.

Not saying that we shouldn't address it, i know a lot of this stuff isn't perfect, but we need to fix it first before perfectionizing it. I have many refactor ideas for stuff but i'm holding my self for now.

PR #1191, opened 2026-03-20
fix: disable instance reconciliation with annotation instead of label

michaelhtm:
In kubernetes, annotations are used to change the behaviour of the
controller, while labels are mainly meant to be used as identifiers for
the resources.

NOTE: This is not a breaking change, since the label based approach has not been
tagged/released yet.

### DIFF
--- a/pkg/dynamiccontroller/dynamic_controller.go
+++ b/pkg/dynamiccontroller/dynamic_controller.go
@@ -463,10 +463,10 @@ func (dc *DynamicController) enqueueFromInformer(parentGVR schema.GroupVersionRe
 }
 
 func reconcileEnabledInUpdate(oldMeta, newMeta metav1.Object) bool {
-	oldLbls := oldMeta.GetLabels()
-	newLbls := newMeta.GetLabels()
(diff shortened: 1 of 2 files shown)


pkg/controller/instance/controller.go
```diff
@@ -224,10 +224,10 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 	rcx.Mark.GraphResolved()
 
 	//--------------------------------------------------------------
-	// 7. Reconcile nodes (SSA + prune) and update runtime state, only if the suspend label is not present.
+	// 7. Reconcile nodes (SSA + prune) and update runtime state, only if the suspend annotation is not present.
 	//--------------------------------------------------------------
-	labels := inst.GetLabels()
-	reconcileState, ok := labels[metadata.InstanceReconcileLabel]
+	annotations := inst.GetAnnotations()
+	reconcileState, ok := annotations[metadata.InstanceReconcileAnnotation]
```
a-hilaly: can we move `metadata.InstanceReconcileAnnotation` to the api package?

PR #1200, opened 2026-03-23
Watch owned `GraphRevisions` from the RGD controller

a-hilaly:
- Set `Controller: true` on `GraphRevision` `OwnerReferences` so the `RGD` controller can watch them
- Add a watch on owned `GraphRevisions` in the `RGD` controller                                                                                                    
- Add integration tests verifying the `RGD` controller reacts to `GraphRevision` lifecycle events (create, update, delete, status changes)

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller_reconcile_test.go
+++ b/pkg/controller/resourcegraphdefinition/controller_reconcile_test.go
@@ -25,8 +25,10 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


jakobmoellerdev:
The .Owns() watch has no predicates, so every GR create/update/delete/status-change triggers an RGD reconcile. Given the RGD reconciler already has generation-skip caching and is designed for idempotent reconciles, this should be fine, but it will increase reconcile frequency for RGDs with many active revisions. 
Should we add a predicate filter (e.g. only on delete + status changes)?

a-hilaly:
>Should we add a predicate filter (e.g. only on delete + status changes)?

One could manually create GRs and that's valid, we react to those as well

PR #1201, opened 2026-03-23
fix(instance): surface graph resolution failures in conditions and metrics

a-hilaly:
The instance controller silently returned errors when graph resolution failed, leaving stale conditions from the last successful reconcile. Users had no signal on the instance about why it wasn't progressing.

Now `resolveCompiledGraph` failures mark `GraphResolved=False` with the resolution error and persist conditions via a minimal status write path that works without a full `ReconcileContext`. The instance state is set to Error so the failure is visible in kubectl.

Adds `instance_graph_resolution_total` (by gvr) and
`instance_graph_resolution_failures_total` (by gvr, reason) metrics. The reason label distinguishes not_available, pending, failed, and unknown_state cases. The metric is emitted inside resolveCompiledGraph so it fires on every failure path regardless of caller.

Removes the `ReconciliationSuspended` condition - it was a non-terminal condition set independently of the Ready rollup, and the suspend annotation still works without it (it just skips node reconciliation).

### DIFF
--- a/pkg/controller/instance/controller.go
+++ b/pkg/controller/instance/controller.go
@@ -207,6 +207,11 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 	compiledGraph, err := c.resolveCompiledGraph()
 	if err != nil {
 		log.Error(err, "failed to resolve graph revision")
+		mark := NewConditionsMarkerFor(inst)
+		mark.GraphResolutionFailed("%v", err)
+		if statusErr := c.updateConditionsStatus(ctx, inst); statusErr != nil {
+			log.Error(statusErr, "failed to update conditions status after graph resolution failure")
+		}
 		return err
 	}
 	runtimeObj, err := runtime.FromGraph(compiledGraph, inst, c.reconcileConfig.RGDConfig)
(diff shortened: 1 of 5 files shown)


pkg/controller/instance/controller.go
```diff
@@ -207,6 +207,9 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
 	compiledGraph, err := c.resolveCompiledGraph()
 	if err != nil {
 		log.Error(err, "failed to resolve graph revision")
+		mark := NewConditionsMarkerFor(inst)
+		mark.GraphResolutionFailed("%v", err)
+		_ = c.updateConditionsStatus(ctx, inst)
```
jakobmoellerdev: The status update error is silently discarded. While the reconcile still returns the original graph resolution error (triggering a requeue), if the status write fails repeatedly, users never see GraphResolved=False — maybe worth a log?

pkg/controller/instance/metrics.go
```diff
@@ -61,4 +63,20 @@ var (
 		},
 		[]string{"gvr"},
 	)
+
+	instanceGraphResolutionTotal = prometheus.NewCounterVec(
```
jakobmoellerdev: nit: The success counter instanceGraphResolutionTotal name should arguably be instance_graph_resolution_success_total to follow Prometheus naming conventions where the _total suffix
  implies a counter but doesn't distinguish success/failure. As-is, pairing _total (success) with _failures_total (failure) is slightly asymmetric. This is cosmetic.

a-hilaly:
>When a revision is Pending, it's a transient state (compilation in progress)

correct

> Counting it in failures_total could be misleading for alerting right?

correct

PR #1202, opened 2026-03-23
feat(perf): use DefaultUnstructuredConverter to eliminate JSON round-trips in conditions

spattk:
Replaced `json.Marshal`/`json.Unmarshal` round-trips in `unstructuredWrapper.GetConditions()`, `SetConditions()`, and `initialStatus()` with `runtime.DefaultUnstructuredConverter`, which converts directly between `map[string]interface{}` and typed structs without intermediate JSON byte serialization.

### Why

The JSON round-trip was executed 10-20+ times per reconciliation (every `ConditionSet` operation calls `GetConditions()`/`SetConditions()` internally). `DefaultUnstructuredConverter` avoids the allocation and parsing overhead of JSON bytes.

### Future Work

If profiling shows condition conversion remains a bottleneck, a deeper refactor could store `[]v1alpha1.Condition` directly in `ReconcileContext`, parse once at reconciliation start, and serialize once at the end — eliminating repeated conversions entirely. This would require reworking the `Object` interface used by `ConditionSet`.

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -71,6 +71,10 @@ spec:
               value: {{ .Values.config.dynamicControllerDefaultResyncPeriod | quote }}
             - name: KRO_DYNAMIC_CONTROLLER_DEFAULT_QUEUE_MAX_RETRIES
               value: {{ .Values.config.dynamicControllerDefaultQueueMaxRetries | quote }}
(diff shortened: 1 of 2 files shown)


spattk:
Got into some mess of branch conflicts. Created another PR: https://github.com/kubernetes-sigs/kro/pull/1204

PR #1203, opened 2026-03-23
Update helm chart and values.yaml

a-hilaly:


helm/values.yaml
```diff
@@ -173,7 +183,7 @@ metrics:
     create: false
     type: "ClusterIP"
     # Port for the metrics service
-    port: 8080
+    port: 8078
```
jakobmoellerdev, line 186: this is a breaking change, are you sure you wanna do this?

a-hilaly: it was wrong - https://github.com/kubernetes-sigs/kro/blob/main/helm/values.yaml#L53-L54 - https://github.com/kubernetes-sigs/kro/blob/main/helm/templates/deployment.yaml#L49-L50

PR #1210, opened 2026-03-25
fix: make applyset orphan pruning respect reverse topological order

### LINKED ISSUE #1135: Should pruning resources honor reverse topological order?
### Feature Description

**Problem Statement**:
ApplySet orphan pruning is not DAG aware today. Resources that fall out of the desired graph are pruned in a flat concurrent pass, so there is no reverse-topology guarantee during orphan deletion.

This is easiest to observe when `includeWhen` flips from `true` to `false`, but it is not specific to `includeWhen`. Any previously managed resource that becomes an orphan through applyset pruning can be deleted out of dependency order. Full instance deletion already uses a DAG-aware reverse deletion path; orphan pruning does not.

This looks like an applyset/controller pruning gap rather than a dependency-inference issue in PR #1104. The main open design questions are whether orphan pruning should be strictly reverse-topological, whether deletes can still be concurrent within the same dependency layer, and whether pruning should happen in a single pass or in waves across reconciles.

jakobmoellerdev:
ApplySet orphan pruning previously deleted all orphaned resources in a flat concurrent pass with no dependency ordering. This caused resources to be deleted before their dependents, breaking includeWhen/readyWhen conditions and stalling RGDs (e.g. when includeWhen flips from true to false, the dependency could be deleted before the dependent).

Split ApplySet.Prune() into ListOrphans() + DeleteOrphan() on the Interface so callers can impose ordering. The instance controller's pruneOrphans() now:

1. Lists orphan candidates via ListOrphans()
2. Maps each candidate to its graph node via the kro.run/node-id label
3. Groups candidates into waves by topological position
4. Deletes waves in reverse order (dependents before dependencies), with concurrent deletion within each wave

Unmapped orphans (nodes removed from the graph entirely) are deleted first since they have no known dependents in the current graph.

Fixes: https://github.com/kubernetes-sigs/kro/issues/1135

### DIFF
--- a/pkg/controller/instance/applyset/applyset.go
+++ b/pkg/controller/instance/applyset/applyset.go
@@ -60,12 +60,6 @@ type Interface interface {
 	// Batch metadata contains only GKs/namespaces from THIS apply (not parent memory).
 	Apply(ctx context.Context, resources []Resource, mode ApplyMode) (*ApplyResult, Metadata, error)
 
-	// Prune deletes orphaned resources (those with applyset label but not in KeepUIDs).
-	// Pass Project().PruneScope() to search both current batch locations AND parent memory.
-	// Prune deletes all orphans concurrently without ordering guarantees.
-	// For DAG-aware deletion, use ListOrphans + DeleteOrphan instead.
-	Prune(ctx context.Context, opts PruneOptions) (*PruneResult, error)
-
 	// ListOrphans discovers orphaned resources without deleting them.
 	// Returns candidates that the caller can order (e.g. reverse-topological)
 	// before issuing deletes via DeleteOrphan.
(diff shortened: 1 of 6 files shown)


a-hilaly:
I think this is fine as a tactical fix for now, there are def bigger battles to win wrt to deploying/deleting resources in waves and level apply things.

@jakobmoellerdev do we need a 0.9.1 for this one?

jakobmoellerdev:
I think since includeWhen was shipped with this in mind it would be a candidate for 0.9.1.
I believe we will need quite a bit of turnaround to get 0.10.0 out due to the large instance controller work we have to do

michaelhtm:
@jakobmoellerdev the changes lgtm, feel free to unhold when it's ready to merge :)

Issue #1209, opened 2026-03-25
Non-Helm manifest installation instructions should link to access control docs


coderanger:
### Documentation Issue

The current installation instructions include a non-Helm option, however while the Helm chart defaults to the "unrestricted" RBAC layout, the manifests give you an "aggregation" setup. This leads to the Quick Start and all other examples failing out of the box. While the access control options are documented under the "Advanced" section, it was not clear that this was actually a required step while using the non-Helm path.

PR #1214, opened 2026-03-27
Add missing information about required RBAC into Quick start guide.

remm:
Addresses the issue described in the issue https://github.com/kubernetes-sigs/kro/issues/1209

### DIFF
--- a/website/docs/docs/getting-started/02-deploy-a-resource-graph-definition.md
+++ b/website/docs/docs/getting-started/02-deploy-a-resource-graph-definition.md
@@ -31,6 +31,35 @@ Before you begin, make sure you have the following:
   cluster.
 - `kubectl` installed and configured to interact with your Kubernetes cluster.
 
+:::info[RBAC Configuration for an Aggregation mode]
+
+```yaml title="kro-quickstart-rbac.yaml"
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.kro.run/aggregate-to-controller: "true"
+  name: kro:controller:my-application
+rules:
+  - apiGroups: ["kro.run"]
+    resources: ["applications"]
+    verbs: ["*"]
+  - apiGroups: ["apps"]
+    resources: ["deployments"]
+    verbs: ["*"]
+  - apiGroups: [""]
+    resources: ["services"]
+    verbs: ["*"]
+  - apiGroups: ["networking.k8s.io"]
+    resources: ["ingresses"]
+    verbs: ["*"]
+```
+
+```bash
+kubectl apply -f kro-quickstart-rbac.yaml
+```
+:::
+
 ## Create your Application RGD
 
 Let's create an RGD that combines a `Deployment`, a `Service` and


website/docs/docs/getting-started/02-deploy-a-resource-graph-definition.md
```diff
@@ -31,6 +31,57 @@ Before you begin, make sure you have the following:
   cluster.
 - `kubectl` installed and configured to interact with your Kubernetes cluster.
 
+:::info[Raw manifest installation: RBAC required]
```
jakobmoellerdev: I believe we can generalize here and say this is required when aggregation mode is disabled and then also mention that for HELM it is enabled by default, whereas with manifests it is not (@a-hilaly this probably should be unified to use aggregation mode for both WDYT)

website/docs/docs/getting-started/02-deploy-a-resource-graph-definition.md
```diff
@@ -31,6 +31,57 @@ Before you begin, make sure you have the following:
...
+
+```yaml title="kro-quickstart-rbac.yaml"
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.kro.run/aggregate-to-controller: "true"
+  name: kro:controller:my-application
+rules:
+  - apiGroups:
```
jakobmoellerdev: do you mind condensing this with some smarter yaml syntax arrays `[]` ?
that way it doesnt take as much space and is an easier read

website/docs/docs/getting-started/02-deploy-a-resource-graph-definition.md
remm: Please check the updated changes.

jakobmoellerdev: I think this is still very specific to Raw manifest installation. Why dont you reframe this around `RBAC configuratio when aggregation is disabled`

remm: I hope my last change is more general. Please let me know your feedback.

remm: @jakobmoellerdev, or do you prefer a snippet only with the title `RBAC configuration when aggregation is disabled`?

jakobmoellerdev: I think I would like to make this snippet read more generic since you can read it from both installation modes (you could also install the helm chart without aggregation)

remm: It takes many iterations. I fixed it again according to your last comment. Please check it.
And thank you for the feedback.

website/docs/docs/getting-started/02-deploy-a-resource-graph-definition.md
```diff
@@ -31,6 +31,35 @@ Before you begin, make sure you have the following:
   cluster.
 - `kubectl` installed and configured to interact with your Kubernetes cluster.
 
+:::info[RBAC Configuration with Aggregation Disabled]
```
jakobmoellerdev: I just checked the docs again for a final review and believe this may be misleading after all :D https://kro.run/docs/advanced/access-control shows two modes: `unrestricted` and `aggregation`. My understanding is that the ClusterRole is only needed when the mode is `aggregation` and not when its disabled right?

remm: Yes, you are right, the issue https://github.com/kubernetes-sigs/kro/issues/1209, we want to cover is the manifests with an "aggregation" setup. Our discussion ended in an incorrect final :D
So, we need a title "RBAC Configuration for an Aggregation mode"  for the snippet.

remm: Done :)

PR #859, opened 2025-11-24
KREP-005: Level-based Topological Sorting for ResourceGraphDefinitions

bschaatsbergen:
Relates #827
Relates #822

PR #1215, opened 2026-03-27
KREP-023: Level-Aware Wavefront Instance Controller

jakobmoellerdev:
Proposal to link our various KREPs ideas for 0.10 for Instance Controller Work to work with Graph Revisions and Levelled Apply. This will be a way for KRO to do proper eventual consistent graph reconciliation

Suggests a plan to replace the sequential instance controller model with a **level-aware wavefront synchronizer**.

The core idea:

1. Group resources into **dependency levels** using topological sorting.
   Level 0 has no dependencies, level 1 depends on level 0, and so on. This is inspired by #859 
2. Apply all resources within a level **in parallel**. Wait for the level
   to finish. Then move to the next level. Inspired by #827 
3. When deleting, go in **reverse order** — remove level 2 before level 1
   before level 0. This prevents dangling references. Inspired by #1210
4. Track which **GraphRevision** each resource was last applied at, so the
   controller knows exactly what to update when the RGD changes.

Comparing to other KREPs:

- #859 (Level Based Apply): The direct predecessor of this PR as originally thought through by @bschaatsbergen and me when we first thought about pruning
- #1174 (Graph Revisions): Baseline for this proposal to succeed for N-1 to N Graph Migration. This engine allows this migration to happen seamlessly.
- #861 (propagation control): defines when a mutation may begin propagating (a new gate semantic, propagateWhen). KREP-023 defines how the reconcile pipeline is structured (level-ordered wavefront)
- #1161 (managedResource status): KREP-022 defines the status API surface — what the user sees in instance.status.managedResources. KREP-023 defines the reconcile engine that produces the data KREP-022 surfaces.
- #1091 (resource lifecycles): KREP-023 is about reconcile ordering. KREP-023's diff phase
  uses ActionAdopt/ActionOrphan concepts from KREP-014 when classifying what changed between graph revisions, but the full adoption/orphan semantics (how to take ownership of an
  existing resource, what happens during orphan on delete) are deferred to KREP-014.
- KEP-3659 (ApplySet, upstream Kubernetes): The key difference is that KEP-3659's ApplySet has no concept of ordering among managed resources (flat membership only). KREP-023 uses the annotation extension point to layer ordering metadata on top of an otherwise flat inventory.


docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+  with different behaviors. Instance deletion (`reconcileDeletion()` →
+  `planNodesForDeletion()`) is already reverse-topological: it iterates
+  nodes in forward order and keeps overwriting `deletionNode`, so it always
+  selects the deepest surviving dependent — one node per reconcile cycle.
+  The unordered path is orphan pruning: `pruneOrphans()` → `ApplySet.Prune()`
+  lists all resources labeled with the applyset ID but absent from the
+  current desired set, then deletes them in a flat concurrent pass with no
+  dependency ordering. A Service could be pruned before its Deployment, or
+  a Namespace before the resources inside it. This is a known bug tracked
+  in PR #1210.
```
NicholasBlaskey, line 43: kinda an edge case but we have seen a report about two ACK EC2 security group resources that referenced each other and blocked on each other deletion

The only way to delete these resources was to delete them in parallel

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+  Nodes with `Included == false` are excluded from the projected DAG entirely.
+- `forEach` expressions are evaluated to produce the expansion set. Each
+  combination of bindings produces a distinct `NodeID`.
+- After projection, Kahn's algorithm ([KREP-003]) groups included nodes into
+  levels.
+
+The projected DAG can change while reconciling. When a node becomes Ready,
+its status fields become available, which may flip another node's
+`includeWhen` from false to true. The controller handles this by re-running
+projection up to 5 times per reconcile. Cycle detection on
```
NicholasBlaskey, line 227: is 5 times here arbitrary?

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+    ActionDelete                      // In cluster, not in projected DAG
+    ActionOrphan                      // KREP-014: remove labels, keep resource
+
+    // Gating states
+    ActionBlocked                     // Dependencies not ready
+    ActionGated                       // Dependencies ready, propagateWhen false (KREP-006)
+)
+```
+
+Node classification logic:
```
NicholasBlaskey, line 271: This logic classification seems valuable and useful 

from this resource lifecycle configuration feels pretty clear

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+  then reads from the updated `n.observed` on its dependencies via
+  `buildContext()`. Context refresh is implicit — there is no separate
+  synchronization step.
+- **Concurrency bound.** A semaphore limits parallel SSA applies to
+  `maxConcurrency` (default: 10).
+
+**Reverse wavefront (delete/prune)**
+
+When resources need to be removed (instance deletion or nodes removed by a
+new GraphRevision), levels are processed in descending order (2, 1, 0):
```
NicholasBlaskey, line 369: From this frame I wonder if we can avoid evaluating the graph on deletion.

All we would need to do is to delete resources labeled to that instance going level by level

This avoids a resource with a runtime cel expression error causing the instance to get stuck in deleting 
```
kind: ...
name: "${string(1/0)}"
```

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+
+```
+DAG:  ConfigMap ──► Deployment-A ──► Service-A
+      Secret    ──► Deployment-B ──► Service-B
+
+Deployment-A fails (SSA conflict). Deployment-B is Ready.
+
+  Level 1 terminal states: [Error, Ready] - level complete, advance.
+
+  Level 2 dependency check:
```
NicholasBlaskey, line 445: This stands out as a an architecture smell. Its almost as if we need to bypass our level data structure to be able to infer the real graph order.

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+    ActionGated                       // Dependencies ready, propagateWhen false (KREP-006)
+)
+```
+
+Node classification logic:
+
+```
+  Resource not in cluster?              --> ActionCreate
+  kro.run/revision < target revision?   --> ActionUpdate (force)
+  Template differs from live?           --> ActionUpdate
```
NicholasBlaskey, line 276: how do we do this difference? I do think this is a valuable thing to address for performance and for allowing things like ```recreateOnUpdate``` eventually 

I do think this is a valuable problem to solve but I don't there is a great answer due to mutating webhooks

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+## Other solutions considered
+
+**Per-node dependency graph walker instead of levels**
+
+Instead of grouping nodes into levels and processing level-by-level, the
+wavefront could track each node's dependencies individually. A node would
+start as soon as all its specific dependencies are Ready, regardless of level.
+
+Not chosen: level-based execution is simpler to reason about, debug, and
+observe. "Level 1 is done, level 2 is starting" is easier to follow than
```
NicholasBlaskey, line 744: I'm not fully sold on the proposal being simpler to reason about than a true parallel graph traversal. I think there is some potential for confusion around unrelated parts of the graph impacting each other.

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+represent the mixed state. The controller wouldn't know which nodes to retry
+on the next reconcile. Per-node labels give exact progress visibility and
+enable resumption from any failure point.
+
+**Block the entire graph on any node error**
+
+Instead of error isolation, the wavefront could halt entirely when any node
+fails.
+
+Not chosen: independent branches have no reason to stop. Blocking Service-B
```
NicholasBlaskey, line 773: I do wonder if this would cause unexpected behavior if users had implicit dependencies (maybe to get around a circular dependency they hard code some values that could have been dependencies) 

We could say model your dependencies better but stopping on errors feels like a safer default

docs/design/proposals/level-aware-wavefront-instance-controller.md
jakobmoellerdev: Yeah, it frankly is. It is a simple recursion break and we could make this adjustable.

docs/design/proposals/level-aware-wavefront-instance-controller.md
jakobmoellerdev: This is a very smart thing! I like this idea. It does conflict a little with our last community call thought of having the entire template be expressable in CEL though no?

docs/design/proposals/level-aware-wavefront-instance-controller.md
jakobmoellerdev: Hmm. Are you suggesting we block the branch even if it itself does not depend on the level then?

docs/design/proposals/level-aware-wavefront-instance-controller.md
jakobmoellerdev: I believe the biggest issue comes from scheduling. if we have a true per-node walk, then it will be very hard to reason about a sane batching / concurrency mechanism from what I know. Do you have a suggestion on a sane bounded scheduling for such a graph traversal that is not based on Kahns or other levelling schemes?

docs/design/proposals/level-aware-wavefront-instance-controller.md
Skarlso: >kinda an edge case but we have seen a report about two ACK EC2 security group resources that referenced each other and blocked on each other deletion

This is kinda a thing in AWS since the begin of time and has been reported and tried to be resolved in several other communities across the ecosystem. :D ( I was a maintainer of cluster-api-provider-aws, we ran into this problem CONSTANTLY ).

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+
+The synchronizer operates in four phases per reconcile cycle:
+
+```mermaid
+flowchart LR
+    P["Phase 1\nProject"] --> D["Phase 2\nDiff"]
+    D --> W["Phase 3\nExecute\nwavefront"]
+    W --> S["Phase 4\nStatus\nupdate"]
+    W -- "node becomes ready" --> P
+    S -. "watch event\ntriggers new\nreconcile" .-> P
```
Skarlso, line 129: What happens if the event is dropped due to backpressure?

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+
+#### Design details
+
+##### Phase 1: Project
+
+Projection takes the structural DAG from the current
+[GraphRevision](https://kro.run/api/crds/graphrevision) and the instance's
+input values, and produces a **runtime DAG**: which resources to create, at
+which dependency levels, with fully-rendered templates. Dynamic elements --
+`includeWhen`, `forEach`, CEL expressions - are all resolved here.
```
Skarlso, line 150: As far as I understand the DAG is _eventually_ lazy evaluated anyways, so do I understand it correctly, that this would switch the lazy evaluation to an eager one so front-load the necessary data already when dealing with defining the DAG with ALL the data and evaluated expressions already?

If yes, does that mean that we will be front-heavy in reconcile? I believe the cost is the same, just want to make sure I get that right.

So combining this with the above thing where we requeue on evaluation if a node is not ready and reading `includeWhen` I have to ask this... Does this need to be in a loop? The original, as far as I understand it now, does this weird ErrDataPending loop. This now, would generate the levels and the DAG beforehand and includeWhen would decide is a node is part of that dag or not. It evaluates it, turns out it's not part, so it evaluates only level 1 for example. But now level 2 is ready, but it's not in the graph so we reloop.

Originally this is when ErrDataPending happens and the whole thing is re-evaluated. Could this benefit from that? So what I'm trying to ask is is the in-reconcile re-project loop worth it versus just requeuing and re-projecting on the next reconcile loop upon an event change?

docs/design/proposals/level-aware-wavefront-instance-controller.md
Skarlso: I wonder if it is necessary really.. I'm working through a thought of mine here with events and requeuing but maybe it's not worth it. :)

Skarlso: Okay, so I'm wondering, since node watches exist can't we do something like this:

Project -> Diff -> Execute -> Level 0 -> SetObserved; Level 1 -> SetObserved ( these are in parallel ) -> But then level 2 is not `ready` yet, return and `watch`.  -> Status update IN_PROGRESS, ACTIVE, GATED.

So, basically, we combine the requeue on watch event from the original with the parallel processing of the new. I know that you would like to switch this requeue from being something that happens on deeply nested graphs but I feel like this arbitrary 5* loop is just a bandaid that aims to shift the cost of waiting to a CPU bound operation that's faster. But I'm wondering if it's justified?

I feel like, if we recompute the DAG on each reconcile anyways, we might as well utilize the event watch and that would simplify the caching problems as well. Or am I missing some data somewhere that tells some kind of different truth about the real consequence?

docs/design/proposals/level-aware-wavefront-instance-controller.md
Skarlso: How would reordering the RGD dependencies work in this case? 🤔 Wouldn't that mean that there would be a bunch of create/delete operations since the object name changed? Or did I misunderstand the `name: ` here?

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+```
+
+Deletion does not respect `propagateWhen`. When a user deletes an instance,
+all resources are cleaned up promptly regardless of gates.
+
+**Mixed forward and reverse in the same reconcile**
+
+When a new GraphRevision adds some nodes and removes others, both wavefronts
+run in the same reconcile cycle. The forward wavefront runs first, then the
+reverse wavefront prunes stale resources:
```
Skarlso, line 414: Forgive my ignorance here, is it possible that the same revision adds a node then removes the same node?  Even if yes, I would think that's not a technical problem, right?

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+When a node fails, only its dependents are blocked — independent branches
+at the same or deeper levels continue unaffected:
+
+```
+DAG:  ConfigMap ──► Deployment-A ──► Service-A
+      Secret    ──► Deployment-B ──► Service-B
+
+Deployment-A fails (SSA conflict). Deployment-B is Ready.
+
+  Level 1 terminal states: [Error, Ready] - level complete, advance.
```
Skarlso, line 443: How would this be rolled back...? I mean, how would I fix this? 

For example, one of the legs fails, but the others succeeded. Let's say, I have an RGD with a database migration and a deployment that would use it... Technically, they are defined separately for a faster deployment and my deployment just retries the connection until it's ready. The application manages to come online expecting a new schema in the database. But the database deployment fails for some reason to apply.

The application is now running because it detected the connection to the previous database that was untouched and now the application is actually runtime failing, expecting a different schema in the database and potentially, broke something by overwriting some data it expected to be there but wasn't since the database failed to deploy but the application did not ( since they are processed in parallel ).

I _think_ the answer to that is that in that case maybe the application should do a pre-check ( meaning it's on the developer to make sure you don't break stuff ) or that you should then maybe think about making the app dependent on the db deployment after all and not process them in parallel.

I'm not sure however. What's you thought?

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+│    fail  Error
+│           │
+│           │ SSA success,
+│           │ readyWhen = false
+│           ▼
+└──────► WaitReady ─── readyWhen = true ──► Ready ───► Deleting ───► Deleted
+                                            │                        ▲
+                                            │  node removed or       │
+                                            │  instance deleted      │
+                                            └────────────────────────┘
```
Skarlso, line 497: This would definitely be more easy to follow / view as a mermaid diagram. :))) The ASCII code is falling apart and is MUCH harder to refactor if you want to change something. :D WDYT?

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+    Message: "6/8 nodes Ready. 1 Error: [deployment-a].
+             1 Blocked: [service-a] (depends on deployment-a).
+             deployment-a: SSA conflict (field manager 'helm'
+             owns .spec.replicas)."
+```
+
+**Node state machine**
+
+```
+                      includeWhen = false
```
Skarlso, line 468: I don't get this statemachine. :D Maybe a mermaid would be better? As I read it now, it looks like, there is no path forward from Blocked. includeWhen true -> Blocked. includeWhen becomes true -> Blocked.

Ready -> Blocked. Excluded -> Blocked. :D

And Applying apparently can only `Error`? :D

docs/design/proposals/level-aware-wavefront-instance-controller.md
```diff
@@ -0,0 +1,932 @@
...
+| **Node moves level** | `kro.run/level` label updated during SSA. Processed at new level.     |
+| **Edge added**       | Node waits for the new dependency. Cycles are caught at compile time. |
+| **Edge removed**     | Node may move to a lower level. Processed there.                      |
+
+If a new GraphRevision becomes active mid-reconcile, the current reconcile
+finishes against the in-progress revision. The controller picks up the new
+one on the next cycle. The default behavior is **skip-to-latest** — each
+GraphRevision is a complete snapshot, so there is nothing in revision N that
+must be applied before N+1. An opt-in `revisionPolicy: Serialized` is
+available for users who need per-revision health validation.
```
Skarlso, line 648: Are all revisions stored...?
Nevermind, I just see the open question for `skip-to-latest`.

So we already have `LastIssuedRevision` in status. Can't we use that if it exists and rebuild from annotation if it doesn't?

Skarlso (commented):
Fantastic write-up honestly. I think this is a massive benefit for the current implementation even with the given short commings.

I have a couple of questions for clarification. :) Otherwise, this is amazing! I really like the per level parallel processing of the DAG.

docs/design/proposals/level-aware-wavefront-instance-controller.md
NicholasBlaskey: I do think there is an argument for deletion that it should be unique and protected in a way that updates aren't. Blocking deletion due to graph errors feels preventable to me.

I might not have full context here on what the fallouts of having the entire template be expressible are. 

>How would reordering the RGD dependencies work in this case? 🤔 Wouldn't that mean that there would be a bunch of create/delete operations since the object name changed? Or did I misunderstand the name: here?

Yeah, if the name changes. This is already supported today, just the delete path is at risk if for whatever reason we start failing to evaluate expressions in identity fields.

jakobmoellerdev:
this will be superceded by our new graph engine (see https://github.com/kubernetes-sigs/kro/pull/1355)
/close

PR #1218, opened 2026-03-28
fix: block RGD deletion while instances still exist

### LINKED ISSUE #1067: You can delete RGDs keeping Applications dangeling
### Description

~~~
# Create a RGD
$ kubectl apply -f nginx-simple.yaml 
resourcegraphdefinition.kro.run/my-application created
# Create Application
$ kubectl apply -f instance.yaml 
application.kro.run/my-app-instance created
$ kubectl get applications.kro.run 
NAME              STATE    READY   AGE
my-app-instance   ACTIVE   True    31s
# Delete RGD
$ kubectl delete -f nginx-simple.yaml 
resourcegraphdefinition.kro.run "my-application" deleted
Now everyting is gone.
$ kubectl delete -f  instance.yaml 
application.kro.run "my-app-instance" deleted

Will never finishes as no one picks the finalizer

~~~

- kro version: kro-v0.8.5
- Kubernetes Version 
-Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.34.0

_No response_

officialasishkumar:
When a `ResourceGraphDefinition` is deleted, the controller immediately
shuts down the instance micro-controller and optionally deletes the CRD.
If instances of the generated CRD still exist at that point, they are
left with a dangling `kro.run/finalizer` that no controller can remove,
causing `kubectl delete` on the instance to hang indefinitely.

**Root cause:** The RGD deletion path proceeds through cleanup
(controller shutdown + CRD deletion) without checking whether instances
still exist. Once the instance controller is gone, the finalizer on
existing instances can never be removed.

**Fix:** Before proceeding with cleanup in the RGD reconciler's deletion
path, list instances of the generated CRD using the dynamic client. If
any instances are found, log a message and requeue after 10 seconds.
The user must delete all instances first before the RGD can be fully
cleaned up. If the CRD is already gone (deleted externally or not yet
created), the check is skipped and cleanup proceeds normally.

**What changed:**
- `pkg/controller/resourcegraphdefinition/controller.go`: Added instance
  existence check before `cleanupResourceGraphDefinition()` in the
  deletion path. Uses `r.clientSet.Dynamic().Resource(gvr).List()` with
  `Limit: 1` for efficiency.
- `pkg/controller/resourcegraphdefinition/controller_test.go`: Added
  `newFakeSetForDeletion` helper and a new test case
  `blocks_deletion_while_instances_still_exist`. Updated existing
  deletion test cases to include the `clientSet` field.

**How to test:**
```bash
go test ./pkg/controller/resourcegraphdefinition/ -run TestReconcile -v
```

Fixes #1067

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller.go
+++ b/pkg/controller/resourcegraphdefinition/controller.go
@@ -17,11 +17,14 @@ package resourcegraphdefinition
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


pkg/controller/resourcegraphdefinition/controller.go
michaelhtm, line 1: should this be behind a flag? called `retain-instances` or `abandon-instances`? It can default to the current behavior

AndrewCharlesHay:
Security pass on this, looks clean overall. Two non-blocking notes for the maintainers:

1. **TOCTOU between the existence check and cleanup**: there's a race where a new instance could be created after `List(Limit: 1)` returns 0 but before the CRD/finalizer teardown begins. Not a security bug — k8s GC orphans instances when the CRD disappears — but worth a sentence in the PR description or a comment in the code so future readers don't try to "fix" it by removing the check.

2. **Indefinite pin**: a stuck instance finalizer (buggy downstream controller, revoked RBAC, etc.) will hold the RGD in Terminating forever at the 10s requeue cadence. That's intentional per the PR goal, but probably deserves a release-note callout so operators know to look at instance finalizers when an RGD appears stuck.

The `Limit: 1` + narrow error classification (`IsNotFound` / `IsNoMatchError` only) is the right shape — nice.

Issue #1060, opened 2026-02-19
Multicluster Support for KRO


mjudeikis:
### Feature Description

# Feature Request: Multicluster Support for KRO

**Problem Statement**:

KRO currently operates on a single Kubernetes cluster. Users who manage multiple clusters must deploy and configure KRO separately on each cluster. This creates operational overhead and prevents centralized management of ResourceGraphDefinitions across a fleet of clusters.

Organizations run workloads across multiple clusters for geographic distribution, environment separation, team isolation, or high availability. These users need a way to define resources once and deploy them across multiple clusters from a single control plane.

**Proposed Solution**:

Add multicluster support to KRO using a hub-spoke model:

- **Hub cluster**: Runs KRO controller and stores ResourceGraphDefinitions (RGDs)
- **Spoke clusters**: Receive generated CRDs and run instances with their child resources

Key design points:
- Use `multicluster-runtime` (MCR) library as a drop-in replacement for `controller-runtime`
- Cluster discovery via pluggable providers (starting with kubeconfig secrets)
- RGDs defined only in hub, CRDs and instances live in spokes
- Opt-in via `--enable-multicluster` flag (backward compatible)

**Alternatives Considered**:

- **Separate KRO per cluster**: Current approach. Simple but creates operational overhead and prevents centralized management. And clusters can't be just controlplanes, and not all clusters can be controlplanes, so this is not a viable long-term solution.
- **GitOps-based distribution**: Use ArgoCD to deploy RGDs to multiple clusters. Adds complexity and another tool dependency. No unified status visibility. Still required per-cluster KRO instances.

**Additional Context**:

**What is multicluster for KRO?**

> Proposed definition: Multicluster for KRO means the ability to manage ResourceGraphDefinition instances and their child resources across multiple Kubernetes clusters from a single KRO control plane. Keeping RGDs centralized in a hub cluster while distributing CRDs and instances to spoke clusters where workloads run.

Specifically:
- **One hub, many spokes**: A single KRO controller runs in a hub cluster and connects to multiple spoke clusters
- **Definitions stay central**: RGDs are authored and stored only in the hub cluster
- **Resources are distributed**: CRDs and instances live in spoke clusters where workloads run
- **Same-cluster locality**: All child resources of an instance are created in the same spoke cluster as the instance itself (no cross-cluster dependencies)

What multicluster is **not** for KRO:
- Federation or replication of RGDs across clusters
[...]

PR #1223, opened 2026-03-31
KREP-013: Multicluster support via multicluster-runtime

mjudeikis:
## Summary                                                                                                                                                                               
                                                                                                                                                                                           
  Multicluster support for KRO using `sigs.k8s.io/multicluster-runtime` (MCR) as a drop-in for `controller-runtime`. Hub-spoke model: RGDs on hub, CRDs + instances on spokes. No RGD API  
  changes.                                                                                                                                                                                 
                                                                                                                                                                                           
  Addresses #1060.                                                                                                                                                                         
   
  ### Key points                                                                                                                                                                           
  - MCR operates at the reconcile loop level, transparent to the RGD API
  - Pluggable cluster discovery (kubeconfig Secrets first, Cluster API later)                                                                                                              
  - Per-cluster `DynamicController` instances via `MulticlusterDynamicController`                                                                                                          
  - Opt-in via `--enable-multicluster` flag, zero change without it                                                                                                                        
  - Same-cluster locality: all child resources stay in the instance's cluster                                                                                                              
                                                                                                                                                                                           
  ### Relationship to KREP-012                                                                                                                                                             
  Compares with #1064 (Cluster Targets). Different layer — KREP-012 is API-level (Target field in RGD), KREP-013 is runtime-level (reconcile loop). They're complementary: KREP-013 handles
   "where do instances live", KREP-012 handles "where do specific resources go". KREP-012's Target can be built on top of KREP-013's `ClusterClientFactory`.                               
                                                            
  ### Also covers                                                                                                                                                                          
  - Code restructuring needed to make controller-runtime vs MCR pluggable (5 interface extractions, incremental migration path)
  - How KREP-012 Target functionality can be implemented within KREP-013's foundation                                                                                                      
                                                                                                                                                                                           
  POC: https://github.com/mjudeikis/kro/tree/mcr.poc

### DIFF
--- a/docs/design/proposals/multicluster-runtime.md
+++ b/docs/design/proposals/multicluster-runtime.md
@@ -420,9 +420,19 @@ health probe system.
 (`controller-runtime/pkg/metrics`), GraphRevision controller
 (`ctrl.NewControllerManagedBy` builder).
 
+**Medium-high coupling** - `DynamicController` has bidirectional coupling
+with the instance controller via the `WatchCoordinator`: the DynamicController
+dispatches reconcile events to instance controllers, and instance controllers
+feed `WatchRequest`s back to the `WatchCoordinator` (which manages informers
+via `WatchManager`). The coordinator maintains per-instance watch state,
+reverse indexes for event routing, and handles informer lifecycle — all of
+which are cluster-scoped. In multicluster mode, the `WatchManager` needs
+per-cluster informers (watching child resources on the correct spoke), and
+the `WatchCoordinator`'s indexes need to be cluster-aware (routing events
+from spoke A only to instances on spoke A).
+
 **Low coupling** - Instance controller (already uses custom abstractions,
-resolves graphs from the in-memory `revisions.Registry`),
-`DynamicController` (only needs `Runnable` interface and `ctrl.Request` type).
+resolves graphs from the in-memory `revisions.Registry`).
 
 ### Required Changes
 
@@ -486,12 +496,36 @@ The only change is updating `SetupWithManager()` to accept the kro manager
 interface and use the appropriate builder pattern, consistent with the RGD
 controller restructuring.
 
-#### 4. DynamicController Abstraction
+#### 4. DynamicController and Watch System Abstraction
 
-The `DynamicController` currently implements controller-runtime's `Runnable`
-interface and uses `ctrl.Request`. For multicluster, the
-`MulticlusterDynamicController` wraps per-cluster instances. Both need to
-satisfy a common interface:
+The `DynamicController` has deeper coupling than just the `Runnable` interface
+and `ctrl.Request` type. Instance controllers feed `WatchRequest`s back to
+the `WatchCoordinator`, which manages per-GVR informers via the
+`WatchManager`. This creates a bidirectional dependency:
+
+```
+DynamicController → instance controller (dispatches reconcile events)
+instance controller → WatchCoordinator → WatchManager (feeds back watch requests)
+```
+
+The `WatchManager` creates informers using `metadatainformer.NewFilteredMetadataInformer`
+with a single metadata client. The `WatchCoordinator` maintains per-instance
+watch state and reverse indexes (scalar + collection) for event routing — all
+currently scoped to a single cluster.
+
+For multicluster, the watch system needs cluster awareness:
+
+- **`WatchManager`**: needs per-cluster metadata clients and per-cluster
+  informers. A child resource watch on spoke A must use spoke A's metadata
+  client, not the hub's. The `createInformer` func needs a cluster parameter.
+- **`WatchCoordinator`**: the reverse indexes (`scalarIndex`, `collectionIndex`)
+  must include cluster identity so events from spoke A route only to instances
+  on spoke A. The `instanceKey` struct gains a `clusterName` field.
+- **`EnqueueFunc`**: the callback `func(parentGVR, instance)` needs to include
+  cluster name so the DynamicController enqueues to the correct per-cluster
+  work queue.
+
+The abstraction interface for the RGD controller:
 
 ```go
 // pkg/dynamiccontroller/interface.go
@@ -502,8 +536,14 @@ type Interface interface {
         parentGVK schema.GroupVersionKind, handler Handler,
         childGVRs ...schema.GroupVersionResource) error
     Deregister(ctx context.Context, gvr schema.GroupVersionResource) error
+    Coordinator() *WatchCoordinator
 }
 ```
+
+The `MulticlusterDynamicController` wraps per-cluster `DynamicController`
+instances, each with its own `WatchManager` and cluster-scoped informers.
+The `WatchCoordinator` is shared but cluster-aware — it routes events to
+the correct instance based on both GVR and cluster name.
 
 The RGD controller depends on this interface, not the concrete type. At startup,
 `main.go` creates either a `DynamicController` or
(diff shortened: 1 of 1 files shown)


docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,811 @@
...
+- Deploy KRO N times for N clusters, duplicating RGDs and configuration
+- Wrap KRO with GitOps tools like ArgoCD, adding complexity and losing unified
+  status visibility
+- Build custom multi-cluster orchestration outside KRO
+
+Example: a platform team managing 50 edge clusters needs the same set of RGDs
+available on all clusters. Today this requires 50 KRO installations, 50 copies
+of each RGD, and no centralized way to observe instance health across the fleet.
+
+The hub-spoke model solves this: define RGDs once in the hub, KRO automatically
```
jakobmoellerdev, line 31: I am lacking a support for spoke-to-spoke communication and Meta RGDs that reconcile around multiple clusters. maybe out of scope for this proposal but should probably be mentioned right?

docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,811 @@
...
+- **Per-cluster controllers**: Each spoke gets its own `DynamicController`
+  instance with cluster-specific clients and informers.
+- **Opt-in**: Enabled via `--enable-multicluster` flag. Without the flag, KRO
+  behaves exactly as it does today (single-cluster mode).
+
+### Core Concepts
+
+**Hub and Spoke Separation**
+
+RGDs are authored and stored only in the hub cluster. The RGD controller runs
```
jakobmoellerdev: We will need to update this a little because GraphRevisions are now a thing!

docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,811 @@
...
+
+### CRD Distribution
+
+When an RGD is reconciled and a CRD is generated, KRO distributes the CRD to
+all engaged spoke clusters:
+
+1. RGD controller generates CRD (existing behavior)
+2. CRD is applied to the hub cluster (existing behavior)
+3. CRD is applied to all engaged spoke clusters via `ClusterClientFactory`
+4. When a new cluster is engaged, all existing CRDs are distributed to it
```
jakobmoellerdev: how would we control a rollout between multiple runtime clusters? (i.e. phased)
Also how would we control lifecycle differentiation (as in spoke A gets Graph A, spoke B gets Graph B)? Is such a thing possible with mcr?

docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,811 @@
...
+
+CRD distribution uses server-side apply to handle conflicts gracefully. If a
```
jakobmoellerdev, line 325: nit: I feel like this is irrelevant to the Proposal

docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,811 @@
...
+over which cluster receives resources. The cluster target is part of the
+application definition - for example, "this database goes to the production
+cluster" or "deploy to the cluster specified by the user in spec.targetCluster".
+
+**KREP-013** is suited for fleet management where the same RGDs should be
+available across many clusters. The cluster topology is an operational concern
+decoupled from the RGD definition - for example, "make all our platform RGDs
+available on every cluster in the fleet".
+
+### Complementary, Not Competing
```
jakobmoellerdev, line 373: I think this is a good overview, thanks for this

docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,811 @@
...
+`client.Client`). The key coupling points fall into three categories:
+
+**High coupling** - Manager creation/lifecycle, RGD controller's embedded
+client and builder pattern, health probe system.
+
+**Medium coupling** - Logging (`logr`), metrics registry
+(`controller-runtime/pkg/metrics`).
+
+**Low coupling** - Instance controller (already uses custom abstractions),
+`DynamicController` (only needs `Runnable` interface and `ctrl.Request` type).
```
jakobmoellerdev: There is now stronger coupling in dynamic controller because the instance feeds back watch requests to DynamicController

jakobmoellerdev (commented):
I believe this is a good proposal, has an implementation backed proof of concept, and can work without maintaining any debt, building on existing primitives. 

I'm in support of the proposal, well done.

docs/design/proposals/multicluster-runtime.md
mjudeikis: Spoke to spoke is out of scope. This would require dependencies, like network connectivity between them. One could build this using mcr, but this is out of scope. 

> Meta RGDs that reconcile around multiple clusters
what do you mean by this? RGDs produced by RGDs?

docs/design/proposals/multicluster-runtime.md
mjudeikis: No, this would require to pull logic from KREP-012 in a way. One could build something like this (like deployment rollout strategies), but because it works on the reconcile loop level, it does not have it. We would need to add something at the API level.

docs/design/proposals/multicluster-runtime.md
mjudeikis: I think we need to be explicit for this. We don't have a CRD distribution problem as it's a singleton cluster. After this was implemented, this is not the case anymore, and it becomes closer to a distributed system in traits than an operator.

docs/design/proposals/multicluster-runtime.md
mjudeikis: We can add this, but I would do this as follow-up

docs/design/proposals/multicluster-runtime.md
mjudeikis: Ah, wrote this looking to old version of KRO :D updating to represent how this is handled

mjudeikis: TL:DR:

```
GraphRevisions are never distributed to spokes. They are an internal hub
concern — spokes only receive the generated CRDs. The revision registry is
hub-local in-memory state that instance controllers on the hub read from
directly.
```

docs/design/proposals/multicluster-runtime.md
jakobmoellerdev: Maybe for MCR it is, but not for KRO. we need to put this into the proposal

docs/design/proposals/multicluster-runtime.md
jakobmoellerdev: For me the KREP will be incomplete without a vision on how to move to phased rollout between clusters TBH. even if it is just an outlook

docs/design/proposals/multicluster-runtime.md
mjudeikis: Added a section about the watch system.

TL:DR:
```
For multicluster, the watch system needs cluster awareness (keys, funcs)
```

mjudeikis:
This KREP now has many technical implementation details. I can split this a bit, but tried to keep it tied "to the ground" so impact is clearer

docs/design/proposals/multicluster-runtime.md
mjudeikis: Can you give me a use case for spoke-to-spoke? Im having a hard time wrapping my head around why one would want this in the initial version. I think im missing something here.

docs/design/proposals/multicluster-runtime.md
mjudeikis: How would you imagine this? The idea I was thinking of is a separate object, like `RGDRollout`, which is based on selectors/labels, that would orchestrate it. Easiest, would not overload existing API and would be opt-in only when using multi-cluster mode.

docs/design/proposals/multicluster-runtime.md
jakobmoellerdev: If you wanna maintain a graph across clusters this becomes interesting: 

Example: Service Mesh Configuration between 2 clusters.

You could use KRO to:
1. Setup both clusters with certificate / mTLS information
2. Configure Ingress/Egress on both clusters
3. Then establish connectivity between both

all from one graph

mjudeikis: But this is hub-to-spoke, as the graph still comes from a central hub cluster as orchestration for both leaf clusters. There would be no KRO running in any of the leaf/spoke clusters, so no KRO operations as such. 
The connection establishment is a side-effect of how KRO can be used, not direct spoke-to-spoke communication by KRO itself.

docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,1322 @@
...
+clients are created when `Engage` is called and cleaned up when the cluster's
+context is cancelled.
+
+### Instance Controller Changes
+
+The instance controller becomes cluster-aware while continuing to resolve
+compiled graphs from the hub-local in-memory registry:
+
+```go
+func (c *Controller) Reconcile(ctx context.Context, clusterName string, req ctrl.Request) error {
```
ntnn, line 258: `clusterName` should be the typed `multicluster.ClusterName`

docs/design/proposals/multicluster-runtime.md
```diff
@@ -0,0 +1,1322 @@
...
+In multicluster mode, it returns cluster-specific clients. The instance
+controller uses the factory interface, making it cluster-aware without
+importing MCR.
+
+#### 6. Reconcile Signature Alignment
+
+The instance controller's `Reconcile(ctx, req) error` and the RGD controller's
+`Reconcile(ctx, obj) (Result, error)` differ. For multicluster, both need a
+cluster name in the reconcile path. Rather than changing signatures, the cluster
+name is carried in the context:
```
ntnn, line 586: This conflicts with the seciont "Instance Controller Changes" above, or?
There is says the instance controller should expect the cluster name in its signature now.
Or are these two different controllers?

PR #1230, opened 2026-04-01
Enable full CEL expressions support for external collection selectors

tbarizien:
Changes the type of `resources[].externalRef.metadata.selector` from `metav1.LabelSelector` to `extv1.JSON` in order to solve https://github.com/kubernetes-sigs/kro/issues/1228

Writing something like the snippet below in a RGD is not stopped anymore by the api server and gets reconciled nicely: 
```yaml
resources:
- id: refs
    externalRef:
      metadata:
        selector: ${schema.spec.selector}
```

This enables RGD writers to give control to the RGD users on the resources that are being referenced.

A first validation layer during RGD validation makes sure the selector either contains CEL or is a valid LabelSelector. 
The logic resolving the selector was already present. No changes had to be made here.


api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -120,7 +124,6 @@ type CRDMetadata struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
-// +kubebuilder:validation:XValidation:rule="(has(self.name) && !has(self.selector)) || (!has(self.name) && has(self.selector))",message="exactly one of name or selector must be provided"
```
tbarizien, line 145: I had to move this check to the RGD validation step. It wasn't keen on the new selector type 😬

jakobmoellerdev:
Thanks a lot for your contribution! Before we continue review, would you please sign the CLA? Thanks!

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -136,10 +139,43 @@ type ExternalRefMetadata struct {
 	Namespace string `json:"namespace,omitempty"`
 	// Selector is a label selector for collection external references.
 	// When set, all resources matching the selector are included.
+	// The selector may be a literal Kubernetes LabelSelector object or a full CEL
+	// expression that resolves to one at reconcile time.
 	// Mutually exclusive with Name.
 	//
 	// +kubebuilder:validation:Optional
-	Selector *metav1.LabelSelector `json:"selector,omitempty"`
+	Selector *extv1.JSON `json:"selector,omitempty"`
```
jakobmoellerdev, line 147: I believe elsewhere in the spec we use the runtime version of this extension. I think both would work here but why did you choose specifically this one?

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -136,10 +139,43 @@ type ExternalRefMetadata struct {
 	Namespace string `json:"namespace,omitempty"`
 	// Selector is a label selector for collection external references.
 	// When set, all resources matching the selector are included.
+	// The selector may be a literal Kubernetes LabelSelector object or a full CEL
```
jakobmoellerdev: I believe we should link to the normal selector spec here

pkg/graph/validation.go
```diff
@@ -267,6 +270,44 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
 	return nil
 }
 
+func validateExternalRefMetadata(metadata v1alpha1.ExternalRefMetadata) error {
```
jakobmoellerdev, line 273: Im wondering here, is this not a block that should belong to the standard metadata? could it be templated normally as well or do we really need such an extension function?

jakobmoellerdev (commented):
First Pass. I believe @a-hilaly you mentioned there might a simpler way here. Please chime in as needed!

api/v1alpha1/resourcegraphdefinition_types.go
jakobmoellerdev: Why cant this be an OR selector that validates on object or string? are we sure this doesnt work?

tbarizien:
What is the policy regarding dryness of test files? [The linter is failing on duplicates in `Expect` statements](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kro/1230/presubmits-verify-lint/2041536195715928064) , I could abstract it but it doesn't feel like improving readability/maintainability by much.

api/v1alpha1/resourcegraphdefinition_types.go
tbarizien: I used it because `runtime.RawExtension` marks the field as an object when controller-gen runs while `extv1.JSON` doesn't. 
After looking into it, I could also add both `Schemaless` and `PreserveUnknownFields` markers on the field and have it be a `RawExtension`. 

It's important that the field is not declared as an object in order to support the `selector: ${cel.expression}` syntax.

Should I use `RawExtension` with markers or keep it this way?

api/v1alpha1/resourcegraphdefinition_types.go
tbarizien: I have tried this and `ExaclyOneOf` without success. Since the field has no type it isn't added to the backing cel object. The following error then happens on CRD creation: `undefined field 'selector'`

I haven't been able to find a workaround using xvalidation :/

pkg/graph/validation.go
tbarizien: I don't understand what you mean by "templating normally" here, could you elaborate?

jakobmoellerdev:
in this case I would appreciate you actually addressing the lint

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -136,10 +139,43 @@ type ExternalRefMetadata struct {
...
+func (m ExternalRefMetadata) HasSelector() bool {
+	if m.Selector == nil {
+		return false
+	}
+
+	raw := bytes.TrimSpace(m.Selector.Raw)
+	return len(raw) > 0 && string(raw) != "null"
+}
+
+func MarshalJSONValue(value any) (*extv1.JSON, error) {
```
jakobmoellerdev: these functions should not belong to our API package, please move them into internal helpers

jakobmoellerdev: on second skim, not even sure theyre used outside of tests.

jakobmoellerdev (commented):
While I do like the change (this should have been like this from the start), this does require testing that the original created RGDs are not broken if they are applied. If they are then this would be a breaking change and could not be accepted easily. Do you mind creating a test for this?

tbarizien:
@jakobmoellerdev I don't really understand what you mean by "testing that the original created RGDs are not broken if the proposed changes are applied". Could you elaborate by providing a test scenario and maybe point me to where similar tests have been implemented before. 

Those changes don't have any impact on the CRD the RGD ends up creating, only on the behavior of the controller when reconciling them based on the RGD they belong to. The stricter selector behavior was integration tested before and those tests are still present and passing. 
Maybe I'm missing something here :see_no_evil:

jakobmoellerdev:
> @jakobmoellerdev I don't really understand what you mean by "testing that the original created RGDs are not broken if the proposed changes are applied". Could you elaborate by providing a test scenario and maybe point me to where similar tests have been implemented before.
> 
> Those changes don't have any impact on the CRD the RGD ends up creating, only on the behavior of the controller when reconciling them based on the RGD they belong to. The stricter selector behavior was integration tested before and those tests are still present and passing. Maybe I'm missing something here 🙈

Im talking about a RGD created with 0.9.0 for example, that then must be changed. The reason for this is that you are breaking the openAPI schema of the RGD CRD here with a different field schema for the selector. So it would be best to test that this does not cause regressions

jakobmoellerdev (commented):
/lgtm

pkg/graph/validation.go
```diff
@@ -267,6 +270,44 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
...
+	}
+
+	raw := bytes.TrimSpace(metadata.Selector.Raw)
+	if len(raw) == 0 || string(raw) == "null" {
+		return nil
+	}
+
+	if bytes.Contains(raw, []byte("${")) {
+		return nil
+	}
```
michaelhtm: Why does the existence of `${` make it valid? should we validate it's a valid CEL expression as well?

pkg/graph/validation_test.go
```diff
@@ -782,6 +784,131 @@ func TestValidateCombinableResourceFields(t *testing.T) {
...
+		{
+			name: "selector object is valid",
+			metadata: v1alpha1.ExternalRefMetadata{
+				Selector: toRawExtension(t, map[string]interface{}{
+					"matchLabels": map[string]string{"app": "demo"},
+				}),
+			},
+		},
+		{
+			name: "selector CEL string is valid",
```
michaelhtm, line 873: can we add a test with invalid CEL expression (maybe a missing `}`)?

michaelhtm (approved):

pkg/graph/validation.go
tbarizien: The logic I was following here was too superficial. I dug a bit more on how the controller validates fields and
am pushing some changes that will add some more integration tests on such use cases and make the whole validation process in line with other unstructured fields.

tbarizien:
Is there anything needed here from my side to move forward? 
My team and I are needing this to avoid some costly workaround. I'd like reassurance there is agreement on the targeted behavior before we start to maintain a fork waiting for this to be merged.

pkg/graph/validation.go
```diff
@@ -267,6 +270,90 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
...
+	raw := bytes.TrimSpace(metadata.Selector.Raw)
+	if len(raw) == 0 || string(raw) == "null" {
+		return nil
+	}
+
+	if bytes.Contains(raw, []byte("${")) {
+		if raw[0] != '{' {
+			// The entire selector is a string (e.g. a standalone CEL expression
+			// that resolves to a LabelSelector at runtime). Allow it; the
+			// downstream parser will validate the expression.
```
michaelhtm: this lets something like `selector: "${schema.spec.selector"` pass as a valid CEL expression. we should instead use `IsStandaloneExpression` in parser.go

pkg/graph/validation.go
```diff
@@ -267,6 +270,90 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
...
+	if raw[0] != '{' {
+		return fmt.Errorf("selector must resolve to a Kubernetes LabelSelector object")
+	}
+
+	var selector metav1.LabelSelector
+	if err := json.Unmarshal(raw, &selector); err != nil {
+		return fmt.Errorf("invalid selector object: %w", err)
+	}
+
+	if _, err := metav1.LabelSelectorAsSelector(&selector); err != nil {
```
michaelhtm: should we use `metav1validation.ValidateLabelSelector` here instead?

pkg/graph/validation.go
tbarizien: It wasn't exported so I restrained using it. I exported it and incorporated it in the validation process.

pkg/graph/validation.go
tbarizien: Wasn't aware of that method. Used it instead

pkg/graph/builder.go
```diff
@@ -1227,6 +1232,14 @@ func resolveSchemaAndTypeName(c *schema.Cache, segments []fieldpath.Segment, roo
 // expectedTypeForField computes the expected CEL type for a field descriptor
 // by deriving it from the OpenAPI schema at the path.
 func expectedTypeForField(bc *buildContext, descriptor *variable.FieldDescriptor, rootSchema *spec.Schema, resourceID string) *cel.Type {
+	// Paths under metadata.selector come from ExternalRef synthetic resources.
```
Skarlso, line 1215: Unless I misunderstand here something, there is a real problem with this approach.

Previously `resolveExternalCollectionSelector` was relying on a structure that was a proper... structure and the api server could give you an error in case there was some malformed data around.

Unless I didn't trace the validation properly, there are no hard stops on parsing something like:
```
spec.selector: {app: demo}
```

which a user could think is the right syntax. Or as far as I saw even a typo might get away here. Something like:
```json
{"matchLabel": {"app":"demo"}}
```

without the `s`. And if that happens, we fall back to 
```go
	if err != nil || !found {
		return labels.Everything(), nil
	}
```

meaning, if I configure this correctly, I basically could select ALL configmaps from a namespace or multiple namespaces depending on how the thing is configured. 🤔

pkg/graph/validation.go
```diff
@@ -267,6 +273,121 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
...
+	var selectorMap map[string]interface{}
+	if err := json.Unmarshal(raw, &selectorMap); err != nil {
+		return fmt.Errorf("invalid selector object: %w", err)
+	}
+	if ml, ok := selectorMap["matchLabels"]; ok {
+		switch v := ml.(type) {
+		case map[string]interface{}:
+			// valid
+		case string:
+			if !(strings.HasPrefix(v, "${") && strings.HasSuffix(v, "}")) {
```
Skarlso: I think this is just `IsStandaloneExpression`, no?

pkg/graph/validation.go
```diff
@@ -267,6 +273,121 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
...
+	}
+
+	raw := bytes.TrimSpace(metadata.Selector.Raw)
+	if len(raw) == 0 || string(raw) == "null" {
+		return nil
+	}
+
+	if bytes.Contains(raw, []byte("${")) {
+		return validateCELSelector(raw)
+	}
```
Skarlso: I think this is a bit prone to errors. I'm sure we have some better ways around to test for CEL expressions.

I'm pretty sure this will not work with literal values in the expression. 🤔

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -135,12 +137,27 @@ type ExternalRefMetadata struct {
...
+	// The selector may be a literal Kubernetes [metav1.LabelSelector] object or a full CEL
+	// expression that resolves to one at reconcile time.
+	//
 	// Mutually exclusive with Name.
 	//
 	// +kubebuilder:validation:Optional
-	Selector *metav1.LabelSelector `json:"selector,omitempty"`
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Selector runtime.RawExtension `json:"selector,omitempty"`
```
Skarlso, line 172: This should be a pointer, otherwise, we will start littering things with the empty value.

pkg/graph/validation.go
```diff
@@ -267,6 +273,121 @@ func validateCombinableResourceFields(res *v1alpha1.Resource) error {
...
+		switch v := me.(type) {
+		case []interface{}:
+			// valid
+		case string:
+			if !(strings.HasPrefix(v, "${") && strings.HasSuffix(v, "}")) {
+				return fmt.Errorf("expected array type for path metadata.selector.matchExpressions, got string")
+			}
+		default:
+			return fmt.Errorf("expected array type for path metadata.selector.matchExpressions, got string")
+		}
```
Skarlso: Hardcoded `got string`.

pkg/graph/builder.go
```diff
@@ -1227,6 +1232,14 @@ func resolveSchemaAndTypeName(c *schema.Cache, segments []fieldpath.Segment, roo
...
+	// The selector is always a LabelSelector, whose structure is known, but
+	// it doesn't exist in the target resource's OpenAPI schema. Return the
+	// concrete types so the CEL type checker can catch mismatches.
+	if t := selectorFieldType(descriptor.Path); t != nil {
```
Skarlso: This is on EVERY resource... right? Not just selectors. 🤔  I need to write a unit test for this, but I'm pretty sure, this is not desired. I'm thinking a schema less crd would potentially get selector schemantics. But I need to verify that.

test/integration/suites/core/externalref_test.go
```diff
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/ptr"
```
Skarlso: This was an accidental revert I believe?

Skarlso (commented):
I have some concerns here and there. :)

api/v1alpha1/resourcegraphdefinition_types.go
tbarizien: I agree with you on a theoritical level, however, I aligned with other optional `runtime.RawExtension` fields in this file that are not defined as pointers. It didn't seem a big deal since emptiness can be checked via the `Raw` array length. 
Should I stray from the repo standards with this field only or fix it across the board?  
I don't really want to break consistency with the neighboring fields and fixing it everywhere seems out of scope here.

Skarlso: Ah, no, yeah. I didn't notice that other fields are similar in that approach. I agree with you then. Thanks 🙇‍♀️

pkg/graph/builder.go
Skarlso: @tbarizien I see you had 👀  on this, would you like me to write a reproducer unit test for this scenario or did you get that? :)

tbarizien: @michaelhtm commit addressed it partially and I added some additional validation on top in the latest commits. Some unit tests have been added to cover the use cases you described. 
If you have any other concerns, I'd be happy to address it ASAP :)

Skarlso: Ah okay, let me check that out. :) I didn't see that, sorry 🙇‍♀️

Skarlso: Okay, so:

```go
func TestResolveExternalCollectionSelector_FailsOpen(t *testing.T) {
	tests := []struct {
		name     string
		selector interface{}
	}{
		{
			name:     "bare label map without matchLabels",
			selector: map[string]interface{}{"app": "demo"},
		},
		{
			name:     "misspelled matchLabels key",
			selector: map[string]interface{}{"matchLabel": map[string]interface{}{"app": "demo"}},
		},
		{
			name:     "null selector",
			selector: nil,
		},
		{
			name:     "empty selector object",
			selector: map[string]interface{}{},
		},
		{
			name:     "empty matchLabels",
			selector: map[string]interface{}{"matchLabels": map[string]interface{}{}},
		},
	}

	for _, tt := range tests {
			cm := &unstructured.Unstructured{Object: map[string]interface{}{
				"apiVersion": "v1",
				"kind":       "ConfigMap",
				"metadata": map[string]interface{}{
					"name":      "configs",
					"namespace": "default",
					"selector":  tt.selector,
				},
			}}

			selector, err := resolveExternalCollectionSelector("configs", cm)

			// we MUST have a hard fail here
			require.Error(t, err,
				"selector %#v resolve to %#v, which lists every ConfigMap in scope",
				tt.selector, selector)
		})
	}
}

```

EACH of these should fail, right? But they don't right now.

tbarizien: Ok, so we've been focusing on pre cel resolution on the previous commits and blocking invalid selectors there, making the RGD unhealthy if some are present. 
We've omitted checking for extra fields on runtime, the code was dropping them silently. 
I've pushed a fix for this, addressing the 2 first cases in your tests.  

The later cases I didn't address. We cannot reject selecting everything, this is a valid use case supported by k8s selectors and we should support it.

Skarlso: Understood. :) Thanks very much! :)

Skarlso:
/lgtm

Thank you for working on this!

Skarlso:
@tbarizien Hey Thomas. Just checking if you did actually test with pre current release RGDs if this is a breaking change or not? Jakob had a question about that, but I don't see a followup on it. :)

tbarizien:
> @tbarizien Hey Thomas. Just checking if you did actually test with pre current release RGDs if this is a breaking change or not? Jakob had a question about that, but I don't see a followup on it. :)

I've added an automated upgrade test a few month ago in `test/upgrade/fixtures/legacy-selector`. It is passing but feel free to take a second look since I'm not that accustomed to writing such tests :)

tbarizien:
@matthchr is there anything missing before review?

Skarlso:
/lgtm

still looks good

test/upgrade/legacy_selector_compat_test.go
```diff
@@ -0,0 +1,189 @@
...
+var _ = ginkgo.Describe("Post-Upgrade Legacy Selector Compatibility", ginkgo.Ordered, func() {
+	var rgdGenBefore int64
+
+	ginkgo.BeforeAll(func() {
+		if !isPostUpgrade() {
+			ginkgo.Skip("Legacy selector compatibility checks only run in post-upgrade mode")
+		}
+	})
+
+	ginkgo.It("should have the legacy-selector RGD in Active state after upgrade", func() {
```
cheeseandcereal, line 67: `upgrade-kro` waits for Deployment `Available`, which can be true immediately while the old leader still exists before it's been updated. Probably update this test to validate rollout completion and candidate leadership before these assertions, otherwise the compatibility test can pass and mutate the RGD when it shouldn't.

tbarizien: Interesting :thinking: 
Doesn't it make more sense to fix it globally for all upgrade tests in the makefile? Is this then still in scope for this PR? 
I can fix it in this test only but it feels like writing targeted boilerplate.

cheeseandcereal: Yeah that's fair. I guess it's just more of a potential problem here, even though it is a pre-existing issue. I wouldn't be opposed to fixing it globally, but I also agree that it's not necessarily within the scope of this specific PR to solve since it is a global config/test issue.

Skarlso:
This looks good! I did verify backwards compatibility. Anyone obviously importing this will break, but I doubt that's the case. So Looks good from my side and can be merged. :) 🎉

jakobmoellerdev (approved):

PR #1234, opened 2026-04-02
KREP-021: kro CLI

jlbutler:
Supersedes #1156 (original by @NicholasBlaskey).

Expands the kro CLI proposal with lint, fmt, diff commands, ORAS-based OCI distribution, and a cluster dependency matrix.

### DIFF
--- a/docs/design/proposals/kro-cli.md
+++ b/docs/design/proposals/kro-cli.md
@@ -1,0 +1,192 @@
+# KREP-021: kro CLI
+
+## Summary
+
+The kro CLI exists today with basic validate and generate commands, all requiring a cluster connection and only available by building from source. This proposal expands the CLI into a comprehensive tool for authoring, validating, distributing, and previewing kro ResourceGraphDefinitions — offline where possible, with cluster access when needed. The CLI ships as part of kro releases.
+
+The CLI is the primitive layer for working with kro compositions. It is deterministic, scriptable, suitable for CI, and agent-friendly.
+
+## Problem Statement
+
+The current CLI is not very accessible today as it must be built from source. It should be readily available, and versioned along with the project.
+
+It is also fairly limited in scope, in contrast to broad use cases for kro and what users may need a CLI for.
+
+**Dry run**
+Updating an RGD affects all instances immediately with no preview of impact. Teams need to see what will change, both to the RGD itself and to the resources it manages, before committing.
+
+**Distribution**
+There is no standard way to package and distribute RGDs. Users resort to git repositories and kubectl apply. The Kubernetes ecosystem thrives on reusable artifacts (Helm charts, OCI images) — RGDs need the same distribution model.
+
+**Validation**
+The only way to know if an RGD is valid today is to apply it to a cluster. Errors surface at reconciliation time, not authoring time. Users need fast feedback in their local workflow and CI pipelines, without requiring a live cluster.
+
+**Authoring tools**
+The CLI does not currently offer diff, linter, or other authoring tools to help users adopt and succeed with kro.
+
+## Proposal
+
+### Overview
+
+Expand the existing `kro` CLI with new commands and improve existing ones. All changes are to the existing `cmd/kro/` codebase. The CLI is published as part of kro releases via GoReleaser.
+
+### Command Summary
+
+| Command | Description | Status |
+|---------|-------------|--------|
+| `kro validate rgd` | Validate an RGD file | Existing, updated for offline use |
+| `kro validate instance` | Validate an RGD and instance together | New |
+| `kro lint` | Check RGDs against conventions and best practices | New |
+| `kro fmt` | Normalize RGD YAML formatting | New |
+| `kro diff` | Structural diff between two RGD files | New |
+| `kro preview` | Preview changes against live cluster state | New |
+| `kro push` | Validate and push an RGD to an OCI registry | New |
+| `kro pull` | Pull an RGD from an OCI registry | New |
+| `kro install` | Pull an RGD and apply it to the cluster | New |
+| `kro registry login` | Authenticate to OCI registries | New |
+| `kro generate crd` | Generate CRD from an RGD | Existing |
+| `kro generate instance` | Generate sample instance from an RGD | Existing |
+| `kro generate diagram` | Generate HTML dependency graph | Existing |
+
+### Design Details
+
+#### Validate
+
+Update existing `validate rgd` to work offline by default. The current implementation requires a cluster connection for CRD schema resolution. The updated command resolves schemas from local CRD files or bundled Kubernetes schemas, with cluster discovery as an opt-in flag.
+
+```bash
+kro validate rgd -f rgd.yaml \
+  --crds ./crds/              \  # Optional: directory of CRD files for schema resolution
+  --kubernetes-version 1.30   \  # Optional: built-in schema version (default: latest bundled)
+  --from-cluster                 # Optional: discover CRDs and version from API server
+```
+
+New `validate instance` command validates an RGD and a sample instance together — confirming the instance conforms to the schema the RGD would generate.
+
+```bash
+kro validate instance --rgd-file rgd.yaml --instance-file instance.yaml
+```
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+  --from-cluster                 # Optional: discover CRDs and version from API server
+```
+
+New `validate instance` command validates an RGD and a sample instance together — confirming the instance conforms to the schema the RGD would generate.
+
+```bash
+kro validate instance --rgd-file rgd.yaml --instance-file instance.yaml
+```
+
+#### Lint
```
NicholasBlaskey, line 70: I think lint may be a bit premature

I wonder if we should hold off until we identify issues that are substantial and should be checked by the linter

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+#### Fmt
+
+Normalize RGD YAML. Consistent field ordering, indentation, and formatting for readability and consistency. Pure text transformation — no cluster, no schema knowledge.
+
+```bash
+kro fmt -f rgd.yaml        # in-place
+kro fmt -f rgd.yaml --check # exit non-zero if changes needed (for CI)
+```
+
+#### Diff
```
NicholasBlaskey, line 89: I'm not completely sold on diff

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+- Publishing the CLI as part of releases via GoReleaser
+- Offline validation with local CRD files and bundled Kubernetes schemas
+- Instance validation against RGD schemas
+- Lint and fmt commands
+- Structural diff between RGD files
+- Preview of changes against live cluster
+- OCI packaging and distribution via ORAS
+- Registry authentication
+
+### Not In Scope
```
NicholasBlaskey, line 178: We talked a bit about this offline but helm has a ton of global flags

we mentioned this offline a bit but helm has a ton of global flags that deal
```
Flags:
      --burst-limit int                 client-side default throttling limit (default 100)
      --debug                           enable verbose output
  -h, --help                            help for helm
      --kube-apiserver string           the address and the port for the Kubernetes API server
      --kube-as-group stringArray       group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --kube-as-user string             username to impersonate for the operation
      --kube-ca-file string             the certificate authority file for the Kubernetes API server connection
      --kube-context string             name of the kubeconfig context to use
      --kube-insecure-skip-tls-verify   if true, the Kubernetes API server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --kube-tls-server-name string     server name to use for Kubernetes API server certificate validation. If it is not provided, the hostname used to contact the server is used
      --kube-token string               bearer token used for authentication
      --kubeconfig string               path to the kubeconfig file
  -n, --namespace string                namespace scope for this request
      --qps float32                     queries per second used when communicating with the Kubernetes API, not including bursting
```

I think we don't need to support all of these initially, but ```--namespace``` would be especially important.

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+
+The CLI is the primitive layer for working with kro compositions. It is deterministic, scriptable, suitable for CI, and agent-friendly.
+
+## Problem Statement
+
+The current CLI is not very accessible today as it must be built from source. It should be readily available, and versioned along with the project.
+
+It is also fairly limited in scope, in contrast to broad use cases for kro and what users may need a CLI for.
+
+**Dry run**
```
NicholasBlaskey, line 15: dry run we will hold off on the implementation as Kro fundamentals may change soon

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+
+Examples: resource ID naming conventions, CEL expression style, required annotations, status field coverage, SimpleSchema usage patterns.
+
+```bash
+kro lint -f rgd.yaml
+```
+
+#### Fmt
+
+Normalize RGD YAML. Consistent field ordering, indentation, and formatting for readability and consistency. Pure text transformation — no cluster, no schema knowledge.
```
michaelhtm, line 82: curious to see how this will work

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+
+```bash
+kro fmt -f rgd.yaml        # in-place
+kro fmt -f rgd.yaml --check # exit non-zero if changes needed (for CI)
+```
+
+#### Diff
+
+Structural diff between two RGD files. Surfaces field additions, removals, type changes, and CEL expression changes. No cluster required.
```
michaelhtm, line 91: should diff also be used to compare against different GraphRevisions in the cluster?

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+| `kro registry login` | Authenticate to OCI registries | New |
+| `kro generate crd` | Generate CRD from an RGD | Existing |
+| `kro generate instance` | Generate sample instance from an RGD | Existing |
+| `kro generate diagram` | Generate HTML dependency graph | Existing |
+
+### Design Details
+
+#### Validate
+
+Update existing `validate rgd` to work offline by default. The current implementation requires a cluster connection for CRD schema resolution. The updated command resolves schemas from local CRD files or bundled Kubernetes schemas, with cluster discovery as an opt-in flag.
```
spattk, line 55: will we give an option to connect to cluster?

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+-   replicas: 1
++   replicas: 3
+```
+
+#### OCI Distribution
+
+RGDs are packaged as OCI artifacts and published to repositories in OCI-compliant registries. Uses ORAS for registry interaction.
+
+```bash
+kro push -f rgd.yaml registry.io/org/my-rgd:v1.0.0
```
michaelhtm, line 133: does this have to be just an RGD? could we see an RGD and instance being published?

docs/design/proposals/kro-cli.md
spattk: answered below. it's configurable.

docs/design/proposals/kro-cli.md
```diff
@@ -0,0 +1,192 @@
...
+
+Structural diff between two RGD files. Surfaces field additions, removals, type changes, and CEL expression changes. No cluster required.
+
+```bash
+kro diff -f old-rgd.yaml -f new-rgd.yaml
+```
+
+This is distinct from `preview`, which diffs against live cluster state.
+
+#### Preview
```
spattk, line 99: do we really wanna tackle this right now? changes to graph like propagation control would make some part of it non-deterministic. we should probably nail it down a bit more before going to the implementation.

docs/design/proposals/kro-cli.md
NicholasBlaskey: lets wait for a usecase on this one. 

Can be added in a future revision but not planning on adding diff immediately

docs/design/proposals/kro-cli.md
NicholasBlaskey: could potentially look into adding this as a default instance that comes with an RGD

This package and publish commands need more scoping. Will focus on offline aspects of CLI first

Skarlso:
Hey there peeps.

I would like to take over this Krep and bring it to conclusion if everyone is okay with that.

I talked with Jesse over slack and Adam as well and we have another idea for the CLI as well which would be folded into this Krep. It's enabling bootstrapping like flux or capi/capa does.

And I would like to fold that into this Krep instead of creating a new one which would also target the client anyways.

What do you all think about that?

Thanks.

Skarlso:
I talked this over with Nick and Jesse and going to take over. The format is TBD :)

Skarlso:
/close


This has been supersedes by https://github.com/kubernetes-sigs/kro/pull/1421.

PR #1241, opened 2026-04-08
Prevent deleting other RGDs resources on CRD conflict

### LINKED ISSUE #1169: Deleting a conflicted RGD stops reconciliation for the original RGD
## Summary

  When two RGDs define the same Kind, the second one is rejected (Inactive). However, deleting the **rejected RGD** causes the **original Active RGD** to stop reconciling instances.

  ## Reproduction

1. create a RGD A points to CRD A
2. create a RGD B also point to CRD A

Note B will be inactive as expected

3. delete RGD B

4. instances of CRD A will no longer reconicle

  Controller logs after deletion:
```
  DEBUG   dynamic-controller      Successfully unregistered GVR   {"gvr": "kro.run/v1alpha1/myapps"}
  INFO    rgd-controller  skipping CRD deletion because allowCRDDeletion is disabled
  DEBUG   rgd-controller  setting resourcegraphdefinition as unmanaged {"name": "myapp-advanced"}
```

  Expected Behavior

  Deleting an Inactive/conflicted RGD should have no effect on the Active RGD that owns the CRD. The original RGD should continue reconciling instances normally.

  Impact

  - Deleting rejected/conflicted RGDs breaks original RGDs
  - Restarting should fix this

ResourceGraphDefinition (Create, Update, Deletion)

NicholasBlaskey:
Fixes https://github.com/kubernetes-sigs/kro/issues/1169

Avoid calling Deregister if our RGD hasn't called register on it

Avoid deleting CRDs if our RGD does not own them

### DIFF
--- a/pkg/controller/resourcegraphdefinition/controller.go
+++ b/pkg/controller/resourcegraphdefinition/controller.go
@@ -17,6 +17,7 @@ package resourcegraphdefinition
 import (
 	"context"
 	"errors"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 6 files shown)


jakobmoellerdev (commented):
/ok-to-test
/lgtm

jakobmoellerdev (approved):

PR #1244, opened 2026-04-10
Validate we do not attempt to apply duplicate k8s resources

NicholasBlaskey:
Addresses 1/2 of this issue https://github.com/kubernetes-sigs/kro/issues/957
```
    different graph nodes in the same instance produce identical resources
    different collection items from the same collection producing identical resources
```

If we define an RGD with duplicate resources 
```
  resources:
    - id: configmap1
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: cm1
        data:
          element: "ok"
    - id: configmap2
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: cm1
        data:
          element: "ok"
```

we previously would get an infinite reconcile as we fought out the resourceID label

Now throw a validation error

### DIFF
--- a/pkg/controller/instance/resources.go
+++ b/pkg/controller/instance/resources.go
@@ -105,6 +105,9 @@ func (c *Controller) buildApplyInputs(rcx *ReconcileContext) (*reconcileResult, 
 
 	r.supersetPatch, err = r.applier.Project(resources)
 	if err != nil {
+		if errors.Is(err, applyset.ErrDuplicateResource) {
+			return nil, err
+		}
 		return nil, rcx.delayedRequeue(fmt.Errorf("project failed: %w", err))
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/instance/resources.go
```diff
@@ -101,8 +102,11 @@ func (c *Controller) buildApplyInputs(rcx *ReconcileContext) (*reconcileResult,
 		}
 		r.unresolvedErr = err
 	}
-	r.resources = resources
+	if err := duplicateResourceCheck(resources); err != nil {
```
jakobmoellerdev: agreed this is worth it to have, although to me this feels like a layer issue. maybe this should be inside the applier project preparation. WDYT?

NicholasBlaskey: updated, I do think it works out cleaner.

We do have to check the error to avoid requeue but other than is better

jakobmoellerdev (commented):
/ok-to-test
/lgtm

jakobmoellerdev (approved):

PR #1248, opened 2026-04-13
Document forEach dependent iterations pattern with map().flatten()

NicholasBlaskey:
Add some docs calling out a pattern for using nested iterators


jakobmoellerdev (approved):
/ok-to-test
/lgtm

Issue #1247, opened 2026-04-12
forEach resource cleanup regression in v0.9.1: orphaned resources when dependent forEach iterates empty array


vterdunov:
**Observed Behavior**:

When deleting a custom resource instance backed by an RGD with two `forEach` resources — where resource A iterates an empty array and resource B depends on A via `size(a)` — the child resources created by B are not cleaned up. The ConfigMap created by resource B remains present indefinitely (observed for 120s+) after the instance is deleted.

This occurs on v0.9.1. On v0.9.0, cleanup completes correctly within seconds.

**Expected Behavior**:

Deleting the instance should clean up all child resources, including those created by resource B, regardless of whether resource A produced zero instances.

**Reproduction Steps**:

1. Apply the ClusterRole:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: kro:controller:platform-deletetest
  labels:
    rbac.kro.run/aggregate-to-controller: "true"
rules:
  - apiGroups: ["platform.internal"]
    resources: ["deletetests", "deletetests/status", "deletetests/finalizers"]
    verbs: ["*"]
  - apiGroups: [""]
    resources: ["configmaps"]
    verbs: ["*"]
```

2. Apply the RGD and wait for it to become `Active`:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: deletetest
spec:
  schema:
    group: platform.internal
    apiVersion: v1alpha1
    kind: DeleteTest

    types:
      Entry:
        id: string

    spec:
      preDeploy: "[]Entry"
      items: "[]Entry"

  resources:
    - id: a
      forEach:
        - step: ${schema.spec.preDeploy}
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${step.id}

    - id: b
      forEach:
        - el: ${schema.spec.items}
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${el.id}
          annotations:
            dep: ${string(size(a))}
```

3. Create the instance:

```yaml
apiVersion: platform.internal/v1alpha1
kind: DeleteTest
metadata:
  name: test-delete
spec:
  preDeploy: []
  items:
    - id: x
```

4. Verify ConfigMap `x` is created with annotation `dep: "0"`.

5. Delete the instance:

```
kubectl delete deletetest test-delete
```

6. Check whether ConfigMap `x` is cleaned up:
[...]

PR #1249, opened 2026-04-13
Fix resource leak during deletion when collections reference empty collection

NicholasBlaskey:
Resolved issue here
https://github.com/kubernetes-sigs/kro/issues/1247

Cause seems to be when we updated identity paths we missed a node type
https://github.com/kubernetes-sigs/kro/pull/1232/changes#diff-515649dd828384245d841ca308c508cb1d69aa0f743736c40c8f7ddf601d2b63R60-R65

Make this be an override so new node types get this for free

### DIFF
--- a/pkg/runtime/node.go
+++ b/pkg/runtime/node.go
@@ -56,15 +56,23 @@ type Node struct {
 	resourceSchema *spec.Schema
 }
 
-// identityPaths are the template field path prefixes resolved in identity mode,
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (commented):
Good catch!

/ok-to-test

jakobmoellerdev (commented):
/lgtm

jakobmoellerdev (approved):

PR #1251, opened 2026-04-15
feat: add metadata propagation for labels & annotations

### LINKED ISSUE #652: Support downstream propagation of labels/annotations
### Feature Description

**Problem Statement**:
In order to properly manage resources created by KRO in frameworks like ArgoCD we need to be able to provide all downstream resources with specific annotations/labels. 

Right now unless done manually per each resource there is no way to achiever that

**Proposed Solution**:
From talking to @jakobmoellerdev a good approach could be to expose in the controller capabilities of replicating towards the RGD instances whitelisted labels and/or annotations. 

```
... think about propagating any metadata such as labels or annotations via selector based on a Spec field in the RGD. 

It should likely be an opt-in thing and by default nothing would be propagated
```

**Alternatives Considered**:
The only alternative to having such solution is manually adding the reference to annotations/labels in the resources created by RGD

**Additional Context**:
Originally coming from various discussions around https://github.com/kro-run/kro/pull/366

**Contribution to developing**: I would be more than happy to kick off some implementation of a feature like that if we would agree on specifics how to best implement such. 

KatharinaSick:
This adds support for propagating allowlisted labels & annotations from RGIs down to their resources.

Closes #652 

---

_AI Disclaimer (because it was discussed in the linked issue): I used AI to understand the codebase better, format my comments and to write tests. The "main" code has been written by me as it's important to me to understand how things are working before contributing (you could just hand the issue over to Claude/Copilot/whatever yourself so I don't see value in that 😅)_


pkg/controller/instance/resources.go
```diff
@@ -359,6 +361,10 @@ func (c *Controller) processRegularNode(
 		node.SetObserved([]*unstructured.Unstructured{current})
 	}
 
+	// Add allowlisted labels & annotations from the instance first, so that
+	// kro's own decorator labels always take precedence on key conflicts.
+	c.applyMetadataPropagation(rcx, desired)
```
KatharinaSick, line 366: This is an interesting edge case. If somebody wants to propagate `kro.run` labels down to all resources, they'll be overwritten by line 369 again. As those labels are improtant for kro to work properly, I decided to go for this order

pkg/dynamiccontroller/dynamic_controller.go
```diff
@@ -445,7 +446,8 @@ func (dc *DynamicController) enqueueFromInformer(parentGVR schema.GroupVersionRe
 			// This covers an edge case where the user only updates the reconcile annotation from disabled to enabled,
 			// which will trigger this func, but the generation won't change since it's a metadata-only update.
 			// We want to make sure this transition still triggers a reconciliation.
-			if !reconcileEnabledInUpdate(oldMeta, newMeta) {
+			// Similarly, enqueue if labels or annotations changed so that metadata propagation is applied.
+			if !reconcileEnabledInUpdate(oldMeta, newMeta) && !metadataNeedsReconcile(oldMeta, newMeta) {
```
KatharinaSick, line 450: Added this because otherwise labels/annotations would not get propagated down if added to the RGI. It leads to more reconciles though. I hope this is ok.

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -35,6 +35,11 @@ type ResourceGraphDefinitionSpec struct {
 	//
 	// +kubebuilder:validation:Optional
 	Resources []*Resource `json:"resources,omitempty"`
+	// MetadataPropagation defines which labels and annotations from an instance are
+	// propagated to all child resources. If not set, nothing is propagated.
+	//
+	// +kubebuilder:validation:Optional
+	MetadataPropagation *MetadataPropagation `json:"metadataPropagation,omitempty"`
```
jakobmoellerdev, line 42: was there a specific reason you chose to go for a CRD field in favor of an instance level annotation this time? the reason im asking is because now the propagation can not be controlled by instance but is for all instances. this may or may not be a good thing.

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -120,6 +125,19 @@ type CRDMetadata struct {
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
+// MetadataPropagation defines which labels and annotations from an RGI should be
+// propagated to all child resources created from that instance.
+type MetadataPropagation struct {
```
jakobmoellerdev, line 130: I believe in the very distant future, there is the chance we dont just want to propagate metadata but want to be more flexible. Im wondering if it makes sense that instead of hardcoding this, we expose the instance node and are able to code against the instance metadata node.

Think of something like this:

```yaml
resources:
  - id: job
    readyWhen:
      - ${job.status.completionTime != null}
    template:
      apiVersion: batch/v1
      kind: Job
      metadata:
        name: ${schema.metadata.name}-job
        labels: 
          propagated: "${instance.metadata.labels['mylabel']}"
```

This would be much more explicit and correlate more to the explicit dependency modelling that is prevalent in KRO. It also allows mutation of the labels

We dont have to go this way immediately but im wondering how the design decision was made.

api/v1alpha1/resourcegraphdefinition_types.go
KatharinaSick: Honestly, I just followed the outcomes of the discussions in #652. To me, this approach makes sense because that way the platform team can choose which labels/annotations get propagated down to all resources (e.g. the ones they enforce via Kyverno policies).

But I'm also happy to add support on both levels (or ditch this PR because of the next comment :D)

api/v1alpha1/resourcegraphdefinition_types.go
KatharinaSick: Oh wow, that's actually a much cleaner approach! I'm also good to close this PR and implement your suggestion.

But if we go for this way, your concern from above will still hold true. It means, propagation can only be controlled on RGD level again (if I understood everything correctly so far)

jakobmoellerdev: I think instance level propagation is possible then because one can then control if the label should be propagated based on a Cel expression that uses the instance spec. I think we should pivot to this approach due to its flexibility but also looking for other maintainer input here!

KatharinaSick: Oh true - I haven't thought about that. That's cool! :)

-> I'll let you close the PR in case you want to wait for inputs from other maintainers.

I'm also happy to contribute the new/improved solution but will only be able to start early May as I have some deadlines coming up.

AndrewCharlesHay:
Metadata propagation is useful, but I want to flag a few security-adjacent concerns with allowing user-controlled labels/annotations to flow through to child resources.

**Main concern: no denylist for reserved / privileged keys.**

Kubernetes (and its ecosystem) treats several label and annotation prefixes as privileged or semantically meaningful — propagating them from an instance onto a child resource can trip other controllers into unintended behavior. Specifically dangerous if someone configures `labels: ["*"]` or a broad wildcard:

- `kubernetes.io/*`, `k8s.io/*` — reserved prefixes.
- `pod-security.kubernetes.io/*` — PSA enforcement levels. A user annotating their RGI with `pod-security.kubernetes.io/enforce: privileged` could have that land on a namespace-like child (if the template admits it) and downgrade security posture.
- `node-role.kubernetes.io/*`, `topology.kubernetes.io/*` — could affect scheduling / taints.
- `iam.gke.io/gcp-service-account`, `eks.amazonaws.com/role-arn`, `azure.workload.identity/client-id` — cloud-provider Workload Identity annotations. If one of these lands on a `ServiceAccount` child resource, the child picks up the cloud IAM identity. This is a privilege escalation surface — an instance author who can set annotations on an RGI + can choose which RGD to use can effectively mint tokens for a chosen cloud IAM principal.
- `kubectl.kubernetes.io/last-applied-configuration` — large opaque blobs, could bypass quotas.

Recommendations:

1. Add a hard-coded denylist of known-privileged prefixes (`kubernetes.io/`, `k8s.io/`, `pod-security.kubernetes.io/`, `iam.gke.io/`, `eks.amazonaws.com/`, `azure.workload.identity/`, `rbac.authorization.k8s.io/`) that is applied regardless of allowlist pattern. The allowlist can then be user-controlled without footguns.
2. Consider whether propagation should be per-resource-type — a label that's safe on a `ConfigMap` is not safe on a `ServiceAccount`.
3. Enforce a max count / max total size (e.g., 64 propagated labels, 16 KB total annotation bytes) to prevent an RGI author from bloating every child with thousands of inherited annotations.

**Smaller notes:**

- Wildcard matching uses `strings.HasSuffix` on `strings.TrimSuffix(pattern, "*")` — worth confirming `myorg.com/*` doesn't accidentally match `bad-myorg.com/foo`. Quick unit test with adversarial pattern/key pairs would be valuable.
- Precedence comment says "kro's own decorator labels always take precedence on key conflicts" and the call order (propagation → decorator) matches. Good. Consider locking this with a test that asserts kro labels survive when the instance tries to set the same key.
- RBAC: the controller now reads `rcx.Instance.GetLabels()` / `GetAnnotations()`, which was already readable. No RBAC expansion. Good.

jakobmoellerdev:
@AndrewCharlesHay I will warn you here that I don't want to read obviously AI-generated paragraphs that have not been condensed down by you. This is not acceptable for proper interaction with humans.

KatharinaSick:
Thanks for the input @AndrewCharlesHay - I'm happy to implement it but would first like to know if we're going to continue with this approach or pivoting to expsing the instance node ([ref](https://github.com/kubernetes-sigs/kro/pull/1251#discussion_r3095144008))

api/v1alpha1/resourcegraphdefinition_types.go
michaelhtm: I like the idea of exposing instance labels using CEL. Maybe still using `schema.metadata.labels`.

jakobmoellerdev:
closing as stale, feel free to reopen / rebase!

PR #1254, opened 2026-04-15
Add an option to disable using field selectors

NicholasBlaskey:
Some Kubernetes environments may not have access to or want to use field selectors

Add a CLI flag to allows this to be disabled.

Note explicitly avoiding add this as a Helm chart option because this feature is rare

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -94,6 +94,7 @@ func main() {
 		leaderElectionRenewDeadline   time.Duration
 		leaderElectionRetryPeriod     time.Duration
 		featureGatesFlag              string
+		useFieldSelectors             bool
 	)
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8078", "The address the metric endpoint binds to.")
@@ -170,6 +171,8 @@ func main() {
 		"Duration the acting leader will retry refreshing leadership before giving up.")
 	flag.DurationVar(&leaderElectionRetryPeriod, "leader-election-retry-period", 2*time.Second,
 		"Duration between each leader election action.")
+	flag.BoolVar(&useFieldSelectors, "use-field-selectors", true,
+		"Use field selectors when true, otherwise fall back to label selectors. Set to false for Kubernetes <= 1.30 compatibility.")
 
 	opts := zap.Options{
 		Development: true,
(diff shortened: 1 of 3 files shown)


jakobmoellerdev (commented):
do we really wanna support clusters below 1.30?

NicholasBlaskey:
> do we really wanna support clusters below 1.30?

EKS supports 1.30 in extended support until July.

I agree as an open source project we probably do not want to officially support end of life Kubernetes versions or commit resources to validating Kro works on EOL Kubernetes versions, but I think in this case since Kro does not function at all on 1.30 and the workaround is relatively simple it could provide value for users.

jakobmoellerdev:
I believe in Open Source this becomes an issue because officially K8s 1.32 is already EOL. It is very hard to accept in the project without us having an extended support lifecycle as well. We should probably add one as a proposal so we can get your demands met in upstream. WDYT?

NicholasBlaskey:
Yeah agree on needing a proposal here long term. I know GKE and other providers also have their own form of long term support so I would be curious what routes other projects have taken.

This PR is not explicitly adding support for an EOL version, just adding an option to disable a mechanism used in the feature and fallback to another one. Though I do agree there is not much value in this feature beyond that use case.

cmd/controller/main.go
```diff
@@ -94,6 +94,7 @@ func main() {
 		leaderElectionRenewDeadline   time.Duration
 		leaderElectionRetryPeriod     time.Duration
 		featureGatesFlag              string
+		useFieldSelectors             bool
```
michaelhtm, line 97: Can we expose this in the helm chart as well?

AndrewCharlesHay:
Flagging a trust-boundary shift worth calling out in review:

Field selectors on `spec.snapshot.name` are server-validated against the actual object field. Labels are user-settable. In the `--use-field-selectors=false` path, `listGraphRevisions` now trusts `kro.run/resourcegraphdefinition-name` to identify revisions for an RGD.

Confused-deputy shape: anyone with create/update on `GraphRevision` (normally just kro's SA, but easy to over-grant in shared clusters) could craft a revision with that label pointing at a victim RGD. Depending on what the orphan-adoption path does with returned items, the controller could end up claiming (`ownerReference`) a spec chosen by the label writer under the victim RGD's identity.

Defense-in-depth suggestions for the label-mode path:
1. After the `List`, verify each item's `spec.snapshot.name == rgd.Name` before treating it as adoptable — makes label-mode parity with field-mode in terms of what actually gets matched.
2. Expand the flag help text to note that label mode widens the trust boundary vs. the field-selector default, and is intended for clusters on k8s ≤ 1.30 that can't use CRD selectable fields.
3. Consider a `klog.Warningf` at startup when `UseFieldSelectors == false` so it's visible in operator logs (not just config).

None of this blocks the feature — the default stays `true` so existing users are fine. Just want the fallback path to be deliberate rather than a silent downgrade.

NicholasBlaskey:
closing due to this no longer being needed

PR #1257, opened 2026-04-16
feat(runtime): add conditional dependency filtering with caching

NicholasBlaskey:
Filter out unneeded dependencies with CEL partial evaluation

This is more for correctness than an optimization. Consider the expression, 

`'schema.spec.useCache ? cache : database'`, we should only ever have a depencey on cache or database not both as it works today.
 
Based off ideas of this KREP very loosely 

https://github.com/kubernetes-sigs/kro/pull/1125

Limiting to schema allows us to avoid needing to rewrite how we build graphs. The general solution can not be solved easily with the current architecture. Processing nodes bottom up does not allow us to easily optimize out nodes.

### DIFF
--- a/pkg/graph/graph.go
+++ b/pkg/graph/graph.go
@@ -15,6 +15,8 @@
 package graph
 
 import (
+	"sync"
+
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/kube-openapi/pkg/validation/spec"
 
@@ -47,4 +49,9 @@ type Graph struct {
 	// Includes resource schemas (keyed by resource ID) and the instance schema
 	// (keyed by InstanceNodeID). Used at runtime for schema-aware CEL value conversion.
 	ResourceSchemas map[string]*spec.Schema
+
+	// SchemaHashToProcessedDeps caches dependency filtering results.
+	// Key: "specHash:nodeID"
+	// Value: []string (filtered dependencies for that node with that spec)
+	SchemaHashToProcessedDeps sync.Map
 }
--- a/pkg/graph/node.go
+++ b/pkg/graph/node.go
@@ -117,6 +117,22 @@ type ForEachDimension struct {
 // It contains the template, variables, and conditions for a resource.
 // No CRD/schema references are kept here - schemas are only used
 // during build for CEL type-checking.
+//
+// EXPRESSION TYPES: Node contains 4 types of CEL expressions:
+//   - Variables: Template fields (e.g., ${db.host})
+//   - IncludeWhen: Conditions for whether to create resource
+//   - ReadyWhen: Conditions for resource readiness
+//   - ForEach: Iterator expressions for collections
+//
+// When adding new expression types, update these locations:
+//  1. pkg/graph/builder.go: extractTemplateDependencies, extractForEachDependencies, extractConditionDependencies
+//  2. pkg/graph/builder.go: compileNode (expression compilation)
+//  3. pkg/runtime/filter_dependency.go: filterDependencies (runtime dep filtering)
+//  4. pkg/runtime/runtime.go: Phase 3 of FromGraph (expression wiring)
+//  5. pkg/runtime/node_resolve.go: ExpandCollection (if relevant to collections)
+//
+// Note: The need to update expressions across multiple fields is becoming repetitive.
+// Future refactor could consolidate into a single Expressions list that each field references.
 type Node struct {
 	// Meta contains identification metadata (ID, Type, GVR, etc.)
 	Meta NodeMeta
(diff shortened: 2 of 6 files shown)


pkg/runtime/filter_dependency.go
```diff
@@ -0,0 +1,292 @@
...
+		exprs = append(exprs, expr.Original)
+	}
+	for _, iter := range node.ForEach {
+		exprs = append(exprs, iter.Expression.Original)
+	}
+	return exprs
+}
+
+// shouldOptimizeDependencies returns true if the node has non-schema dependencies and expressions.
+func shouldOptimizeDependencies(node *graph.Node) bool {
```
michaelhtm, line 175: should some of the preprocessing happen at compile time? (RGD reconciler)
Eg. whether or not we should optimize dependencies

NicholasBlaskey: Yeah open to moving it, but since it is just reading some resources I don't think it is a huge performance gain

We do have to wait for runtime in order to do the partial evaluation so we know the schema values

pkg/graph/node.go
```diff
@@ -141,6 +157,46 @@ type Node struct {
 	ForEach []ForEachDimension
 }
 
+// cloneExpression creates a deep copy of an Expression.
+func cloneExpression(expr *krocel.Expression) *krocel.Expression {
```
jakobmoellerdev, line 161: nit: I believe you can generate deepcopy functions with k8s for expression to not have to do this manually.

jakobmoellerdev:
Its still a large change even with this mode and I feel like this is a stop-gap. Which current use cases are blocked by this? Since currently the node state can be skipped, would there be a large functional impact if we leave code as is until we have a permanent solution in place?

NicholasBlaskey:
> Its still a large change even with this mode and I feel like this is a stop-gap. Which current use cases are blocked by this? Since currently the node state can be skipped,  would there be a large functional impact if we leave code as is until we have a permanent solution in place?

From the krep this is the functional impact. 

> ```yaml
> spec:
>   endpoint: ${schema.spec.useCache
>       ? cache.status.endpoint
>       : database.status.endpoint}
> ```
> 
> If `cache` is guarded by `includeWhen: ${schema.spec.useCache}` and
> `schema.spec.useCache` is `false`, the node above should still be valid. The
> expression does not need `cache` in that case. Under the current model, the
> compiler still infers dependencies on both `cache` and `database`, and the
> runtime treats both as required. The dependent node is skipped even though the
> `database` branch was sufficient.


What do you mean by currently the node state being skipped here? 
>Since currently the node state can be skipped,

jakobmoellerdev:
In the current model, even though the dependency would be inferred (as stated correctly), it would get evaluated to false from the spec immediately and have no obvious impact beside a topology change without impact. I think this is an edge case so I would like to see a real use case that actually impacts people significantly.

NicholasBlaskey:
> In the current model, even though the dependency would be inferred (as stated correctly), it would get evaluated to false from the spec immediately and have no obvious impact beside a topology change without impact. I think this is an edge case so I would like to see a real use case that actually impacts people significantly.

In this example resource if `api` has 

```spec:
  endpoint: ${schema.spec.useCache
      ? cache.status.endpoint
      : database.status.endpoint}
``` 

Then if ```schema.spec.useCache=true``` and ```cache``` has an ```includeWhen``` that is false then we would also exclude `api` due to contagious exclusion and the dependencies existing.

I do think we have gotten this feedback that this is really surprising behavior if not being an outright bug report.

jakobmoellerdev:
> In this example resource if `api` has
> 
> ```
>   endpoint: ${schema.spec.useCache
>       ? cache.status.endpoint
>       : database.status.endpoint}
> ```
> 
> Then if `schema.spec.useCache=true` and `cache` has an `includeWhen` that is false then we would also exclude `api` due to contagious exclusion and the dependencies existing.
> 
> I do think we have gotten this feedback that this is really surprising behavior if not being an outright bug report.

Why is this surprising behavior? if useCache is true then it should obviously only respect the true branch and then wait if cache doesn not exist. Are you sure this is the correct example here?

NicholasBlaskey:
> > In this example resource if `api` has
> > ```
> >   endpoint: ${schema.spec.useCache
> >       ? cache.status.endpoint
> >       : database.status.endpoint}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then if `schema.spec.useCache=true` and `cache` has an `includeWhen` that is false then we would also exclude `api` due to contagious exclusion and the dependencies existing.
> > I do think we have gotten this feedback that this is really surprising behavior if not being an outright bug report.
> 
> Why is this surprising behavior? if useCache is true then it should obviously only respect the true branch and then wait if cache doesn not exist. Are you sure this is the correct example here?

The opposite sorry. 

This situation
```useCache=true``` 
```database, includeWhen=false```
```cache, includeWhen=true```

`api` has a dependency on both database and cache and `api` would be excluded due to contagious exclusion.

File: pkg/graph/graph.go [hunk 3]
```diff
@@ -47,4 +49,9 @@ type Graph struct {
 	// Includes resource schemas (keyed by resource ID) and the instance schema
 	// (keyed by InstanceNodeID). Used at runtime for schema-aware CEL value conversion.
 	ResourceSchemas map[string]*spec.Schema
+
+	// SchemaHashToProcessedDeps caches dependency filtering results.
+	// Key: "specHash:nodeID"
+	// Value: []string (filtered dependencies for that node with that spec)
+	SchemaHashToProcessedDeps sync.Map
```
jakobmoellerdev, line 56: I really think we should avoid schema hashes at all costs at the moment due to schema pointers and their lookup not being reliable enough.

PR #1259, opened 2026-04-17
KREP-024: Instance Propagation Control

michaelhtm:
This KREP discusses the implementation plan for instance propagation
control accross GraphRevisions.
    
builds on:
* https://github.com/kubernetes-sigs/kro/pull/861
* https://github.com/kubernetes-sigs/kro/pull/1174

### DIFF
--- a/docs/design/proposals/instance-propagation-control.md
+++ b/docs/design/proposals/instance-propagation-control.md
@@ -1,0 +1,467 @@
+# KREP: Instance Propagation Control
+
+## Summary
+
+This proposal introduces `propagateWhen`, an RGD-level mechanism to control
+**when** instances of a ResourceGraphDefinition are allowed to reconcile against
+a new GraphRevision. When an RGD spec changes and a new revision is issued,
+instead of all instances immediately reconciling, CEL expressions are evaluated
+across the instance population to determine which instances may proceed. This
+enables gradual rollout, canary validation, and blast radius control.
+
+## Motivation
+
+When an RGD spec changes today, kro issues a new GraphRevision and immediately
+re-enqueues **all** instances via `DynamicController.Register()`. Every instance
+unconditionally resolves the latest compiled graph and reconciles. There is no
+mechanism to gate, batch, or order how instances adopt a new revision.
+
+This is dangerous for platform teams managing many instances of a single RGD. A
+bad change to an `Application` RGD that manages 200 instances immediately
+affects all 200. There is no way to validate the change on a subset first, no
+way to gate propagation on health signals, and no way to slow down the rate of
+change.
+
+### Concrete example
+
+A platform team runs an RGD defining a `WebApp` custom resource. 200 instances
+exist across the cluster. The team updates the RGD to change the deployment
+strategy from RollingUpdate to Recreate.
+
+Without propagation control: all 200 instances immediately reconcile. Every
+WebApp's deployment is replaced simultaneously, causing cluster-wide downtime.
+
+With propagation control: the team configures `propagateWhen` to require that
+50% of already-propagated instances must be healthy before the next batch
+proceeds. The first instance updates, proves healthy, then the next batch
+follows. If the first instance fails its readiness checks, propagation halts
+and the team is alerted.
+
+### Relationship to KREP-006
+
+[KREP-006](https://github.com/kubernetes-sigs/kro/pull/861) introduced the
+concept of `propagateWhen` broadly — covering both intra-instance (collection
+resources within a graph) and cross-instance (instances of an RGD) propagation.
+This proposal focuses exclusively on the **cross-instance** case and defines the
+concrete implementation primitives needed to achieve it.
+
+### Relationship to KREP-013
+
+[KREP-013](graph-revisions.md) introduced `GraphRevision` as an immutable
+snapshot of a compiled RGD spec with monotonically increasing revision numbers.
+This proposal builds directly on that foundation — `GraphRevision` provides the
+stable revision identity that propagation control gates against.
+
+## Proposal
+
+### Overview
+
+The core design introduces five primitives:
+
+1. **Propagation Orchestrator** — a sub-step of the RGD controller that
+   evaluates `propagateWhen` and selectively enqueues instances.
+2. **RevisionReconciled Condition** — an instance-level condition tracking which
+   GraphRevision an instance last successfully reconciled against.
+3. **Propagated Condition** — an RGD-level condition reporting propagation
+   progress.
+4. **Propagation CEL Environment** — a cross-instance CEL context with full
+   object access and helper methods.
+5. **Dynamic Parent Informer** — full object cache for instance CRDs, enabling
+   the orchestrator to read instance state without API calls.
+
+### Design details
+
+#### 1. All Instances Target Latest — No Revision Pinning
+
+Instances always move towards the latest revision. The orchestrator controls
+*when* an instance is re-enqueued, not *which* revision it targets. Once an
+instance reconciles, it gets the latest.
+
+The gate is: "should this instance be re-enqueued yet?" not "which revision
+should it use?"
+
+This simplifies the model: there is exactly one target state (latest revision),
+and the orchestrator decides the pace at which instances converge to it.
+
+#### 2. Propagation Orchestrator
+
+The propagation orchestrator runs as a sub-step of the RGD controller
+reconciliation, after `ensureServingState()` succeeds.
+
+Rationale for this placement:
+- The RGD controller already owns the "activate revision for serving" lifecycle.
+- It has access to the compiled graph and revision number.
+- A single point of coordination avoids consistency issues (multiple instances
+  independently deciding they're in the same batch).
+- It reads instance state from the parent informer cache (O(N) in-memory, no
+  API calls).
+
+The orchestrator's algorithm:
+
+```
+1. Read all instances from the parent informer cache (full objects)
+2. For each instance, read the RevisionReconciled condition
+3. Determine which instances are at the latest revision vs pending
+4. For each pending instance, evaluate propagateWhen CEL expressions
+5. Instances where all expressions return true → enqueue for reconciliation
+6. Update the RGD's Propagated condition with progress
+7. If propagation is still in-flight → return RequeueAfter
+```
+
+#### 3. Instance Propagation Gating
+
+When `propagateWhen` is configured, instances must not blindly reconcile against
+the latest revision. Two scenarios need distinct handling:
+
+**Scenario A — RGD spec changes (new GraphRevision issued):**
+
+The orchestrator decides which instances advance. Only cleared instances are
+enqueued. Un-cleared instances are not enqueued and do not reconcile against
+the new revision.
+
+**Scenario B — Individual instance spec changes (user updates their instance):**
+
+A user may update their instance directly (e.g., changing a field in their
+Application CR). In this case, the instance should reconcile using its
+**current** GraphRevision — not the latest — unless the orchestrator has already
+cleared it to propagate.
+
+Resolution: The propagation registry (in-memory, maintained by the orchestrator)
+tracks which instances are cleared to propagate. The instance controller
+consults this registry in `resolveCompiledGraph()`:
+
+- If the instance is marked "cleared" in the registry → use latest revision
+- If the instance is NOT cleared → use the revision from its
+  RevisionReconciled condition (stay on current revision while reconciling the
+  user's spec change)
+
+This keeps the instance controller stateless with respect to propagation — it
+delegates the decision to the registry.
+
+#### 4. RevisionReconciled Condition
+
+A new condition on instance status:
+
+```yaml
+status:
+  conditions:
+    - type: "RevisionReconciled"
+      status: "True"
+      reason: "7"                          # revision number
+      lastTransitionTime: "2026-04-30T12:00:00Z"
+```
+
+Semantics:
+- **type**: `RevisionReconciled`
+- **status**: `True` when the instance has successfully reconciled at the noted
+  revision (state reaches Active)
+- **reason**: The revision number the instance reconciled against
+- **lastTransitionTime**: When this revision was first successfully reconciled
+
+The instance controller sets this condition at the end of a successful
+reconciliation when the instance reaches Active state.
+
+Design considerations:
+- `lastTransitionTime` enables future time-based propagation rate control (e.g.,
+  "propagate 5 instances every 1 hour" by checking transition timestamps).
+- `Ready` condition remains for user-facing readiness; `RevisionReconciled` is
+  kro-internal machinery.
+- The reason field as revision number allows the orchestrator to determine
+  revision currency by comparing against the latest issued revision.
+
+#### 5. Propagated Condition on RGD
+
+A new condition on the RGD status reporting propagation progress:
+
+```yaml
+status:
+  conditions:
+    - type: "Propagated"
+      status: "True"         # All instances have propagated
+      reason: "AllInstancesUpdated"
+      message: "15/15 instances reconciled at revision 7"
+```
+
+States:
+- `True` — all instances have reconciled at the latest revision
+- `False` — propagation is in-flight (some instances pending)
+- `Unknown` — error evaluating propagateWhen expressions
+
+When `propagateWhen` is not configured, this condition is not set (preserving
+existing behavior).
+
+#### 6. Dynamic Parent Informer
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+instances are rolled forward to the latest Active revision according to two
+controls:
+
+1. **`rate`** — controls the pace of propagation: `all` (current behavior),
+   `linear(N)`, or `exponential(N)`.
+2. **`when`** — a list of CEL expressions that gates batch release based on
+   aggregate instance health.
+
+Instances are sorted using a fixed default ordering (Ready first, oldest first,
+name tiebreaker). Custom instance ordering is deferred to a future RGD-level
```
cheeseandcereal, line 74: Is this deterministic? What happens if new instances are created in the middle of a rollout? etc.

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+  target revision directly. The old rollout's progress is not lost — instances
+  that already reached an intermediate revision are counted as "not on target"
+  and enter the next rollout's pending pool.
+
+This is consistent with Kubernetes Deployment behavior: a new rollout supersedes
+the old one, and the system converges toward the latest desired state.
+
+### Instance ordering
+
+Instances are ordered using a fixed default sort:
```
NicholasBlaskey, line 473: how does adding a new instance play into this sort order?

docs/design/proposals/instance-propagation-control.md
michaelhtm: New instances created in the middle of a rollout automatically get the latest revision. Since all instances are converging towards latest, this made sense to me. There's an open question about it at the end of the doc

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+#### Examples
+
+```yaml
+# Exponential rollout (1, 2, 4, 8, 16...)
+propagate:
+  rate: "exponential(2)"
+
+# Linear batches of 10, gated on propagated instance health
+propagate:
+  rate: "linear(10)"
```
NicholasBlaskey, line 193: is this suppose to be a cel expression?

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+instances.total         int    // total instance count
+
+propagated.ready         int    // count of instances on target revision with Ready condition True
+propagated.notReady      int    // count of instances on target revision with Ready condition False
+propagated.readyUnknown  int    // count of instances on target revision with Ready condition Unknown
+propagated.total         int    // total instances on target revision
+```
+
+Two approaches for maintaining aggregate state are under consideration (see
+design question: "Should aggregate counters use incremental updates or per-cycle
```
spattk, line 179: not very clear of what `per-cycle scans` mean here. can you elaborate a bit?

docs/design/proposals/instance-propagation-control.md
michaelhtm: for now, i thought of it as just keywords supported by kro

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+#### Why a struct instead of a CEL expression list?
+
+KREP-006 proposed `propagateWhen` as a list of CEL expression strings. This KREP
+uses a struct with separate `rate` and `when` fields instead. The rationale:
+
+- **Rate is not a predicate.** "Release the next exponential batch" is a
+  scheduling decision, not a boolean condition. Encoding it as a CEL expression
+  (e.g., `exponential(index, propagated.ready)`) forces per-instance
+  evaluation context (`index`) that adds complexity without added expressiveness.
+  Keyword-based rates are simpler to validate, document, and implement.
```
cheeseandcereal, line 154: Does this lock us out of custom rate/release algorithms? Is that something we accept? (i.e. you are stuck with only a pre-defined "linear" and "exponential" mode without any way to define your own, at least not without doing something like layering RGDs, etc).

docs/design/proposals/instance-propagation-control.md
michaelhtm: responded here: https://github.com/kubernetes-sigs/kro/pull/1259/changes#r3103017575

New instances are considered "propagated", so they don't need to be added to the sort order

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+scans?"). Regardless of approach, the propagation flow diagram guarantees
+`instances.total > 0` before any `when` expressions are evaluated (empty
+instance lists exit early as "Propagated").
+
+#### Examples
+
+```yaml
+# Exponential rollout (1, 2, 4, 8, 16...)
+propagate:
+  rate: "exponential(2)"
```
cheeseandcereal, line 189: How does this work without a 'when', wouldn't propagation be effectively instant since the rollout would always immediately continue between waves?

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+                               │
+                               ▼
+                ╔══════════════════════════╗
+                ║  Instances empty?        ║───Yes──▶ Propagated ✓
+                ╚══════════════════════════╝
+                               │ No
+                               ▼
+                ┌──────────────────────────┐
+                │  Load previous           │
+                │  PropagationState        │
```
cheeseandcereal, line 277: What is 'previous PropagationState'? Are we saving interim state somewhere outside of memory (besides just instance current revision, which already exists)

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+    // Supported values:
+    //   "all"              — all eligible instances propagate at once
+    //   "linear(N)"        — fixed batches of N instances
+    //   "exponential(N)"   — batches of 1, N, N^2, N^3...
+    //
+    // The argument N is a scaling factor:
+    //   "exponential(2)" produces batches of 1, 2, 4, 8, 16...
+    //   "exponential(3)" produces batches of 1, 3, 9, 27...
+    //   "linear(5)" produces batches of 5, 5, 5...
+    Rate string `json:"rate"`
```
NicholasBlaskey, line 118: is Rate even necessary? 

Or can we just control everything through propagation conditions?

docs/design/proposals/instance-propagation-control.md
michaelhtm: The RGD reconciler will be requeuing, and waiting for propagation to end. For every reconciliation cycle, we may need to scan the GVR cache for the up-to-date status of all instances in the cache

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+                               │ No
+                               ▼
+                      RequeueAfter (backoff)
+```
+
+#### Propagation requeue interval
+
+When propagation is in progress, the RGD controller requeues itself to evaluate
+the next cycle. The requeue interval uses a backoff: it starts at
+`--rgd-progress-requeue-delay` (default: 3s) and increases when the `when` gate
```
cheeseandcereal, line 362: Is there some way we can detect progress without having to re-queue. If something changes on an instance, can we propagate it to re-evaluating at the RGD level? This would allow for full event-driven progress, and only have to rely on a requeue poll as a fall-back.

With this, we are fundamentally limited to polling.

docs/design/proposals/instance-propagation-control.md
michaelhtm: Without when, the controller would just enqueue the instances exponentially, instead of all at once. 
There may be use cases where the user just doesn't want their dynamic controller to hammer their API server, and they want propagation to happen every 3 seconds for example

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+                ║  All propagated?         ║───Yes──▶ done
+                ╚══════════════════════════╝
+                               │ No
+                               ▼
+                      RequeueAfter (backoff)
+```
+
+#### Propagation requeue interval
+
+When propagation is in progress, the RGD controller requeues itself to evaluate
```
spattk, line 360: what happens let's say an author creates two GR in the following scenario
1. GR1 is currently propagating with it's condition "when 60% instances are ready" is successful.
2. GR2 should proceed when 80% of instances are ready.

There is a short window of time when both of the propagation are ready to execute. how are we controlling the order of in-flight propagations?

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+on each requeue cycle to compute counters fresh. O(N) per cycle, O(M × N) over
+M requeue cycles. Trade-off: no coupling, but more work per cycle at high
+instance counts.
+
+### New RGD condition: `Propagated`
+
+A new informational condition `Propagated` is added to the RGD:
+
+- **True**: All instances are on the latest Active revision.
+- **Unknown**: Propagation is in progress (with message showing progress, e.g.,
```
NicholasBlaskey, line 541: more a question, is unknown the right status here? I thought unknown was used for more error cases and not typical lifecycles

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+
+This label is written by the instance controller in the existing
+`applyManagedFinalizerAndLabels()` SSA patch — no additional API call. It is
+readable from the parent informer cache, which the RGD controller uses to
+determine each instance's current revision during propagation planning.
+
+### Propagation state: in-memory map
+
+Propagation assignments (which instance should use which revision) are stored in
+the revisions registry as an in-memory map alongside the existing per-RGD
```
cheeseandcereal, line 386: Is the generation of this deterministic? How do we handle if another controller takes over mid-prapogation, or another instance joins mid-propagation, etc?

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+- **True**: All instances are on the latest Active revision.
+- **Unknown**: Propagation is in progress (with message showing progress, e.g.,
+  "42/100 instances on latest revision").
+- **False**: Propagation failed (e.g., CEL expression error).
+
+`Propagated` is NOT a dependent of `Ready`. It does not affect the RGD's Active
+state. It is only set when `propagate` is configured and the feature gate is
+enabled.
+
+### Instance count on GraphRevisions
```
NicholasBlaskey, line 549: this seems like a nice to have but might not be essential for a first release for propagation control?

docs/design/proposals/instance-propagation-control.md
michaelhtm: This is loaded from in memory, it mainly tells us which instances did we enqueue in the previous reconcile

docs/design/proposals/instance-propagation-control.md
michaelhtm: Propagation conditions?
like the `when` expressions?

docs/design/proposals/instance-propagation-control.md
michaelhtm: All instances are propagating towards latest all the time. If a new Revision is created, the older propagation stops and the new one begins

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+The restart reconstruction window is handled naturally: instances that were
+released but haven't reconciled yet still carry their old `kro.run/revision`
+label. Post-restart, the planner sees them as pending (not on target, not in the
+Released map). They continue reconciling against their current revision — which
+is correct, since the resolver's fallback for an unreleased instance with a
+non-zero `currentRevision` is "stay on it" — until the planner re-releases them
+in a subsequent batch. Batch boundaries are not preserved across restarts: an
+instance that was in batch 1 pre-restart may end up in a different batch
+post-restart if instance state (e.g., Ready status) has changed. This is
+acceptable because the planner is stateless and the rollout converges to the
```
cheeseandcereal, line 449: "because the planner is stateless" isn't really a valid reason to explain why a behavior is "acceptable".

I wonder if we should do some thinking about how custom ordering is going to work, since that will definitely require state, and will help be more determinstic about the rollout. Maybe we could piggyback on some fundamentals on that solution to ensure we generate state about a rollout once, save it, and now have deterministic rollouts regardless of restarts.

docs/design/proposals/instance-propagation-control.md
NicholasBlaskey, line 189: I wonder if we can pick a sensible default for most users

What is the use case for linear over exponential?

docs/design/proposals/instance-propagation-control.md
michaelhtm: you're right, i'll switch False and Unknown

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+
+### Instance ordering
+
+Instances are ordered using a fixed default sort:
+
+1. Ready instances first (Ready condition = True).
+2. Creation timestamp ascending (older instances first).
+3. Namespace/name lexicographic as tiebreaker.
+
+This default prioritizes propagating to healthy instances first. A Ready
```
cheeseandcereal, line 479: Is this really desirable? Lets say there was another rollout that was _causing_ the unhealthy behavior. Now there's an update with a new revision, and now this new rollout is going to prioritize the remaining untouched healthy instances, effectively specifically making bad changes more likely to propagate quicker.

Also - when we're talking about trying to do anything deterministically, since this is a value that changes, it fundamentally prevents or makes deterministic algorithms for ordering much more difficult.

docs/design/proposals/instance-propagation-control.md
spattk: yeah, found it under "Overlapping propagation" as well. thanks!

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+instance is already serving successfully on the current revision — it is the
+lowest-risk candidate to move to a new revision because any regression is
+immediately detectable (Ready → NotReady). Propagating to NotReady or Unknown
```
cheeseandcereal, line 482: `when` statements can have custom criteria for propagation which may have things like alternative "healthy" criteria during a rollout. In these cases, this "Ready" state is not the most desirable thing to monitor/make decisions off of, right?

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+`propagate` is a struct field on `ResourceGraphDefinitionSpec`:
+
+```go
+type ResourceGraphDefinitionSpec struct {
+    // ... existing fields ...
+
+    // Propagate controls how instances transition between GraphRevisions.
+    // When omitted, all instances immediately adopt the latest Active
+    // revision (current default behavior).
+    // Requires the InstancePropagationControl feature gate.
```
NicholasBlaskey, line 101: the feature gate won't apply to the RGD CRD right? 

So if you have the feature gate disabled your RGD CRD will still have it right?

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
...
+label and `currentRevision = 0`. In `GetRevisionForInstance()`, an instance with
+`currentRevision = 0` that is not in the Released map falls through to the
+latest revision. This is correct: new instances should use the newest graph
+since the system is trying to converge all instances there anyway.
+
+### What happens when an instance is deleted during propagation?
+
+Deletion always wins. On the next propagation cycle, the RGD controller's
+informer cache no longer contains the deleted instance. The planner naturally
+excludes it — aggregate counts adjust, batch sizes may shrink, and the remaining
```
cheeseandcereal, line 622: I wonder if re-computing the batches is always the correct/desirable behavior

docs/design/proposals/instance-propagation-control.md
```diff
@@ -0,0 +1,843 @@
+# KREP-NNN: Instance Propagation Control
```
chrisdoherty4:
```suggestion
# KREP-024: Instance Propagation Control
```

Skarlso:
For the reference, KREP-024 is the Graph KREP that was merged. :)

PR #1260, opened 2026-04-17
fix: instance deletion stuck when SSA field manager owns finalizer

NicholasBlaskey:
Fix issue causing instances to hang on deletion if they were created before this commit 
https://github.com/kubernetes-sigs/kro/commit/7e4aeaffdfdc67a6a6da5242a9d40498a833dac4

If the Kro finalizer is owned by another field manager and we do an empty SSA apply without that finalizer, then we just release our ownership. Since another field manager owns that field k8s won't remove it.

We do specify with force=true, but there is no conflict since our SSA is not specifying what our intent is with that field, just that we don't care about the field.

Switch to just using merge patch to avoid dealing with this edge case.

We also were adding ourselves as an owner to other finalizers which wasn't great. 

Matches this pattern
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/garbagecollector/operations.go#L104

### DIFF
--- a/pkg/controller/instance/deletion.go
+++ b/pkg/controller/instance/deletion.go
@@ -15,6 +15,7 @@
 package instance
 
 import (
+	"encoding/json"
 	"fmt"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -22,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
 	"github.com/kubernetes-sigs/kro/pkg/graph"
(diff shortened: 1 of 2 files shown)


AndrewCharlesHay:
The switch to merge-patch + `retry.RetryOnConflict` looks correct, but the safety of the whole thing is load-bearing on `resourceVersion` being included in the patch body and the API server honoring it as a precondition.

Concretely: JSON merge patch replaces `metadata.finalizers` atomically (arrays aren't merged). If another controller added a finalizer between the `Get` and the `Patch`, *without* the resourceVersion precondition you'd silently wipe their finalizer. With it, you get a 409 and the retry loop re-fetches — which is the behavior you want.

Suggestion: add a one-line comment above the `patchData` marshal explaining why `resourceVersion` has to stay in the body. Otherwise someone doing a "cleanup" refactor could drop it and reintroduce the race without any test catching it (the test here uses a fake client that doesn't enforce resourceVersion conflicts the same way a real apiserver does).

Something like:
```go
// resourceVersion MUST stay in the patch body — it's the precondition
// that protects against silently dropping finalizers added by other
// controllers between the Get above and this Patch. Removing it would
// make merge-patch wipe the finalizers array atomically.
```

NicholasBlaskey:
updated test to check the contents of patch

michaelhtm (approved):

a-hilaly (approved):

a-hilaly (approved):

PR #1262, opened 2026-04-20
Allowing passing through parameter to run specific integration tests

NicholasBlaskey:
Running specific intg tests through go test doesn't work because they need setup

Add an optional parameter to let you pass in a focus to ginkgo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,21 +131,22 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: test
-test: manifests generate fmt vet envtest ## Run tests. Use WHAT=unit or WHAT=integration
+test: manifests generate fmt vet envtest ## Run tests. Use WHAT=unit or WHAT=integration, pass extra args after --
 ifeq ($(WHAT),integration)
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_VERSION) --bin-dir $(LOCALBIN) -p path)" \
 		go tool ginkgo -p -v \
 		--cover \
 		--coverprofile=integration-cover.out \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -131,15 +132,21 @@ vet: ## Run go vet against code.
...
 .PHONY: test
-test: manifests generate fmt vet envtest ## Run tests. Use WHAT=unit or WHAT=integration
+test: manifests generate fmt vet envtest ## Run tests. Use WHAT=unit or WHAT=integration, optionally FOCUS=<pattern> (integration only)
 ifeq ($(WHAT),integration)
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_VERSION) --bin-dir $(LOCALBIN) -p path)" \
 		go tool ginkgo -p -v \
 		--cover \
 		--coverprofile=integration-cover.out \
 		-coverpkg=github.com/kubernetes-sigs/kro/pkg/... \
+		$(if $(FOCUS),--focus "$(FOCUS)") \
```
jakobmoellerdev: why dont we simply pass all arguments to make here?

NicholasBlaskey: updated to `TEST_ARGS`

I don't think make has a way to just allow it to accept any arbitrary args without putting it in a variable

jakobmoellerdev: I believe I saw something like this before: https://gist.github.com/cemdrk/77d70dddc2710778dfc32a589b62d929 maybe take a look if this is more desirable

NicholasBlaskey: updated

jakobmoellerdev (commented):
/lgtm
/ok-to-test

great stuff!

jakobmoellerdev (approved):

PR #1267, opened 2026-04-23
KREP-025 Propagation Rate Limit

michaelhtm:
Add design proposal for propagation rate limiting

### DIFF
--- a/docs/design/proposals/propagation-rate-limit.md
+++ b/docs/design/proposals/propagation-rate-limit.md
@@ -1,0 +1,246 @@
+# Propagation Rate Limit
+
+## Problem
+
+When a new GraphRevision becomes Active, the RGD controller re-registers the
+instance microcontroller, which enqueues every existing instance for
+re-reconciliation. Every instance resolves to the latest revision on its next
+reconcile, and each reconcile applies every resource in the graph via
+server-side apply.
+
+The SSA patch itself is cheap. The problem is what it causes downstream. When
+an RGD image field changes, every instance's Deployment spec changes, which
+triggers a new ReplicaSet, which terminates old pods and schedules new ones,
+which pulls images and fires readiness probes. For an RGD with 500 instances
+and 3 Deployments each, that's 1500 concurrent pod rollouts hitting the
+scheduler, the node autoscaler, and the image registry simultaneously.
+
+There is no mechanism to pace how quickly instances begin transitioning. The
+blast radius is always 100%.
+
+## Proposal
+
+Control propagation with a single parameter that defines how many instances are
+admitted to a new revision per unit of time:
+
+```yaml
+metadata:
+  annotations:
+    kro.run/propagation-rate: "10/m"
+```
+
+Meaning: release up to 10 instances per minute. Every minute, the controller
+releases the next batch of up to 10 pending instances. The downstream effects
+(pod rollouts, scheduling, image pulls) are spread over time instead of hitting
+all at once.
+
+Format: `"<count>/<unit>"` where unit is `s` (second), `m` (minute), or `h`
+(hour). When unset, all instances transition immediately (current behavior).
+
+### Why time-based rate limiting
+
+A concurrency limiter (max-in-flight) bounds how many instances are
+mid-transition simultaneously. But kro's SSA apply completes fast — the
+instance "lands" (label updates) within seconds. Slots free quickly and the
+limiter burns through the entire fleet in minutes. The downstream effects (pod
+scheduling, image pulls, node autoscaling) continue long after the slot was
+freed. A concurrency limiter doesn't bound the actual cluster load.
+
+What matters is the **admission rate** — how quickly new instances start
+transitioning, regardless of how fast prior ones finish. A time-based rate
+directly controls this: "my cluster can handle 10 instance transitions per
+minute" is a statement about scheduler and node capacity, not about kro's
+reconcile throughput.
+
+### Configuration
+
+**Annotation** on the RGD (per-RGD override):
+
+```yaml
+metadata:
+  annotations:
+    kro.run/propagation-rate: "10/m"
+```
+
+**Controller flag** (cluster-wide default when annotation is absent):
+
+```
+--propagation-default-rate=""
+```
+
+**Precedence:**
+
+```
+effective = annotation if present, else flag if non-empty, else unset
+
+unset -> unlimited (current behavior)
+```
+
+The flag acts as a default, not a ceiling — the annotation always wins. A
+ceiling flag can be added later without breaking changes (see Future work).
+
+**Validation:**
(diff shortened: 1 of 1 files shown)


docs/design/proposals/propagation-rate-limit.md
```diff
@@ -0,0 +1,660 @@
...
+
+### Design details
+
+The rate limit is modeled as a bucket with two parameters. The *burst size* is
+the bucket capacity, meaning the maximum number of instances that can be
+in-flight (transitioning) at any time. It also governs how many instances are
+released on the first cycle. The *fill rate* controls how many new instances are
+released per subsequent cycle as in-flight slots free up. Together these produce
+a "fast start, steady pace" shape. When only one parameter is set, the other
+defaults to the same value, collapsing to a flat rate.
```
NicholasBlaskey: I'm not sold on these two parameters being necessary

It feels like we are leaking the token bucket implementation into the customer experience

fill rate vs burst size feels like not super useful or intuitive ways to control this rate. I'm not sure why I would ever want a initial burst, it feels more like I would want the opposite an initial slowdown gradually increasing

docs/design/proposals/propagation-rate-limit.md
```diff
@@ -0,0 +1,660 @@
...
+restart.
+
+#### Instance ordering
+
+Instances are processed in a deterministic order using a rotation scheme seeded
+by the target revision number:
+
+1. Sort all instances by creation timestamp ascending (older instances first).
+2. Namespace/name lexicographic as tiebreaker for identical timestamps.
+3. Rotate the sorted list by `targetRevision % len(instances)` positions.
```
NicholasBlaskey: What is the benefit of this algorithm compared to like sorting by hashing `namespace+name+revision`?

It seems strange we are still having a determined order

docs/design/proposals/propagation-rate-limit.md
michaelhtm: yea you're right. I think maybe just a max-propagation count would be enough?
cc: @JamesGAWS

docs/design/proposals/propagation-rate-limit.md
michaelhtm: i'd say using creationTimestamp is more predictable..

docs/design/proposals/propagation-rate-limit.md
```diff
@@ -0,0 +1,660 @@
...
+Deleted instances are skipped when encountered in the stored order. They do not
+consume batch budget. The propagation loop naturally excludes them because the
+informer cache no longer contains them. Aggregate counts adjust accordingly —
+if the deleted instance was pending, the total pending count decreases.
+
+**Instance updated during propagation.**
+
+User-initiated spec changes to an instance (e.g., changing a field on the
+custom resource) do not affect propagation ordering or batch budget. The
+instance's position in the stored order is unchanged. If the instance is pending,
```
NicholasBlaskey: does this mean we lock out the user updates to an instance applying until it is propagated?

michaelhtm: no, instance updates still occur, the RGD changes are the ones we limit

docs/design/proposals/propagation-rate-limit.md
```diff
@@ -0,0 +1,660 @@
...
+This is an operational concern: the cluster operator understands system capacity
+and API server pressure, and needs a knob to cap concurrent change, independent
+of any future rollout policy the platform team might define.
+
+## Goals
+
+- Give cluster operators a way to cap how many instances transition to a new
+  revision at a time, reducing blast radius during rollouts.
+- Provide two intuitive parameters: a burst size (initial capacity) and a fill
+  rate (steady-state pace), so operators can shape the rollout curve.
```
cheeseandcereal: What do these mean? What are the units? If I put 1,2, how does that actually translate to the rollout, how long there is between, etc?

docs/design/proposals/propagation-rate-limit.md
```diff
@@ -0,0 +1,660 @@
...
+- **Annotations:** `kro.run/propagation-max-burst-size` and
+  `kro.run/propagation-fill-rate` on the RGD resource.
+- **Controller flags:** `--propagation-default-max-burst-size` and
+  `--propagation-default-fill-rate` as cluster-wide defaults when the
+  annotations are absent.
+- **Feature gate:** `PropagationRateLimit` (Alpha, default disabled).
+
+These provide an operator-imposed ceiling on in-flight propagation. When a new
+revision is Active, the burst size caps how many instances can be dispatched in
+the first cycle, and the fill rate caps how many new instances are released per
```
NicholasBlaskey: there feels to be a third parameter which is important but implicit here and that is how often are we running cycles. 

This is now tied to reconciles right? 

Is this a constant time or could a user artificially speed up cycles by issuing updates to an RGD (no ops that trigger reconciles)? 

It almost feels like tying physic calculations in a game to the frame rate. If we double the reconcile rate I don't think users would necessarily want their instances to suddenly release faster.

I wonder if we need to expose this as a fixed configurable time interval too.

JamesGAWS: Rate limiting decouples updates from reconciles.  Reconciles won't do work unless they are allowed to progress by the rate limiter. If reconciliation slows due to resource constraints, rate limiting will allow subsequent reconciliation at a higher rate for a period of time to allow KRO to "catch up".

docs/design/proposals/propagation-rate-limit.md
JamesGAWS: Rate and burst are concepts independent from the token bucket algorithm (although token buckets are a good way to implement them). Burst exists not only to allow a jump start at the beginning, but also to allow some room for reconciliation to "catch up" if it temporarily slows during processing.

docs/design/proposals/propagation-rate-limit.md
```diff
@@ -0,0 +1,660 @@
...
+post-restart cycle dispatches up to `burst_size` instances. If previously
+in-flight instances have already landed (label updated) but the Released map
+was lost, these slots appear free and new instances are released on top of the
+ones that already transitioned. The overshoot is bounded by `burst_size`,
+occurs only once after restart, and self-corrects on the next cycle when
+in-flight counts are recomputed from labels. This is acceptable for an Alpha
+feature and is consistent with Kubernetes Deployment controller behavior on
+restart.
+
+#### Instance ordering
```
JamesGAWS: Can we consider a comparator instead? It's a relatively simple and common pattern to allow users to chose their own ordering. We can also consider providing "built in" comparators for users who don't want fine grained ordering control. Best example here I can think of is a `random` comparator for users who don't care about ordering.

michaelhtm:
Solving rate limiting https://github.com/kubernetes-sigs/kro/pull/1259, by eventually supporting `time.Now()` in CEL
/close

PR #1268, opened 2026-04-23
Fix collection deletion with CEL errors in forEach identity paths

NicholasBlaskey:
Avoid depending on ```GetDesiredIdent``` in deletion path for collections.

For a collection that fails to resolve like 
```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: workers
spec:
  schema:
    apiVersion: v1alpha1
    kind: Workers
    spec:
      count: integer

  resources:
    - id: pods
      forEach:
        - i: ${lists.range(schema.spec.count)}
      template:
        apiVersion: v1
        kind: Pod
        metadata:
          name: ${"pod-" + string(i / 0)}
          namespace: ${schema.metadata.namespace}
        spec:
          containers:
            - name: app
              image: nginx
```
deletion would hang forever

This also occurs if you go over the max collection size. Prevent this by just relying on labels instead of relying on desired identity. 

Note this is also something that may be worth considering for all nodes. We could instead rely on node-id instead of needing to resolve the cel expression in name for regular resources.

### DIFF
--- a/pkg/controller/instance/deletion.go
+++ b/pkg/controller/instance/deletion.go
@@ -16,6 +16,8 @@ package instance
 
 import (
 	"fmt"
+	"slices"
+	"strconv"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


pkg/runtime/node.go
```diff
@@ -284,13 +284,8 @@ func (n *Node) resolve(mode resolveMode) (result []*unstructured.Unstructured, e
 func (n *Node) DeleteTargets() ([]*unstructured.Unstructured, error) {
 	switch n.Spec.Meta.Type {
 	case graph.NodeTypeCollection, graph.NodeTypeResource:
-		desired, err := n.GetDesiredIdentity()
```
NicholasBlaskey, line 287: I wasn't sure of the intention of this original code.

I am not sure if it was just for ordering or also is trying to provide another level of safety that we aren't deleting a collection item that is no longer referenced, but deleting the item feels more correct to me

pkg/controller/instance/deletion.go
```diff
@@ -78,6 +80,26 @@ func (c *Controller) planNodesForDeletion(
...
+			}
+			if len(items) == 0 {
+				state.SetDeleted()
+				continue
+			}
+			sortByCollectionIndex(items)
+			node.SetObserved(items)
+			state.SetInProgress()
+			deletionNode = node
+			continue
```
michaelhtm, line 101: nit: what's the point of setting deletionNode if we continue?

pkg/controller/instance/deletion.go
michaelhtm, line 106: should we consider calling GetDesiredIdentity where it's needed? Then we can move the collection logic in the switch case?

michaelhtm (commented):
Thanks @NicholasBlaskey 
left a few comments below

pkg/controller/instance/deletion.go
NicholasBlaskey: I tried this and it ends up being more complicated because we need to duplicate a lot of logic across switch cases

pkg/controller/instance/deletion.go
NicholasBlaskey: This is matching the logic in the switch. We do return the deletionNode at the end when we are looping so this isn't a noop

PR #1269, opened 2026-04-24
Fix validation marker overwriting other XValidations

NicholasBlaskey:
An RGD like this will overwrite the immutable validation and not respect it
```
 apiVersion: kro.run/v1alpha1
  kind: ResourceGraphDefinition
  metadata:
    name: app-with-immutable-status
  spec:
    schema:
      apiVersion: v1alpha1
      kind: Application
      spec:
        status: string | immutable=true validation="self.matches('^(pending|active|done)$')"
    resources:
      - id: configmap
        template:
          apiVersion: v1
          kind: ConfigMap
          metadata:
            name: ${schema.metadata.name}-config
          data:
            status: ${schema.spec.status}
            name: ${schema.spec.name}
```

Switch to using append to catch this case

### DIFF
--- a/pkg/simpleschema/markers.go
+++ b/pkg/simpleschema/markers.go
@@ -309,9 +309,10 @@ func applyValidationMarker(schema *extv1.JSONSchemaProps, marker *Marker) error 
 	if strings.TrimSpace(marker.Value) == "" {
 		return fmt.Errorf("validation failed")
 	}
-	schema.XValidations = []extv1.ValidationRule{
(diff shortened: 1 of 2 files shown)


michaelhtm (approved):

jakobmoellerdev (commented):
/ok-to-test
/lgtm

jakobmoellerdev (approved):

michaelhtm (approved):

PR #1270, opened 2026-04-24
Validate default marker values match their field type

NicholasBlaskey:
If you created an RGD with ``` count: integer | default="hello"```

you would see the error message ```create CRD: CustomResourceDefinition in version "v1" cannot be handled as a CustomResourceDefinition: invalid character 'h' looking for beginning of value```

Provide a cleaner error message for this case


jakobmoellerdev (commented):
/lgtm
/ok-to-test

jakobmoellerdev (approved):

PR #962, opened 2026-01-21
Modify tracking-id annotation for ArgoCD

allamand:
Updated tracking-id annotation to use orValue for default, as it is now needed with latest versions

PR #1273, opened 2026-04-29
Modify tracking-id annotation for ArgoCD

allamand:
Updated tracking-id annotation to use `.orValue("")` for default, as it is now needed with latest versions.

The `?` optional access operator is kept because `annotations` may be absent on the resource — without it, resources with no annotations would fail reconciliation. The `?` + `.orValue("")` pattern is the canonical kro approach (see docs and test examples throughout the repo).

Supersedes #962 (rebased on latest main, cleaned up merge commit).

### DIFF
--- a/website/docs/docs/faq.md
+++ b/website/docs/docs/faq.md
@@ -80,7 +80,7 @@ sidebar_position: 100
            blockOwnerDeletion: true
            controller: false
       annotations:
-        argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"]}
+        argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"].orValue("")}
    ```
 
    This annotation allows ArgoCD to properly track and manage the resources


allamand:
cc @jakobmoellerdev

website/docs/docs/faq.md
```diff
@@ -80,7 +80,7 @@ sidebar_position: 100
            blockOwnerDeletion: true
            controller: false
       annotations:
-        argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"]}
+        argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"].orValue("")}
```
jakobmoellerdev, line 83: do you really want to set an empty annotation? it looks to me like we dont want to have an optional value at all here. WDYT?

michaelhtm: yea..what if we use omit() instead?

jakobmoellerdev: agreed. much cleaner now with the sentinel but that would require for us to lift the feature flag.

allamand: I tried with .omit() and this works well I'll update the pr. I didn't needed to activate Frazier flag though..

allamand: hmm it is not working in EKS capabilities, maybe an older version ?

allamand: The error with .omit() was:
  
  failed to validate resource "s3bucket": failed to compile template expression 
  "schema.metadata.?annotations[\"argocd.argoproj.io/tracking-id\"].omit()" 
  at path "metadata.annotations[\"argocd.argoproj.io/tracking-id\"]": 
  ERROR: <input>:1:68: found no matching overload for 'omit' applied to 'optional_type(string).()'
   | schema.metadata.?annotations["argocd.argoproj.io/tracking-id"].omit()
   | ...................................................................^

jakobmoellerdev: thats likely because the capabilities have not had this enabled yet.

Skarlso:
Hey @allamand. I see you fixed some things upstream. Is this still relevant here? :)

jakobmoellerdev:
/close 

if you still feel like this is relevant please reopen

PR #1280, opened 2026-05-08
Revert "Emit instance condition telemetry"

michaelhtm:
This reverts commit eeee5ec328d1ca2f2abcff31c2b8d93e8c9921a6.

reverting to unblock v0.9.2 release.


a-hilaly (approved):
Looks like a few wrong decision were made in the original PR. kube-state-metrics emits a timestamp for status ready 
https://docs.cloudera.com/management-console/1.5.4/monitoring-metrics/topics/cdppvc_ds_kube_pod_status_ready_time_trics.html, cert-manager also emits `certmanager_certificate_expiration_timestamp_seconds`. Ideally we should follow the same patterns and not emit a pre calculated "time since"

Reverting this as we need to cut 0.9.2 and we continue discussing the format of metric.

/lgtm

PR #1282, opened 2026-05-09
fix(instance): set ResourcesReady=False with ReconciliationSuspended reason when suspended

### LINKED ISSUE #1263: Suspended instances show no indication of suspension in status conditions
### Description

**Observed Behavior**:

I was testing how kro handles drift specifically whether it would recreate a managed
ConfigMap if I deleted it manually. I deleted the ConfigMap and watched the instance status,
expecting to see it briefly go unhealthy and then recover as kro reconciled it back.

Instead nothing happened. The instance briefly flipped to `Ready=False` with
`"awaiting resource readiness: <nil>"` after a requeue, but `state` stayed `ACTIVE` and
the ConfigMap was never recreated.

Eventually I realized I had annotated that instance with `kro.run/reconcile=suspended`
earlier while testing something else and never removed it.

**Expected Behavior**:

The status conditions should include a `ReconciliationSuspended` condition when the
`kro.run/reconcile=suspended` annotation is present, so that anyone looking at the instance
can immediately tell that kro is intentionally not reconciling it. Something like:

```yaml
status:
  state: ACTIVE
  conditions:
  - type: Ready
    status: "True"
  - type: ReconciliationSuspended
    status: "True"
    reason: Suspended
    message: "reconciliation is suspended via kro.run/reconcile annotation"
```

**Reproduction Steps**:

ResourceGraphDefinition:
```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: suspension-test
spec:
  schema:
    apiVersion: v1alpha1
    kind: SuspensionTest
    spec:
      name: string
  resources:
    - id: configmap
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.spec.name}
        data:
          key: value
```

Instance:
```yaml
apiVersion: kro.run/v1alpha1
kind: SuspensionTest
metadata:
  name: my-instance
  namespace: default
spec:
  name: my-configmap
```

Steps:
1. Apply the RGD and instance above
2. Wait for `state: ACTIVE`
3. Add the suspension annotation:
   ```bash
   kubectl annotate suspensiontest my-instance kro.run/reconcile=suspended --overwrite
   ```
4. Check the status:
   ```bash
   kubectl get suspensiontest my-instance -o yaml | grep -A 30 "status:"
   ```
5. Manually delete the managed ConfigMap:
   ```bash
   kubectl delete configmap my-configmap
   ```
6. Check the status again — still shows `state: ACTIVE` with no indication of suspension

**Versions**:
- kro version: v0.9.1
- Kubernetes Version: v1.35.0

**Involved Controllers**:
- Instance controller (`pkg/controller/instance`)

**Error Logs**:

No logs. That is part of the problem — everything looks ready in the logs and status
even though the controller has stopped reconciling.

---

Instance (Create, Update, Deletion)

RohanKaran:
## What this PR does

When an instance is suspended via the `kro.run/reconcile` annotation, the controller skipped node reconciliation but wrote no signal to status indicating suspension. The instance continued to show `Ready=True` even though kro was not actively verifying resource health — a silent false positive.

This PR sets `ResourcesReady=False` with reason `ReconciliationSuspended` when the suspend annotation is present. This rolls up to `Ready=False`, making suspension visible in kubectl, Lens, and any tool that reads conditions.

## Changes

- `pkg/controller/instance/controller.go` — detect suspension early, mark the condition, return before the normal reconcile path
- `pkg/controller/instance/status.go` — add `ReconciliationSuspended` marker method
- `pkg/controller/instance/controller_test.go` — unit tests for both canonical `suspended` and legacy `disabled` values
- `test/integration/suites/deploymentservice/suite_test.go` — update existing integration test to assert the new condition behavior

Fixes #1263


pkg/controller/instance/controller.go
```diff
@@ -265,17 +265,20 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
...
-			if deletingErr, ok2 := errors.AsType[*resourceDeletingError](err); ok2 {
-				rcx.Mark.ResourcesDeleting("%v", deletingErr)
-				_ = c.updateStatus(rcx)
-				return rcx.delayedRequeue(deletingErr)
-			}
-			rcx.Mark.ResourcesNotReady("resource reconciliation failed: %v", err)
+	if v1alpha1.IsReconcileSuspended(reconcileState) {
+		rcx.Mark.ReconciliationSuspended("reconciliation suspended via %s annotation", v1alpha1.InstanceReconcileAnnotation)
+		return c.updateStatus(rcx)
+	}
```
michaelhtm, line 295: Should we also introduce a new instance state called `Suspended`? Keeping it as is bight be misleading..

michaelhtm (commented):
Thanks @RohanKaran! changes lgtm for the most part. 
Just have one question below

pkg/controller/instance/controller.go
RohanKaran: Yeah I agree

api/v1alpha1/instance_state.go
```diff
@@ -28,6 +28,8 @@ const (
 	InstanceStateDeleting InstanceState = "DELETING"
 	// InstanceStateError means reconciliation hit an error.
 	InstanceStateError InstanceState = "ERROR"
+	// InstanceStateSuspended means the controller is not actively verifying resource health
+	InstanceStateSuspended InstanceState = "SUSPENDED"
```
jakobmoellerdev: Maybe a bit of a radical take but honestly I am currently not willing to move into a more complex state machine for the simple reason that KRO becomes harder to reason about. I dont think it is wise to add so many states to this. Arguably we should even move away from states completely in favor of conditions.

jakobmoellerdev: Note that Im also not in favor of adding more conditions, just that a simple condition reason could state that were suspended so it is ready false

RohanKaran: > Maybe a bit of a radical take but honestly I am currently not willing to move into a more complex state machine for the simple reason that KRO becomes harder to reason about. I dont think it is wise to add so many states to this. Arguably we should even move away from states completely in favor of conditions.

Looking at both state and conditions deeply they feel redundant to an extent. But at present with state being the main thing users see, having it say ACTIVE while suspended seems contradictory to what conditions report. Btw was there a specific reason both coexist?

jakobmoellerdev: I think honestly speaking the project has just grown and we have evolved with adoption patterns. Looking back, Instance state has been around for a long time and the standardization on conditions came after. I think we should move to whatever the ecosystem standardized on, which is the conditions array, not a custom state field

RohanKaran: Makes sense. Reverted the state change commit.

jakobmoellerdev (commented):
/lgtm

jakobmoellerdev (commented):
/lgtm
@cheeseandcereal @michaelhtm ptal

api/v1alpha1/instance_state.go
michaelhtm: In this case, should we remove state from additional printer columns? Because it would be misleading when the reconciliation is suspended. We can replace it with Ready Reason?

jakobmoellerdev (approved):

PR #1286, opened 2026-05-12
Update `gcb-docker-gcloud `builder image to fix push-images job

michaelhtm:
The previous image tag (`v20251110-7ccd542560`) was removed from the
registry, causing `kro-push-images` to fail with "manifest unknown".
Update to `v20260205-38cfa9523f` which is the current tag used across
`kubernetes/test-infra`.


jakobmoellerdev (approved):
/lgtm

PR #1294, opened 2026-05-15
Make CLI validate work offline

NicholasBlaskey:
Make CLI validate default to offline resolution

Offline resolution works by taking a list of CRDs and a kubernetes version.

Generate built in k8s types and use them for offline validation


jakobmoellerdev:
/close

currently way too out of base to be further considered, feel free to revisit

PR #1295, opened 2026-05-16
fix: skip main-object SSA apply for status-only patch nodes

ellistarn:
## Summary

- Status-only patch nodes (like `kindStatus`) no longer register a field manager on the main resource
- Only the status subresource field manager is created
- Fixes ownership conflicts when a Kind is a node in a parent Graph

## Problem

When a patch node contains only identity fields (`apiVersion`, `kind`, `metadata`) plus `status`, `ssaWrite` was splitting it into:
1. A main-object SSA apply (with just identity fields) — registering a field manager
2. A status subresource SSA apply

The main-object apply served no purpose (identity fields don't need ownership) but it registered the kind controller as a field manager on the Kind resource. When the parent graph tried to reapply the same Kind on subsequent reconciles, KRO's ownership check saw the competing field manager and rejected the apply with a `Conflict` error.

## Fix

After stripping `status` from the payload, check if only identity fields remain (`apiVersion`, `kind`, `metadata`). If so, skip the main-object apply entirely — only apply to the status subresource.

## Test

`TestPatch_StatusOnlyDoesNotClaimMainResource` — verifies that a status-only patch node does not register a main-object field manager, allowing the original owner to retain exclusive ownership.


ellistarn:
Sorry rogue agent

PR #1298, opened 2026-05-19
Add kro fmt command for YAML formatting

NicholasBlaskey:
Add a CLI fmt command that uses an opinionated configuration of google/yamlfmt to format rgd files. 

An alternative would be this but it doesn't preserve newlines https://github.com/kubernetes-sigs/yaml/tree/master/yamlfmt

Note we haven't added Kro specific formatting. Open to ideas here but nothing stood out as critical.


jakobmoellerdev:
this is too large of a diff, please rebase and then refactor separately. cheers!

/close

PR #1299, opened 2026-05-19
Fix nil pointer dereference in ExternalDocs conversion

NicholasBlaskey:
Not currently customer-impacting (kro-generated CRDs never have ExternalDocs)

We are going to use this logic for offline validation, saw this here so wanted to put up a fix separately


jakobmoellerdev (commented):
/ok-to-test
/lgtm

jakobmoellerdev (approved):

PR #1300, opened 2026-05-28
feat(graph): allow schema.* references in instance status expressions

jakobmoellerdev:
## Summary

Status expressions previously only allowed references to resource IDs (e.g. `${deployment.status.replicas}`). Referencing `schema.spec.*` — the instance's own spec fields — was blocked at build time with no strong technical reason. This change lifts that restriction, consistent with how `includeWhen` already works.

**Use cases enabled:**
- Echo a spec field into status: `status.host: ${schema.spec.host}`
- Combine resource output with spec input: `status.url: ${svc.status.ip + '/' + schema.spec.path}`

## Changes

**`pkg/graph/builder.go`**
- `buildStatusSchema`: add `SchemaVarName` to the allowed-identifiers list passed to `inspectExpressionRestricted`
- `buildInstanceNode`: relax the zero-deps guard to permit schema-only expressions; wire `InstanceNodeID` into `instanceDeps` when schema is referenced so the runtime can resolve it at evaluation time

**`pkg/runtime/runtime.go`**
- Handle `InstanceNodeID` in the instance dep-wiring loop (`instNode.deps[InstanceNodeID] = instNode`) — previously only regular resource nodes were wired this way

**Tests**
- Two new unit test cases in `TestGraphBuilder_Validation`: schema-only and mixed resource+schema status expressions
- Three new integration tests in `test/integration/suites/core/status_schema_ref_test.go`: schema-only population, mixed resource+schema, and spec-update propagation

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -712,10 +712,17 @@ func buildInstanceNode(
 		if err != nil {
 			return nil, fmt.Errorf("failed to extract dependencies from expression %q: %w", statusVariable.Expression, err)
 		}
-		if len(deps) == 0 {
-			return nil, fmt.Errorf("instance status field must refer to a resource: %s", statusVariable.Path)
+		// Status fields may reference resources, schema, or both.
+		referencesSchema := slices.Contains(statusVariable.Expression.References, SchemaVarName)
+		if len(deps) == 0 && !referencesSchema {
+			return nil, fmt.Errorf("instance status field must refer to a resource or schema: %s", statusVariable.Path)
 		}
 		instanceDeps = append(instanceDeps, deps...)
+		// If this expression references schema, include the instance node itself as a dep
+		// so the runtime wires it into instNode.deps for context building.
+		if referencesSchema && !slices.Contains(instanceDeps, InstanceNodeID) {
+			instanceDeps = append(instanceDeps, InstanceNodeID)
+		}
 
 		instanceStatusVariables = append(instanceStatusVariables, &variable.ResourceField{
 			FieldDescriptor: statusVariable,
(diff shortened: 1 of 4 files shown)


pkg/graph/builder.go
```diff
@@ -805,13 +812,11 @@ func buildStatusSchema(
 		return nil, nil, nil, fmt.Errorf("status fields without expressions are not supported: %v", noExpressionFields)
 	}
 
-	// Instance status expressions can ONLY reference resources, not schema.
-	// At runtime, status is populated after resources are created.
-
-	// Verify status expressions don't reference schema and populate References
+	// Verify status expressions only reference known resources or schema, and populate References.
+	allowedStatusVars := append(nodeNames, SchemaVarName)
```
Skarlso: I'm pretty sure if nodeNames every acquires larger capacity this will mutate that array.
I think to be save you should use 
```go
slices.Concat(nodeNames, []string{SchemaVarName})
```

instead?

pkg/runtime/runtime.go
```diff
@@ -126,7 +125,10 @@ func FromGraph(g *graph.Graph, instance *unstructured.Unstructured, rgdConfig gr
 
 	// Wire up instance node dependencies.
 	for _, depID := range instNode.Spec.Meta.Dependencies {
-		if dep, ok := rt.nodes[depID]; ok {
+		if depID == graph.InstanceNodeID {
+			// Status expressions referencing schema need the instance node itself as a dep.
+			instNode.deps[graph.InstanceNodeID] = instNode
```
Skarlso: Hm, you are setting the same thing on the above line at 118, not? Shouldn't you be doing the same here and then just proceed as is? 

so something like
```go
        instNode.deps[graph.InstanceNodeID] = instNode
        for _, depID := range instNode.Spec.Meta.Dependencies {
                if dep, ok := rt.nodes[depID]; ok {
                        instNode.deps[depID] = dep
                }
        }
```

Or am I misunderstanding the logic? :)

Skarlso (commented):
Just two tiny questions. :)

pkg/graph/builder.go
cheeseandcereal: This comment is no longer true, so probably needs an update.

pkg/runtime/runtime.go
cheeseandcereal: +1 was looking and thought the same thing. Doesn't seem like this needs to be conditional in the loop, right?

jakobmoellerdev:
@cheeseandcereal / @michaelhtm could one of you lgtm please?

michaelhtm (approved):

cheeseandcereal (approved):
lgtm, thanks!

PR #1301, opened 2026-06-01
Add author-defined custom status conditions to RGD instances

lalitg327:
## Summary

Adds support for author-defined custom status conditions on RGD instances.
Authors declare a `conditions:` block under their RGD's `status:`, and kro
evaluates the expressions on every reconcile and writes the result to
`.status.conditions[]`.

## Why

Today every kro-managed instance carries the same four built-in conditions
(InstanceManaged, GraphResolved, ResourcesReady, Ready). These reflect kro's
reconciliation lifecycle ("all child resources were applied") but say nothing
about whether the application is actually healthy.

An RGD author cannot express domain-specific health signals. "The database is
ready when the primary is in Running phase and all replicas are caught up" has
to be tracked in a separate controller, or not at all. Authors also can't
define what `Ready` means for their custom resource, since kro's `Ready` is the
conjunction of the lifecycle built-ins, not the application's actual health.

This PR closes both gaps. Authors can:

- Surface domain-specific conditions like `PrimaryReady`, `AllReplicasReady`,
  or `BackupRecent` on `.status.conditions[]`.
- Override the default `Ready` with their own definition (e.g. "ResourcesReady
  AND replicas serving traffic").
- Read kro's internal lifecycle conditions inside their own CEL via
  `runtime.condition(schema, 'X')`, so they can compose lifecycle signals with
  domain checks.

When an RGD declares author conditions, only those conditions appear on
`.status.conditions[]`; kro's built-ins are filtered out so the author retains
full ownership of their status surface. The built-ins remain computed and
accessible to author CEL through `runtime.condition`.

### Basic: one author condition reading a child resource

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: webapp
spec:
  schema:
    apiVersion: v1alpha1
    kind: WebApp
    spec:
      replicas: integer
    status:
      conditions:
        - ${runtime.newCondition({
            type: 'AppReady',
            status: deployment.status.readyReplicas > 0 ? 'True' : 'False',
            reason: 'ReplicaCount',
            message: ''
          })}
```

### Author `Ready` combining kro's lifecycle signal with a domain check

`runtime.condition(schema, 'ResourcesReady')` reads kro's internal value for
any of the four built-in condition types, so authors keep access to those
signals even after they take ownership of the wire. Here `Ready` is `True`
only when kro has finished applying resources AND the deployment is serving
traffic:

```yaml
status:
  conditions:
    - ${runtime.newCondition({
        type: 'Ready',
        status: runtime.condition(schema, 'ResourcesReady').status == 'True'
          && deployment.status.readyReplicas > 0 ? 'True' : 'False',
        reason: deployment.status.readyReplicas > 0 ? 'AllHealthy' : 'NotReady',
        message: ''
      })}
```

### Collection expansion: one condition per item

A single `conditions:` entry can return `list(Condition)`. kro flattens the
result, so the instance ends up with one condition per element and the
condition count tracks the collection size. Mapping over a `forEach` collection
resource gives each element's full object, so the per-item status can be read
from the resource itself (field access is type-checked, so a misspelled status
field is rejected at build time):

```yaml
spec:
  schema:
    apiVersion: v1alpha1
    kind: SubnetGroup
    spec:
      name: string
      cidrBlocks: "[]string"
    status:
      conditions:
        - ${subnets.map(s, runtime.newCondition({
            type: 'Subnet-' + s.metadata.name + '-Ready',
            status: s.status.state == 'available' ? 'True' : 'False',
            reason: 'SubnetState',
            message: ''
          }))}
  resources:
    - id: subnets
      forEach:
        - cidr: ${schema.spec.cidrBlocks}
      template:
        apiVersion: ec2.services.k8s.aws/v1alpha1
        kind: Subnet
        metadata:
          name: ${schema.spec.name}-${cidr}
        spec:
          cidrBlock: ${cidr}
          vpcID: vpc-123
```

## What's new

- New CEL functions registered on the `runtime` variable:
  - `runtime.newCondition({type, status, reason, message})` constructs a Condition.
  - `runtime.condition(obj, type)` looks up a condition by type from
    `obj.status.conditions[]`.
- Authors write the bare-identifier syntax
  `runtime.newCondition({type: 'X', status: 'True', ...})`. A parse-time CEL
  macro rewrites identifier keys to string literals and rejects unknown or
  duplicate keys and missing required fields with source-position errors.
- Both functions return a typed Condition, so field access (`cond.status`) is
  type-checked and a misspelled field is rejected at build time rather than at
  runtime.
- Build-time validation rejects expressions where one author condition reads
  back another author condition (`runtime.condition(schema, 'X')` where `X` was
  defined by `runtime.newCondition({type: 'X', ...})` in the same RGD), and
  rejects invalid literal status values.
- Per-condition wire shape includes `lastTransitionTime` (preserved across
  reconciles when the condition's status hasn't changed) and
  `observedGeneration` (set to the instance's current generation).
- When evaluation degrades (one or more author expressions fail or return
  duplicate types), surviving conditions still land on the wire and the
  instance is marked `state=Error`.

## Test plan

- Unit tests pass:
  `go test ./pkg/cel/library/... ./pkg/runtime/... ./pkg/graph/... ./pkg/controller/instance/...`
- Integration tests pass: `go test ./test/integration/suites/core/...` (envtest required)
- Manually applied an RGD with a `conditions:` block; verified author
  conditions appear on `.status.conditions[]` and built-ins are absent.
- Manually triggered a graph-resolve failure; verified author conditions remain
  on the wire and `state=Error` is set.


Skarlso:
Joined 7 hours ago. Created a +3000 line PR with zero existing issues. No KREP, no change discussion and clearly generated content. :) Do you honestly think any of the maintainers would even give this a second thought without immediately pointing you to the contribution guide and the llm policy? :)

lalitg327:
The diff totals ~3,533 lines across 20 files. Here is the breakdown:
Tests: 2,106 lines (60%) across 8 test files.
Production code: ~860 lines across 12 files.
Comments in production files: ~434 lines documenting design choices.
The actual feature implementation is ~860 lines of code; the rest is verification and documentation.

pkg/controller/instance/controller.go
```diff
@@ -123,10 +123,24 @@ type Controller struct {
...
 	metricsEnabled bool
+
+	// hasAuthorConditions reports whether the RGD this controller serves
+	// declared an author `conditions:` block. When true, the early-exit
+	// path on graph-resolve failure (updateConditionsStatus) does not
+	// write kro's built-in conditions to the wire, so .status.conditions[]
+	// only ever contains author-declared conditions. The next successful
+	// reconcile through updateStatus will evaluate the author's CEL and
+	// repopulate the wire.
+	hasAuthorConditions bool
```
michaelhtm, line 131: do we want to make this feature all or nothing?
What if we allowed users define their own conditions, and also let them keep the existing conditions, unless the condition type overlaps?

pkg/controller/instance/controller.go
```diff
@@ -123,10 +123,24 @@ type Controller struct {
...
 }
 
 // NewController constructs a new controller that resolves the newest issued
 // graph revision for the RGD from a GraphRevisionResolver.
+//
+// hasAuthorConditions should be true when the RGD this controller serves
+// declared a `conditions:` block. It gates whether updateConditionsStatus
+// filters kro's built-in conditions from the wire on graph-resolve
+// failures (so authors keep ownership of .status.conditions[]).
```
michaelhtm: nit: is this comment necessary here?

pkg/controller/instance/controller_test.go
```diff
@@ -248,6 +248,7 @@ func TestReconcileGraphResolutionFailureMarksCondition(t *testing.T) {
 				metadata.NewKROMetaLabeler(),
 				newControllerTestCoordinator(t),
 				record.NewFakeRecorder(100),
+				false, // hasAuthorConditions: this test is for RGDs with no author conditions
```
michaelhtm: same here

pkg/graph/builder.go
```diff
@@ -690,11 +703,16 @@ func extractForEachDependencies(
 // buildInstanceNode creates the instance node from pre-computed status components.
 // This is called after spec schema, status schema, and CRD have been built separately.
 // Uses the shared inspectorEnv for AST inspection.
+//
+// conditions are pre-compiled author-defined condition expressions (the
+// optional `conditions:` block). Their dependencies are folded into the
+// instance node's Dependencies so the topological sort accounts for them.
```
michaelhtm: nit: same here

pkg/graph/builder.go
```diff
@@ -775,8 +809,17 @@ func buildInstanceSpecSchema(rgSchema *v1alpha1.Schema) (*extv1.JSONSchemaProps,
...
-// Returns: (schema, fieldDescriptors, statusTemplate, error)
+// using CEL type checking. Uses the shared inspectorEnv for validation and
+// typed env for compilation.
+//
+// If the status block contains a `conditions:` key, its raw expression
+// strings are returned separately (and the key is removed from the inferred
+// status template). Author conditions don't participate in CEL output-type
+// inference; their schema is the standard []metav1.Condition shape, which
+// kro injects via crd.SetCRDStatus' default-conditions behavior when the
+// inferred status has no `conditions` field.
```
michaelhtm: these type comments don't seem to describe what the function does. Can we remove them? or maybe place them somewhere more appropriate?

pkg/runtime/conditions.go
```diff
@@ -0,0 +1,293 @@
...
+			out = append(out, *cond)
+		}
+		return out, nil
+	}
+
+	return nil, fmt.Errorf("condition %q must return a Condition or list(Condition), got %v", exprText, val.Type().TypeName())
+}
+
+// setKeys returns the keys of a string-keyed set as a slice.
+func setKeys(s map[string]struct{}) []string {
```
michaelhtm: is this used anywhere?

pkg/runtime/conditions.go
```diff
@@ -0,0 +1,293 @@
...
+// were returned. The reconcile itself is not failed.
+//
+// logger is used to record skipped expressions and dropped duplicates;
+// pass logr.Discard() in unit tests where logging is irrelevant.
+func (n *Node) EvaluateConditions(logger logr.Logger, kroBuiltins []v1alpha1.Condition) ([]library.Condition, error) {
+	if len(n.conditionExprs) == 0 {
+		return nil, nil
+	}
+
+	ctx := n.buildContext()
```
michaelhtm, line 79: should this ensure we build the context of the resources that conditions depends on?

pkg/runtime/conditions.go
```diff
@@ -0,0 +1,293 @@
...
+		dups = append(dups, t)
+	}
+	// Sort so log output and error messages are deterministic.
+	sortStrings(dups)
+	return kept, dups
+}
+
+// sortStrings sorts in-place. Inlined to avoid adding a sort import for
+// a single use; the slice is always small (≤ #author conditions).
+func sortStrings(s []string) {
```
michaelhtm: is there a reason we don't want to import?

pkg/controller/instance/status.go
```diff
@@ -257,3 +324,86 @@ func (rcx *ReconcileContext) updateInstanceState() {
...
+
+	out := make([]v1alpha1.Condition, 0, len(authored))
+	now := metav1.Now()
+
+	for _, a := range authored {
+		cond := v1alpha1.Condition{
+			Type:               v1alpha1.ConditionType(a.ConditionType),
+			Status:             metav1.ConditionStatus(a.Status),
+			Reason:             stringPtr(a.Reason),
+			Message:            stringPtr(a.Message),
```
michaelhtm, line 353: why not just set these to nil instead of empty string?

pkg/graph/builder.go
```diff
@@ -841,10 +897,115 @@ func buildStatusSchema(
...
+	inspectorEnv *cel.Env,
+	nodeNames []string,
+) ([]*krocel.Expression, error) {
+	if len(conditionExprStrings) == 0 {
+		return nil, nil
+	}
+
+	// Strip ${...} wrappers and produce *krocel.Expression values with
+	// Original set; References and Program are populated below.
+	conditions, err := parser.ParseConditionExpressions(conditionExprStrings)
```
michaelhtm: maybe we can rename `ParseConditionExpressions` here. it is the correct use case, but the name may be confusing. Maybe it should just be `extractExpressions`?

pkg/controller/instance/controller.go
lalitg327: Author opts in to ownership: when a `conditions:` block is present, kro assumes the author wants full control over `.status.conditions[]`. Built-ins remain accessible via `runtime.condition(schema, 'X')`, so an author who wants kro's lifecycle signals on the wire can pull them in explicitly through their own conditions. This keeps the wire surface predictable. Either kro owns it, or the author does.

pkg/controller/instance/controller.go
lalitg327: removed

pkg/controller/instance/controller_test.go
lalitg327: removed comment

pkg/graph/builder.go
lalitg327: removed

pkg/graph/builder.go
lalitg327: agree, removed

pkg/runtime/conditions.go
lalitg327: not used. removed

pkg/runtime/conditions.go
lalitg327: Yes. buildContext() walks n.deps, which for the instance node is populated from Meta.Dependencies in runtime.FromGraph. buildInstanceNode folds each condition expression's expr.References into instanceDeps at build time, so by the time we hit buildContext() here, the resources referenced by author conditions are already in n.deps

pkg/runtime/conditions.go
lalitg327: Replaced with sort.Strings. Was trying to avoid the import for one use. but you're right, the cost is negligible.

pkg/controller/instance/status.go
lalitg327: stringPtr already does this. It returns nil for empty strings

pkg/graph/builder.go
lalitg327: Renamed to UnwrapExpressions. Couldn't reuse ExtractExpressions directly because the same package already has a public function with that name (in cel.go) doing different work.

michaelhtm (approved):

michaelhtm (commented):
/lgtm

pkg/testutil/generator/resourcegraphdefinition.go
```diff
@@ -158,6 +158,49 @@ func WithScope(scope krov1alpha1.ResourceScope) SchemaOption {
...
+// Each element must be a CEL expression string of the form `${...}` returning
+// a Condition (or list(Condition) for collection expansion).
+//
+// Must be applied AFTER WithSchema, since it mutates the same Status raw
+// extension. Caller responsibility: ensure WithSchema appears earlier in the
+// option list.
+//
+// This is a ResourceGraphDefinitionOption (not a SchemaOption) so it can
+// re-encode the Status raw extension after merging the new key in.
+func WithStatusConditions(conditions ...string) ResourceGraphDefinitionOption {
```
a-hilaly: I think WithSchema is enough, no need to add a specialized function that overrides what WithSchema does.

test/integration/suites/core/instance_custom_conditions_test.go
```diff
@@ -0,0 +1,630 @@
...
+		if t, _ := m["type"].(string); t == condType {
+			return m
+		}
+	}
+	return nil
+}
+
+// conditionTypes returns the set of condition.type values on an instance.
+// Useful for asserting which conditions are (and aren't) present.
+func conditionTypes(inst *unstructured.Unstructured) []string {
```
a-hilaly:
```suggestion
// getConditionTypes returns the set of condition.type values on an instance.
// Useful for asserting which conditions are (and aren't) present.
func getConditionTypes(inst *unstructured.Unstructured) []string {
```

pkg/runtime/node.go
```diff
@@ -49,6 +49,11 @@ type Node struct {
 	templateExprs    []*expressionEvaluationState
 	templateVars     []*variable.ResourceField
 
+	// conditionExprs are the per-reconcile evaluation states for the
+	// instance's author-defined `conditions:` expressions. Only the
+	// instance node populates this; resource nodes leave it nil.
+	conditionExprs []*expressionEvaluationState
```
a-hilaly, line 55: Why store condition expressions seperately? can't we just consider them a status variable part of the template?

pkg/runtime/runtime.go
```diff
@@ -164,6 +164,13 @@ func FromGraph(g *graph.Graph, instance *unstructured.Unstructured, rgdConfig gr
 		instNode.templateExprs = append(instNode.templateExprs, state)
 	}
 
+	// Instance condition expressions. Each Expression carries its own
+	// References, populated by the builder.
+	for _, expr := range instNode.Spec.Conditions {
+		state := getOrCreateExpr(expr, variable.ResourceVariableKindDynamic, expr.References)
+		instNode.conditionExprs = append(instNode.conditionExprs, state)
+	}
+
```
a-hilaly: ditto, those are just status variables

pkg/graph/node.go
```diff
@@ -139,6 +139,15 @@ type Node struct {
...
 	ForEach []ForEachDimension
+
+	// Conditions holds compiled CEL expressions for the author-defined
+	// `conditions:` block on the instance (NodeTypeInstance only). Each
+	// expression is expected to return a Condition (or list(Condition)
+	// for collection expansion) at evaluation time.
+	//
+	// Set only on the instance node when the RGD's status block contains
+	// a `conditions:` list. Resource nodes leave this nil.
+	Conditions []*krocel.Expression
```
a-hilaly, line 147: ditto

pkg/graph/node.go
```diff
@@ -160,6 +169,7 @@ func (n *Node) DeepCopy() *Node {
 		IncludeWhen: slices.Clone(n.IncludeWhen),
 		ReadyWhen:   slices.Clone(n.ReadyWhen),
 		ForEach:     slices.Clone(n.ForEach),
+		Conditions:  slices.Clone(n.Conditions),
```
a-hilaly, line 169: ditto

pkg/graph/parser/conditions.go
```diff
@@ -21,10 +21,16 @@ import (
...
+// UnwrapExpressions takes a list of CEL expressions wrapped in ${...} (e.g.
+// readyWhen entries, includeWhen entries, or author condition entries) and
+// returns Expression objects with Original set to the unwrapped inner text.
+// References and Program are populated later by the builder.
+//
+// Each input must be exactly one standalone ${...} — interpolation forms
+// like "prefix-${expr}-suffix" or concatenations like "${a}${b}" are
+// rejected. For interpolation-style strings, see ExtractExpressions in
+// cel.go.
+func UnwrapExpressions(conditions []string) ([]*krocel.Expression, error) {
```
a-hilaly, line 33: why are you renaming this function? was this change intentional?

pkg/graph/builder.go
```diff
@@ -723,6 +737,21 @@ func buildInstanceNode(
...
+	for _, expr := range conditions {
+		for _, ref := range expr.References {
+			if ref == SchemaVarName || ref == library.RuntimeVarName {
+				continue
+			}
+			if !slices.Contains(instanceDeps, ref) {
+				instanceDeps = append(instanceDeps, ref)
+			}
+		}
+	}
```
a-hilaly: I think we shouldn't add a dependency at all. Dedup/filter later on feels artificial. Let's make sure that earlier logic does that.

pkg/graph/builder.go
```diff
@@ -841,10 +886,115 @@ func buildStatusSchema(
...
+	exprs := make([]string, 0, len(rawList))
+	for i, elem := range rawList {
+		s, ok := elem.(string)
+		if !ok {
+			return nil, fmt.Errorf("status.conditions[%d] must be a CEL expression string, got %T", i, elem)
+		}
+		exprs = append(exprs, s)
+	}
+	return exprs, nil
+}
```
a-hilaly, line 932: why are we removing them from the status map? to be just added to it again in runtime?

pkg/graph/builder.go
```diff
@@ -841,10 +886,115 @@ func buildStatusSchema(
...
+
+// buildConditions parses, validates, type-checks, and compiles the author's
+// condition expressions. The returned slice is in the same order as the
+// input. References (the deps each expression touches) are populated on
+// each Expression for use by buildInstanceNode.
+//
+// Returns nil if conditionExprStrings is empty.
+func buildConditions(
```
a-hilaly, line 937: please reduce the amout of comments inside this function, feels like a lot of AI slop

pkg/controller/instance/status.go
```diff
@@ -182,6 +215,40 @@ func (c *Controller) updateStatus(rcx *ReconcileContext) error {
 		}
 	}
 
+	// If the RGD declared author conditions, evaluate them and replace
```
a-hilaly: ditto

pkg/cel/library/runtime.go
```diff
@@ -0,0 +1,506 @@
...
+		// validation layers into one parse-time pass.
+		//
+		// The return type is dyn (rather than the typed Condition) because
+		// the *Condition runtime value already implements traits.Indexer
+		// for .type/.status/.reason/.message field access; declaring a
+		// full struct type would not add expressiveness.
+		cel.Function("newCondition",
+			cel.MemberOverload("runtime_newCondition_map",
+				[]*cel.Type{runtimeType, mapType},
+				cel.DynType,
```
a-hilaly: can you do better than dynType?

pkg/cel/library/runtime.go
```diff
@@ -0,0 +1,506 @@
...
+			return nil, eh.NewError(key.ID(),
+				fmt.Sprintf("runtime.newCondition: unknown key %q (allowed: type, status, reason, message)", name))
+		}
+		if _, dup := seen[name]; dup {
+			return nil, eh.NewError(key.ID(),
+				fmt.Sprintf("runtime.newCondition: duplicate key %q", name))
+		}
+		seen[name] = struct{}{}
+
+		if name == "status" && val.Kind() == ast.LiteralKind {
```
a-hilaly: why is newCondition doing checks on status field? IMO this kind of validation is incorrect at level

pkg/cel/library/runtime.go
```diff
@@ -0,0 +1,506 @@
...
+}
+
+// expectedConditionKeys are the only keys allowed in the input to
+// runtime.newCondition.
+var expectedConditionKeys = map[string]struct{}{
+	"type":    {},
+	"status":  {},
+	"reason":  {},
+	"message": {},
+}
```
a-hilaly, line 100: i would make those consts somewhere

pkg/cel/library/runtime_test.go
a-hilaly, line 1: please keep it to one test file

a-hilaly (commented):
Thanks @lalitg327 - i dropped a few initial comments to shape this PR. Are there any other examples/use cases that you can mention in the description?


@jakobmoellerdev @michaelhtm do you think we need a KREP for this one?

lalitg327:
> Thanks @lalitg327 - i dropped a few initial comments to shape this PR. Are there any other examples/use cases that you can mention in the description?

Thanks @a-hilaly for reviewing the PR. Added a "Why" section and two new examples on top of the basic one: an author-defined Ready composing a kro built-in, and collection expansion (one condition per forEach item). Let me know if you'd like other cases covered

pkg/testutil/generator/resourcegraphdefinition.go
lalitg327: Dropped WithStatusConditions and inlined the conditions directly into the status map of each WithSchema call.

test/integration/suites/core/instance_custom_conditions_test.go
lalitg327: Renamed

pkg/runtime/node.go
lalitg327: templateExprs can't model conditions because each templateExprs entry resolves to one unique path in the template (status.url, metadata.name) and is written there directly. Author conditions are the opposite: many expressions all converge on the single .status.conditions[] slice, and that slice needs dedup-by-type, list-flattening (for .map()), and lastTransitionTime/observedGeneration stamping before it's written. None of that fits the one-expression-one-path template machinery, so conditions get their own conditionExprs list and their own evaluation path (EvaluateConditions).

pkg/runtime/runtime.go
lalitg327: Same reason as on the conditionExprs field declaration: condition results converge on a single shared destination (.status.conditions[]) after dedup and stamping, while templateExprs entries each write to a unique path

pkg/graph/node.go
lalitg327: Same reason I gave in earlier [reply](https://github.com/kubernetes-sigs/kro/pull/1301#discussion_r3437291689) for having a separate Conditions field

pkg/graph/node.go
lalitg327: same reason as [this](https://github.com/kubernetes-sigs/kro/pull/1301#discussion_r3437291689)

pkg/graph/parser/conditions.go
lalitg327: Yes, @michaelhtm asked for this rename. Couldn't reuse ExtractExpressions directly because the same package already has a public function with that name (in cel.go) doing different work.

pkg/graph/builder.go
lalitg327: Moved the runtime CEL identifier filter into extractDependencies (next to the existing SchemaVarName filter). buildInstanceNode now uses extractConditionDependencies and appends the result directly, matching how status-field deps are folded a few lines up

pkg/graph/builder.go
lalitg327: Short answer is that we are not adding them back the same way we took them out. Removing conditions from the status map is a routing decision, not a round trip.

The status map feeds ParseSchemalessResource, which walks the map and turns every ${...} it finds into a status field: one expression, one path, "compute this and write it here." If we leave conditions in the map, the parser recurses into the list and assigns each entry a positional path like conditions[0], conditions[1], and treats it as a normal status field. That breaks in two places. First, schema inference runs over those paths and tries to produce a CRD property literally named conditions[0], which is not a valid field. Second, at runtime softResolve writes each value to its path with UpsertValueAtPath, so a condition lands in a fixed positional slot with no dedup by type, no lastTransitionTime/observedGeneration stamping, and the indices shuffle between reconciles depending on which expressions resolved. It also can't handle the collection case (servers.map(s, runtime.newCondition(...))), where one expression returns many conditions: the whole list gets written into conditions[0] as a nested list instead of being flattened.

This is why we pull conditions out before parsing and hand the expressions to buildConditions instead. At runtime they go through EvaluateConditions, which is the machine that actually understands conditions: it flattens the collection case, dedups by type, and stamps the timestamps and generation before writing to .status.conditions[]. Pulling the key out also leaves the conditions slot absent in the inferred status, so SetCRDStatus injects the standard conditions array schema instead of whatever malformed property inference would have produced from the conditions[0] path.

pkg/graph/builder.go
lalitg327: Trimmed, kept only the comments that explain non-obvious behavior

pkg/controller/instance/status.go
lalitg327: Trimmed

pkg/cel/library/runtime.go
lalitg327: Agreed it was at the wrong layer. The macro is meant to be a pure syntactic rewriter. I pulled the status-value check out of it. Literal status values are now validated in ValidateConditionExpressions alongside the other build-time condition rules

pkg/cel/library/runtime.go
lalitg327: Go can't make a map a const, so I pulled the field keys and status values into const blocks and derived the lookup sets from them. Also replaced the scattered "type"/"status"/... literals across the macro, Get, conditionFromMap, and the provider with the consts so there's one source of truth.

pkg/cel/library/runtime_test.go
lalitg327: Consolidated the runtime CEL feature to one source (runtime.go) and one test (runtime_test.go) file

pkg/cel/library/runtime.go
lalitg327: Updated newCondition and condition to return the typed conditionType instead of dyn. This will help catch misspelled fields (e.g. .stauts) at build time.

jakobmoellerdev (commented):
before this can be merged, you will need to add documentation to the website alongside this feature as it is too large to go undocumented

lalitg327:
> before this can be merged, you will need to add documentation to the website alongside this feature as it is too large to go undocumented

@jakobmoellerdev Thanks for the feedback. Yeah, makes sense to have it documented on the website. I've updated the PR with documentation for custom status conditions.

a-hilaly (approved):
LGTM, overall. @michaelhtm @jakobmoellerdev can you PTAL?

a-hilaly:
and @cheeseandcereal

michaelhtm (approved):
/lgtm

Issue #330, opened 2025-02-27
Skip Kind verification for resources that rely on CRDs that are not yet present in the cluster


alicvsroxas:
### Feature Description

**Problem Statement**:
We run a particular use case where we deploy Kubernetes resources from specific Custom Resource Definitions that are being applied to the cluster from different controller's helm chart.
For example,  a typical flow we do is:
1. Apply helm chart of an Operator
2. Helm chart installs CRD for the resources the operator manages
3. Deploy a resource using these CRDs

When trying to translate the above flow with Kro, 
Kro firsts checks all the resources it needs to apply in the RGDs before allowing to create underlying resources,
the moment it sees a resource that requests a CRD that is not yet in the cluster, it fails the RGD.
Which logically makes sense, because it have no way of knowing I'm creating the CRDs in step 1.

Here is an example RGD, I created that Kro have failed due to this logic above:
```
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: kafka-zookeeper
spec:
  schema:
    apiVersion: v1alpha1
    kind: KafkaZookeeper
    spec:
      name: string
      namespace: string | default="strimzi"
      replicas: integer | default=3
  resources:
    - id: ns
      template:
        apiVersion: v1
        kind: Namespace
        metadata:
          name: ${schema.spec.namespace}
    - id: kafka
      template:
        **# Issue comes from the following apiVersion:**
        apiVersion: kafka.strimzi.io/v1beta2
        kind: Kafka
        metadata:
          name: ${schema.spec.name}
          namespace: ${ns.metadata.name}
        spec:
          kafka:
            version: 3.7.0
            replicas: ${schema.spec.replicas}
     .....
```

**Proposed Solution**:
Wondered if we could add a feature to Kro where we can specify a flag like `skip-verification:true`
or add this as an addition to the feature request of adding [depends on](https://github.com/kro-run/kro/issues/110).
To tell Kro to skip Kind verification if it gets the CRD from a different component in a different RGD.
For example:
* An RGD that deploys an Operator with verification -> will install Kind <kafka.strimzi.io/v1beta2>
* An RGD that deploys a resource using the Kind <kafka.strimzi.io/v1beta2> without verification.
* An RGD that calls both above RGD to create the overall system with a depends on logic.

**Alternatives Considered**:
Way to mitigate this, is getting the CRDs and applying them to the clusters prior to using Kro.
[...]

PR #1302, opened 2026-06-04
KREP-024: Graph — atomic composition primitive

ellistarn:
## Summary

Proposes Graph as a new `kro.run/v1alpha1` Kind — the atomic runtime primitive for composing
Kubernetes resources. Graph provides resource composition alone, separable from Kind definition
and instance management. Higher-level abstractions (RGD, Kind) rebuild from Graph via recursive
nested composition.

Validated by implementing the full RGD system as three levels of nested Graphs (L0 controller →
L1 per-RGD → L2 per-instance), passing 23/36 upstream compatibility tests with remaining gaps
due to prototype immaturity.

## Contents

- `docs/design/proposals/graph.md` — the KREP
- `examples/graph/rgd.yaml` — RGD implemented as nested Graphs
- `examples/graph/singleton.yaml` — fan-in pattern (usage + controller)

## Context

This formalizes work from the [krocodile](https://github.com/ellistarn/kro/tree/krocodile) prototype branch.

### DIFF
--- a/docs/design/proposals/graph.md
+++ b/docs/design/proposals/graph.md
@@ -19,6 +19,49 @@ composition rather than imperative Go code. This implementation passes 23 of 36 
 compatibility test files, with remaining gaps due to prototype immaturity rather than structural
 limitations (see [examples/graph/rgd.yaml](../../../examples/graph/rgd.yaml)).
 
+Beyond the RGD proof, Graph enables patterns that are simpler than what RGD can express today:
+- [Namespace decorator](../../../examples/graph/namespace-decorator.yaml) — watch namespaces, create
+  NetworkPolicies. No CRD, no schema, no instance. (The decorator pattern from KREP-003.)
+- [Ingress fan-in](../../../examples/graph/ingress-fanin.yaml) — aggregate Services into a single
+  Ingress with dynamic routes. (The aggregated-resource pattern from KREP-003.)
+- [KRO installation](../../../examples/graph/kro-install.yaml) — install kro itself as a Graph.
+  The static-bundle pattern: dependency-ordered, health-aware, one object replaces a Helm chart.
+- [Singleton](../../../examples/graph/singleton.yaml) — fan-in with priority-based resolution
+  when multiple actors claim the same resource.
+
+### What this KREP covers
+
+**Proposed:** The Graph Kind — node types (`template`, `patch`, `ref`, `watch`, `def`), dependency
+inference from CEL expressions, status conditions (`Compiled`, `Ready`), self-references, and nested
+composition. These are new primitives that do not exist in KRO today.
+
+**Inherited unchanged from RGD:** `includeWhen`, `forEach`, and CEL expression syntax. These
+mechanisms carry forward with the same semantics and are not redefined by this KREP.
+
+**Directional (defers to respective KREPs):** `propagateWhen` semantics (KREP-006), collection-level
+rollout budgets (KREP-006), and `readyWhen` behavioral differences from the current RGD
+implementation. Sections in this document illustrate how these features compose with Graph's
+recursive structure, but their final API and semantics are defined by their respective KREPs.
+
+**Relationship to RGD:** Graph is proposed as a sibling primitive. RGD continues to work unchanged.
+We have the option to implement RGD's internals on top of Graph in the future, but for the immediate
+term both implementations live as siblings sharing significant code in the underlying graph engine.
+Graph is additive — it does not replace or deprecate RGD.
+
+**Security posture:** Creating a Graph gives its author full control over everything KRO can do —
+creating, patching, deleting, and watching arbitrary Kubernetes resources within the permissions
+granted to the KRO controller's service account. A Graph author operates at the same privilege level
+as today's RGD author: the "infrastructure author" persona who defines what resources exist and how
+they relate.
+
+KRO's dual-persona model is preserved at higher layers. Authors (broad permissions) write
+RGDs/Graphs; consumers (narrow permissions scoped to the CRD the author created) interact only with
+the Kinds those Graphs implement, never with Graph objects directly. Graph is author-facing — end
+users never create or modify Graphs. Future work on credential scoping — short-lived tokens, caller
+credentials (analogous to how CloudFormation uses the caller's IAM role rather than the service's),
+or per-Graph service accounts — applies uniformly to all KRO primitives and is out of scope for this
+KREP.
+
 ## Motivation: Graph as Scope
 
 A Graph is an **isolated scope**. Its nodes form a flat namespace — private, visible only within the
@@ -230,6 +273,10 @@ available). `readyWhen` determines:
 
 ### propagateWhen
 
+> *This section defines `propagateWhen`'s core gating semantics within Graph. Rate-limited rollouts,
+> reactive controls, and manual approval gates are proposed separately in KREP-006. The examples
+> here illustrate composability but do not prescribe the final rollout API.*
+
 `propagateWhen` is the complement to `readyWhen`. Where `readyWhen` signals when a node is healthy,
 `propagateWhen` gates when a node may be evaluated at all. Together they bookend a node's lifecycle:
 `propagateWhen` controls when mutation _can start_; `readyWhen` signals when it is _complete_.
@@ -299,6 +346,9 @@ the effects of items earlier — this is how propagation budgets tighten within 
 pass.
 
 This makes collection-level propagation budgets expressible:
+
+> *The budget pattern below illustrates how `propagateWhen` composes with `forEach`. The rollout
+> API — including built-in strategies and budget syntax — is defined by KREP-006.*
 
 ```yaml
 # Exponential rollout — budget doubles each wave
--- a/docs/design/proposals/graph.md
+++ b/docs/design/proposals/graph.md
@@ -2,11 +2,11 @@
 
 ## Summary
 
-Graph is a new `kro.run/v1alpha1` Kind — the atomic runtime primitive for composing Kubernetes
-resources. A Graph is a set of nodes evaluated in topological order. Create it and its resources
-converge; delete it and they cascade away. It is the simplest possible unit of composition in kro.
+Graph is a new `kro.run/v1alpha1` Kind — the atomic primitive for composing Kubernetes resources. A
+Graph is a set of nodes evaluated in topological order. Create it and its resources converge; delete
+it and they cascade away. It is the simplest possible unit of composition in kro.
 
-RGD today combines Kind definition, instance management, and resource composition into one object.
+RGD today combines CRD management, instance management, and resource composition into one object.
 Graph separates these — providing resource composition alone — making each concern independently
 usable. This enables patterns RGD cannot express: static resource bundles (no CRD needed),
 singletons (multiple contributors, one resource), and decorators (react to existing resources
@@ -60,26 +60,15 @@ the Kinds those Graphs implement, never with Graph objects directly. Future work
 scoping (short-lived tokens, caller credentials, per-Graph service accounts) applies uniformly to
 all KRO primitives and is out of scope for this KREP.
 
-## Motivation: Graph as Scope
+## Graph as a Scope
 
-A Graph is an **isolated scope**. Its nodes form a flat namespace — private, visible only within the
-Graph — with spec as input and status as output. Nothing inside a Graph is reachable from outside
-except through the Kubernetes resource boundary: the Graph object's own spec and status fields.
-
-This isolation has three structural effects:
-
-1. **Concurrency is a consequence of scope boundaries.** A `forEach` over 1000 instances stamps 1000
-   independent Graphs, each with its own scope and lifecycle, each reconciled independently.
-   Parallelism isn't a feature of the evaluation loop — it emerges from the isolation.
-
-2. **Propagation is scope-local.** A `propagateWhen` gate within a child Graph is invisible to the
-   parent. The parent sees only the child's `Ready` condition. Rollout control composes without
-   leaking — a canary strategy inside one instance doesn't slow or block sibling instances.
-
-3. **Communication is explicit.** Graphs talk to each other only through the resources that bind
-   them: the spec of a child Graph stamped by a parent, a `ref` to a resource managed by another
-   Graph, or a `watch` on a Kind whose instances are managed elsewhere. There is no shared memory,
-   no implicit coupling, no cross-scope references.
+A Graph is analogous to a **scope** in traditional programming languages. Its nodes form a flat
+namespace — private, visible only within the Graph — with spec as input and status as output.
+Nothing inside a Graph is reachable from outside except through the Kubernetes resource boundary:
+the Graph object's own spec and status fields. Concurrency is a structural consequence of this
+isolation: a `forEach` over 1000 instances stamps 1000 independent Graphs, each reconciled
+independently. Parallelism isn't a feature of the evaluation loop — it emerges from the scope
+boundary.
 
 ## Proposed API
 
@@ -203,14 +192,11 @@ Pure computation — no Kubernetes resource created or read. The result enters s
       app: ${deployment.metadata.labels['app']}
 ```
 
-#### Everything is a CEL expression
+#### Dynamic GVKs
 
-Every string field in a node body — including `apiVersion`, `kind`, `metadata.name`, and selector
-values — can be a CEL expression. Dynamic evaluation is the general case; statically-known GVKs are
-an optimization that enables compile-time type checking. When the compiler encounters a dynamic GVK,
-it uses a deferred-type path: the node compiles permissively (untyped) until the first reconcile
-resolves the concrete GVK. This is what makes the RGD-from-Graph pattern possible — the user's Kind
-isn't known at compile time.
+A node's `apiVersion`, `kind`, and `metadata.name` can be CEL expressions. This enables patterns
+where the target resource type isn't known at author time — for example, watching instances of a
+Kind defined by an RGD whose schema is only known at runtime:
 
 ```yaml
 - id: watchInstances
@@ -228,11 +214,11 @@ isn't known at compile time.
 stamps a node once per item in a collection. Both retain the same semantics as in today's RGD and
 are not redefined here — see KREP-008 and KREP-002 respectively.
 
-| Modifier | Question it answers | When false | Defined in |
-| --- | --- | --- | --- |
-| `includeWhen` | Should this node exist? | Prune — resource deleted | KREP-008 |
-| `propagateWhen` | May this node mutate now? | Freeze — last-applied state persists | KREP-006 |
-| `readyWhen` | Is this node healthy? | Signal only — Graph not Ready | (inherited from RGD) |
+| Modifier        | Question it answers       | When false                           | Defined in           |
+| --------------- | ------------------------- | ------------------------------------ | -------------------- |
+| `includeWhen`   | Should this node exist?   | Prune — resource deleted             | KREP-008             |
+| `propagateWhen` | May this node mutate now? | Freeze — last-applied state persists | KREP-006             |
+| `readyWhen`     | Is this node healthy?     | Signal only — Graph not Ready        | (inherited from RGD) |
 
 ### Dependencies
 
@@ -287,12 +273,11 @@ KREP-006 for semantics, strategies, and collection-level defaults.
 ### Nested Graphs
 
 A `template:` node can contain any Kind — including Graph itself. Nesting isn't an opt-in feature;
-it emerges from the Kubernetes API. This KREP defines deferral and recursive-compilation semantics
-because nesting cannot be prevented, only left undefined.
+it emerges from the Kubernetes API and Graph semantics.
 
 When a node's template is a Graph, you get nested composition: a parent Graph that stamps a child
-Graph as a real cluster object. The child is persisted, reconciled independently, and carries its own
-conditions and revision history — providing full debuggability via `kubectl get graph`. The
+Graph as a real cluster object. The child is persisted, reconciled independently, and carries its
+own conditions and revision history — providing full debuggability via `kubectl get graph`. The
 combination of `forEach` + template:{kind: Graph} stamps one child Graph per item in a collection.
 This is how an RGD-equivalent is expressed — a per-RGD Graph stamps a per-instance Graph, each with
 its own scope, its own revisions, and its own lifecycle.
@@ -366,33 +351,33 @@ intervention.
 If compilation fails, Ready is False with reason NotCompiled. A single alarm on Ready covers both
 compilation failures and runtime issues.
 
-| Ready Reason | Status  | Meaning                           |
-| ------------ | ------- | --------------------------------- |
-| Ready        | True    | All nodes applied and ready       |
-| NotReady     | Unknown | readyWhen not yet met             |
-| Pending      | Unknown | Waiting for upstream data         |
-| NotCompiled  | False   | Spec is invalid                   |
-| Conflict     | False   | SSA field ownership contested     |
-| Error        | False   | Client request failed (4xx)       |
-| SystemError  | False   | Infrastructure failure (5xx)      |
+| Ready Reason | Status  | Meaning                       |
+| ------------ | ------- | ----------------------------- |
+| Ready        | True    | All nodes applied and ready   |
+| NotReady     | Unknown | readyWhen not yet met         |
+| Pending      | Unknown | Waiting for upstream data     |
+| NotCompiled  | False   | Spec is invalid               |
+| Conflict     | False   | SSA field ownership contested |
+| Error        | False   | Client request failed (4xx)   |
+| SystemError  | False   | Infrastructure failure (5xx)  |
 
 User-defined status does not live on the Graph object. It lives on custom resources and is written
 via `patch:` nodes. The Graph's status contains only controller-managed conditions.
 
 ## Relationship to Existing KREPs
 
-| KREP | Relationship |
-| --- | --- |
-| KREP-001 (Status Conditions) | System conditions (`Compiled`, `Ready`) exist on Graph objects — never on user resources. Users define their own status via `patch:` nodes. |
-| KREP-002 (Collections) | Adopted unchanged. Graph extends it: when a forEach node's template is a Graph, you get recursive composition. |
-| KREP-003 (Decorators) | A Decorator is naturally a Graph with `watch:` + `forEach`. No special runtime support needed. |
-| KREP-006 (Propagation Control) | Graph adopts KREP-006's lifecycle signals and `propagateWhen` gating. Because Graph is recursive, propagation composes at every nesting level. KREP-006 defines the API. |
-| KREP-008 (includeWhen) | Graph implements `includeWhen` as a first-class modifier. Dependency inference works naturally — no special handling needed. |
-| KREP-011 (Variables) | `def:` is Graph's implementation. Same semantics. |
-| KREP-013 (Graph Revisions) | Applies unchanged. Each nested Graph gets independent revisions. |
-| KREP-014 (Resource Lifecycles) | Graph's node types implicitly define lifecycle: `template:` = delete-on-prune, `patch:` = release-fields-on-prune. Per-node lifecycle policies are a natural follow-on. |
-| KREP-018 (Partial Dependencies) | Soft dependencies (`?.`) address the taken-branch problem — the untaken branch uses optional chaining rather than creating a hard gate. |
-| KREP-019 (Deferred Fields) | Graph's `?.` optional chaining is the implementation of this KREP. |
+| KREP                            | Relationship                                                                                                                                                             |
+| ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| KREP-001 (Status Conditions)    | System conditions (`Compiled`, `Ready`) exist on Graph objects — never on user resources. Users define their own status via `patch:` nodes.                              |
+| KREP-002 (Collections)          | Adopted unchanged. Graph extends it: when a forEach node's template is a Graph, you get recursive composition.                                                           |
+| KREP-003 (Decorators)           | A Decorator is naturally a Graph with `watch:` + `forEach`. No special runtime support needed.                                                                           |
+| KREP-006 (Propagation Control)  | Graph adopts KREP-006's lifecycle signals and `propagateWhen` gating. Because Graph is recursive, propagation composes at every nesting level. KREP-006 defines the API. |
+| KREP-008 (includeWhen)          | Graph implements `includeWhen` as a first-class modifier. Dependency inference works naturally — no special handling needed.                                             |
+| KREP-011 (Variables)            | `def:` is Graph's implementation. Same semantics.                                                                                                                        |
+| KREP-013 (Graph Revisions)      | Applies unchanged. Each nested Graph gets independent revisions.                                                                                                         |
+| KREP-014 (Resource Lifecycles)  | Graph's node types implicitly define lifecycle: `template:` = delete-on-prune, `patch:` = release-fields-on-prune. Per-node lifecycle policies are a natural follow-on.  |
+| KREP-018 (Partial Dependencies) | Soft dependencies (`?.`) address the taken-branch problem — the untaken branch uses optional chaining rather than creating a hard gate.                                  |
+| KREP-019 (Deferred Fields)      | Graph's `?.` optional chaining is the implementation of this KREP.                                                                                                       |
 
 ## Future Work
 
--- a/docs/design/proposals/graph.md
+++ b/docs/design/proposals/graph.md
@@ -54,17 +54,6 @@ We have the option to implement RGD's internals on top of Graph in the future, b
 term both implementations live as siblings sharing significant code in the underlying graph engine.
 Graph is additive — it does not replace or deprecate RGD.
 
-**Security posture:** A Graph author operates at the same privilege level as today's RGD author: the
-"infrastructure author" persona who defines what resources exist and how they relate. Graph authors
-have full control over everything KRO can do within the permissions granted to the KRO controller's
-service account.
-
-KRO's dual-persona model is preserved at higher layers. Authors (broad permissions) write
-RGDs/Graphs; consumers (narrow permissions scoped to the CRD the author created) interact only with
-the Kinds those Graphs implement, never with Graph objects directly. Future work on credential
-scoping (short-lived tokens, caller credentials, per-Graph service accounts) applies uniformly to
-all KRO primitives and is out of scope for this KREP.
-
 ## Proposed API
 
 ### The Graph Kind
@@ -115,10 +104,9 @@ cannot evaluate until the Deployment has been applied and its observed state is 
 
 ### Node Types
 
-Graph has five node types. Each is declared by a keyword on the node object. The types are split
-along two axes: **operation** (own, contribute, read, compute) and **scope shape** (single object
-vs. list). We choose explicit keywords over fewer types with strategy flags so that a reader knows
-what a node does — and what type it produces in scope — at a glance from the top-level keyword.
+Graph has five node types. Each is declared by a keyword on the node object. We choose explicit
+keywords over fewer types with strategy flags so that a reader knows what a node does — and what
+type it produces in scope — at a glance from the top-level keyword.
 
 #### `template:`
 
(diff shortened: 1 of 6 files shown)


examples/graph/rgd.yaml
```diff
@@ -0,0 +1,428 @@
...
+                  priority: 1
+                  jsonPath: .status.topologicalOrder
+                - name: AGE
+                  type: date
+                  jsonPath: .metadata.creationTimestamp
+
+    # ─── L0: Watch all RGD objects (gated on CRD being established) ──────────
+    - id: watchRgds
+      includeWhen:
+        - ${rgdCrd.?status.?conditions.orValue([]).exists(c, c.type == 'Established' && c.status == 'True')}
```
Skarlso, line 123: Oh this will solve this: https://github.com/kubernetes-sigs/kro/issues/330

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,512 @@
+# KREP-024: Graph
+
+## Summary
+
+Graph is a new `kro.run/v1alpha1` Kind — the atomic runtime primitive for composing Kubernetes
```
Skarlso: If i understand the premise properly, this will also enable parallel processing of individual layers because of the recursive processing nature of them and the topological sorting, right?

ellistarn: Yes! Added some language around how this parallelism emerges.

Skarlso: Nice! Thank you! :) 🙇

pnz1990:
This is a really strong direction. The RGD-as-Graph example is the part that makes it click for me: Graph is not just a smaller API surface, it looks like the right primitive underneath the current abstractions.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,519 @@
...
+RGD today conflates Kind definition, instance management, and resource composition into one object.
+Graph separates these — providing resource composition alone — making each concern independently
+usable. This enables patterns RGD cannot express: static resource bundles (no CRD needed),
+singletons (multiple contributors, one resource), and decorators (react to existing resources without
+defining a new Kind). Higher-level abstractions like RGD are built _from_ Graph, not alongside it;
+every feature added to Graph automatically benefits every abstraction layered on top.
+
+Graph was validated by building the full RGD system from it. A single Graph implements the RGD
+controller — creating CRDs, watching instances, managing resources, writing status — all through
+composition rather than imperative Go code. This implementation passes 23 of 36 upstream kro
```
pnz1990, line 23: This is a great validation point. It would be useful to add a small taxonomy for the remaining 13 failures: prototype gaps, unsupported existing behavior, behavior intentionally changed by Graph, etc. That would make it easier to tell whether the remaining work is implementation depth or API design risk.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,519 @@
...
+
+The ownerReference triggers self-deletion when the instance enters Terminating. The `patch:` node
+holds the instance until teardown completes — all template resources are pruned in reverse DAG
+order, then the patch releases the finalizer as its last act, and the instance finishes deletion.
+This pattern is load-bearing for the RGD implementation: each per-instance Graph (L2) holds a
+finalizer on its instance, ensuring no managed resources are orphaned on deletion.
+
+## How RGD Emerges from Graph
+
+The RGD system is three levels of nested Graphs:
```
pnz1990, line 311: The recursive model is compelling, but it raises the obvious performance question: do we have a target latency/API-call budget for the RGD shape at scale? For example, 1000 instances with L0/L1/L2 reconciliation in play.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+## Motivation: Graph as Scope
+
+A Graph is an **isolated scope**. Its nodes form a flat namespace — private, visible only within the
+Graph — with spec as input and status as output. Nothing inside a Graph is reachable from outside
+except through the Kubernetes resource boundary: the Graph object's own spec and status fields.
+
+This isolation has three structural effects:
+
+1. **Concurrency is a consequence of scope boundaries.** A `forEach` over 1000 instances stamps 1000
```
Skarlso: I feel like I'm missing something here. If I understand correct what it means that the healthy is no longer a gate to evaluation it means that it will not wait for an object to become healthy, but for the data to be available and that's supposed to make a lot of things a lot faster so we don't need to think about problems during evaluating a potentially huge graph and it's topologically sorted layers.

But, if I do understand this correctly...

Data is being observed instead of health. But in most cases, data follows health closely doesn't it? And if in the other cases, data was already there because it was provided through the Spec field, wouldn't that have been not a problem in the previous implementation as well? If I understand the flow in the previous iteration it's sequential as well so it was waiting on health not data. But the data almost arrived at the same time as health unless the health condition was waaay too broad. But even if it was way too broad, it would have meant that all the data in the status was required after all.

So what I'm saying here is that I don't understand what the gain here is supposed to be? Waiting for the data doesn't seem to be that big of an improvement and the serial nature of the process would yet again fall into the same loop of reconcile, wait, reconcile, wait, reconcile, wait... Right?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+Defines raw data for use by other nodes. No Kubernetes resource is created or read — `def:` is pure
+computation. The result enters scope under the node's `id` like any other node.
+
+```yaml
+- id: naming
+  def:
+    prefix: ${deployment.metadata.name + '-' + deployment.metadata.namespace}
+    labels:
+      app: ${deployment.metadata.labels['app']}
```
Skarlso: Oh this is nice. These are reusable values, right?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+```
+
+`def:` is Graph's implementation of the variables concept proposed in KREP-011. Graphs that compose
+at scale — like the RGD implementation — require named intermediate computations to remain
+maintainable; `def:` provides them without creating cluster resources.
+
+#### `includeWhen` and `forEach`
+
+Graph retains `includeWhen` and `forEach` as node-level modifiers with the same semantics as in
+today's RGD. `includeWhen` conditionally excludes a node (making it a prune candidate when false).
```
Skarlso: I understand what you want to say here but the English is weird. :D 

~~~
`includeWhen` conditionally excludes a node
~~~
-> would be nicer as conditionally includes, not?

docs/design/proposals/graph.md
Skarlso: This comes from me reading this part: https://github.com/kubernetes-sigs/kro/pull/1302/changes#diff-df9de290b47bfc9b438bad0d71ded6a824033fa9cdcdceaf3af8d2329d69f7b4R205

Skarlso: Further... `forEach` here says forEach over a 1000 instances, each reconciled independently. But `forEach` later is explained to be sequential. Which one is it? :) Or is that a different forEach?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+# Parent strips outer ${}, child Graph (L1) evaluates the inner expression:
+group: ${${rgd.spec.schema.group}}
+```
+
+This composes to arbitrary depth. `${${${...}}}` defers two levels.
+
+#### Recursive Compilation
+
+When the child template is a Graph, the compiler extracts the child spec, strips one deferral level,
```
Skarlso: Except for dynamic GVK types, right?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+Every string field in a node body — including `apiVersion`, `kind`, `metadata.name`, and selector
+values — can be a CEL expression. This enables dynamic GVKs:
+
+```yaml
+- id: watchInstances
+  watch:
+    apiVersion: ${${schema.spec.schema.group}}/${${schema.spec.schema.apiVersion}}
+    kind: ${${schema.spec.schema.kind}}
+    selector: {}
```
Skarlso: I assume there will be documentation on how to provide properly scoped RBAC for dynamic types. :)

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+### readyWhen
+
+`readyWhen` is a list of CEL boolean expressions evaluated against the node's live state in the
+cluster. When all are true, the node is considered _ready_.
+
+```yaml
+- id: deployment
+  readyWhen:
+    - ${deployment.status.availableReplicas == deployment.spec.replicas}
```
Skarlso: What happens if fields are empty? Since we are "waiting on data" and not ready I assume we fail or retry or provide some kind of default when fields are empty? This isn't discussed here unless I missed it. :)

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+Graph — with spec as input and status as output. Nothing inside a Graph is reachable from outside
+except through the Kubernetes resource boundary: the Graph object's own spec and status fields.
+
+This isolation has three structural effects:
+
+1. **Concurrency is a consequence of scope boundaries.** A `forEach` over 1000 instances stamps 1000
+   independent Graphs, each with its own scope and lifecycle, each reconciled independently.
+   Parallelism isn't a feature of the evaluation loop — it emerges from the isolation. Within a
+   single scope, serial evaluation is correct and sufficient: CEL evaluates in microseconds, each
+   node is one API call, and the API server — not the graph — is the throughput constraint.
```
Skarlso: Errr... at least two, right? GET and Apply, not?

Skarlso (commented):
Couple of questions remain in my mind, otherwise very nice! :)

docs/design/proposals/graph.md
ellistarn: > Further... forEach here says forEach over a 1000 instances, each reconciled independently. But forEach later is explained to be sequential. Which one is it? :) Or is that a different forEach?

There are some implementation options here, but I was imagining that a single graph executes serially. i.e., the nested graphs are printed one by once. But once they are created, they naturally are evaluated in parallel by KRO, since they are just normal graphs.

docs/design/proposals/graph.md
ellistarn: Yeah exactly. I am a bit conflicted on the name `def` vs `let`.

docs/design/proposals/graph.md
ellistarn: Agree

docs/design/proposals/graph.md
ellistarn: So for a dynamic GVK, you're blocked on compiling until the type is known. As soon as that type enters scope, the graph has the opportunity to validate. This is strictly more powerful than KRO is today, and was necessary to implement the RGD controller in pure YAML.

docs/design/proposals/graph.md
ellistarn: I see RBAC as orthogonal scope, but yes, we need to solve this :)

docs/design/proposals/graph.md
ellistarn: an empty readyWhen is just immediately ready after resolved and applied.

docs/design/proposals/graph.md
ellistarn: Yeah -- in practice, once the graph is converged, it's just a single get. There is a whole world of [optimizations](https://github.com/ellistarn/kro/blob/krocodile/experimental/docs/design/007-optimizations.md) I explored, but I consider out of scope for this KREP.

docs/design/proposals/graph.md
ellistarn: I didn't spend too much time on this because the performance is (in theory) identical to how KRO scales today.

docs/design/proposals/graph.md
ellistarn: > Data is being observed instead of health. But in most cases, data follows health closely doesn't it? And if in the other cases, data was already there because it was provided through the Spec field, wouldn't that have been not a problem in the previous implementation as well? If I understand the flow in the previous iteration it's sequential as well so it was waiting on health not data. But the data almost arrived at the same time as health unless the health condition was waaay too broad. But even if it was way too broad, it would have meant that all the data in the status was required after all.

Consider defining a deployment and a service object. For convenience you set the service's selector to the deployment's selector. You want to define a readyWhen on the deployment so that the graph is not considered ready until the replicas are available. In KRO's current semantic, you'd be forced to wait for the deployment to become ready before the service could even be created.

Imagine you are orchestrating EC2 Instances and EC2 Network Interfaces. You can create and attach the NetworkInterfaces as soon as the Instance's ID is known (<1 sec after creation). You don't want to wait for the instance to become ready, and realistically, the readiness of the instance might depend on the network interface's set.

This pattern emerges broadly. Gating on data enables the graph to flow as quickly as possible, and draw wait conditions wherever is strictly necessary.

Skarlso: Ahha, I think I get it. You could somewhat get into this by defining loose readyWhen conditions on both nodes though, right? If I understand correctly.

docs/design/proposals/graph.md
Skarlso: I really like `let`, if only it wouldn't trigger my Javascript PTSD. :DDDD

docs/design/proposals/graph.md
Skarlso: Got it.

docs/design/proposals/graph.md
Skarlso: Yes, but the field will be empty... so does that fail to parse?

docs/design/proposals/graph.md
Skarlso: Ah got it, okay.

docs/design/proposals/graph.md
ellistarn: `var` is an option as well :), it's just slightly off since vars are typically mutable.

docs/design/proposals/graph.md
ellistarn: Yes, but then you lose the true readiness signal. The graph's readiness signal is derived by the sum of its nodes. Systems like Helm actually watch these signals when using `--wait`, so it is semantically quite meaningful to get the ordering correct. There is more discussion on decoupling readyWhen and propagateWhen in https://github.com/kubernetes-sigs/kro/pull/861

docs/design/proposals/graph.md
ellistarn: I'm  confused. `readyWhen: []` proceeds immediately. `readyWhen: [""]` is a compiliation error (must be a boolean)

docs/design/proposals/graph.md
Skarlso: Okay makes sense. Thanks!!

docs/design/proposals/graph.md
Skarlso: I think that's what I was trying to get at. 🤣 That being a compile time error is good enough I believe.

docs/design/proposals/graph.md
Skarlso: Fun fact. :D I did some research because this is really interesting topic. So turns out that def is actually rarely used for variable definition in most programming languages. It's mostly used to define functions, function types and things you can call literally so `def bla` would be `bla()` usually.

And `let` wins out overwhelmingly if you want to define a _variable_ or some piece of reusable data. It's used by javascript, rust, swift, ocaml, f#, haskell, elm, scheme, lisp, nix, cue, dhall...

However! Go templates, as we know, uses `define` keyword for reusable pieces of template bits. ` {{ define "bla" }}`. So, if this is a piece of reusable template that is parameterized and called after that, then `def` sounds good. :) So `def` defines something that can be invoked and `let` defines something that you use.

ellistarn: Completely open to `let` or `var` or `const`

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+```yaml
+- id: watchInstances
+  watch:
+    apiVersion: ${${schema.spec.schema.group}}/${${schema.spec.schema.apiVersion}}
+    kind: ${${schema.spec.schema.kind}}
+    selector: {}
+```
+
+When the compiler encounters a dynamic GVK, it uses a deferred-type path: the node compiles
```
NicholasBlaskey: what is a dynamic GVK? 

Any cel expression? 

Any cel expression with a variable in it?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+    metadata:
+      name: my-webapp
+    status:
+      endpoint: ${service.status.loadBalancer.ingress[0].hostname}
+      ready: ${deployment.status.availableReplicas > 0}
+```
+
+#### `ref:`
+
+Reads a single named resource into scope. No write, no ownership, no cleanup. The resource is
```
cheeseandcereal: I'm curious about semantic choice of separating ref and watch. Don't these effectively do the same thing, but with different selection mechanisms? Or am I missing something else.

If it is just the selection mechanism that differs, I guess I would ask if it would make sense to combine these two.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+#### Why ref and watch replace externalRef
+
+`externalRef` in today's RGD overloads a single field to mean "read one named resource" and "watch a
+collection by selector." `ref:` and `watch:` make the operation — and the scope type (object vs
+list) — visible at a glance from the top-level keyword.
+
+#### Everything is a CEL expression
+
+Every string field in a node body — including `apiVersion`, `kind`, `metadata.name`, and selector
```
cheeseandcereal: Just string fields? What about numbers/objects/other types of fields/etc?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+Graph has five node types. Each is declared by a keyword on the node object.
+
+#### `template:`
+
+Creates and manages a Kubernetes resource. On creation the resource is applied via server-side
+apply. On deletion of the Graph, the resource is deleted. This is the direct equivalent of a
+`template:` resource in today's RGD.
+
+#### `patch:`
```
michaelhtm, line 116: will patch allow users to access the data on the existing resource using CEL?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+a per-instance Graph, each with its own scope, its own revisions, and its own lifecycle.
+
+#### Deferral Boundaries
+
+Child Graph CEL expressions live as literal strings inside the parent's template. The parent
+compiler sees them as opaque data, not as CEL. The evaluation model uses a nesting convention to
+manage this:
+
+- `${...}` — evaluated at the current level
+- `${${...}}` — the outer `${}` is stripped (producing a literal `${...}` string), evaluated one
```
NicholasBlaskey: I understand this syntax and think it is somewhat elegant

I still think this is very cryptic on first reading / trying to understand the behavior here

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+| KREP-018 (Partial Dependencies)            | CEL branching (`${cond ? A.field : B.field}`) and Graph's soft dependency mechanism (`?.`) together address partial dependency scenarios. Soft deps allow referencing a node without creating a hard gate — the taken-branch problem becomes less acute when the untaken branch uses optional chaining.                                                                                    |
+| KREP-019 (Deferred Fields & Soft Deps)     | Graph's `?.` optional chaining is the implementation of this KREP. Soft deps allow progressive field population across reconcile passes.                                                                                                                                                                                                                                                   |
+
+## Future Work
```
NicholasBlaskey, line 390: I don't have a good sense of how this proposal fits into the current project

Is this just additive and new resource? Are we planning to implement RGD on a graph? Are most users going to be using RGD or graph?

docs/design/proposals/graph.md
ellistarn: This should say "every field in the node body". String is just impl detail here, since CEL requires it to be a string, even if it evaluates to another type -- that said, the existing cases are all just strings.

docs/design/proposals/graph.md
ellistarn: I haven't run into a use case for this, but I don't see why not.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+Two conditions on orthogonal axes. Both use standard Kubernetes condition fields;
+`lastTransitionTime` is preserved when status is unchanged; `observedGeneration` advances
+independently per condition.
+
+**`Compiled`** — is the spec valid? Set once when the spec is processed, permanent until the spec
+changes. True on success; False on expression errors, dependency cycles, or malformed node
+declarations.
+
+**`Ready`** — has the graph converged? `True` means all nodes are applied and their `readyWhen`
```
cheeseandcereal: So the only way to control control custom "ready" condition, is to use `readyWhen` expressions (in combination with the other signals that role into this. I think this makes sense, I just want to be clear that this will be treated in a "special" way compared to any other (user-defined) status.

So: Modifying ready condition = modify readyWhen on the same graph object
Modifying any other condition = create different graph patch object referencing this one

docs/design/proposals/graph.md
ellistarn: Agree - @a-hilaly has some improvements using scope shadowing. This is just what falls out naturally from how the graph is evaluated.

docs/design/proposals/graph.md
ellistarn: I suspect that users will use multiple interfaces, depending on what they're trying to do. The 4 main patterns I've needed are in the intro (graph, singleton, decorator, and graphdefinition/rgd/kind/choose-your-name-for-the-templating-pattern-here)

examples/graph/rgd.yaml
```diff
@@ -0,0 +1,428 @@
...
+            - id: checks
+              def:
+                resources: "${${schema.?spec.?resources.orValue([])}}"
+                reserved: '${${["apiVersion","context","dependency","dependencies","each","externalRef","externalReference","externalRefs","externalReferences","graph","instance","item","items","kind","kro","metadata","namespace","object","resource","resourcegraphdefinition","resourceGraphDefinition","resources","root","runtime","schema","self","serviceAccountName","spec","status","this","variables","vars","version"]}}'
+                allIDs: "${${schema.?spec.?resources.orValue([]).map(r, r.id)}}"
+
+            - id: validation
+              def:
+                result: >-
+                  ${${
```
NicholasBlaskey, line 168: this example shows a very high level of complexity here to me.

I think a risk here is that we provide users something that technically can do everything but in practice ends up being akin to writing pure functional programming languages like haskell. I do think there is a market for haskell, but the market for something simpler like Python seems larger

We can say just use RGD for users that want a simpler experience, but I do worry if we implement a bunch of features in graph RGD will just turn into a getting starting experience instead of the main way users interact with Kro.

docs/design/proposals/graph.md
ellistarn: As defined, yes. There is an open question that's not covered in this doc as to whether or not we should make .ready() respect resources that have a `Ready` condition, as that could simplify integration a bit. It's just sugar, but I think it's semantically correct.

examples/graph/rgd.yaml
ellistarn: Agree -- I suspect that for the nesting use case, users will come in at higher order abstractions. Not dissimilar from programming languages + their standard libraries.

docs/design/proposals/graph.md
ellistarn: Happy to discuss -- this was more for my / claude's own sanity when reading.

docs/design/proposals/graph.md
ellistarn: Any cel expression that resolves to a string.

docs/design/proposals/graph.md
cheeseandcereal: Is this necessarily true? Because a lot of the DAG state was being internally computed and only saved in memory before, while they're now broken out into their own persisted separate k8s resources for each graph object now, right?

The things to process doesn't change, but it seems like we are doing a lot more management of the DAG state within etcd that was previously happening exclusively in memory.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+Creates and manages a Kubernetes resource. On creation the resource is applied via server-side
+apply. On deletion of the Graph, the resource is deleted. This is the direct equivalent of a
+`template:` resource in today's RGD.
+
+#### `patch:`
+
+Contributes fields to a resource you don't own. The resource must already exist — `patch:` does not
+create it. Fields are applied via server-side apply with a distinct field manager. On prune, the
+field claims are released (the field manager is removed), but the resource is not deleted.
```
cheeseandcereal: We are implicitly adopting the SSA-behavior around field management with this. I think this is worth calling out. For example, adding a patch could add a field to an object, some other system could touch that field, and now removing this patch and removing the field owner doesn't necessarily remove the field. This is intentional behavior as defined by SSA, but isn't the most intuitive. If we want to adopt all those quirks, we should just do it with eyes wide open.

ellistarn: I am open minded to alternatives, but since KRO uses SSA today, I kept that as the basis of this feature for the purposes of the doc.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+without defining a new Kind). Higher-level abstractions like RGD are built _from_ Graph, not
+alongside it; every feature added to Graph automatically benefits every abstraction layered on top.
+
+Graph was validated by building the full RGD system from it. A single Graph implements the RGD
+controller — creating CRDs, watching instances, managing resources, writing status — all through
+composition rather than imperative Go code. This implementation passes 23 of 36 upstream kro
+compatibility test files, with remaining gaps due to prototype immaturity rather than structural
+limitations (see [examples/graph/rgd.yaml](../../../examples/graph/rgd.yaml)).
+
+## Motivation: Graph as Scope
```
JamesGAWS: Graph looks like it's evolving KRO's mission from kubernetes resource orchestration to a lower level yaml/CEL based DSL that is linked to kubernetes.  What are the tradeoffs? Can we develop the motivation section a bit more to explore this evolution of KRO's purpose? It's not clear to me at the moment how data, mission, and action lead to Graph as the best solution.

docs/design/proposals/graph.md
jakobmoellerdev, line 116: IMHO we should think about merging template and patch.

I would suggest to have template as a combination of what to apply and then a strategy that defines how to apply (i.e. Apply but dont create, or Apply and allow Create). Same for pruning

docs/design/proposals/graph.md
jakobmoellerdev: ref is different in that it is a single resource reference. we will need to decide if kro should watch single objects in the future. the current dynamic reconciler opionates on this too and does watch them. as such im suggesting to have them both merged into a single watch primitive that works for both selectors and single objects

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+#### `includeWhen` and `forEach`
+
+Graph retains `includeWhen` and `forEach` as node-level modifiers with the same semantics as in
+today's RGD. `includeWhen` conditionally excludes a node (making it a prune candidate when false).
+`forEach` stamps a node once per item in a collection. These are not redefined here — see KREP-008
+and KREP-002 respectively.
+
+#### Why ref and watch replace externalRef
+
+`externalRef` in today's RGD overloads a single field to mean "read one named resource" and "watch a
```
jakobmoellerdev: to me this was never an issue tbh, not sure why we want to change this. the most important part here is that there is no efficient concept for single resource watches anyways so the lower level functionality needs to watch more specifically anyhow

docs/design/proposals/graph.md
jakobmoellerdev: I would argue that everything is anyhow dynamic and we just check if there is a prerequisite calculation that needs to happen

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+- id: deployment
+  readyWhen:
+    - ${deployment.status.availableReplicas == deployment.spec.replicas}
+  template:
+    apiVersion: apps/v1
+    kind: Deployment
+    ...
+```
+
+**readyWhen is a health signal. It does not gate downstream nodes.**
```
jakobmoellerdev: how would you now gate downstream nodes? with includeWhen => ready() ?

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+proceeds as soon as dependencies are _in scope_ (have been applied and their observed state is
+available). `readyWhen` determines:
+
+1. Whether `.ready()` returns true for this node in CEL expressions
+2. Whether the **Graph itself** is Ready — the Graph's Ready condition is the conjunction of all
+   nodes' `readyWhen` results
+
+### propagateWhen
+
+`propagateWhen` is the complement to `readyWhen`. Where `readyWhen` signals when a node is healthy,
```
jakobmoellerdev: propagateWhen is semantically equivalent to includeWhen in many cases and it needs distinction (see above) Its very hard for me to make a modelling recommendation based on this

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+When any `propagateWhen` expression evaluates to false, the node and all its dependents are not
+re-evaluated — they remain in their previous state. When all expressions are true, evaluation
+proceeds normally. The default is `[]` (no gate — evaluate immediately). Where `includeWhen: false`
+prunes a node (deleting its resource), `propagateWhen: false` freezes it — the resource persists in
+its last-applied state.
+
+### Self-References
+
+Before evaluating a node's expressions, kro GETs the target resource from the API server. The live
```
jakobmoellerdev: I dont understand this entire section, probably needs some more concrete rewording

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+pass.
+
+This makes collection-level propagation budgets expressible:
+
+```yaml
+# Exponential rollout — budget doubles each wave
+- id: deployments
+  forEach:
+    - app: ${apps}
+  propagateWhen:
```
jakobmoellerdev: here propagateWhen is used on the deployments node (itself) and above it is used for another node. when to use include vs propagate for these cases?

docs/design/proposals/graph.md
jakobmoellerdev: I think this should be scoped out into a separate proposal. Child Graphs can also be included by referral imho and its not a hardblocker. If we do implement it we should introduce scopes like @a-hilaly suggested, but for me this can be split from the main proposal tbh

docs/design/proposals/graph.md
jakobmoellerdev: In theory having more idomatic primitives with Graphs makes it much more efficient for caching. Historically we have problems with this in the current codebase. I think its important to define caching / invalidation principles here and then we can hash out exact implementation later.

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,497 @@
...
+
+**`Compiled`** — is the spec valid? Set once when the spec is processed, permanent until the spec
+changes. True on success; False on expression errors, dependency cycles, or malformed node
+declarations.
+
+**`Ready`** — has the graph converged? `True` means all nodes are applied and their `readyWhen`
+conditions pass. `Unknown` means the controller is still making progress. `False` means something
+requires human intervention.
+
+Compiled rolls into Ready: if the spec fails compilation, Ready is `False` with reason
```
jakobmoellerdev: Sorry but this just reads like one huge AI section, not gonna read into this without rewording

docs/design/proposals/graph.md
jakobmoellerdev: This needs more attention. We basically have to rewire half of the codebase to make this reasonable. I also dont see any way to migrate from RGDs effectively if im honest (which is acceptable but a boundary condition for me)

jakobmoellerdev (commented):
While there are certainly details to hash out I believe this has the chance to become our saving grace for the simple fact that we have continuous maintenance issues on the current primitives and it has become quite hard to push for extensive features in KRO.

Im +1 for the initiative

docs/design/proposals/graph.md
ellistarn: I started here and weighed a single `template` + `strategy` based approach, same for `ref` vs `watch`. Ultimately I was convinced by the readability of having separate keys. I tried it out, and Claude immediately stopped making category error mistakes on the nodes. It's worth a verbal debate, I think, but leaving as is for now.

docs/design/proposals/graph.md
ellistarn: Adding some language that covers these differences in use case. Though ultimately I think we probably need this to live in KREP 006.

docs/design/proposals/graph.md
ellistarn: I'll cut this -- this was for enabling future use cases like status conditions, time, etc.

docs/design/proposals/graph.md
ellistarn: #861

docs/design/proposals/graph.md
ellistarn: Let's debate!

khaled:
FWIW I'm a big fan of this KREP and can barely wait for it to land.  I've wrestled with the existing KRO implementation quite a bit and quickly ran up against its limits.  This seems to address all/most of the issues I've had in one fell swoop.

jakobmoellerdev (commented):
/approve

docs/design/proposals/graph.md
```diff
@@ -0,0 +1,410 @@
...
+These mechanisms carry forward with the same semantics and are not redefined by this KREP.
+
+**Defers to KREP-006:** `propagateWhen` gating semantics, `.ready()` and `.updated()` lifecycle
+signals, collection-level rollout strategies, and budget syntax. Graph supports all of these; their
+API and behavior are defined by KREP-006.
+
+**Relationship to RGD:** Graph is proposed as a sibling primitive. RGD continues to work unchanged.
+We have the option to implement RGD's internals on top of Graph in the future, but for the immediate
+term both implementations live as siblings sharing significant code in the underlying graph engine.
+Graph is additive — it does not replace or deprecate RGD.
```
justinsb, line 55: I think this is the right call, adding more code and keeping RGD on the v1 train.

I do think we need a clearer use-case for _why_ Graph then, e.g. what use cases are we unblocking.  It will help people decide whether they want to use Graph or RGD (or even Graph vs controller-runtime), and help the developers to decide what is in and out of scope

ellistarn: I tried to balance keeping the KREP focused while also pointing at the motivation. The singleton/decorator examples I posted in this PR are a good demonstration for why. I also cover this in the intro:

> This enables patterns RGD cannot express: static resource bundles (no CRD needed),
singletons (multiple contributors, one resource), and decorators (react to existing resources
without defining a new Kind).

---

Personally, I am using the raw graph to replace helm in a lot of my workflows. Instead of rendering a set of resources from a chart, I'm using graph to do that. They're all one-offs, so I don't need RGD's define-and-then-instantiate pattern. The key bit I need is a single thing that can manage all of the resources of an install/uninstall/update and graph does that perfectly. The other pattern I'm using heavily is decorator, for things like labelling nodes and defaulting resources on namespace creation (see the example).

Is there something specific you're looking for in the motivation beyond that?

examples/graph/rgd.yaml
jlbutler: +1 to Nick's concern. The primitive is powerful and obviously something we should pursue. But the RGD example is kind of horrifying... while we may assume we wrap it up in a stdlib, that doesn't yet exist and is only implied here. 

I'd feel a lot better if the higher-level abstraction (Kind, or whatever we land on) were at least sketched out - could be follow-up work, not blocking the direction, just being realistic about Graph being a primitive truly. I'm concerned we're pointed at shipping assembly, hoping the compilers and SDKs show up later.

docs/design/proposals/graph.md
jlbutler: I'm leaning toward Jakob's take here. Fewer primitives feels better directionally (template/patch feel effectively like apply, with or without create+own... ref/watch seems even more clearcut (read one vs read many).

We don't need to resolve this in one go, but acknowledging there is an open question on API shape for a verbal discussion does seem correct.

docs/design/proposals/graph.md
jlbutler: My attempt to summarize this thread lands a mild concern for me;  agreed this in theory gives us more granularity and opportunities for efficiency, but I'd like to call this out directly.  I can't see how the performance would be identical to how the current implementation scales, other than in pure theoretical terms. For posterity, it might be worth naming the concerns and speaks to how we'd plan to measure and monitor going forward.

docs/design/proposals/graph.md
jlbutler: +1 agree with Jakob. This is a boundary condition not a detail IMO. Sharing an engine between similar siblings is safe framing, but the RGD-on-graph makes the migration/parity callout tangible.  Not blocking on direction but another spot to maybe drive clarity around what we're stepping into?

jlbutler:
I agree with consensus on direction, and am inclined to land this as a working proposal. Graph looks to be a flexible, expressive, and portable primitive that broadly extends the value of kro. 

KREPs should be well thought-out and defensible, you're solidly checking boxes there. They also can adapt as design work proceeds, and this one will adapt no doubt. +1 aligned with Jakob's take, approval of the direction, not an endorsement of a complete spec.

The central tension persists. We need a way to freely evolve, and Graph looks to be it.  At the same time, we can't have two full engines, indefinitely driving two experiences within the project. We need these to converge. I'm more comfortable making this a stated goal, so for posterity will record it here. 

Thank you for the contribute @ellistarn !

/lgtm

PR #1304, opened 2026-06-04
fix(cel): accept whole-valued float64 for integer-typed schema fields

### LINKED ISSUE #1303: UnstructuredToVal rejects float64 input for type: integer schema fields (asymmetric with type: number)
### Description

**Observed Behavior**:

`UnstructuredToVal` in `pkg/cel/unstructured/valunstructured.go` rejects every value that arrives as `float64` when the destination OpenAPI schema declares `type: integer`. The integer switch only handles Go `int`, `int32`, `int64`:

```go
if schema.Type() == "integer" {
    switch v := unstructured.(type) {
    case int:
        return types.Int(v)
    case int32:
        return types.Int(v)
    case int64:
        return types.Int(v)
    default:
        return types.NewErr("invalid data, expected int, got %T", unstructured)
    }
}
```

In practice every JSON number coming out of `encoding/json` (the default decoder behind `unstructured.Unstructured`) is `float64`. So any integer-typed CRD field flowing through this adapter triggers the error message `invalid data, expected int, got float64`.

The `type: number` branch immediately above is asymmetric — it explicitly accepts both ints and floats and returns a `types.Double`.

**Expected Behavior**:

For a `type: integer` schema, accept a `float64` whose value is a whole number representable as `int64`, returning `types.Int`. Continue to reject NaN, infinities, and non-integer floats explicitly so no silent truncation occurs. This is consistent with PR #1080's stated intent, quoting its description:

> "This ensures CEL runtime values match their compile-time types — Secret data evaluates as bytes, timestamps as proper CEL timestamps, **integers as ints rather than floats**, etc."

The current integer branch never reaches that goal in the common path because the input it sees from JSON-decoded resources is `float64`, not `int*`.

**Reproduction Steps**:

The simplest reproduction is via `crossplane-contrib/function-kro@v0.2.1`, which embeds `kro@v0.9.0` code. The same bug is present on `main` (commit `6df2c7b` at the time of filing).

1. Define an XRD with an `integer` field, e.g. `numNodeGroups: { type: integer }`.
2. Drive an XR through a Composition that uses `function-kro` and references the field as `${schema.spec.numNodeGroups}` (or any integer-typed field that flows through `UnstructuredToVal`).
3. The composition reconciler reports:

   ```
   cannot get desired state for resource "primary": node "primary":
   eval "schema.spec.numNodeGroups": invalid data, expected int, got float64
   ```

4. User-side workarounds — `${int(schema.spec.numNodeGroups)}` — don't help because `UnstructuredToVal` runs at value-adapter time (when CEL reads the input), before any user CEL expression is evaluated.

**Versions**:
- kro version: bug reproduced via `function-kro@v0.2.1` (bundles upstream `kro@v0.9.0`); present on `kubernetes-sigs/kro@main` (commit `6df2c7b`).
- Kubernetes Version: AWS EKS 1.34 (any modern version exhibits the same JSON decoding behavior).

**Involved Controllers**:
- Crossplane v2.x with `crossplane-contrib/function-kro:v0.2.1` in `crossplane-system/function-kro-*`.
- The same code path also affects the standalone kro controller, since the file is shared.

**Error Logs** (sample from a real reconcile):
```
cannot compose resources: pipeline step "render" returned a fatal result:
cannot get desired state for resource "primary":
node "primary": eval "schema.spec.?numNodeGroups.orValue(1)":
invalid data, expected int, got float64
```

Common Expression Language (CEL)

RanMarkovich:
Fixes #1303.

UnstructuredToVal rejected every value arriving as float64 when the destination OpenAPI schema declared type: integer. The integer switch only handled Go int / int32 / int64 - but every JSON number decoded by encoding/json (the default behind unstructured.Unstructured) is float64. So any integer-typed CRD field that flowed through this adapter triggered:

` invalid data, expected int, got float64`

This is asymmetric with the number branch immediately above, which accepts both ints and floats.

This completes the work of #1080, whose stated goal was to make "integers as ints rather than floats" - by handling the case where the incoming value is already a float64 from JSON decoding.

Accept float64 if and only if it is a whole number that fits in int64; reject NaN, infinities, and non-integer floats explicitly so no silent truncation can occur.

Tests: `TestUnstructuredToVal_Integer` extended with a `float64 whole` case; new `TestUnstructuredToVal_Integer_RejectsNonInteger` for fractional / NaN / ±Inf / string. No change for existing `int / int32 / int64` callers.

### DIFF
--- a/pkg/cel/unstructured/valunstructured.go
+++ b/pkg/cel/unstructured/valunstructured.go
@@ -21,6 +21,7 @@ package unstructured
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"sync"
 	"time"
@@ -180,6 +181,14 @@ func UnstructuredToVal(unstructured interface{}, schema common.Schema) ref.Val {
 			return types.Int(v)
 		case int64:
 			return types.Int(v)
+		case float64:
+			// encoding/json decodes every JSON number as float64; accept whole
+			// values that fit in int64 so integer-typed CRD fields work end-to-end.
+			// Reject non-integer floats explicitly rather than silently truncating.
+			if !math.IsNaN(v) && !math.IsInf(v, 0) && v == math.Trunc(v) && v >= math.MinInt64 && v <= math.MaxInt64 {
+				return types.Int(int64(v))
+			}
+			return types.NewErr("invalid data, expected int, got non-integer float %v", v)
 		default:
 			return types.NewErr("invalid data, expected int, got %T", unstructured)
 		}
--- a/pkg/cel/unstructured/valunstructured_test.go
+++ b/pkg/cel/unstructured/valunstructured_test.go
@@ -18,6 +18,7 @@ package unstructured
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


jakobmoellerdev (commented):
/ok-to-test

jakobmoellerdev (commented):
/lgtm
/approve

PR #1305, opened 2026-06-05
pkg/graph: expose options for NewBuilder()

neolit123:
i am building a "kro aware" controller that works with [kcp](https://github.com/kcp-dev/kcp), but i faced a blocker where the function NewBuilder() had to be forked to work with this setup. instead of forking it to support kcp and aggregated api servers, i propose this minimal backwards compatible change to NewBuilder().

(copied from the commit message)

```
Expose the resolver and rest mapper of NewBuilder()
as options following a 'builder' pattern.

In NewBuilder(), NewCombinedResolver() returns
a resolver which is entirely /openapi based.
Passing a custom schema resolver is useful
when the API types being referenced
in the RGD are not served from the cluster's standard
/openapi discovery endpoint e.g. types served
through kcp virtual workspaces or aggregated API servers.

A custom rest mapper is useful when resource discovery
should not use the host cluster's standard discovery
endpoint (e.g. vs a kcp virtual workspaces) or in tests
where a fake mapper is preferred over a live
API server call.

Add minimal UT updates too.
```

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -49,23 +49,32 @@ import (
 	"github.com/kubernetes-sigs/kro/pkg/simpleschema"
 )
 
-// NewBuilder creates a new GraphBuilder instance.
-func NewBuilder(clientConfig *rest.Config, httpClient *http.Client) (*Builder, error) {
-	schemaResolver, err := schemaresolver.NewCombinedResolver(clientConfig, httpClient)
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (commented):
/ok-to-test

jakobmoellerdev (approved):
/lgtm

PR #1309, opened 2026-06-16
fix: reject CEL bytes-typed expressions at RGD compile time

### LINKED ISSUE #1308: Instance controller panics ("cannot deep copy …") on an RGD whose CEL expression resolves to a non-JSON-representable type
## Description

**Observed Behavior:**

A `ResourceGraphDefinition` whose status field or resource-template field uses a CEL expression that resolves to a value outside the set of types a Kubernetes unstructured object can hold causes the kro controller to panic with `panic: cannot deep copy <type>`. The RGD is accepted and its CRD is installed, but the first time an instance is reconciled the controller panics while deep-copying the resolved object.

The two values that trigger this are CEL **`uint`** and CEL **`bytes`**, as well as any `list`, `map`, or `optional` that contains them. For example:

- `uint`: an unwrapped `uint(...)` conversion.
- `bytes`: an unwrapped hash call — `hash.sha256(...)`, `hash.md5(...)`, `hash.fnv64a(...)`.

**Root cause:**

kro's `GoNativeType` (`pkg/cel/conversion/conversions.go`) converts CEL output into Go values. The resolved object is then deep-copied through apimachinery's `runtime.DeepCopyJSONValue` (`unstructured.Unstructured.DeepCopy()` for resource templates; `unstructured.NestedMap` → `runtime.DeepCopyJSON` for status). `DeepCopyJSONValue` only accepts JSON-representable Go types:

```
string, int64, bool, float64, nil, json.Number, map[string]interface{}, []interface{}
```

Anything else hits its `default` case and panics. Two of the Go types `GoNativeType` can emit are not in that set:

| CEL output type | Go type from `GoNativeType` | Deep-copy result |
|---|---|---|
| `bool` | `bool` | ✅ ok |
| `int` | `int64` | ✅ ok |
| `double` | `float64` | ✅ ok |
| `string` | `string` | ✅ ok |
| `duration` / `timestamp` | `string` | ✅ ok |
| `list(int)` | `[]interface{}` | ✅ ok |
| **`uint`** | **`uint64`** | ❌ **panic: cannot deep copy uint64** |
| **`bytes`** | **`[]byte`** | ❌ **panic: cannot deep copy []uint8** |
| **`optional_type(uint)`** | **`uint64`** | ❌ **panic** |
| **`list(uint)` / `list(bytes)`** | `[]interface{}{…}` | ❌ **panic** |
| **`map(string, uint)` / `map(string, bytes)`** | `map[string]interface{}{…}` | ❌ **panic** |

**Expected Behavior:**

An RGD with a field or status expression that resolves to a non-representable type should be **rejected at build time** (e.g. `GraphAccepted=False` with an actionable message), the same way condition expressions are already required to return `bool`. It must never produce an installable CRD or reach the controller. Authors who want a hash or an unsigned value in their object should wrap it to return a representable type, e.g. `base64.encode(hash.sha256(...))` (string) or `int(...)`.

Two things make this more subtle than a single hardcoded type check:

1. **It is more than one type.** The check must cover the full non-representable set — `uint`, `bytes`, and any `list`/`map`/`optional` containing them — not a single type, ideally derived from the same JSON-representable allowlist that `DeepCopyJSONValue`/`json.Marshal` enforce, so the two cannot drift apart.

2. **The static CEL type is not the runtime Go value.** A field with schema `format: byte` is typed as CEL `bytes`, but at runtime its value arrives as a base64 **string** from the unstructured object. An expression like `${secret.data.token}` is statically `bytes` yet resolves to a valid string and works today — it must **not** be rejected. So the validation cannot simply reject on "static CEL type is `bytes`/`uint`"; it needs to reason about the value that will actually be produced, and/or be paired with a runtime coercion/guard in the conversion path so a value that slips past static typing is coerced or fails gracefully rather than panicking.

### Example 1 — `uint` in a status field

`ResourceGraphDefinition`:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: uintbug
spec:
  schema:
    apiVersion: v1alpha1
    kind: UintBug
    spec:
      name: string
    status:
      digest: ${uint(configmap.metadata.name.size())}   # CEL uint -> Go uint64
  resources:
    - id: configmap
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.spec.name}
        data:
          key: value
```

`Instance`:

```yaml
apiVersion: kro.run/v1alpha1
kind: UintBug
metadata:
  name: demo
  namespace: default
spec:
  name: my-config
```

### Example 2 — `bytes` in a resource-template field

`ResourceGraphDefinition`:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: bytesbug
spec:
  schema:
    apiVersion: v1alpha1
    kind: BytesBug
    spec:
      name: string
  resources:
    - id: configmap
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.spec.name}
          annotations:
            digest: ${hash.sha256(schema.spec.name)}   # CEL bytes in a template field
        data:
          key: value
```

`Instance`:

```yaml
apiVersion: kro.run/v1alpha1
kind: BytesBug
metadata:
  name: demo
  namespace: default
spec:
  name: my-config
```

Steps (either example):

```console
$ kubectl apply -f rgd.yaml        # RGD is accepted, CRD installed
$ kubectl apply -f instance.yaml   # first reconcile panics
$ kubectl get pods -n kro-system   # controller restarts
```

## Versions

- kro version: 0.9.2
- Kubernetes Version (`kubectl version`): not version-specific; also reproduces in unit/envtest

## Involved Controllers

- kro instance controller (`pkg/controller/instance`)

## Error Logs (if applicable)

```
panic: cannot deep copy uint64        # Example 1 (or: cannot deep copy []uint8 for Example 2)

k8s.io/apimachinery/pkg/runtime.DeepCopyJSONValue(...)
    .../apimachinery/pkg/runtime/converter.go:646
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.NestedMap(...)
    .../apimachinery/pkg/apis/meta/v1/unstructured/helpers.go:235
github.com/kubernetes-sigs/kro/pkg/controller/instance.(*Controller).updateStatus(...)
    .../pkg/controller/instance/status.go:176
```

cheeseandcereal:
Fixes #1308

## Problem

An RGD whose status (or resource template) field uses a CEL expression that evaluates to the CEL **`bytes`** type (e.g. an unwrapped `hash.sha256(...)`/`hash.md5(...)`/`hash.fnv64a(...)`) is accepted and installs its CRD, but the first instance reconcile panics with `cannot deep copy []uint8` and crashloops the controller. Since the panic is in the shared dynamic-controller worker, a single malformed RGD takes down the controller for every RGD it serves. Full repro, panic stack, and root-cause walkthrough are in #1308.

In short: `bytes` resolves to a Go `[]byte`, which is not a valid JSON/unstructured value and panics in apimachinery's `DeepCopyJSONValue` when the resolved status is deep-copied during `updateStatus`.

## Fix

Reject `bytes`-typed expressions at **RGD compile/validation time**, mirroring the existing `validateConditionExpression` (bool) precedent, instead of letting the value reach the runtime and crash the controller. The error surfaces as `GraphAccepted=False` on the RGD (no controller changes needed), so a malformed RGD is rejected cleanly and never produces an installable CRD or instance.

We deliberately do **not** change `GoNativeType`: it is a shared conversion helper used on multiple paths, and silently re-encoding bytes (e.g. to base64) there would mask the user error and change behavior for other callers. Validation belongs at the type-checking layer.

Changes:

- **`pkg/cel/conversion/conversions.go`**: add `IsBytesOrOptionalBytes(t *cel.Type) bool`, mirroring `IsBoolOrOptionalBool`.
- **`pkg/graph/builder.go`**: reject bytes at the two output-type chokepoints:
  - `buildStatusSchema` (status fields are type-*inferred*, so they aren't caught by the expected-type check).
  - `validateExpressionType` (covers all resource template fields).

  Both return a clear, actionable error:

  ```
  ... returns type "bytes", which cannot be represented in a Kubernetes object;
  wrap it in base64.encode(...) or string(...)
  ```

Well-formed RGDs are unaffected; the idiomatic `${base64.encode(hash.sha256(...))}` (which returns `string`) continues to work.

## Behavior change

Using the repro from #1308 (`status.digest: ${hash.sha256(configmap.metadata.name)}`):

**Before:** the RGD is accepted, the CRD is installed, and the first instance reconcile panics and crashloops the controller.

**After:** the RGD is rejected at build time and never installs a CRD or touches the controller:

```console
$ kubectl get rgd hashbug -o jsonpath='{.status.state}'
Inactive

$ kubectl get rgd hashbug \
    -o jsonpath='{range .status.conditions[*]}{.type}={.status}: {.message}{"\n"}{end}'
GraphAccepted=False: failed to build instance status schema: status expression \
  "hash.sha256(configmap.metadata.name)" at path "digest" returns type "bytes", \
  which cannot be represented in a Kubernetes object; wrap it in \
  base64.encode(...) or string(...)
Ready=False: failed to build instance status schema: ...
```

Wrapping the expression so it returns a `string` (`${base64.encode(hash.sha256(configmap.metadata.name))}`) makes the RGD `Active` and the instance reconciles with a populated status:

```console
$ kubectl get hashok demo -o jsonpath='{.status.digest}{"  "}{.status.state}'
MnFaN4sAFC2DRpyox4IxI6gl9UB7bt6OlUuUq2fb67o=  ACTIVE
```

## Tests

- **`pkg/cel/conversion/conversions_test.go`**: `TestIsBytesOrOptionalBytes` (bytes / optional(bytes) -> true; string/int/bool/list-of-bytes -> false).
- **`pkg/graph/builder_test.go`**: added to the CEL type-checking table:
  - status field returning bytes -> rejected;
  - resource field returning bytes -> rejected;
  - status field with `base64.encode(hash.sha256(...))` -> accepted (positive control).
- **`test/integration/suites/core/status_test.go`** (envtest): an RGD with a bytes status field becomes `Inactive` with `Ready=False` carrying the "cannot be represented in a Kubernetes object" message.

Confirmed before the fix: a controller-level test driving a `${hash.sha256(...)}` status field through `updateStatus` panics with `cannot deep copy []uint8`. After the fix the RGD is rejected at build time and the controller is never handed the bytes value.

```
make test WHAT=unit            # pass
make test WHAT=integration     # pass (Ran 1 of 139, new spec included)
make lint                      # 0 issues
```

## Live validation (real EKS cluster)

Built the patched image with `ko`, pushed to ECR, deployed via the Helm chart to an EKS 1.34 cluster, and exercised both paths:

- **Negative (the bug):** the bytes-status RGD produced `Inactive`/`GraphAccepted=False` with the expected message; **no CRD created; controller pod stayed `Running` with 0 restarts and no panic in logs.**
- **Positive:** the `base64.encode(...)` RGD produced `Active`/`GraphAccepted=True`; an instance reconciled to `ACTIVE` with `status.digest` populated as the correct base64 SHA-256 string, verified against an independent computation.

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -835,7 +835,15 @@ func buildStatusSchema(
 			return nil, nil, nil, fmt.Errorf("failed to type-check status expression %q at path %q: %w", expression.UserExpression(), fieldDescriptor.Path, err)
 		}
 
-		statusTypeMap[fieldDescriptor.Path] = checkedAST.OutputType()
+		outputType := checkedAST.OutputType()
+		if conversion.IsBytesOrOptionalBytes(outputType) {
+			return nil, nil, nil, fmt.Errorf(
+				"status expression %q at path %q returns type %q, which cannot be represented in a Kubernetes object; wrap it in base64.encode(...) or string(...)",
+				expression.UserExpression(), fieldDescriptor.Path, outputType.String(),
+			)
+		}
+
+		statusTypeMap[fieldDescriptor.Path] = outputType
 	}
 
 	// convert the CEL types to OpenAPI schema - best effort.
@@ -1227,6 +1235,17 @@ func validateAndCompileTemplates(
 // validateExpressionType verifies that the CEL expression output type matches
 // the expected type. Returns an error if there is a type mismatch.
 func validateExpressionType(outputType, expectedType *cel.Type, expression, resourceID, path string, typeProvider *krocel.DeclTypeProvider) error {
+	// CEL `bytes` cannot be represented in a Kubernetes object: it resolves to a
+	// Go []byte and panics ("cannot deep copy []uint8") in apimachinery's
+	// DeepCopyJSONValue when the resolved object is deep-copied or persisted.
+	// Reject it here rather than letting it crash the controller at runtime.
+	if conversion.IsBytesOrOptionalBytes(outputType) {
+		return fmt.Errorf(
+			"expression %q at path %q in resource %q returns type %q, which cannot be represented in a Kubernetes object; wrap it in base64.encode(...) or string(...)",
+			expression, path, resourceID, outputType.String(),
+		)
+	}
+
 	// Try CEL's built-in nominal type checking first
 	if expectedType.IsAssignableType(outputType) {
 		return nil
--- a/pkg/cel/conversion/conversions.go
+++ b/pkg/cel/conversion/conversions.go
@@ -176,3 +176,21 @@ func IsBoolOrOptionalBool(t *cel.Type) bool {
 	optionalBool := cel.OptionalType(cel.BoolType)
 	return optionalBool.IsAssignableType(t)
 }
+
+// IsBytesOrOptionalBytes checks if a CEL type is bytes or optional_type(bytes).
+// CEL `bytes` (e.g. the result of hash.sha256/hash.md5/hash.fnv64a) cannot be
+// represented in a Kubernetes unstructured object: it resolves to a Go []byte,
+// which is not a valid JSON value and panics ("cannot deep copy []uint8") in
+// apimachinery's runtime.DeepCopyJSONValue when the object is deep-copied or
+// persisted. kro therefore rejects bytes-typed field/status expressions at RGD
+// compile time instead of letting them crash the controller at runtime.
+func IsBytesOrOptionalBytes(t *cel.Type) bool {
+	// Note: A.IsAssignableType(B) means "A accepts B", so we check if bytes accepts t.
+	if cel.BytesType.IsAssignableType(t) {
+		return true
+	}
+
+	// Check if it's optional_type(bytes)
+	optionalBytes := cel.OptionalType(cel.BytesType)
+	return optionalBytes.IsAssignableType(t)
+}
--- a/pkg/cel/conversion/conversions_test.go
+++ b/pkg/cel/conversion/conversions_test.go
@@ -225,3 +225,25 @@ func TestGoNativeType_Timestamp(t *testing.T) {
 	assert.NoError(t, err, "Should be JSON marshallable")
 	assert.NotEmpty(t, marshalled)
 }
+
+func TestIsBytesOrOptionalBytes(t *testing.T) {
+	tests := []struct {
+		name string
+		t    *cel.Type
+		want bool
+	}{
+		{name: "bytes", t: cel.BytesType, want: true},
+		{name: "optional bytes", t: cel.OptionalType(cel.BytesType), want: true},
+		{name: "string", t: cel.StringType, want: false},
+		{name: "optional string", t: cel.OptionalType(cel.StringType), want: false},
+		{name: "int", t: cel.IntType, want: false},
+		{name: "bool", t: cel.BoolType, want: false},
+		{name: "list of bytes", t: cel.ListType(cel.BytesType), want: false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, IsBytesOrOptionalBytes(tt.t))
+		})
+	}
+}
--- a/pkg/graph/builder_test.go
+++ b/pkg/graph/builder_test.go
@@ -2217,6 +2217,81 @@ func TestGraphBuilder_CELTypeChecking(t *testing.T) {
 			wantErr: true,
 			errMsg:  "type mismatch",
 		},
+		{
+			name: "status field returning bytes is rejected",
+			resourceGraphDefinitionOpts: []generator.ResourceGraphDefinitionOption{
+				generator.WithSchema(
+					"Test", "v1alpha1",
+					map[string]interface{}{
+						"name": "string",
+					},
+					map[string]interface{}{
+						"digest": "${hash.sha256(configmap.metadata.name)}",
+					},
+				),
+				generator.WithResource("configmap", map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "${schema.spec.name}",
+					},
+					"data": map[string]interface{}{
+						"key": "value",
+					},
+				}, nil, nil),
+			},
+			wantErr: true,
+			errMsg:  "cannot be represented in a Kubernetes object",
+		},
+		{
+			name: "resource field returning bytes is rejected",
+			resourceGraphDefinitionOpts: []generator.ResourceGraphDefinitionOption{
+				generator.WithSchema(
+					"Test", "v1alpha1",
+					map[string]interface{}{
+						"name": "string",
+					},
+					nil,
+				),
+				generator.WithResource("configmap", map[string]interface{}{
+					"apiVersion": "v1",
+					"kind":       "ConfigMap",
+					"metadata": map[string]interface{}{
+						"name": "test-config",
+					},
+					"data": map[string]interface{}{
+						"digest": "${hash.sha256(schema.spec.name)}",
+					},
+				}, nil, nil),
+			},
+			wantErr: true,
+			errMsg:  "cannot be represented in a Kubernetes object",
+		},
+		{
+			name: "status field with base64-encoded hash is accepted",
+			resourceGraphDefinitionOpts: []generator.ResourceGraphDefinitionOption{
+				generator.WithSchema(
+					"Test", "v1alpha1",
+					map[string]interface{}{
+						"name": "string",
+					},
+					map[string]interface{}{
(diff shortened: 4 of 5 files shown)


pkg/graph/builder.go
```diff
@@ -1227,6 +1235,17 @@ func validateAndCompileTemplates(
 // validateExpressionType verifies that the CEL expression output type matches
 // the expected type. Returns an error if there is a type mismatch.
 func validateExpressionType(outputType, expectedType *cel.Type, expression, resourceID, path string, typeProvider *krocel.DeclTypeProvider) error {
+	// CEL `bytes` cannot be represented in a Kubernetes object: it resolves to a
+	// Go []byte and panics ("cannot deep copy []uint8") in apimachinery's
+	// DeepCopyJSONValue when the resolved object is deep-copied or persisted.
+	// Reject it here rather than letting it crash the controller at runtime.
+	if conversion.IsBytesOrOptionalBytes(outputType) {
```
NicholasBlaskey, line 1242: This seems like it is also an issue for other types and not just bytes

uints in status also panic, probably a bunch others that would fail too
```
 apiVersion: kro.run/v1alpha1
  kind: ResourceGraphDefinition
  metadata:
    name: uintbug
  spec:
    schema:
      apiVersion: v1alpha1
      kind: UintBug
      spec:
        name: string
      status:
        digest: ${uint(configmap.metadata.name.size())}   # returns CEL uint -> Go uint64 -> panics
    resources:
      - id: configmap
        template:
          apiVersion: v1
          kind: ConfigMap
          metadata:
            name: ${schema.spec.name}
          data:
            key: value
```

pkg/graph/builder.go
```diff
@@ -1227,6 +1235,17 @@ func validateAndCompileTemplates(
 // validateExpressionType verifies that the CEL expression output type matches
 // the expected type. Returns an error if there is a type mismatch.
 func validateExpressionType(outputType, expectedType *cel.Type, expression, resourceID, path string, typeProvider *krocel.DeclTypeProvider) error {
+	// CEL `bytes` cannot be represented in a Kubernetes object: it resolves to a
+	// Go []byte and panics ("cannot deep copy []uint8") in apimachinery's
+	// DeepCopyJSONValue when the resolved object is deep-copied or persisted.
+	// Reject it here rather than letting it crash the controller at runtime.
```
NicholasBlaskey, line 1241: This might be too broad, it seems like this specifically is an issue for unstructured types but some other types can be ok with taking bytes

This passes before this commit, but fails the validation
```
 # /tmp/rgd-bytes-raw.yaml
  apiVersion: kro.run/v1alpha1
  kind: ResourceGraphDefinition
  metadata:
    name: bytesraw
  spec:
    schema:
      apiVersion: v1alpha1
      kind: BytesRaw
      spec:
        name: string
      status:
        raw: ${secret.data.token}      # <-- CEL bytes (format: byte field)
    resources:
      - id: secret
        template:
          apiVersion: v1
          kind: Secret
          metadata:
            name: ${schema.spec.name}
          type: Opaque
          data:
            token: aGVsbG8td29ybGQ=
```

edit: this example seems to be more along with resolved typing of cel expressions but I do wonder about the check being too broad too

pkg/graph/builder.go
```diff
@@ -1227,6 +1235,17 @@ func validateAndCompileTemplates(
...
+	if conversion.IsBytesOrOptionalBytes(outputType) {
+		return fmt.Errorf(
+			"expression %q at path %q in resource %q returns type %q, which cannot be represented in a Kubernetes object; wrap it in base64.encode(...) or string(...)",
+			expression, path, resourceID, outputType.String(),
+		)
+	}
```
NicholasBlaskey, line 1247: unrelated, but worth calling out there is a separate arguably more pressing bug where a panic in the dynamic controller crashes the whole process

pkg/graph/builder.go
cheeseandcereal: Yeah this is a good callout. There probably needs to be a more general improvement to the type checking, not specifically for bytes

pkg/graph/builder.go
cheeseandcereal: Yeah, this is a good callout, and should probably be tracked separately. I cut a separate issue for this:

https://github.com/kubernetes-sigs/kro/issues/1310

pkg/graph/builder.go
cheeseandcereal: I'm going to close this PR based on this comment. I have updated the related issue with more investigation, highlighting this is a generic type checking problem, and laying out some of the considerations for implementation that would make more sense there. If/when the issue needs to be addressed, it can be restarted from there.

PR #1313, opened 2026-06-19
Implementation of KREP-024 - Graph CRD

a-hilaly:
Implements KREP-024 (#1302): the Graph Kind and its controller — the compile-and-execute engine that composes Kubernetes resources from a single Graph CRD. A Graph is a set of nodes (templates, refs, watches, defs, and nested graphs) wired by CEL references; the controller derives a dependency DAG, statically type-checks expressions against target schemas, and reconciles in topological order with managed-resource tracking.

The stable RGD controller and its packages are untouched.


Skarlso:
Please correct me if I'm wrong, but technically the KREP has not been accepted yet and still has comments that need to be adjusted and addressed. So will you just keep updating this PR with the findings that the KREP surfaces as more comments come in? And what is krokodile and how is +80.000 lines of code related to that single KREP? :)

jakobmoellerdev:
KREP is now accepted but this needs heavy editorial changes, should we keep a long running branch or a separate code path open?

jakobmoellerdev:
/close 

we will still follow up on this but we need to break this down into more manageable PRs without as much code duplication. We will start on getting this in place with some first PRs building on these ideas here

PR #1315, opened 2026-06-26
docs(website): fix broken relative links in overview and resource-basics

numb86:
Internal links are broken on https://kro.run/docs/overview.

## Reproduction Steps

- Open https://kro.run/docs/overview
- Click the link labeled `schema` under the `How it works` section
- Attempt to navigate to https://kro.run/docs/overview/concepts/rgd/schema, resulting in a 404 error
    - The intended destination should be https://kro.run/docs/concepts/rgd/schema/

## Fix

Specifying the filename as the link destination will correctly link to the intended page.
The same fix is applied to both the current docs and the version-0.9.2 snapshot.
I verified this by building locally.

### DIFF
--- a/website/docs/docs/overview.mdx
+++ b/website/docs/docs/overview.mdx
@@ -24,10 +24,10 @@ runtime failures.
 
 ## How it works
 
-A ResourceGraphDefinition has two parts: a [**schema**](./concepts/rgd/schema) that defines your API
-surface (the fields users fill in), and [**resource templates**](./concepts/rgd/resource-definitions/resource-basics) that reference
-those fields with [CEL expressions](./concepts/rgd/cel-expressions). kro parses the expressions, infers the
-[dependency graph](./concepts/rgd/dependencies-ordering), generates a CRD, and stands up a controller - all at runtime.
+A ResourceGraphDefinition has two parts: a [**schema**](./concepts/rgd/01-schema.md) that defines your API
+surface (the fields users fill in), and [**resource templates**](./concepts/rgd/02-resource-definitions/01-resource-basics.md) that reference
+those fields with [CEL expressions](./concepts/rgd/03-cel-expressions.md). kro parses the expressions, infers the
+[dependency graph](./concepts/rgd/04-dependencies-ordering.md), generates a CRD, and stands up a controller - all at runtime.
 
 <IntroFlow />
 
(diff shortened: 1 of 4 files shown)


website/docs/docs/overview.mdx
```diff
@@ -24,10 +24,10 @@ runtime failures.
 
 ## How it works
 
-A ResourceGraphDefinition has two parts: a [**schema**](./concepts/rgd/schema) that defines your API
-surface (the fields users fill in), and [**resource templates**](./concepts/rgd/resource-definitions/resource-basics) that reference
-those fields with [CEL expressions](./concepts/rgd/cel-expressions). kro parses the expressions, infers the
-[dependency graph](./concepts/rgd/dependencies-ordering), generates a CRD, and stands up a controller - all at runtime.
+A ResourceGraphDefinition has two parts: a [**schema**](./concepts/rgd/01-schema.md) that defines your API
```
cheeseandcereal, line 27: I'm not sure this actually fixes the issue. I believe the issue you're encountering is actually due to the fact that these are _relative_ redirects.

If you are looking at the kro docs from https://kro.run/docs/overview/ - clicking on these links are broken, because the generated link is something like https://kro.run/docs/overview/concepts/rgd/schema (which is not a real page).

If you click around on the sidebar of the docs page, and go back to the first page, it takes you to the URL: https://kro.run/docs/overview (note NO trailing slash). From there, if you click on the `schema` link, it actually _does_ work (taking you to https://kro.run/docs/concepts/rgd/schema which is the valid page).

cheeseandcereal: I just double checked and it looks like this change _would_ resolve the issue. Mainly because:

docusaurus treats these links as literal since they aren't a specific file (i.e. `href:"./concepts/rgd/schema"`). Linking them to a literal file, causes docusaurus to resolve these links at build time, and makes them absolute (i.e. `href:"/docs/concepts/rgd/schema"`)

Are there other areas that should be considered for a similar update? I think we should avoid adding inconsistency in the way we are handling these links in the doc files.

website/docs/docs/overview.mdx
```diff
@@ -162,8 +162,8 @@ in custom controller code.
 
 ## Get started
 
-- [Install kro](./getting-started/Installation) to run the controller in your cluster
-- [Deploy your first RGD](./getting-started/deploy-a-resource-graph-definition) to create a new API end to end
+- [Install kro](./getting-started/01-Installation.md) to run the controller in your cluster
+- [Deploy your first RGD](./getting-started/02-deploy-a-resource-graph-definition.md) to create a new API end to end
 - [Browse examples](../../examples/) to see more patterns
```
cheeseandcereal: This should be updated for consistency too:

`[Browse examples](../examples/examples.md)`

numb86:
Thanks!

I checked for other affected pages by building locally with `trailingSlash: true`, which surfaces the broken relative links.

Besides `overview`, the only other affected page in the current docs and `0.9.2` was Resource Basics (`concepts/rgd/resource-definitions/resource-basics`) — fixed it the same way, and updated `Browse examples` as suggested.

Rebuilding now reports zero broken links for the current docs. Left the older frozen versions untouched.

jakobmoellerdev (approved):
/lgtm

Issue #1293, opened 2026-05-15
Proposal: Deferred schema resolution for resources with unresolved CRDs (bootstrapping RGD composition)


n3rdc4ptn:
### Feature Description

**Problem Statement**:

When an RGD manages multiple child resources that have interdependencies, kro requires **all** referenced resource types (CRDs) to exist at RGD reconciliation time. This makes it impossible for a single "parent" RGD to orchestrate resources where one child creates the CRD that another child depends on.

Example:

```
Platform RGD (MyPlatform)
├── resource A: InstallOperator       ← deploys an operator that registers new CRDs
└── resource B: InstallAddon          ← uses a CRD that only exists after A is running
```

The intended flow is:
1. Platform RGD is applied
2. kro creates instance of A (`InstallOperator`)
3. A becomes ready → the operator is running and registers its CRDs (e.g. `addons.example.io`)
4. Now kro creates instance of B (`InstallAddon`) which depends on that CRD

What actually happens: kro resolves the schema for **all** resources upfront. Since B's resource type doesn't exist yet, the entire RGD fails:

```
failed to build resource "addon": failed to get schema for resource addon:
cannot resolve group version kind "example.io/v1alpha1, Kind=InstallAddon": schema not found
```

The RGD's own CRD (`MyPlatform`) is never created, so no instance can be created at all — even though resource A has no missing dependencies and could proceed independently.

A full example RGD:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: my-platform
spec:
  schema:
    apiVersion: v1alpha1
    group: platform.example.io
    kind: MyPlatform
    spec:
      repoName: string

  resources:
    # A: deploys an operator — this operator registers CRDs like Addon
    - id: operator
      template:
        apiVersion: platform.example.io/v1alpha1
        kind: InstallOperator
        metadata:
          name: default
        spec:
          repoName: ${schema.spec.repoName}

    # B: uses a CRD that only exists after A's operator is running
    - id: addon
      template:
        apiVersion: platform.example.io/v1alpha1
        kind: InstallAddon
        metadata:
          name: default
        spec:
          repoName: ${schema.spec.repoName}
```

[...]

PR #1318, opened 2026-07-02
fix: Support ordered deletion without desired-state projection, CEL evaluation, GraphRevision resolution or external observation

### LINKED ISSUE #1316: Root instance stuck in `DELETING` when `externalRef` is absent and children have controller finalizers
### Summary

When an `externalRef` resource is deleted before the KRO root instance that
references it, **and** the root instance's managed children have their own
controller finalizers (e.g. Config Connector resources carry
`cnrm.cloud.google.com/finalizer`), the root instance gets stuck in `DELETING`
indefinitely.

The sequence:

1. externalRef is deleted (e.g. a shared ConfigMap, a pre-provisioned CRD).
2. Root instance is deleted → `deletionTimestamp` set, `kro.run/finalizer`
   retained so KRO can clean up owned children.
3. KRO issues deletes on the owned children. Children get `deletionTimestamp`
   but their controller finalizers keep them alive.
4. KRO requeues. On the next reconcile it tries to re-evaluate CEL expressions
   that reference the now-absent externalRef → evaluation fails.
5. KRO cannot confirm cleanup is complete and cannot remove `kro.run/finalizer`
   → root stays stuck in `DELETING` forever.

**Without a child controller finalizer this does not occur** (verified on 0.9.2):
children disappear immediately on delete, KRO completes cleanup before the
failing requeue, and the root is GC'd cleanly.

The race is deterministic in GitOps workflows where an externalRef and the root
instance are pruned in the same sync phase: because the externalRef has no
finalizer it is GC'd first, leaving the root with no externalRef but live
children.

Verified on KRO 0.9.2 with the self-contained repro script linked below.

### Steps to reproduce

Tested on KRO **0.9.2** (Helm chart `v0.9.2`, `registry.k8s.io/kro/charts/kro`)
using a plain kind cluster (`kindest/node:v1.31.4`).

**RGD:**

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: configwatcher-resourcegraphdefinition
spec:
  schema:
    apiVersion: v1alpha1
    group: repro.kro.run
    kind: ConfigWatcher
    spec:
      sourceRef:
        name: string | required=true
  resources:
    - id: source
      externalRef:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.spec.sourceRef.name}

    - id: mirror
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.metadata.name}-mirror
        data:
          source-name: ${source.metadata.name}
          source-uid: ${source.metadata.uid}
```

**Reproduce the bug (prod-like, child has controller finalizer):**

```bash
NAMESPACE=kro-repro-test

# 1. Create the source ConfigMap (the externalRef target)
kubectl create ns $NAMESPACE
kubectl create configmap my-config -n $NAMESPACE

# 2. Create the root CR
kubectl apply -n $NAMESPACE -f - <<'CREOF'
apiVersion: repro.kro.run/v1alpha1
kind: ConfigWatcher
metadata:
  name: my-watcher
spec:
  sourceRef:
    name: my-config
CREOF

# 3. Wait for KRO to create the mirror ConfigMap, then add a manual finalizer
#    to simulate a KCC-like controller (IAMPartialPolicy, BigQueryTable, etc.)
#    that does async cleanup:
kubectl wait --for=condition=exists configmap/my-watcher-mirror -n $NAMESPACE --timeout=30s 2>/dev/null || \
  kubectl get configmap my-watcher-mirror -n $NAMESPACE
kubectl patch configmap my-watcher-mirror -n $NAMESPACE \
  --type=json -p='[{"op":"add","path":"/metadata/finalizers","value":["repro.test/keep"]}]'

# 4. Delete the externalRef first (the race: GitOps prune deletes it before the root)
kubectl delete configmap my-config -n $NAMESPACE

# 5. Delete the root CR
kubectl delete configwatcher my-watcher -n $NAMESPACE --wait=false

# After ~6 seconds:
kubectl get configwatcher my-watcher -n $NAMESPACE -o yaml
# Observed: deletionTimestamp set, kro.run/finalizer present — stuck indefinitely
# KRO logs: repeated "instance not found (likely deleted)" for ConfigWatcher

# Recovery (break-glass):
kubectl patch configwatcher my-watcher -n $NAMESPACE \
  --type=json -p='[{"op":"remove","path":"/metadata/finalizers"}]'
# Then strip the manual finalizer from the mirror and delete it manually.
```

### Expected behaviour

Deleting a root instance should reliably clean up all owned child resources
and complete without requiring manual intervention, regardless of whether
referenced `externalRef` resources still exist at deletion time.

### Actual behaviour (0.9.2)

- **Without a child controller finalizer (Scenario A):** root instance and
  mirror ConfigMap both cleaned up successfully — no orphan. Even with the
  externalRef absent, KRO completes cleanup on the first reconcile pass before
  the failed requeue can block it.
- **With a child controller finalizer (Scenario B):** root instance stuck in
  `DELETING` indefinitely. The child survives deletion due to its controller
  finalizer; KRO's subsequent reconcile fails because the externalRef is gone.
  KRO logs `"instance not found (likely deleted)"` on every reconcile attempt
  and cannot remove `kro.run/finalizer`. Manual break-glass (`kubectl patch` to
  strip `kro.run/finalizer`) is required to unblock the root, followed by
  manual child cleanup.

### Impact

In GitOps workflows where multiple resources are pruned in parallel (kpt, Argo
CD sync-waves, etc.), the externalRef and the root instance are often in the
same delete phase. The race is non-deterministic: if the externalRef is pruned
first (which is common because it has no finalizer), the root instance gets
stuck or orphans its children. Recovery requires cluster-admin-level
`kubectl patch` access — not self-serviceable for component owners.

### Related

- #483 — same root cause (CEL evaluation fails during deletion initialization), different trigger (incomplete creation rather than absent externalRef)
- #1231 — externalRef cleanup mis-behaviour on deletion (different trigger: externalRef is the first RGD resource; closed, fix unclear)

### Reproduction repository

https://gist.github.com/narape/04ebff547ca41f84268578e15f32f8b7

tomasaschan:
## Summary

Fixes #1316.

Instance deletion must remain possible when desired-state projection, CEL evaluation, GraphRevision resolution, or external observation is unavailable. This change:

- Handles deletion before GraphRevision resolution through a dedicated runtime-free path.
- Discovers managed children from the instance's persisted ApplySet inventory rather than reconstructing identities from the current graph.
- Persists the `internal.kro.run/apply-order` annotation on regular resources and collection members, then deletes one highest-order wave at a time.
- Keeps terminating objects in their wave until they disappear, so lower dependency waves cannot advance early.
- Uses UID-preconditioned DELETEs to avoid deleting an object recreated between LIST and DELETE.
- Keeps `externalRef` resources read-only and excludes them from deletion inventory.

## Rollout and failure behavior

Normal reconciliation backfills apply-order annotations through SSA. Children with missing, malformed, zero, or negative order annotations share an unordered fallback wave that runs after every valid ordered wave. This prioritizes deletion liveness for instances deleted before annotation backfill completes.

Deletion validates the persisted ApplySet parent ID, kro tooling ownership, group-kind and namespace inventory, and an inventory checksum when present. Invalid inventory, LIST failures, DELETE failures, and UID conflicts retain the root finalizer and are surfaced through the instance's `ResourcesReady` condition rather than being treated as empty inventory.

## Release note

After upgrading, allow existing instances to reconcile once before deleting them. An instance deleted before that reconciliation may lack the ApplySet inventory required for safe cleanup and remain terminating. If this happens, manually delete or verify its managed resources before removing the kro finalizer. Do not replace unknown inventory with an empty inventory, because that can orphan children.

## Design

See [CEL-independent ordered instance deletion](https://github.com/kubernetes-sigs/kro/blob/fix/1316/deletion-with-externalref/docs/design/proposals/ordered-instance-deletion.md).

## Test plan

- [x] Apply-order annotations on regular resources and collection members.
- [x] External references remain unannotated and absent from ApplySet inventory.
- [x] Reserved-metadata validation and SSA backfill.
- [x] Persisted namespaced and cluster-scoped ApplySet discovery, including removed graph nodes.
- [x] Strict highest-wave deletion, terminating blockers, wave advancement, and same-wave collections.
- [x] Fallback deletion for missing or invalid order annotations.
- [x] Inventory validation and finalizer retention on invalid metadata.
- [x] UID conflicts and DELETE-error visibility.
- [x] Focused integration reproduction of #1316.
- [x] Integration coverage for `A -> B` ordered deletion.

Verified with:

```text
go test ./pkg/metadata ./pkg/graph ./pkg/controller/instance/... -count=1
golangci-lint run ./...
```

### DIFF
--- a/docs/design/proposals/ordered-instance-deletion.md
+++ b/docs/design/proposals/ordered-instance-deletion.md
@@ -1,0 +1,263 @@
+# CEL-independent ordered instance deletion
+
+## Problem statement
+
+Instance deletion currently reconstructs managed-resource identities from the
+compiled graph. That makes cleanup depend on the same desired-state and CEL
+evaluation pipeline used during normal reconciliation.
+
+[Issue #1316](https://github.com/kubernetes-sigs/kro/issues/1316) demonstrates
+why that dependency is unsafe. An instance creates a managed child whose
+identity or desired state references an external object. The child remains
+alive behind a finalizer owned by another controller. The external reference is
+then deleted, followed by the root instance. During deletion, kro can no longer
+evaluate the child's CEL expressions because the external data is unavailable.
+The child is not deleted and the kro finalizer on the root instance is stuck.
+
+This is one example of a broader lifecycle invariant: deleting an instance must
+remain possible when desired-state projection, identity CEL, readiness CEL, or
+external observation is unavailable. Cleanup needs durable identities and
+ordering information from the resources that normal reconciliation already
+created.
+
+## Proposal
+
+Persist each managed child's apply order and use the instance's ApplySet
+inventory as the authority for deletion. Instance deletion lists that inventory
+without resolving desired objects, validates every member's persisted order,
+and deletes only the highest remaining order. It waits for that entire wave to
+disappear before advancing.
+
+External references remain read-only inputs. They are not finalized, mutated,
+or deleted by kro.
+
+#### Overview
+
+The lifecycle is governed by these invariants:
+
+- Managed-resource identity during deletion comes from persisted ApplySet
+  inventory, not from reconstructed desired objects or current graph nodes.
+- Every managed child carries `internal.kro.run/apply-order`. Its value is the
+  child's one-based position in the complete runtime DAG's total topological
+  order.
+- External nodes occupy positions in that total order but are not applied, so
+  managed-resource order values may have gaps.
+- Deletion processes only the highest order still present in inventory.
+- A wave remains active until all of its objects are absent, including objects
+  that already have a deletion timestamp.
+- The instance finalizer is removed only after the inventory is empty.
+- Missing or invalid order metadata is a visible migration error. The root
+  finalizer remains, and deletion never silently changes to unordered cleanup.
+
+#### Normal reconciliation
+
+`processNodes` enumerates every runtime node in topological order. Regular
+managed resources receive the node ID and the one-based apply order as
+controller-owned labels. Every expanded member of a `forEach` collection
+receives the same order because the collection is one graph node. External
+references and external collections are observed but never applied and receive
+no apply-order label.
+
+The existing server-side apply path writes the label alongside the desired
+resource. This also backfills an unchanged child: SSA still submits controller
+metadata even when the resource spec has not changed. A later successful
+reconcile against a new GraphRevision can update an object's order. An object
+removed from the graph keeps its last persisted order until normal pruning
+deletes it.
+
+Both `kro.run/` and `internal.kro.run/` are reserved label prefixes in resource
+templates. RGD authors therefore cannot override the node identity or deletion
+order owned by the controller.
+
+#### Deletion inventory
+
+The instance is already the ApplySet parent. Its annotations persist the union
+of group-kinds and additional namespaces that can contain members. Deletion
+creates the same ApplySet, calls `Project(nil)` to reconstruct that scope from
+the parent annotations, and calls `ListOrphans` with an empty keep-UID set. All
+ApplySet members are consequently deletion candidates.
+
+This inventory has two important properties. First, external references are
+absent because kro never applies them and they are not ApplySet members. Second,
+resources retained from an older GraphRevision remain discoverable even if
+their group-kind, namespace, or node ID is absent from the current graph.
+
+The deletion path does not call `IsIgnored`, `GetDesired`,
+`GetDesiredIdentity`, `DeleteTargets`, readiness evaluation, or external
+observation. Deletion is handled before compiled GraphRevision resolution.
+This is intentional: resolving the current revision can itself require
+unavailable dependencies or invalid CEL, which is exactly the failure mode in
+#1316. The early path only needs the parent object, persisted ApplySet
+annotations, and dynamic REST mappings. The tradeoff is that deletion status
+cannot describe current graph nodes when no revision can be resolved; the
+controller still reports the instance-level deleting condition and preserves
+the finalizer on inventory or ordering failures. Keeping deletion after graph
+resolution would preserve richer node status, but would leave the root
+finalizer stuck whenever resolution fails.
+
+#### Ordered deletion waves
+
+Before issuing any DELETE, the controller parses every candidate's
+`internal.kro.run/apply-order`. Only positive base-10 integers are accepted.
+Missing, malformed, zero, or negative values fail the reconciliation before any
+cluster mutation. The error identifies the resource's GVK, namespace, and name.
+When the candidate has a current `kro.run/node-id`, that node is marked Error.
+There is no inference from node ID, graph shape, collection index, or LIST
+order.
+
+After validation, the controller finds the highest remaining order and selects
+only candidates at that order. It skips a DELETE for a candidate that already
+has a deletion timestamp, but that candidate remains in the active wave and
+blocks every lower order until it is actually absent. Other active-wave
+candidates are passed to `ApplySet.DeleteOrphan`. Its UID precondition prevents
+a LIST/DELETE race from deleting a different object recreated with the same
+name.
+
+DELETE calls within a wave are sequential. All candidates remain visible on
+the next LIST, so every reconciliation with non-empty inventory returns a
+delayed requeue. A UID conflict also requeues and does not permit a lower wave
+to advance. Only a later reconciliation that observes no higher-order members
+can begin the next order. Empty inventory is the sole condition for removing
+the root finalizer and cleaning up coordinator watches.
+
+Node status during deletion is derived without CEL:
+
+- External nodes are Skipped.
+- Current managed nodes represented by any live candidate are Deleting,
+  including lower-order nodes that have not received DELETE yet.
+- Current managed nodes with no candidates are Deleted.
+- A DELETE or ordering error marks the relevant current node Error when its
+  node ID is available.
+
+#### Rollout behavior
+
+When an instance controller registers a GVR at startup, it explicitly enqueues
+the instances already present in its cache. Healthy, unsuspended instances
+therefore run normal reconciliation and acquire the order label through SSA,
+even when their child specs are unchanged. This proposal adds neither a
+separate migration job nor a forced restart loop.
+
+Some instances can miss that backfill. An instance already deleting when the
+new version starts bypasses normal child apply. Suspended instances, instances
+whose desired resources cannot resolve, and instances deleted before startup
+reconciliation completes can also retain unlabeled children. These cases fail
+safely during deletion: the ordering error is reported and the root finalizer
+is retained.
+
+Operators have two break-glass choices:
+
+1. Determine the correct historical order and manually assign valid
+   `internal.kro.run/apply-order` labels before allowing deletion to continue.
+2. Remove the root finalizer and manually clean up all remaining managed
+   children.
+
+This is an accepted upgrade limitation, not an unordered compatibility mode.
+The controller must not guess an order for an unlabeled resource.
+
+#### Failure behavior
+
+ApplySet LIST failures and RESTMapping failures retain the root finalizer and
+retry through normal reconciliation error handling. Malformed ordering metadata
+is reported before any DELETE. DELETE errors retain the finalizer, mark the
+associated node Error when possible, and propagate. A UID precondition conflict
+causes a delayed requeue with the active wave unchanged.
+
+A child with a deletion timestamp is expected progress rather than an error.
+It remains Deleting and continues to block lower orders until the API server no
+longer lists it. No error path is allowed to issue deletion for a lower-order
+resource while a higher-order member remains.
+
+## Other solutions considered
+
+#### Finalize external references
+
+Rejected. External references are read-only, can be shared by unrelated
+instances, and can be managed by another authority. Giving kro a finalizer on
+them would change that ownership contract and could block their deletion.
+
+#### Unordered label-based deletion
+
+Rejected. Finding children by an ownership label avoids CEL but deletes all
+resources together. That is a breaking change to kro's dependent-before-
+dependency lifecycle and can remove resources while their dependents are still
+terminating.
+
+#### Reconstruct identity from CEL and current desired state
+
+Rejected. This is the failure mode in #1316. Deletion must survive missing
+dependencies, invalid expressions, unavailable observation, and desired state
+that no longer describes resources created by an older revision.
+
+#### Infer missing order from current node IDs
+
+Rejected for this rollout. It complicates migration and can be wrong when the
+current GraphRevision differs from the revision that created an older resource.
+Failing visibly preserves ordering rather than hiding uncertainty.
+
+## Scoping
+
+#### What is in scope for this proposal?
+
+- Persisting total topological apply positions on regular and collection
+  resources.
+- Reserving the public and internal kro label prefixes in RGD templates.
+- ApplySet-inventory discovery for instance deletion.
+- Strict, UID-preconditioned, highest-order deletion waves.
+- Startup-reconcile backfill and documented manual remediation for missed
+  migration.
+- Unit and focused integration coverage for the lifecycle.
+
+#### What is not in scope?
+
+A future level-aware reconciliation engine may replace total positions with
+dependency levels, as discussed in
+[PR #1215](https://github.com/kubernetes-sigs/kro/pull/1215#discussion_r3164471497).
+The durable-inventory and wait-for-wave semantics remain applicable: store the
+level on each member, delete the highest remaining level, and wait for absence
+before advancing.
+
+This proposal does not add an upgrade migration controller or infer metadata for missed
+backfills, or change normal orphan-pruning order.
+
+## Testing strategy
+
+#### Requirements
+
+Fake-client deletion fixtures carry a stable UID, the computed
+`applyset.kubernetes.io/part-of` value, `kro.run/node-id`, and
+`internal.kro.run/apply-order`. Their parent instance carries the ApplySet
+group-kind and namespace annotations. Integration tests use synthetic child
+finalizers to make ordering observable rather than timing-dependent.
+
+#### Test plan
+
+Unit tests verify regular-resource order, shared collection order, external
+exclusion, SSA label application to existing objects, and validation of both
+reserved prefixes. ApplySet tests cover namespaced and cluster-scoped discovery
+from parent annotations, including older resources no longer represented by the
+current graph.
+
+Deletion unit tests verify that only the highest remaining order receives
+DELETE, a terminating highest-order resource blocks lower orders, the next wave
+starts only after the higher wave disappears, and all resources at one order
+are handled together. They also cover empty-inventory finalizer removal;
+missing, malformed, zero, and negative labels before any DELETE; DELETE error
+state; UID-conflict requeue; and deletion with an absent external reference.
+
+The core integration suite reproduces #1316 by creating an external ConfigMap
+and a managed child derived from it, asserting the child's persisted order,
+holding the child with a synthetic finalizer, deleting both the external object
+and root, and observing that the child receives a deletion timestamp. Removing
+the child finalizer must allow both child and root to disappear.
+
+A second integration scenario creates `A -> B`, holds B with a synthetic
+finalizer, and deletes the root. B must become terminating while A has no
+deletion timestamp. Only after B disappears may A and then the root disappear.
+
+## Discussion and notes
+
+The persisted order is a total topological position starting at one, not a
+dependency depth. Gaps caused by external nodes are intentional and harmless.
+The central contract is not that orders are contiguous; it is that deletion
+uses only persisted inventory and never advances below the highest remaining
+value.
--- a/docs/design/proposals/ordered-instance-deletion.md
+++ b/docs/design/proposals/ordered-instance-deletion.md
@@ -24,9 +24,10 @@ created.
 
 Persist each managed child's apply order and use the instance's ApplySet
 inventory as the authority for deletion. Instance deletion lists that inventory
-without resolving desired objects, validates every member's persisted order,
-and deletes only the highest remaining order. It waits for that entire wave to
-disappear before advancing.
+without resolving desired objects, groups members by persisted order, and
+deletes only the highest remaining order. It waits for that entire wave to
+disappear before advancing. Members without a usable persisted order share a
+fallback wave that runs after every valid order.
 
 External references remain read-only inputs. They are not finalized, mutated,
 or deleted by kro.
@@ -37,17 +38,17 @@ The lifecycle is governed by these invariants:
 
 - Managed-resource identity during deletion comes from persisted ApplySet
   inventory, not from reconstructed desired objects or current graph nodes.
-- Every managed child carries `internal.kro.run/apply-order`. Its value is the
-  child's one-based position in the complete runtime DAG's total topological
-  order.
+- Normal reconciliation gives every managed child
+  `internal.kro.run/apply-order`. Its value is the child's one-based position
+  in the complete runtime DAG's total topological order.
 - External nodes occupy positions in that total order but are not applied, so
   managed-resource order values may have gaps.
 - Deletion processes only the highest order still present in inventory.
 - A wave remains active until all of its objects are absent, including objects
   that already have a deletion timestamp.
+- Members with missing or invalid order metadata share one fallback wave. That
+  wave runs last and has no ordering guarantee within it.
 - The instance finalizer is removed only after the inventory is empty.
-- Missing or invalid order metadata is a visible migration error. The root
-  finalizer remains, and deletion never silently changes to unordered cleanup.
 
 #### Normal reconciliation
 
@@ -91,19 +92,19 @@ unavailable dependencies or invalid CEL, which is exactly the failure mode in
 annotations, and dynamic REST mappings. The tradeoff is that deletion status
 cannot describe current graph nodes when no revision can be resolved; the
 controller still reports the instance-level deleting condition and preserves
-the finalizer on inventory or ordering failures. Keeping deletion after graph
+the finalizer on inventory or deletion failures. Keeping deletion after graph
 resolution would preserve richer node status, but would leave the root
 finalizer stuck whenever resolution fails.
 
 #### Ordered deletion waves
 
-Before issuing any DELETE, the controller parses every candidate's
-`internal.kro.run/apply-order`. Only positive base-10 integers are accepted.
-Missing, malformed, zero, or negative values fail the reconciliation before any
-cluster mutation. The error identifies the resource's GVK, namespace, and name.
-When the candidate has a current `kro.run/node-id`, that node is marked Error.
-There is no inference from node ID, graph shape, collection index, or LIST
-order.
+The controller parses every candidate's `internal.kro.run/apply-order`.
+Positive base-10 integers retain their persisted order. Missing, malformed,
+zero, or negative values are assigned internal order zero, which is lower than
+every valid order. All such candidates therefore share a fallback wave that is
+selected only after every valid ordered wave is absent. There is no inference
+from node ID, graph shape, collection index, or LIST order, and no
+reverse-order guarantee among members of the fallback wave.
 
 After validation, the controller finds the highest remaining order and selects
 only candidates at that order. It skips a DELETE for a candidate that already
@@ -126,8 +127,8 @@ Node status during deletion is derived without CEL:
 - Current managed nodes represented by any live candidate are Deleting,
   including lower-order nodes that have not received DELETE yet.
 - Current managed nodes with no candidates are Deleted.
-- A DELETE or ordering error marks the relevant current node Error when its
-  node ID is available.
+- A DELETE error marks the relevant current node Error when its node ID is
+  available.
 
 #### Rollout behavior
 
@@ -140,27 +141,22 @@ separate migration job nor a forced restart loop.
 Some instances can miss that backfill. An instance already deleting when the
 new version starts bypasses normal child apply. Suspended instances, instances
 whose desired resources cannot resolve, and instances deleted before startup
-reconciliation completes can also retain unlabeled children. These cases fail
-safely during deletion: the ordering error is reported and the root finalizer
-is retained.
+reconciliation completes can also retain unlabeled children. Deletion still
+processes every labeled wave in reverse order, then deletes all remaining
+unlabeled or invalidly labeled children in the shared fallback wave.
 
-Operators have two break-glass choices:
-
-1. Determine the correct historical order and manually assign valid
-   `internal.kro.run/apply-order` labels before allowing deletion to continue.
-2. Remove the root finalizer and manually clean up all remaining managed
-   children.
-
-This is an accepted upgrade limitation, not an unordered compatibility mode.
-The controller must not guess an order for an unlabeled resource.
+This compatibility behavior deliberately prioritizes deletion liveness over
+reverse-order guarantees for resources whose historical order was never
+persisted. It avoids leaving an instance and its children indefinitely stuck in
+deletion during rollout while preserving strict ordering wherever the metadata
+is available.
 
 #### Failure behavior
 
 ApplySet LIST failures and RESTMapping failures retain the root finalizer and
-retry through normal reconciliation error handling. Malformed ordering metadata
-is reported before any DELETE. DELETE errors retain the finalizer, mark the
-associated node Error when possible, and propagate. A UID precondition conflict
-causes a delayed requeue with the active wave unchanged.
+retry through normal reconciliation error handling. DELETE errors retain the
+finalizer, mark the associated node Error when possible, and propagate. A UID
+precondition conflict causes a delayed requeue with the active wave unchanged.
 
 A child with a deletion timestamp is expected progress rather than an error.
 It remains Deleting and continues to block lower orders until the API server no
@@ -175,12 +171,12 @@ Rejected. External references are read-only, can be shared by unrelated
 instances, and can be managed by another authority. Giving kro a finalizer on
 them would change that ownership contract and could block their deletion.
 
-#### Unordered label-based deletion
+#### Fully unordered label-based deletion
 
-Rejected. Finding children by an ownership label avoids CEL but deletes all
-resources together. That is a breaking change to kro's dependent-before-
-dependency lifecycle and can remove resources while their dependents are still
-terminating.
+Rejected as the general deletion strategy. Finding children by an ownership
+label avoids CEL but deleting all resources together discards usable persisted
+ordering. The compatibility fallback limits unordered deletion to members that
+have no usable order and runs it only after every ordered member is absent.
 
 #### Reconstruct identity from CEL and current desired state
 
@@ -192,7 +188,8 @@ that no longer describes resources created by an older revision.
 
 Rejected for this rollout. It complicates migration and can be wrong when the
 current GraphRevision differs from the revision that created an older resource.
-Failing visibly preserves ordering rather than hiding uncertainty.
+The fallback wave represents that uncertainty explicitly without inventing an
+order that may be wrong.
 
 ## Scoping
 
@@ -203,8 +200,7 @@ Failing visibly preserves ordering rather than hiding uncertainty.
 - Reserving the public and internal kro label prefixes in RGD templates.
 - ApplySet-inventory discovery for instance deletion.
 - Strict, UID-preconditioned, highest-order deletion waves.
-- Startup-reconcile backfill and documented manual remediation for missed
-  migration.
+- A final compatibility wave for resources that missed order-label backfill.
 - Unit and focused integration coverage for the lifecycle.
 
 #### What is not in scope?
@@ -216,8 +212,8 @@ The durable-inventory and wait-for-wave semantics remain applicable: store the
 level on each member, delete the highest remaining level, and wait for absence
 before advancing.
 
-This proposal does not add an upgrade migration controller or infer metadata for missed
-backfills, or change normal orphan-pruning order.
+This proposal does not add an upgrade migration controller, infer metadata for
+missed backfills, or change normal orphan-pruning order.
 
 ## Testing strategy
 
@@ -241,8 +237,9 @@ Deletion unit tests verify that only the highest remaining order receives
 DELETE, a terminating highest-order resource blocks lower orders, the next wave
 starts only after the higher wave disappears, and all resources at one order
 are handled together. They also cover empty-inventory finalizer removal;
-missing, malformed, zero, and negative labels before any DELETE; DELETE error
-state; UID-conflict requeue; and deletion with an absent external reference.
+missing, malformed, zero, and negative labels in a shared final wave; DELETE
+error state; UID-conflict requeue; and deletion with an absent external
+reference.
 
 The core integration suite reproduces #1316 by creating an external ConfigMap
 and a managed child derived from it, asserting the child's persisted order,
@@ -260,4 +257,5 @@ The persisted order is a total topological position starting at one, not a
 dependency depth. Gaps caused by external nodes are intentional and harmless.
 The central contract is not that orders are contiguous; it is that deletion
 uses only persisted inventory and never advances below the highest remaining
-value.
+valid value. The fallback order zero is selected only when no valid ordered
+members remain.
--- a/docs/design/proposals/ordered-instance-deletion.md
+++ b/docs/design/proposals/ordered-instance-deletion.md
@@ -70,6 +70,14 @@ Both `kro.run/` and `internal.kro.run/` are reserved label prefixes in resource
 templates. RGD authors therefore cannot override the node identity or deletion
 order owned by the controller.
 
+The write that first installs the instance finalizer also creates a valid empty
+ApplySet inventory when none exists. If the instance already has valid
+inventory, that write preserves it; partial or malformed inventory prevents
+finalizer installation rather than being replaced with an empty scope. An
+automatic replacement could hide previously managed children and orphan them
+during deletion. Recovery therefore requires repairing the metadata, or
+removing it only after confirming that no managed members remain.
+
 #### Deletion inventory
 
 The instance is already the ApplySet parent. Its annotations persist the union
--- a/docs/design/proposals/ordered-instance-deletion.md
+++ b/docs/design/proposals/ordered-instance-deletion.md
@@ -39,9 +39,10 @@ The lifecycle is governed by these invariants:
 - Managed-resource identity during deletion comes from persisted ApplySet
   inventory, not from reconstructed desired objects or current graph nodes.
 - Normal reconciliation gives every managed child
-  `internal.kro.run/apply-order`. Its value is the child's one-based position
-  in the complete runtime DAG's total topological order.
-- External nodes occupy positions in that total order but are not applied, so
+  `internal.kro.run/apply-order`. Its value is the child's one-based reverse
+  topological deletion wave. Dependents have higher values than their
+  dependencies, while nodes that can be deleted together share a value.
+- External nodes participate in the graph layers but are not applied, so
   managed-resource order values may have gaps.
 - Deletion processes only the highest order still present in inventory.
 - A wave remains active until all of its objects are absent, including objects
@@ -52,12 +53,13 @@ The lifecycle is governed by these invariants:
 
 #### Normal reconciliation
 
-`processNodes` enumerates every runtime node in topological order. Regular
-managed resources receive the node ID and the one-based apply order as
-controller-owned labels. Every expanded member of a `forEach` collection
-receives the same order because the collection is one graph node. External
-references and external collections are observed but never applied and receive
-no apply-order label.
+Before processing nodes, the controller groups the compiled DAG into reverse
+topological layers. Regular managed resources receive the node ID and the
+layer's one-based apply order as controller-owned labels. Every expanded member
+of a `forEach` collection receives the same order because the collection is one
+graph node. External references and external collections participate in layer
+calculation but are observed rather than applied and receive no apply-order
+label.
 
 The existing server-side apply path writes the label alongside the desired
 resource. This also backfills an unchanged child: SSA still submits controller
@@ -214,7 +216,7 @@ order that may be wrong.
 
 #### What is in scope for this proposal?
 
-- Persisting total topological apply positions on regular and collection
+- Persisting reverse topological deletion waves on regular and collection
   resources.
 - Reserving the public and internal kro label prefixes in RGD templates.
 - ApplySet-inventory discovery for instance deletion.
@@ -225,15 +227,10 @@ order that may be wrong.
 
 #### What is not in scope?
 
-A future level-aware reconciliation engine may replace total positions with
-dependency levels, as discussed in
-[PR #1215](https://github.com/kubernetes-sigs/kro/pull/1215#discussion_r3164471497).
-The durable-inventory and wait-for-wave semantics remain applicable: store the
-level on each member, delete the highest remaining level, and wait for absence
-before advancing.
-
-This proposal does not add an upgrade migration controller, infer metadata for
-missed backfills, or change normal orphan-pruning order.
+The deletion-wave calculation does not change normal reconciliation to apply
+resources concurrently. This proposal also does not add an upgrade migration
+controller, infer metadata for missed backfills, or change normal
+orphan-pruning order.
 
 ## Testing strategy
 
@@ -274,9 +271,10 @@ deletion timestamp. Only after B disappears may A and then the root disappear.
 
 ## Discussion and notes
 
-The persisted order is a total topological position starting at one, not a
-dependency depth. Gaps caused by external nodes are intentional and harmless.
-The central contract is not that orders are contiguous; it is that deletion
-uses only persisted inventory and never advances below the highest remaining
-valid value. The fallback order zero is selected only when no valid ordered
-members remain.
+The persisted order is a reverse topological layer starting at one. Sink nodes
+receive the highest value, and removing that layer exposes the next set of
+nodes whose dependents are gone. Gaps caused by external-only layers are
+intentional and harmless. The central contract is not that orders are
+contiguous; it is that deletion uses only persisted inventory and never
+advances below the highest remaining valid value. The fallback order zero is
+selected only when no valid ordered members remain.
--- a/docs/design/proposals/ordered-instance-deletion.md
+++ b/docs/design/proposals/ordered-instance-deletion.md
@@ -132,6 +132,11 @@ Deletion status is intentionally instance-level. The runtime-free path cannot
 derive meaningful current-graph node states, and those transient states were
 never persisted independently. `ResourcesReady=Unknown` with reason
 `UnderDeletion` reports progress or the error currently blocking deletion.
+When an RGD defines author conditions, deletion preserves their last persisted
+values and overlays this kro-owned lifecycle condition because author
+conditions cannot be reevaluated without a runtime. If the author defines a
+condition with the same type, the deletion condition temporarily takes
+precedence so cleanup failures remain observable.
 
 #### Rollout behavior
 
--- a/docs/design/proposals/ordered-instance-deletion.md
+++ b/docs/design/proposals/ordered-instance-deletion.md
@@ -158,6 +158,15 @@ retry through normal reconciliation error handling. DELETE errors retain the
 finalizer, mark the associated node Error when possible, and propagate. A UID
 precondition conflict causes a delayed requeue with the active wave unchanged.
 
+Deletion validates the ApplySet parent ID, kro tooling ownership, and the
+presence and syntax of the persisted group-kind and namespace annotations
+before listing members. Normal reconciliation also writes a checksum over that
+inventory; when present, deletion verifies it to detect accidental partial
+mutation. Invalid inventory retains the root finalizer and reports an
+actionable reconciliation error rather than treating the inventory as empty
+and orphaning children. Parents created by older kro versions may omit the
+checksum, but must still carry the standard ApplySet inventory annotations.
+
 A child with a deletion timestamp is expected progress rather than an error.
 It remains Deleting and continues to block lower orders until the API server no
 longer lists it. No error path is allowed to issue deletion for a lower-order
--- a/docs/design/proposals/ordered-instance-deletion.md
+++ b/docs/design/proposals/ordered-instance-deletion.md
@@ -10,9 +10,10 @@ evaluation pipeline used during normal reconciliation.
 why that dependency is unsafe. An instance creates a managed child whose
 identity or desired state references an external object. The child remains
 alive behind a finalizer owned by another controller. The external reference is
-then deleted, followed by the root instance. During deletion, kro can no longer
-evaluate the child's CEL expressions because the external data is unavailable.
-The child is not deleted and the kro finalizer on the root instance is stuck.
+then deleted, followed by (or at the same time as) the root instance. During
+deletion, kro can no longer evaluate the child's CEL expressions because the
+external data is unavailable. The child is not deleted and the kro finalizer on
+the root instance is stuck.
 
 This is one example of a broader lifecycle invariant: deleting an instance must
 remain possible when desired-state projection, identity CEL, readiness CEL, or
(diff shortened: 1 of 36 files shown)


tomasaschan:
Converting to draft while I figure out how to get CI to pass.

tomasaschan:
Some investigation by Claude into the failing CI job:

> ```
> # One-time setup
> make envtest
>
> # Full core suite (parallel)
> KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" \
>   go tool ginkgo -p -v ./test/integration/suites/core/
> 
> # Specific test(s)
> KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.35.0-linux-amd64" \
>  go tool ginkgo -v --focus 'test name pattern' ./test/integration/suites/core/
> ```
>
> The two CRD lifecycle test failures (`should delete CRD when ResourceGraphDefinition is deleted` and `should detect externally deleted CRD and recreate it`) reproduce locally under parallel execution on both main and this branch, but pass if run one at a time.

pkg/controller/instance/deletion.go
```diff
@@ -21,194 +21,122 @@ import (
...
+		return c.removeFinalizer(rcx)
 	}
 
-	return c.removeFinalizer(rcx)
-}
+	// nodesDeleting tracks which nodes had at least one resource deleted or
+	// already terminating. Absent means no live resources found for that node.
+	// Delete errors set state inline and return early, so this set only
+	// contains successful outcomes.
+	nodesDeleting := map[string]struct{}{}
```
jakobmoellerdev: this is an implicit change that removes kros ability to delete resources in reverse order. as such this is a breaking change and should be handled with care because the previous order is no longer respected. not in general against it in my view, just pointing this out

tomasaschan: In my first approach, I had it instead build the DAG and delete only leaf nodes, waiting for them to be successfully garbage collected before processing any dependents. That, however, still didn't solve #1316, since the resource dependency there is an `externalRef` where kro doesn't control the lifecycle. When looking for ways to find and delete the resources even after the `externalRef` is gone, the `LIST`-with-label-selector seemed natural, and then didn't strictly require the reverse-order deletions anymore (since all resources are going away anyway).

Another, completely different, approach, would be to have kro add a finalizer to the `externalRef` resource (at least if it is needed to determine the identity of any resources in the graph), and not remove that finalizer until all resources that depend on it for identity resolution have also been deleted. Would you prefer that approach?

jakobmoellerdev: I believe one way to get around this is to have the externalRef resolved normally even if its 404. that way you can preserve the reverse-order deletion.

I dont think finalizers are the way here, we should not update spec/status of referenced resources to not pollute the design

tomasaschan: The problem outlined by issue #1316 is that sometimes we need the externalRef resource to determine the identity of nodes in the graph, and so if they cannot be fetched, any CEL evaluation that uses their properties fails. How do we get around that?

jakobmoellerdev: imho, its completely possible right now to only compute the identity of that node because in the example given you calculate it based on schema so the node should compute. We even have separate compute options for this in the runtime just for this case. Does that solve your concern?

tomasaschan: For argument's sake (because I think we should aim for a solution where that is possible), what if the name of a generated resource is truly dependent on data from the externalRef? Is it still possible?

jakobmoellerdev: If you reverse order delete by definition the expression should be able to be resolved right? in this case this should work just fine then. Or what am I missing in your view?

tomasaschan: Deletion of the externalRef isn't controlled by kro (unless we set a finalizer on it). So if you create, say, an RGD with an externalRef (say, a config map) which creates another resource (say, a role binding) named by a field on the externalRef, there is nothing (?) that helps us find the generated resource if the configmap is gone. 

If the user uses some gitops tool to manage their config, and remove the root and the referenced configmap in the same change set, the tool is likely to issue DELETEs to both resources in parallel. When kro comes around to reconcile the deletion, the configmap is already gone.

jakobmoellerdev: > Deletion of the externalRef isn't controlled by kro (unless we set a finalizer on it). So if you create, say, an RGD with an externalRef (say, a config map) which creates another resource (say, a role binding) named by a field on the externalRef, there is nothing (?) that helps us find the generated resource if the configmap is gone.

Correct, but this is by design. KRO's entire purpose is to respect dependency ordering expressed by implicit fields. If you code in a dependency on an identity field, I believe this is intended behavior.

tomasaschan: So, you're saying that _by design_ kro can't be used with gitops tools that don't know about kro's internal dependency resolution mechanisms?

The situation I'm trying to solve for here is not that kro is managing a set of resources and the user deletes the root. It's that _the user_ manages a set of resources where kro creates a dependency between a kro-defined resource and something else. That dependency is currently too implicit to be visible when the user issues a deletion, so the fully standard method of "issue a delete call to all resources I want to get rid of" no longer works - because kro's reconciliation of that delete calling on the kro-defined resource is broken.

I'm very much open to finding a different approach to addressing this, but I don't think the current state is good enough to dismiss any attempt at changing it with "this behavior is by design."

tomasaschan: To be clear:

This is a real problem that real users of kro have, that are having them reconsider whether kro is the right tool for them. I think it is, given that we don't just classify this as "by design" but rather actually fix the broken user experience.

This can be solved either by making the externalRef dependency explicit (with a finalizer on the external object), or by breaking the dependency (by not using anything from the external resource when reconciling, in particular, deletions). So it's not like it's a problem that's too complex to have a reasonable solution - it has several.

The solution proposed here doesn't seem like it'll be accepted, and that's fine! There's at least two or three other approaches that could work too. But before I spend any effort implementing any of them, I'd like to at least understand what, specifically, would be required for them to be acceptable, since this approach obviously isn't.

jakobmoellerdev: I mentioned above in my first review - I am not in opposition to the proposal:

> as such this is a breaking change and should be handled with care because the previous order is no longer respected. not in general against it in my view, just pointing this out

I do think however that there are some issues in the design of RGDs that impose implicit dependencies on node identity as you are suggesting them. It is very rarely the case that you need node identity to be fully dynamic and you cant make it based on the instance or the schema, but you have to use external ref.

> This can be solved either by making the externalRef dependency explicit (with a finalizer on the external object), or by breaking the dependency (by not using anything from the external resource when reconciling, in particular, deletions).

I do not think it is valid for us to add finalizers to externalRefs as they are read only until now and I dont think that we want to change this. I think breaking the dependency is thus the correct behavior.

Happy to get overruled on this by other maintainers on this though, im not in a strong stance to oppose and am also okay with changing our generic delete behavior to be unordered. @michaelhtm @jlbutler @a-hilaly PTAL maybe provide a second opinion?

NicholasBlaskey: I wasn't tagged, but saw this thread. I do think this is a footgun and I think a real risk is expecting too much from RGD authors to do things in a way that works around Kro's limitations instead of the way they want to do it based on a reasonable understanding of Kro's features.

In a perfect world, I think deletion should never need CEL evaluation or references to other resources since they can fail. 

I do think changing the delete order is not feasible (without an opt in flag or setting).  

https://github.com/kubernetes-sigs/kro/pull/1215#discussion_r3164471497

I think this ties into this discussion, not necessarily blocked by it. Thoughts on this change but with a new monotonically increasing label like ```internal.kro.run/apply-order: 1``` that we could use for ensuring a best effort reverse order deletion? 

This label would be incremented in the order we apply the resource each reconile, so if we apply a deployment then a service the deployment would get 1 and the service get 2. Then on deleting we can tell the service was applied after the deployment for ordering.

tomasaschan: I pushed another variant, which uses `internal.kro.run/apply-order` labels, created with SSA on new and existing resources, to ensure the deletion order is preserved. To not over-complicate things I didn't include a mechanism to gracefully handle ongoing deletions while this feature is being rolled out; in other words, if one deploys a version that has this change while there are ongoing (possibly stuck, but even normal-operations) deletions, those will require some manual work to resolve. If that's unacceptable, we can add code that gracefully handles that (where possible) too, but this PR is already biiig...

pkg/controller/instance/status.go
```diff
@@ -199,36 +199,42 @@ func (c *Controller) updateStatus(rcx *ReconcileContext) error {
...
-		return err
-	}
-	if resolved, found, _ := unstructured.NestedMap(desired[0].Object, "status"); found {
-		for k, v := range resolved {
-			if k == "conditions" || k == "state" {
-				continue
+	// During early deletion there may be no resolved runtime. In that case
+	// preserve the existing instance status fields and update only kro-owned
+	// deletion state and conditions.
+	if rcx.Runtime != nil {
```
NicholasBlaskey: non blocking: runtime being nullable feels like a footgun

I don't know if I have a great solution here. Creating separate methods for like Deletion order also seems like it has the same potential issue where it could be easy for someone to accidentally introduce a code path calling a function that needs runtime. 

Potentially could try and split the types into one requirement a reconcile context and one that doesn't have it, but not sure how large of change that would end up being / something that can be done later.

pkg/controller/instance/deletion.go
```diff
@@ -225,7 +178,9 @@ func (c *Controller) removeFinalizer(rcx *ReconcileContext) error {
 		return err
 	}
 	rcx.Instance = patched
-	rcx.Runtime.Instance().SetObserved([]*unstructured.Unstructured{patched})
+	if rcx.Runtime != nil {
```
NicholasBlaskey: can this ever be true?

do we still need this call to `SetObserved` here?

pkg/controller/instance/deletion.go
```diff
@@ -17,198 +17,151 @@ package instance
...
+func parseDeletionOrders(candidates []applyset.OrphanCandidate) ([]int, int, *deletionOrderError) {
+	orders := make([]int, len(candidates))
+	highest := 0
+	for i, candidate := range candidates {
+		obj := candidate.Object
+		raw := obj.GetLabels()[metadata.ApplyOrderLabel]
+		validDigits := raw != ""
+		for _, digit := range raw {
+			validDigits = validDigits && digit >= '0' && digit <= '9'
 		}
```
NicholasBlaskey: Do we gain anything by having this validDigits check, or is the error that should be returned for `strconv.Atoi()` sufficient?

pkg/controller/instance/deletion.go
```diff
@@ -17,198 +17,151 @@ package instance
...
+	inventory, err := applier.Project(nil)
 	if err != nil {
-		state.SetError(err)
-		return err
+		return nil, nil, fmt.Errorf("project deletion inventory: %w", err)
 	}
-	if len(targets) == 0 {
-		state.SetDeleted()
-		return nil
+	candidates, err := applier.ListOrphans(rcx.Ctx, applyset.PruneOptions{
```
NicholasBlaskey: This does now require us to issue a list calls for each GVK present in the graph for each reconcile here for deletion

I suppose we could think about reading from one of the informer caches that Kro already runs, but there might be some tradeoffs there (stale data maybe?). 

I'm really not sure if this cost will be an issue or not.

pkg/controller/instance/deletion.go
```diff
@@ -17,198 +17,151 @@ package instance
...
 	}
 
-	if deletionNode != nil {
-		state := rcx.StateManager.NodeStates[deletionNode.Spec.Meta.ID]
-		if err := c.deleteTarget(rcx, deletionNode, state); err != nil {
-			return err
-		}
-		// Deletion is in progress; requeue.
-		return rcx.delayedRequeue(fmt.Errorf("deleting resource %s", deletionNode.Spec.Meta.ID))
+	if len(candidates) == 0 {
```
NicholasBlaskey, line 46: Thinking out loud here. Any edge cases with changing the graph and deleting it? 

The GVK annotation ```applyset.kubernetes.io/contains-group-kinds``` that we use to know what GVK kinds we should list seems robust, we won't update it unless we successfully apply all resources.

If we fail after applying but before pruning, we could end up with two `apply-order` set to the same value, but that seems ok here.

I can't seem to think of any issues here.

pkg/controller/instance/deletion.go
```diff
@@ -17,198 +17,151 @@ package instance
...
+func (e *deletionOrderError) Error() string { return e.err.Error() }
+func (e *deletionOrderError) Unwrap() error { return e.err }
+
+// parseDeletionOrders validates the complete inventory before any mutation.
+func parseDeletionOrders(candidates []applyset.OrphanCandidate) ([]int, int, *deletionOrderError) {
+	orders := make([]int, len(candidates))
+	highest := 0
+	for i, candidate := range candidates {
+		obj := candidate.Object
+		raw := obj.GetLabels()[metadata.ApplyOrderLabel]
```
NicholasBlaskey: This will fail for previous versions.

We discussed this in the community meeting a bit. Should we 

1. delete in any order
2. fallback to the old delete logic if these labels aren't present (for a couple versions)
3. fail the delete

Personally I think deleting in any order is probably good enough with a call out in release notes.

pkg/controller/instance/deletion.go
matthchr: This discussion in this thread is great and the conclusion (label order, use that labelled order to unroll rather than CEL) seems a good solution to me.

My concern is this:

>To not over-complicate things I didn't include a mechanism to gracefully handle ongoing deletions while this feature is being rolled out; in other words, if one deploys a version that has this change while there are ongoing (possibly stuck, but even normal-operations) deletions, those will require some manual work to resolve.

Stuff that is actively deleting getting stuck if it was deleting during the upgrade (say from 0.9.0 -> 0.10.0) feels bad to me. I think we need to decide which of these options is most palatable to us:

1. Add the `internal.kro.run/apply-order` label in one release such as 0.9.x, and call out in release notes that users need to upgrade to that version, then delay this actual deletion logic change to 0.10.0 (or 0.11.0 if we don't put the label in til 0.9) so that we don't get resources getting stuck deleting if they were in deleting during the upgrade.
2. Accept a fallback deletion behavior (unordered?) if the label is missing. Presumably this is pretty easy to implement here? Or is it possible to just fall back to the old deletion behavior if any of the entities in the graph are missing the label (means we would need to preserve it for a while and could remove that fallback in a later release).

and do one of them.

NicholasBlaskey (commented):
thanks for the change. I think this looks pretty good to me.

pkg/controller/instance/deletion.go
```diff
@@ -17,198 +17,151 @@ package instance
...
+}
 
-	if !anyDeleted {
-		// All targets were NotFound, so the node is fully deleted.
-		state.SetDeleted()
-		return nil
+func (c *Controller) updateDeletionNodeStates(rcx *ReconcileContext, candidates []applyset.OrphanCandidate) {
+	if rcx.Runtime == nil {
+		return
+	}
```
cheeseandcereal: `reconcileDeletion` is only ever called from the early-deletion path in `Reconcile`, which always passes `nil` as the runtime. So the `rcx.Runtime == nil` guard here means `updateDeletionNodeStates` (and the `skippedState`/`deletingState`/`deletedState` writes, plus the `SetNodeState` calls on the error paths) never do anything. Node states also aren't persisted anywhere from `StateManager`.

The unit tests aren't catching this because they build the context with a runtime (`newControllerAndContext`), so they assert per-node states the controller can never produce (e.g. `TestReconcileDeletionDeletesAllAndRequeues`, `TestReconcileDeletionExternalRefGoneDoesNotBlock`, and probably some others).

I would recommend either (a) resolve the graph revision best-effort and pass it in for status purposes (falling back to nil, which is what the proposal's tradeoff paragraph implies), or (b) delete `updateDeletionNodeStates` + the nil-guards in `removeFinalizer`/`updateStatus` and make the tests construct the context the way production does (nil runtime).

pkg/controller/instance/deletion.go
cheeseandcereal: +1, I think there needs to be a good alternative/fallback for the case the labels are not there.

Without something being done, this will be a major rough edge. Definitely need to pick one.

pkg/controller/instance/deletion.go
cheeseandcereal: A note here:

We are taking a hard dependency on these annotations now, if they don't exist for any reason (webhook, modified, kubectl replaced, etc), kro's deletion behavior now breaks where it wouldn't before. Worth calling out explicitly, making sure that this is the behavior we want/expect. It may also be worth documenting.

pkg/controller/instance/status.go
cheeseandcereal: The comment says existing status fields are preserved, but `initialStatus()` builds a fresh `{conditions, state}` map and the write replaces `.status` so all author-defined status fields are dropped the moment deletion starts. You probably need to merge these.

You may also want to use `filterOutKroBuiltinConditions` if `c.hasAuthorConditions` is set, so ensure that custom conditions don't get leaked internal state during deletion when runtime is nil.

pkg/controller/instance/deletion.go
```diff
@@ -17,198 +17,151 @@ package instance
...
-			}
-			state.SetError(err)
-			return nil, err
+	orders, highest, orderErr := parseDeletionOrders(candidates)
+	if orderErr != nil {
+		if nodeID := orderErr.nodeID; nodeID != "" {
+			rcx.StateManager.SetNodeState(nodeID, errorState(orderErr))
 		}
+		return orderErr
+	}
```
cheeseandcereal: When deletion is permanently stuck on an invalid/missing order label, the only signal comes from `ResourcesReady=Unknown` condition. Can we surface it a bit better, i.e. `rcx.Mark.ResourcesUnderDeletion("%v", orderErr)` (and similarly for DELETE errors) so `kubectl describe` shows why nothing is progressing.

pkg/controller/instance/deletion.go
tomasaschan: I've now pushed a change to create a "fallback wave" - any resource that matches the instance but has no valid apply-order label, is added to the fallback wave and deleted last.

Effectively, this means that RGs being deleted while rolling this change out will execute unordered rather than in reverse-order; this was discussed at some length in the community meeting on July 22 and accepted as a reasonable trade-off.

pkg/controller/instance/deletion.go
tomasaschan: Yeah, the error check is sufficient. Removed this in my latest push.

pkg/controller/instance/deletion.go
tomasaschan: 40df0c5 should make this better, I hope.

pkg/controller/instance/deletion.go
tomasaschan: Is 72a61b9 in line with what you're after?

pkg/controller/instance/status.go
tomasaschan: 72a61b9 should fix this.

pkg/controller/instance/deletion.go
tomasaschan: This now falls back to unordered deletion of all resources with no label (or a label with an invalid value). In practice, this should only happen while rolling this out, and in _most_ cases unordered deletion is _probably_ fine.

pkg/controller/instance/deletion.go
tomasaschan: This is addressed in the latest version.

pkg/controller/instance/status.go
tomasaschan: Addressed by 72a61b9.

The runtime-free deletion status path now:

  - Copies all existing wire-status fields except state and conditions, preserving fields such as `status.endpoint`.
  - Sets deletion state from kro’s `StateManager`.
  - When author conditions are configured, copies the persisted author conditions exactly, replacing the temporary built-in conditions added by condition markers.

Using `filterOutKroBuiltinConditions` is unnecessary here because the code copies conditions from the pre-marker wire snapshot rather than filtering the marker-mutated instance.

`TestReconcileDeletionPreservesAuthorStatusWithoutRuntime` verifies that:

  - An author-defined `status.endpoint` survives deletion.
  - Only the `AuthorHealthy` condition remains.
  - Its transition timestamp is preserved.

pkg/controller/instance/deletion.go
tomasaschan: Yes, this was unreachable. 72a61b9 removes the `SetObserved` call and makes deletion use a dedicated runtime-free context, so runtime-dependent behavior can no longer appear on this path.

pkg/controller/instance/deletion.go
tomasaschan: My assumption is that this won't be an issue, given that it doesn't run on every reconcile - only when reconciling deletions specifically. That should be a lot rarer than reconciliations as a whole.

pkg/controller/instance/deletion.go
```diff
@@ -17,218 +17,127 @@ package instance
...
-// deleteTarget issues delete requests for the node's delete targets and updates state.
-func (c *Controller) deleteTarget(
-	rcx *ReconcileContext,
-	node *runtime.Node,
-	state *NodeState,
-) error {
-	targets, err := node.DeleteTargets()
+// discoverDeletionInventory reconstructs the deletion search scope solely from
+// the parent ApplySet metadata. It does not evaluate the current graph or CEL.
+func (c *Controller) discoverDeletionInventory(
```
Skarlso, line 76: So the finalizer is written in ensureManaged, and this fails on ValidateParentInventory here immediately. This means, that if there is an error in-between, the finalizer will not be cleared. And there could be several errors as I step through this and one of them is actually related to another problem. ApplySet can fail if an RGD references a CRD that is not there yet. And that can happen often because of how ordering works and if there is RGD chaining. On duplicate resources and some other process node errors.

Skarlso: Further, while this existed before and https://github.com/kubernetes-sigs/kro/issues/1293 kinda addresses it, here, the same error / problem now also blocks the finalizer deletion.

pkg/controller/instance/status.go
```diff
@@ -228,25 +230,62 @@ func (c *Controller) updateStatus(rcx *ReconcileContext) error {
...
+func (c *Controller) updateDeletionStatus(dcx *DeletionContext) error {
+	previousState, _ := dcx.WireStatus["state"].(string)
+	status := initialStatus(dcx.Instance, dcx.StateManager)
+	for k, v := range dcx.WireStatus {
+		if k != "conditions" && k != "state" {
+			status[k] = v
+		}
+	}
+	if dcx.Config.HasAuthorConditions {
+		if conditions, found := dcx.WireStatus["conditions"]; found {
```
Skarlso: The PR says that failures are surfaced through the ResourceReady condition. But here, you are just overwriting the entire conditions section and all errors are now gone, right?

pkg/controller/instance/applyset/inventory.go
```diff
@@ -0,0 +1,139 @@
...
+	if actual, exists := annotations[ApplySetInventoryHashAnnotation]; exists {
+		expected := inventoryHash(expectedID, groupKinds, namespaces)
+		if actual != expected {
+			return fmt.Errorf("invalid %s annotation: got %q, want %q", ApplySetInventoryHashAnnotation, actual, expected)
+		}
+	}
+	return nil
+}
+
+func parseGroupKinds(raw string) (sets.Set[schema.GroupKind], error) {
```
Skarlso, line 89: I think this should align with `parentAnnotationSets` right...? 🤔

pkg/graph/validation.go
```diff
@@ -349,8 +349,8 @@ func isRequiredIdentityField(path string, resourceNamespaced, instanceNamespaced
 	}
 }
 
-// validateNoKROOwnedLabels enforces that the resource template doesn't define any label with
-// LabelKROPrefix (kro.run/). These labels are reserved for internal use ONLY.
+// validateNoKROOwnedLabels enforces that resource templates do not define
+// labels in either controller-owned namespace.
 func validateNoKROOwnedLabels(resourceID string, resourceObject map[string]interface{}) error {
```
Skarlso, line 357: This is a breaking change, right?

pkg/controller/instance/controller.go
```diff
@@ -222,7 +225,25 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (err error
...
+	// 2. Handle deletion before graph resolution
+	//--------------------------------------------------------------
+	// Deletion must not depend on resolving the current GraphRevision or CEL.
+	// Build a context without a runtime and use persisted ApplySet inventory.
+	if inst.GetDeletionTimestamp() != nil {
+		dcx = NewDeletionContext(
+			ctx, log, c.gvr, c.namespaced, c.client.Dynamic(), c.client.RESTMapper(),
+			c.reconcileConfig, inst,
+		)
+		dcx.Watcher = watcher
```
Skarlso: Huh. Is this `Watcher` used internally somewhere that I can't spot? This Watcher is never read anywhere.

pkg/controller/instance/context.go
```diff
@@ -57,6 +57,28 @@ type ReconcileContext struct {
...
+	RestMapper meta.RESTMapper
+
+	Instance *unstructured.Unstructured
+	Config   ReconcileConfig
+
+	WireStatus map[string]interface{}
+
+	Mark         *ConditionsMarker
+	StateManager *StateManager
+	Watcher      dynamiccontroller.InstanceWatcher
```
Skarlso: Yeah, this watcher is never read from anywhere. I'm not sure what the purpose is for it?

pkg/controller/instance/context.go
```diff
@@ -78,6 +100,9 @@ func NewReconcileContext(
 	config ReconcileConfig,
 	instance *unstructured.Unstructured,
 ) *ReconcileContext {
+	if rt == nil {
+		panic("NewReconcileContext requires a runtime; use NewDeletionContext for deletion")
```
Skarlso: I'm not sure this should be a panic. I'm pretty sure an error is enough.

test/integration/suites/core/crd_test.go
```diff
@@ -751,16 +760,20 @@ var _ = Describe("CRD", func() {
...
 			Eventually(func(g Gomega, ctx SpecContext) {
 				err := env.Client.Get(ctx, types.NamespacedName{Name: crdName}, recreatedCRD)
 				g.Expect(err).ToNot(HaveOccurred())
 
-				// A different UID proves this is a new object, not the original
-				g.Expect(recreatedCRD.UID).NotTo(Equal(originalUID))
+				g.Expect(
+					recreatedCRD.UID != originalUID ||
+						recreatedCRD.DeletionTimestamp != nil,
+				).To(BeTrue(), "old CRD still present without deletionTimestamp — waiting for new CRD")
```
Skarlso, line 774: I'm 95% sure this is not needed and that the comment is actually not valid 🤔

pkg/controller/instance/applyset/inventory.go
```diff
@@ -0,0 +1,139 @@
+// Copyright 2026 The Kubernetes Authors.
```
Skarlso, line 1: Nit: We should remove the Year from the boilerplate. This is a non-matching header license now with the rest of the files. 🤔

Skarlso (commented):
I have some concerns over deletion locking in the finalizer in case of errors on the resource graph processing.

test/integration/suites/core/crd_test.go
tomasaschan: This was added to address a flaky test. So it's not _always_ needed, but there is a timing issue in the this test that this wait helps alleviate.

pkg/controller/instance/applyset/inventory.go
tomasaschan: Thx, will fix.

Is there tooling somewhere to ensure the license headers are correct? Wouldn't be bad to cover this in a linter + a command to automatically wrote the right one.

pkg/controller/instance/context.go
tomasaschan: This is mainly just guarding against developer mistakes; we could also remove the guard and have it potentially panic with a nilref if a reconcile context is initialized without a runtime and then tries to use it. I felt this was better.

If this condition is ever true (I don't think there are any code paths where it currently is) it seems unlikely to me that it can be fixed without changing the source code.

pkg/controller/instance/context.go
tomasaschan: Thanks for noticing, this must be left over from some local iteration. Will remove.

pkg/controller/instance/applyset/inventory.go
Skarlso: Yep, there is one here: https://github.com/external-secrets/external-secrets/blob/main/.github/workflows/ci.yml#L52

https://github.com/apache/skywalking-eyes

We use this for license header validation. Though this is a github action change which will increase the scope of this PR. Maybe introduce it separately? :)

pkg/graph/validation.go
tomasaschan: Technically yes, but it's [not a new idea](https://github.com/kubernetes-sigs/kro/blob/ed93efcc8f61ab2bd6f3f89942ec27ed6cc97d39/docs/design/proposals/label-migration.md) and shouldn't really affect any (well-behaved) users.

pkg/controller/instance/applyset/inventory.go
tomasaschan: That link is to a separate repo, though. And in this one, it seems most files _do_ have a year.

I agree it might be good to remove the year across the repo, and I do think we should add linting + codegen capabilities to ensure things are consistently the way we want them, but I'm going to consider all of that out of scope for this PR.

Skarlso: Sorry, I misunderstood your question then. :D I thought you are asking if it could be introduced. :) Yeah, it's not something that exists in Kro right now. Sorry for the confusion. :) 🙇‍♀️

pkg/controller/instance/applyset/inventory.go
tomasaschan: I extracted a common function now used in both places.

pkg/controller/instance/status.go
tomasaschan: Thanks, now fixed!

pkg/controller/instance/deletion.go
tomasaschan: I didn't do anything about addressing the other issue, but this should no longer be a problem for this specific case.

The "hard" part here is if there is an _invalid_ metadata annotation; that requires manual user intervention to ensure resources aren't orphaned. I hope the new error message here is informative enough to, at least, enable them to ask for help with enough context for the community to help them with how to unblock.

Skarlso: Cool. I'll try to see later on what I can do about that. :) I picked that issue up.

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+evaluate the child's CEL expressions because the external data is unavailable.
+The child is not deleted and the kro finalizer on the root instance is stuck.
+
+This is one example of a broader lifecycle invariant: deleting an instance must
+remain possible when desired-state projection, identity CEL, readiness CEL, or
+external observation is unavailable. Cleanup needs durable identities and
+ordering information from the resources that normal reconciliation already
+created.
+
+## Proposal
```
jakobmoellerdev, line 24: FYI This is in effect already a partial implementation of https://github.com/kubernetes-sigs/kro/pull/1215/changes#diff-05b43e39cd557133579c0e7f06aa575ea6f7edd64d95e35ee6bf101e8acc3635R79 which only deletes 1 at a time and only triggers on delete. This is a good thing for me 👍

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+External references remain read-only inputs. They are not finalized, mutated,
+or deleted by kro.
+
+#### Overview
+
+The lifecycle is governed by these invariants:
+
+- Managed-resource identity during deletion comes from persisted ApplySet
+  inventory, not from reconstructed desired objects or current graph nodes.
+- Normal reconciliation gives every managed child
```
jakobmoellerdev: please lets not hard code topological order to children anywhere in preparation to your out of scope statement (https://github.com/kubernetes-sigs/kro/pull/1318/changes#diff-4b136391081fd9a6d71b7aba854439406d8b07ef84be7400424b96e554823db8R226). It is important to understand that topological order is a form of serialization of a graph, thus it is unfit to optimize later. setting this label here will force us to always write a topological order number for every resource which is not desired mid/longterm or deprecate/remove the label later.

I suggest to use levelling by depth of the graph immediately, not by topological order immediately but leave the rest of the proposal as is. that way the dependency chain is still respected but you can parallelize. it is parallel to https://github.com/kubernetes-sigs/kro/pull/1215/changes#diff-05b43e39cd557133579c0e7f06aa575ea6f7edd64d95e35ee6bf101e8acc3635R551 that way.

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+  `internal.kro.run/apply-order`. Its value is the child's one-based position
+  in the complete runtime DAG's total topological order.
+- External nodes occupy positions in that total order but are not applied, so
```
jakobmoellerdev: this would disappear with levels as well

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+  managed-resource order values may have gaps.
+- Deletion processes only the highest order still present in inventory.
+- A wave remains active until all of its objects are absent, including objects
```
jakobmoellerdev: waves can simply be the levels and you can converge the 2 concepts

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+  that already have a deletion timestamp.
+- Members with missing or invalid order metadata share one fallback wave. That
```
jakobmoellerdev, line 48: I think this is smart and I like that idea!

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+  wave runs last and has no ordering guarantee within it.
+- The instance finalizer is removed only after the inventory is empty.
+
+#### Normal reconciliation
```
jakobmoellerdev: the paragraph seems ai written, what do you want to say with it? it sounds like youre stating things that are already in place?

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+The instance is already the ApplySet parent. Its annotations persist the union
+of group-kinds and additional namespaces that can contain members. Deletion
+creates the same ApplySet, calls `Project(nil)` to reconstruct that scope from
+the parent annotations, and calls `ListOrphans` with an empty keep-UID set. All
+ApplySet members are consequently deletion candidates.
+
+This inventory has two important properties. First, external references are
+absent because kro never applies them and they are not ApplySet members. Second,
+resources retained from an older GraphRevision remain discoverable even if
+their group-kind, namespace, or node ID is absent from the current graph.
```
jakobmoellerdev: this is only the case if the labels are still in place correct?

docs/design/proposals/ordered-instance-deletion.md
```diff
@@ -0,0 +1,282 @@
...
+parent object, persisted ApplySet annotations, dynamic client, and REST
+mappings. Deletion intentionally reports only instance-level status; it does
+not synthesize per-node states that cannot be persisted without a resolved
+runtime. Keeping deletion after graph resolution would preserve richer node
+status, but would leave the root finalizer stuck whenever resolution fails.
+
+#### Ordered deletion waves
+
+The controller parses every candidate's `internal.kro.run/apply-order`.
+Positive base-10 integers retain their persisted order. Missing, malformed,
```
jakobmoellerdev: this also reads very ai-heavy, can probably be condensed

jakobmoellerdev (commented):
Generally Supportive of merging this now with the fallback layer in place. I do think that we should go for the effort now of introducing levels into the delete order instead of topological order as you will be introducing a kro label which is part of our controller contract.

We should also probably explore whether its not more worthwhile to downgrade the label to an annotation btw since its an application hint.

tomasaschan:
@jakobmoellerdev Thanks for reviewing! I've now implemented the changes proposed:

* Deletion waves are now topological _layers_ rather than simple topological _indices_. All sink nodes are given the highest value, their direct dependencies the next, and so on, enabling concurrent deletion of all child resources that have no dependents.

* Apply order is moved to an annotation rather than label. I also took the opportunity to ensure `kro.run/` and `internal.kro.run/` prefixes are reserved for both labels and annotations.

* I also went over the proposal doc with a manual edit pass; I've been driving this implementation with Claude and Codex, and they kept revising the doc along with the code changes, so slop accumulated... sorry about that.

Skarlso:
I'm going to give this a manual run today. :) stay tuned for results. :)

Skarlso:
Ordered chain:

```yaml
apiVersion: kro.run/v1alpha1  
kind: ResourceGraphDefinition  
metadata:  
  name: orderedchain.test.kro.run  
spec:  
  schema:  
    apiVersion: v1alpha1  
    kind: OrderedChain  
    spec:  
      name: string  
    status:  
      chainTail: ${c.metadata.name}  
  resources:  
    # a -> b -> c  
    # Expected internal.kro.run/apply-order: a=1, b=2, c=3 # Expected deletion waves: c, then b, then a.
      template:  
        apiVersion: v1  
        kind: ConfigMap  
        metadata:  
          name: ${schema.spec.name}-a  
        data:  
          rank: "first"  
  
    - id: b  
      template:  
        apiVersion: v1  
        kind: ConfigMap  
        metadata:  
          name: ${schema.spec.name}-b  
        data:  
          # forces b to depend on a  
          upstream: ${a.metadata.name}  
  
    - id: c  
      template:  
        apiVersion: v1  
        kind: ConfigMap  
        metadata:  
          name: ${schema.spec.name}-c  
        data:  
          # forces c to depend on b  
          upstream: ${b.metadata.name}
```

Describe:
```yaml
  Resources:
    Dependencies:
      Id:  a
    Id:    b
    Dependencies:
      Id:  b
    Id:    c
  State:   Active
  Topological Order:
    a
    b
    c
```

Label hash:
```yaml
internal.kro.run/applyset-inventory-hash: sha256:j8S1fBgU6QoepjeMnDn1Rnra2Wv3VsxHCBrhTs-FTsk
```

Deletion should be c -> b -> a.
Blocking deletion on b with a finalizer.
Script output:
```
== parent: OrderedChain/chain (ns kro-deltest)
   applyset id      : applyset-vBK6_ZopdfvwlNToSIr2ZyIcMZRlbx05gw622J87Yss-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:j8S1fBgU6QoepjeMnDn1Rnra2Wv3VsxHCBrhTs-FTsk
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: -
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     True     AllResourcesReady
   condition        : Ready              True     Ready

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
   ConfigMap/chain-a                      -                      1
   ConfigMap/chain-b                      -                      2
   ConfigMap/chain-c                      -                      3
```

Commence deletion.

Worked. Now, let's block chain-b.

Put a finalizer on it:
```
kubectl patch cm chain-b -n kro-deltest --type=merge \  
  -p '{"metadata":{"finalizers":["test.kro.run/block"]}}'
```

Correct:
```
== parent: OrderedChain/chain (ns kro-deltest)
   applyset id      : applyset-vBK6_ZopdfvwlNToSIr2ZyIcMZRlbx05gw622J87Yss-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:j8S1fBgU6QoepjeMnDn1Rnra2Wv3VsxHCBrhTs-FTsk
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: 2026-08-12T07:50:48Z
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     Unknown  UnderDeletion
   condition        : Ready              Unknown  UnderDeletion

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
   ConfigMap/chain-a                      -                      1
   ConfigMap/chain-b                      -                      2      yes
```

b is terminating, a is fine. removing finalizer from b.
All gone:
```
== parent: OrderedChain/chain (ns kro-deltest)
   applyset id      : applyset-vBK6_ZopdfvwlNToSIr2ZyIcMZRlbx05gw622J87Yss-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:j8S1fBgU6QoepjeMnDn1Rnra2Wv3VsxHCBrhTs-FTsk
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: 2026-08-12T07:50:48Z
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     Unknown  UnderDeletion
   condition        : Ready              Unknown  UnderDeletion

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
```

```
k get orderedchain -A
No resources found
```

Next, externalRef:
```
== parent: ExtRefDeletion/extref (ns kro-deltest)
   applyset id      : applyset-zMN89yuYNTOQJrZ187U3U8L6tviMgim4l-uY6rf8qEM-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:i-8iJmBzCB9VaoBq02nozJx_II36lhZfXYN-Pzj7tg4
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: -
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     True     AllResourcesReady
   condition        : Ready              True     Ready

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
   ConfigMap/extref-child                 -                      2
```

Expected Order 2 because 1 is unapplied external.

Add a finalizer.
```
kubectl patch cm extref-child -n kro-deltest --type=merge \  
  -p '{"metadata":{"finalizers":["test.kro.run/block"]}}'
```

try deletion
```
kubectl delete cm deletion-source -n kro-deltest
```

```
== parent: ExtRefDeletion/extref (ns kro-deltest)
   applyset id      : applyset-zMN89yuYNTOQJrZ187U3U8L6tviMgim4l-uY6rf8qEM-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:i-8iJmBzCB9VaoBq02nozJx_II36lhZfXYN-Pzj7tg4
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: -
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     False    NotReady
   condition        : Ready              False    NotReady

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
   ConfigMap/extref-child                 -                      2
```

Good, no external ref.

Delete
```
kubectl delete extrefdeletion extref -n kro-deltest --wait=false
```

Check agian
```
== parent: ExtRefDeletion/extref (ns kro-deltest)
   applyset id      : applyset-zMN89yuYNTOQJrZ187U3U8L6tviMgim4l-uY6rf8qEM-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:i-8iJmBzCB9VaoBq02nozJx_II36lhZfXYN-Pzj7tg4
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: 2026-08-12T08:02:26Z
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     Unknown  UnderDeletion
   condition        : Ready              Unknown  UnderDeletion

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
   ConfigMap/extref-child                 -                      2      yes
```

It's terminating. Removed the finalizer.
```
➜  pr-1318 git:(fix/1316/deletion-with-externalref) ✗ kubectl patch cm extref-child -n kro-deltest --type=merge -p '{"metadata":{"finalizers":[]}}'
configmap/extref-child patched
➜  pr-1318 git:(fix/1316/deletion-with-externalref) ✗ ./show-order.sh ExtRefDeletion extref
== parent: ExtRefDeletion/extref (ns kro-deltest)
   applyset id      : applyset-zMN89yuYNTOQJrZ187U3U8L6tviMgim4l-uY6rf8qEM-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:i-8iJmBzCB9VaoBq02nozJx_II36lhZfXYN-Pzj7tg4
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: 2026-08-12T08:02:26Z
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     Unknown  UnderDeletion
   condition        : Ready              Unknown  UnderDeletion

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
...
== parent: ExtRefDeletion/extref (ns kro-deltest) (not found)
```

Gone.

Collections:

```
== parent: CollectionDeletion/coll (ns kro-deltest)
   applyset id      : applyset-Gc-rMoNmlptpdF5xFXCd_RrooMqG4Bj0gPzVod3N7K4-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:nFw00BthmzFpgmDJ8-LCHbiQjYSvho6VcqJtiNSxUaQ
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: -
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     True     AllResourcesReady
   condition        : Ready              True     Ready

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
   ConfigMap/coll-base                    -                      2
   ConfigMap/coll-one                     -                      3
   ConfigMap/coll-three                   -                      3
   ConfigMap/coll-two                     -                      3
```

Expect `coll-base`=2 and **all three** of `coll-one`/`coll-two`/`coll-three`=3.  
A forEach collection is one graph node, so its members share one order and MUST delete as a single wave.

Blocking one for testing:
```
kubectl patch cm coll-two -n kro-deltest --type=merge \  
  -p '{"metadata":{"finalizers":["test.kro.run/block"]}}'
```

Deleting the collection:
```
kubectl delete collectiondeletion coll -n kro-deltest --wait=false
```

Expectation: one and three go away two is terminating 

```
== parent: CollectionDeletion/coll (ns kro-deltest)
   applyset id      : applyset-Gc-rMoNmlptpdF5xFXCd_RrooMqG4Bj0gPzVod3N7K4-v1
   tooling          : kro/devel
   group-kinds      : 'ConfigMap'
   extra namespaces : ''
   inventory hash   : sha256:nFw00BthmzFpgmDJ8-LCHbiQjYSvho6VcqJtiNSxUaQ
   finalizers       : ['kro.run/finalizer']
   deletionTimestamp: 2026-08-12T08:07:48Z
   condition        : InstanceManaged    True     Managed
   condition        : GraphResolved      True     Resolved
   condition        : ResourcesReady     Unknown  UnderDeletion
   condition        : Ready              Unknown  UnderDeletion

== children carrying an apply-order label in ns kro-deltest
   KIND/NAME                              NODE-ID                ORDER  TERMINATING
   ConfigMap/coll-base                    -                      2
   ConfigMap/coll-two                     -                      3      yes
```

Removed finalizer:
```
== parent: CollectionDeletion/coll (ns kro-deltest)
   (not found)
```

All RGDs were successfully removed.

Testing failure case where things hang because something never got an ApplySet.

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: widgetowner.test.kro.run
spec:
  schema:
    apiVersion: v1alpha1
    kind: WidgetOwner
    spec:
      name: string
    status:
      widgetName: ${widget.metadata.name}
  resources:
    # cm is applied first and always succeeds. widget targets a CRD that the
    # I remove, which is what makes the failure. The instance
    # gets far enough to have a finalizer but not far enough to persist a
    # an ApplySet inventory. Hopefully.
    - id: cm
      template:
        apiVersion: v1
        kind: ConfigMap
        metadata:
          name: ${schema.spec.name}-cm
        data:
          owner: ${schema.spec.name}

    - id: widget
      template:
        apiVersion: test.kro.run/v1alpha1
        kind: Widget
        metadata:
          name: ${schema.spec.name}-widget
        spec:
          value: ${cm.data.owner}

```

The CRD:
```
metadata:  
  name: widgets.test.kro.run  
spec:  
  group: test.kro.run  
  scope: Namespaced  
  names:  
    plural: widgets  
    singular: widget  
    kind: Widget  
    listKind: WidgetList  
  versions:  
    - name: v1alpha1  
      served: true  
      storage: true  
      schema:  
        openAPIV3Schema:  
          type: object  
          properties:  
            spec:  
              type: object  
              properties:  
                value:  
                  type: string  
            status:  
              type: object  
```

Now wait for the RGD:

```
kubectl wait --for=jsonpath='{.status.state}'=Active \  
  resourcegraphdefinition/widgetowner.test.kro.run --timeout=60s
```
```
kubectl wait --for=jsonpath='{.status.state}'=Active \
  resourcegraphdefinition/widgetowner.test.kro.run --timeout=60s
resourcegraphdefinition.kro.run/widgetowner.test.kro.run condition met
```

remove the crd
```
kubectl delete crd widgets.test.kro.run
```

And apply the instance:
```
kubectl apply -f 04-widget-instance-never-applied.yaml
```

```
apiVersion: kro.run/v1alpha1  
kind: WidgetOwner  
metadata:  
  name: never-applied  
  namespace: kro-deltest  
spec:  
  name: never-applied
```

Error from kro as expected...
```
026-08-12T10:34:23+02:00       ERROR   dynamic-controller.watch-manager        Watch error     {"gvr": "test.kro.run/v1alpha1, Resource=widgets", "error": "failed to list *v1.PartialObjectMetadata: the server could not find the requested resource"}
github.com/kubernetes-sigs/kro/pkg/dynamiccontroller.(*WatchManager).newWatch.func1
        /Users/C5353099/Projects/kubernetes/kro/pkg/dynamiccontroller/watch_manager.go:195
k8s.io/client-go/tools/cache.(*sharedIndexInformer).SetWatchErrorHandler.func1
        /Users/C5353099/go/pkg/mod/k8s.io/client-go@v0.35.0/tools/cache/shared_informer.go:497
k8s.io/client-go/tools/cache.(*Reflector).RunWithContext.func1
        /Users/C5353099/go/pkg/mod/k8s.io/client-go@v0.35.0/tools/cache/reflector.go:370
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
        /Users/C5353099/go/pkg/mod/k8s.io/apimachinery@v0.35.0/pkg/util/wait/backoff.go:233
k8s.io/apimachinery/pkg/util/wait.BackoffUntilWithContext.func1
        /Users/C5353099/go/pkg/mod/k8s.io/apimachinery@v0.35.0/pkg/util/wait/backoff.go:255
k8s.io/apimachinery/pkg/util/wait.BackoffUntilWithContext
        /Users/C5353099/go/pkg/mod/k8s.io/apimachinery@v0.35.0/pkg/util/wait/backoff.go:256
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
        /Users/C5353099/go/pkg/mod/k8s.io/apimachinery@v0.35.0/pkg/util/wait/backoff.go:233
k8s.io/client-go/tools/cache.(*Reflector).RunWithContext
        /Users/C5353099/go/pkg/mod/k8s.io/client-go@v0.35.0/tools/cache/reflector.go:368
k8s.io/client-go/tools/cache.(*controller).RunWithContext.(*Group).StartWithContext.func3
        /Users/C5353099/go/pkg/mod/k8s.io/apimachinery@v0.35.0/pkg/util/wait/wait.go:63
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1
        /Users/C5353099/go/pkg/mod/k8s.io/apimachinery@v0.35.0/pkg/util/wait/wait.go:72
```

Okay, hopefully this is expected too:
```
2026-08-12T10:35:15+02:00       DEBUG   dynamic-controller.watch-manager        Watch stopped   {"gvr": "test.kro.run/v1alpha1, Resource=widgets"}
2026-08-12T10:35:15+02:00       ERROR   dynamic-controller.watch-coordinator    Failed to ensure watch  {"gvr": "test.kro.run/v1alpha1, Resource=widgets", "error": "cache sync timeout for test.kro.run/v1alpha1, Resource=widgets"}
```

Looks okay:
```
k get widgetowner -n kro-deltest
NAME            STATE         READY   AGE
never-applied   IN_PROGRESS   False   4m36s
```

```yaml
Metadata:
  Creation Timestamp:  2026-08-12T08:34:15Z
  Finalizers:
    kro.run/finalizer
  Generation:        1
  Resource Version:  13043
  UID:               dd79376e-9f3d-4bb7-a6fb-e9ba1e084c7c
Spec:
  Name:  never-applied
Status:
  Conditions:
    Last Transition Time:  2026-08-12T08:34:15Z
    Message:               instance is properly managed with finalizers and labels
    Observed Generation:   1
    Reason:                Managed
    Status:                True
    Type:                  InstanceManaged
    Last Transition Time:  2026-08-12T08:34:15Z
    Message:               runtime graph created and all resources resolved
    Observed Generation:   1
    Reason:                Resolved
    Status:                True
    Type:                  GraphResolved
    Last Transition Time:  2026-08-12T08:34:15Z
    Message:               resource reconciliation failed: project failed: RESTMapping failed for test.kro.run/v1alpha1, Kind=Widget: no matches for kind "Widget" in version "test.kro.run/v1alpha1"
    Observed Generation:   1
    Status:                False
    Type:                  ResourcesReady
    Last Transition Time:  2026-08-12T08:34:15Z
    Message:               resource reconciliation failed: project failed: RESTMapping failed for test.kro.run/v1alpha1, Kind=Widget: no matches for kind "Widget" in version "test.kro.run/v1alpha1"
    Observed Generation:   1
    Status:                False
    Type:                  Ready
  State:                   IN_PROGRESS
```

Now, deleting:

```
kubectl delete widgetowner never-applied -n kro-deltest --wait=false
```

And, success
```
k describe widgetowner -n kro-deltest
No resources found in kro-deltest namespace.
```


And with that... I think this is done. :D 🙇‍♀️

pkg/graph/validation.go
```diff
@@ -363,8 +366,38 @@ func validateNoKROOwnedLabels(resourceID string, resourceObject map[string]inter
...
+		}
+	}
+
+	return nil
+}
+
+// validateNoKROOwnedAnnotations prevents resource templates from overriding
+// annotations used as persisted controller state.
+func validateNoKROOwnedAnnotations(resourceID string, resourceObject map[string]interface{}) error {
+	annotationsRaw, found, err := unstructured.NestedFieldCopy(resourceObject, "metadata", "annotations")
```
jakobmoellerdev: there is `unstructured.NestedStringMap` that you can use, also since you dont edit we dont need a copy

pkg/graph/validation.go
```diff
@@ -349,8 +352,8 @@ func isRequiredIdentityField(path string, resourceNamespaced, instanceNamespaced
 	}
 }
 
-// validateNoKROOwnedLabels enforces that the resource template doesn't define any label with
-// LabelKROPrefix (kro.run/). These labels are reserved for internal use ONLY.
+// validateNoKROOwnedLabels enforces that resource templates do not define
+// labels in either controller-owned namespace.
 func validateNoKROOwnedLabels(resourceID string, resourceObject map[string]interface{}) error {
 	labelsRaw, found, err := unstructured.NestedFieldCopy(resourceObject, "metadata", "labels")
```
jakobmoellerdev: same here I think, I know you didnt add this initially

pkg/runtime/runtime.go
```diff
@@ -56,12 +62,17 @@ func FromGraph(g *graph.Graph, instance *unstructured.Unstructured, rgdConfig gr
 		metrics.RuntimeCreationDuration.Observe(duration.Seconds())
 		metrics.RuntimeCreationTotal.Inc()
 	}()
+	applyOrders, err := applyOrdersForDAG(g.DAG)
```
jakobmoellerdev: shouldnt this be computed when the runtime is setup, just like g.TopologicalOrder?

pkg/controller/instance/controller.go
```diff
@@ -428,3 +465,21 @@ func (c *Controller) applyManagedFinalizerAndLabels(rcx *ReconcileContext) (*uns
...
+	}
+	annotations := obj.GetAnnotations()
+	for _, key := range []string{
+		applyset.ApplySetToolingAnnotation,
+		applyset.ApplySetGKsAnnotation,
+		applyset.ApplySetAdditionalNamespacesAnnotation,
+		applyset.ApplySetInventoryHashAnnotation,
+	} {
+		if _, found := annotations[key]; found {
+			return true
```
jakobmoellerdev, line 484: shouldnt we return true only if all of them are set or is there a reason to be more lenient here?

pkg/metadata/labels.go
```diff
@@ -31,10 +31,18 @@ import (
...
+	// InternalLabelKROPrefix is reserved for controller-internal metadata.
+	InternalLabelKROPrefix = "internal." + LabelKROPrefix
+	// InternalAnnotationKROPrefix is reserved for controller-internal metadata.
+	InternalAnnotationKROPrefix = "internal." + AnnotationKROPrefix
 )
 
 const (
 	NodeIDLabel = LabelKROPrefix + "node-id"
+	// ApplyOrderAnnotation persists a managed resource's reverse topological deletion wave.
+	ApplyOrderAnnotation = InternalAnnotationKROPrefix + "apply-order"
```
jakobmoellerdev: nit: you store this annotation under labels.go so maybe not the best place

jakobmoellerdev: I think its worth documenting the new deletion behavior under the docs in advanced or similar since it is very custom to kro. Would you mind adding a few words for users now?

pkg/controller/instance/applyset/applyset.go
jakobmoellerdev: `ListOrphans` treats every `RESTMapping` failure as fatal.                                                                                                             
   If a CRD that was previously in the instance's scope has since been deleted,                                                                                                               
   the mapper returns `*meta.NoKindMatchError`, `discoverDeletionInventory`                                                                                                                   
   returns error, and `reconcileDeletion` marks "deletion blocked" indefinitely.                                                                                                              
   Only manual annotation surgery unblocks it. Maybe we should be treating `IsNoMatchError`                                                                                                              
   (only) as "resource type gone, no children of this kind possible" and continuing.

pkg/controller/instance/controller.go
```diff
@@ -386,6 +401,16 @@ func (c *Controller) applyManagedFinalizerAndLabels(rcx *ReconcileContext) (*uns
 	// Fast path: if everything is already correct → no patch
 	hasFinalizer := metadata.HasInstanceFinalizer(obj)
 	needFinalizer := !hasFinalizer
+	hasInventoryMetadata := hasApplySetInventoryMetadata(obj)
+	if needFinalizer && hasInventoryMetadata {
```
jakobmoellerdev, line 405: An instance that already carries kro's finalizer from a previous release has `needFinalizer=false`, so the SSA never writes the ApplySet parent labels/annotations. Because deletion now runs before graph resolution (`controller.go:241`), if such an instance receives a `deletionTimestamp` before its first post-upgrade normal reconcile has run `Project()`, `ValidateParentInventory` fails on the missing `applyset.kubernetes.io/id` label and the finalizer is stuck forever with no automatic recovery.

not sure if we really wanna fix that since its such a rare edge case but worth noting

pkg/metadata/labels.go
```diff
@@ -31,10 +31,18 @@ import (
 const (
 	// LabelKROPrefix is the label key prefix used to identify KRO owned resources.
 	LabelKROPrefix = v1alpha1.KRODomainName + "/"
+	// AnnotationKROPrefix is the annotation key prefix reserved for KRO.
+	AnnotationKROPrefix = v1alpha1.KRODomainName + "/"
```
jakobmoellerdev: why do we have a second prefix if its the same as the label prefix, shouldn't it just be `KROPrefix` ?

pkg/metadata/labels.go
```diff
@@ -31,10 +31,18 @@ import (
...
+	// InternalLabelKROPrefix is reserved for controller-internal metadata.
+	InternalLabelKROPrefix = "internal." + LabelKROPrefix
+	// InternalAnnotationKROPrefix is reserved for controller-internal metadata.
+	InternalAnnotationKROPrefix = "internal." + AnnotationKROPrefix
```
jakobmoellerdev: same here

jakobmoellerdev (commented):
almost good to go!

pkg/metadata/labels.go
tomasaschan: `KROPrefix` isn't defined currently. I can do that - do you want me to refactor so all current label constants also use it?

jakobmoellerdev: Yeah, I think I was trying to say that you can demote `LabelKROPrefix` to be `KROPrefix`, and then just reuse that everywhere

pkg/controller/instance/controller.go
tomasaschan: I added a release note draft to the PR body about this; is there somewhere I should put that to ensure it makes into the release notes when this gets shipped?

pkg/metadata/labels.go
tomasaschan: Added a docs page for this, feel free to suggest changes. Not sure if it's at the right level of abstraction... 😅

pkg/controller/instance/controller.go
tomasaschan: Yeah, if any one is present we want to run normal validation and fail if the metadata is invalid; otherwise, we risk overwriting a corrupt applyset with an empty on, and orphan managed resources.

pkg/runtime/runtime.go
tomasaschan: I thought I made this cached-and-reused, but maybe I didn't move it far enough up the stack. Fixed.

pkg/controller/instance/controller.go
jakobmoellerdev: I think thats good enough, we will add it on next release manually

pkg/graph/validation.go
tomasaschan: I went for `unstructured.NestedFieldNoCopy` to avoid the copy. Unfortunately there isn't an exported `NestedStringMapNoCopy`.

tomasaschan:
@jakobmoellerdev Thanks for taking another pass! I think I've addressed all your feedback, but feel free to push back if you're missing anything.

jakobmoellerdev (approved):
/lgtm

PR #1321, opened 2026-07-06
Fix RGD CRD description examples to match parser behavior

iammerus:
**What this PR does / why we need it:**

The RGD CRD field descriptions show examples that the parsers actually reject, which misleads users (and LLMs) consuming the schema:

- `schema.spec` shows pipe-separated markers with `min=`/`max=` (`"integer | default=1 | min=1 | max=10"`), but `pkg/simpleschema` splits on the first pipe only, markers are space-separated, and the valid marker names are `minimum`/`maximum`. Corrected example: `"integer | default=1 minimum=1 maximum=10"`.
- `readyWhen` and `includeWhen` show bare expressions (e.g. `["database.status.phase == 'Running'"]`), but `ParseConditionExpressions` in `pkg/graph/parser/conditions.go` requires each entry to be a standalone `${...}` expression and rejects the documented form with "only standalone expressions are allowed".

This updates the Go doc comments in `api/v1alpha1/resourcegraphdefinition_types.go` and regenerates the CRDs with `make manifests` (which also refreshes the `website/docs/api/crds` copies).

Doc-only change, no behavior change.

### DIFF
--- a/api/v1alpha1/resourcegraphdefinition_types.go
+++ b/api/v1alpha1/resourcegraphdefinition_types.go
@@ -72,7 +72,8 @@ type Schema struct {
 	// Spec defines the schema for the instance's spec section using SimpleSchema syntax.
 	// This becomes the OpenAPI schema for instances of the generated CRD.
 	// Use SimpleSchema's concise syntax to define fields, types, defaults, and validations.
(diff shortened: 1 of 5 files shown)


jakobmoellerdev (commented):
/ok-to-test

jakobmoellerdev (approved):
/lgtm

PR #1325, opened 2026-07-09
Update condition telemetry to reflect author conditions

lalitg327:
[#1301](https://github.com/kubernetes-sigs/kro/pull/1301) added author-defined custom status conditions. However, the instance condition telemetry (metrics and events) doesn't reflect them.

The emitter reads conditions from the in-memory instance object, which the marker populates with kro's built-in conditions. When an RGD declares author conditions, updateStatus writes only the author's conditions to the wire (built-ins are suppressed), so the in-memory object no longer matches what's persisted. As a result the emitter reports built-in conditions that never reach etcd and misses the author conditions entirely.

This PR mirrors the persisted status onto the in-memory object before the API write, so the emitter reports the conditions kro actually persists — the author's conditions when declared, built-ins otherwise.

### DIFF
--- a/pkg/controller/instance/status.go
+++ b/pkg/controller/instance/status.go
@@ -160,6 +160,9 @@ func (c *Controller) updateConditionsStatus(ctx context.Context, inst *unstructu
 		}
 		status["state"] = string(v1alpha1.InstanceStateError)
 		cur.Object["status"] = status
+		// Mirror persisted status onto the instance so the deferred
+		// emitters read what's on the wire, not the marker's built-ins.
+		inst.Object["status"] = status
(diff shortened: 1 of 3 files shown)


jakobmoellerdev (commented):
/ok-to-test
/lgtm

cheeseandcereal (approved):

michaelhtm (approved):
/lgtm

lalitg327:
@jakobmoellerdev Thanks for reviewing the PR. It is passing all presubmits test. Can you approve the changes?

jakobmoellerdev (approved):

Issue #1054, opened 2026-02-17
externalRef incorrectly requires namespace for cluster-scoped resources


I558191:
### Description

 **Observed Behavior**:                                                                                                                                                
                                                                                                                                                                        
After upgrading to Kro v0.8.5, the ResourceGraphDefinition CRD validation incorrectly requires the `metadata.namespace` field for `externalRef` resources, even when referencing cluster-scoped resources.                                                                                                                                 

When applying an RGD with an `externalRef` to a cluster-scoped resource (without a namespace field), the validation fails with:                                       
`spec.resources[0].externalRef.metadata.namespace: Required value`                                                                      

  Inspecting the CRD schema shows that `namespace` is marked as required:
  ```yaml
  externalRef:
    properties:
      metadata:
        properties:
          name:
            type: string
          namespace:
            type: string
        required:
        - name
        - namespace  # ← This should NOT be required for cluster-scoped resources
````
**Expected Behavior:**

According to the https://kro.run/docs/concepts/rgd/resource-definitions/external-references/:

namespace: Optional: Defaults to instance namespace

And specifically states:
Cluster-scoped resources: StorageClasses, ClusterIssuers (omit namespace)

The namespace field should be:
1. Optional in the CRD schema validation
2. Omitted when referencing cluster-scoped resources

This aligns with https://github.com/kubernetes-sigs/kro/pull/976 which states: "Kro 0.8.0 and above no longer accept metadata.namespace on CRs that are defined at cluster-scope"

**Reproduction Steps:**

1. Create an RGD that references a cluster-scoped custom resource (e.g., Crossplane Subaccount):

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: some-rgd
spec:
  schema:
    apiVersion: v1alpha1
    kind: someKind
    spec:
      name: string
  resources:
    - id: subaccount
      externalRef:
        apiVersion: account.btp.sap.crossplane.io/v1alpha1
        kind: Subaccount
        metadata:
          name: "${schema.spec.name}"
          # namespace intentionally omitted - Subaccount is cluster-scoped
```

2. Verify the referenced resource is cluster-scoped:
```
$ kubectl describe crd subaccounts.account.btp.sap.crossplane.io | grep Scope
Scope:        Cluster
```

3. Apply the RGD:
```
$ kubectl apply -f rgd.yaml
The ResourceGraphDefinition "some-rgd" is invalid:
[...]

PR #1326, opened 2026-07-10
docs: Helm does not update CRDs when upgrading to latest

johanngoltz:
In #1054, an issue was reported that was caused by upgrading Kro via Helm, but not installing the corresponding CRDs. Unlike instructions for upgrading to a specific version, those for upgrading to the latest do not mention that CRDs must be upgraded separately.


cheeseandcereal (approved):
Nice! Yeah this is a quirk of helm. Makes sense to call out to me.

jakobmoellerdev (approved):

PR #1327, opened 2026-07-11
feat: add shortNames and categories in the RGD

tcondeixa:
## Summary

Expose `shortNames` and `categories` as optional fields on the `Schema` struct, allowing RGD authors to control how their generated CRDs appear in kubectl queries. Both fields default to empty, preserving existing behaviour.

## Changes

- `api/v1alpha1/resourcegraphdefinition_types.go` — add `ShortNames []string` and `Categories []string` to `Schema`
- `pkg/graph/crd/crd.go` — thread both fields through `SynthesizeCRD` → `newCRD` → `Names`
- `pkg/graph/crd/crd_test.go` — update direct `newCRD` call with new parameters
- Generated CRD manifests and website docs updated via `make manifests`

## Tests

- [x] Existing `pkg/graph/crd` tests pass unchanged
- [x] New tests added to `TestNewCRD` covering `shortNames` (`["wa", "webapp"]`) and `categories` (`["all", "kro"]`) pass-through from RGD schema to generated CRD `spec.names`

Tests done using local `kind`:
- [x] Existing RGDs deployed on a cluster are not affected after upgrading kro — existing CRDs retain their current `spec.names` without `shortNames` or `categories` being added
- [x] New RGD without `shortNames` and `categories` creates CRD without `shortNames` and `categories` (works as before)
- [x] New RGD with `shortNames` and `categories` creates CRD with correct `spec.names.shortNames` and `spec.names.categories`, verified via `kubectl api-resources` and `kubectl get <shortName>`

## Notes

Adding or changing `shortNames`/`categories` on an existing RGD will not trigger a CRD update unless a schema change also occurs. This is a pre-existing gap that also affects `additionalPrinterColumns` and will be addressed in a separated PR (issue) if there is agreement.

This PR was written in part with the assistance of generative AI


michaelhtm (approved):
Nice!
/ok-to-test

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -105,6 +105,19 @@ type Schema struct {
 	// +kubebuilder:validation:Optional
 	AdditionalPrinterColumns []extv1.CustomResourceColumnDefinition `json:"additionalPrinterColumns,omitempty"`
 
+	// ShortNames are short names for the generated CRD that can be used with kubectl.
+	// Example: ["wa", "webapp"] allows `kubectl get wa` instead of `kubectl get webapp`.
+	//
+	// +kubebuilder:validation:Optional
+	ShortNames []string `json:"shortNames,omitempty"`
```
cheeseandcereal, line 112: Changing only these fields on an existing RGD will never reach the cluster: `CRDWrapper.Ensure` skips the patch when `crdcompat.CompareVersions` reports no changes (pkg/client/crd.go:157-166), and that comparison only looks at `spec.versions` (pkg/graph/crd/compat/compat.go:25-74), never `spec.names`. 

The RGD still reports Active/KindReady=True and the new GraphRevision snapshot contains the new values, so users get silent divergence on the most common flow (adding a short name to an existing RGD). Since `spec.names` changes are never breaking, could we extend the up-to-date check in `Ensure` with something like `!equality.Semantic.DeepEqual(existing.Spec.Names, desired.Spec.Names)` or some other solution? Otherwise this is likely going to introduce unexpected behavior.

pkg/graph/crd/crd.go
```diff
@@ -56,10 +56,12 @@ func newCRD(group, apiVersion, kind string, schema *extv1.JSONSchemaProps, scope
...
 			Names: extv1.CustomResourceDefinitionNames{
-				Kind:     kind,
-				ListKind: kind + "List",
-				Plural:   pluralKind,
-				Singular: strings.ToLower(kind),
+				Kind:       kind,
+				ListKind:   kind + "List",
+				Plural:     pluralKind,
+				Singular:   strings.ToLower(kind),
+				ShortNames: shortNames,
```
cheeseandcereal, line 63: Removal can never propagate.

`CRDWrapper.patch` sends a JSON merge patch of the marshaled desired CRD (pkg/client/crd.go:196-210), and upstream `CustomResourceDefinitionNames.ShortNames`/`Categories` are `json:",omitempty"`.

When a user deletes these fields from the RGD, the keys are absent from the patch document and merge-patch semantics leave the existing values untouched. i.e.: existing `{"names":{"kind":"WebApp","shortNames":["wa"]}}` patched with `{"names":{"kind":"WebApp"}}` still yields `shortNames:["wa"]`.

This differs from additionalPrinterColumns, which live inside the `spec.versions` array and are replaced wholesale on patch. So once set, these fields are stuck on the CRD until manual CRD surgery or RGD delete/recreate (which cascades to instances). Just comparing won't fix it. The patch will need to have an explicit null to properly clear/support removal as expected.

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -105,6 +105,19 @@ type Schema struct {
 	// +kubebuilder:validation:Optional
 	AdditionalPrinterColumns []extv1.CustomResourceColumnDefinition `json:"additionalPrinterColumns,omitempty"`
 
+	// ShortNames are short names for the generated CRD that can be used with kubectl.
+	// Example: ["wa", "webapp"] allows `kubectl get wa` instead of `kubectl get webapp`.
+	//
+	// +kubebuilder:validation:Optional
```
cheeseandcereal, line 111: The API server enforces DNS-1035 on every shortName and category (apiextensions validation), so an invalid value like `WA` is accepted at RGD admission and only fails deep in reconcile as KindReady=False (pkg/controller/resourcegraphdefinition/controller_reconcile.go:389-391), then retries forever.

Other fields already validate at admission (`Kind` and `APIVersion` have Pattern markers at lines 51 and 60). controller-gen is pinned at v0.20.0 which supports item markers, so suggest for both fields: `+kubebuilder:validation:items:Pattern=^[a-z]([-a-z0-9]*[a-z0-9])?$`, `+kubebuilder:validation:items:MaxLength=63`, and a sane `+kubebuilder:validation:MaxItems`. Also consider `+listType=set` or a builder-side dedup check so `["wa","wa"]` is rejected. Then `make generate && make manifests`.

api/v1alpha1/resourcegraphdefinition_types.go
```diff
@@ -105,6 +105,19 @@ type Schema struct {
 	// +kubebuilder:validation:Optional
 	AdditionalPrinterColumns []extv1.CustomResourceColumnDefinition `json:"additionalPrinterColumns,omitempty"`
 
+	// ShortNames are short names for the generated CRD that can be used with kubectl.
+	// Example: ["wa", "webapp"] allows `kubectl get wa` instead of `kubectl get webapp`.
```
cheeseandcereal, line 109: I think this comment is wrong. Both wa and webapp are shortnames, the full name is "webapplications".

Need to update this example/wording to be correct.

helm/crds/kro.run_resourcegraphdefinitions.yaml
```diff
@@ -349,6 +357,13 @@ spec:
                     x-kubernetes-validations:
                     - message: scope is immutable
                       rule: self == oldSelf
+                  shortNames:
```
cheeseandcereal, line 360: You added descriptions to the CRD/api-spec here, but there are no docs (outside of api references) for these new fields.

For a new feature like this which adds new CRD fields, the docs really need to be updated too.

pkg/graph/crd/crd.go
```diff
@@ -29,10 +29,10 @@ import (
...
 // scope must be either extv1.NamespaceScoped or extv1.ClusterScoped; defaults to NamespaceScoped.
 func SynthesizeCRD(group, apiVersion, kind string, spec, status extv1.JSONSchemaProps, statusFieldsOverride bool, scope extv1.ResourceScope, rgSchema *v1alpha1.Schema) *extv1.CustomResourceDefinition {
-	return newCRD(group, apiVersion, kind, newCRDSchema(spec, status, statusFieldsOverride), scope, rgSchema.AdditionalPrinterColumns, rgSchema.Metadata)
+	return newCRD(group, apiVersion, kind, newCRDSchema(spec, status, statusFieldsOverride), scope, rgSchema.AdditionalPrinterColumns, rgSchema.Metadata, rgSchema.ShortNames, rgSchema.Categories)
 }
 
-func newCRD(group, apiVersion, kind string, schema *extv1.JSONSchemaProps, scope extv1.ResourceScope, additionalPrinterColumns []extv1.CustomResourceColumnDefinition, metadata *v1alpha1.CRDMetadata) *extv1.CustomResourceDefinition {
+func newCRD(group, apiVersion, kind string, schema *extv1.JSONSchemaProps, scope extv1.ResourceScope, additionalPrinterColumns []extv1.CustomResourceColumnDefinition, metadata *v1alpha1.CRDMetadata, shortNames, categories []string) *extv1.CustomResourceDefinition {
```
cheeseandcereal, line 35: Nit: The function signature here is getting really large. Might be worth considering how to shorten by passing something like `rgSchema` through (`SynthesizeCRD` already takes to to extract fields from it) and get the fields off of that.

pkg/graph/crd/crd_test.go
```diff
@@ -291,19 +295,51 @@ func TestNewCRD(t *testing.T) {
...
+			expectedPrinterColumns: defaultAdditionalPrinterColumns,
+			expectedCategories:     []string{"all", "kro"},
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			schema := &extv1.JSONSchemaProps{Type: "object"}
-			crd := newCRD(tt.group, tt.apiVersion, tt.kind, schema, tt.scope, tt.printerColumns, nil)
+			crd := newCRD(tt.group, tt.apiVersion, tt.kind, schema, tt.scope, tt.printerColumns, nil, tt.shortNames, tt.categories)
```
cheeseandcereal, line 333: The new cases only cover `newCRD` directly, and neither sets both fields at once, so the actual exported pass-through at crd.go:32 (`rgSchema.ShortNames, rgSchema.Categories`) has no coverage and swapping those two arguments would still pass the suite.

May be worth adding a TestSynthesizeCRD case with both fields set to distinct values, e.g. `Schema{ShortNames: []string{"wa"}, Categories: []string{"kro"}}`, asserting both land in `crd.Spec.Names`. (May not be necessary if the function signature is refactored.)

pkg/graph/crd/crd_test.go
```diff
@@ -291,19 +295,51 @@ func TestNewCRD(t *testing.T) {
 				},
 			},
 		},
+		{
```
cheeseandcereal, line 298: It would be worth adding an integration test(s) for this given it's a whole new feature in addition to these unit tests.

jakobmoellerdev:
@tcondeixa are you planning to address the comments?

jakobmoellerdev:
/close

superceded by https://github.com/kubernetes-sigs/kro/pull/1347

tcondeixa:
sorry, but I was on vacations. I will check the feedback provided at least to learn. thanks

PR #1329, opened 2026-07-13
Compute condition duration metric at scrape time via a Collector

lalitg327:
Compute `instance_condition_current_status_seconds` at scrape time so it stays accurate while the controller is idle.

## Background

`instance_condition_current_status_seconds` was a `GaugeVec` whose value was computed at reconcile time as `time.Since(lastTransitionTime)` and stored. Since kro only runs code during reconciles, once an instance reaches steady state the stored value stops updating. It keeps reporting the duration as of the last reconcile, not the actual elapsed time:

```
T=30s   last reconcile → gauge = 30
T=630s  scrape → still reads 30   (reality: 630s in this state)
```

Metric agents like CloudWatch and Datadog read the gauge as-is, so the value must be correct when scraped. Shortening the resync period would keep it fresh but wastes API-server calls at scale, and the metric shouldn't drive reconciliation.

## Change

Back the metric with a custom `prometheus.Collector` (`durationCollector`) instead of a `GaugeVec`:

- **At reconcile time:** cache each condition's `LastTransitionTime`, keyed by `(gvr, namespace, name, condition_type)` with status and reason stored as the value. Keying by condition type keeps at most one series per condition type per instance: a status/reason change overwrites the previous entry, and a condition that disappears from the wire is evicted.
- **At scrape time:** `Collect()` computes `time.Now() - cachedTimestamp` per entry and emits it as a gauge.

The key deliberately excludes status and reason so the cache holds at most one series per condition type per instance, avoiding stale/phantom series when the cache diverges from etcd (e.g. a failed status write).

The value is derived at scrape time, so it stays accurate with no forced reconciles and no added API-server load. Timestamps are cached as a side effect of reconciles that happen anyway. The cache is in-memory derived state: on controller restart it rebuilds from etcd as instances reconcile, so values self-heal within seconds.

The metric name, labels, and semantics are unchanged. Only the computation moves from reconcile-time-stored to scrape-time-computed. `EmitConditionMetrics` now feeds the collector's cache (cache/evict) rather than setting a gauge; the GaugeVec calls (`WithLabelValues`, `DeleteLabelValues`, `DeletePartialMatch`) are replaced by `Cache`/`EvictKey`/`EvictInstance`/`EvictGVR`.

## Testing

- `duration_collector_test.go`: cache/collect, zero-timestamp handling, negative-duration clamping, the eviction paths, `Describe`, and a `-race` concurrency test (concurrent `Cache` vs `Collect`).
- A regression test asserts the value grows between two scrapes with no reconcile in between.
- A regression test asserts a status/reason change with diverged initial conditions leaves a single series (no phantom).
- `go build ./...` and `go test ./pkg/metrics/... ./pkg/controller/instance/...` pass.

### DIFF
--- a/pkg/metrics/duration_collector.go
+++ b/pkg/metrics/duration_collector.go
@@ -1,0 +1,154 @@
+// Copyright 2026 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// conditionKey identifies a condition on a single instance. It deliberately
+// excludes status and reason so there is at most one cache entry (one emitted
+// series) per condition type per instance: caching a new status/reason
+// overwrites the previous one instead of leaving a stale series behind.
+type conditionKey struct {
+	gvr           string
+	namespace     string
+	name          string
+	conditionType string
+}
+
+// conditionValue is the mutable part of a cache entry: the condition's current
+// status and reason (emitted as labels) and the LastTransitionTime the
+// duration is measured from.
+type conditionValue struct {
+	status string
+	reason string
+	ts     time.Time
+}
+
+func (k conditionKey) labelValues(v conditionValue) []string {
+	return []string{k.gvr, k.namespace, k.name, k.conditionType, v.status, v.reason}
+}
+
+// durationCollector computes instance_condition_current_status_seconds at
+// scrape time. It caches each condition's LastTransitionTime and returns
+// time-since on Collect, so the value stays accurate while the controller
+// is idle — unlike a GaugeVec set at reconcile time, which would freeze.
+type durationCollector struct {
+	desc *prometheus.Desc
+
+	mu      sync.RWMutex
+	now     func() time.Time // injectable for tests
+	entries map[conditionKey]conditionValue
+}
+
+func newDurationCollector(fqName, help string) *durationCollector {
+	labelNames := []string{
+		labelGVR,
+		labelNamespace,
+		labelName,
+		labelConditionType,
+		labelConditionStatus,
+		labelReason,
+	}
+	return &durationCollector{
+		desc:    prometheus.NewDesc(fqName, help, labelNames, nil),
+		now:     time.Now,
+		entries: make(map[conditionKey]conditionValue),
+	}
+}
+
+func (c *durationCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.desc
+}
+
+func (c *durationCollector) Collect(ch chan<- prometheus.Metric) {
+	// Snapshot under the lock, then emit unlocked: channel sends can block
+	// on a slow scrape consumer and must not stall Cache/Evict.
+	c.mu.RLock()
+	snapshot := make([]struct {
+		key conditionKey
+		val conditionValue
+	}, 0, len(c.entries))
+	for k, v := range c.entries {
+		snapshot = append(snapshot, struct {
+			key conditionKey
+			val conditionValue
+		}{k, v})
+	}
+	c.mu.RUnlock()
+
+	now := c.now()
+	for _, e := range snapshot {
+		seconds := now.Sub(e.val.ts).Seconds()
+		if seconds < 0 {
+			seconds = 0 // clamp clock skew / future timestamps
+		}
+		ch <- prometheus.MustNewConstMetric(
+			c.desc,
+			prometheus.GaugeValue,
+			seconds,
+			e.key.labelValues(e.val)...,
+		)
+	}
+}
+
+// Cache records the current status, reason, and transition time for a
+// condition, replacing any previous entry for the same condition type so only
+// the current series is emitted. A real (non-zero) ts always overwrites. A
+// zero ts anchors the series to now on first sighting and is preserved on
+// later re-caches of the same (status, reason), so a condition without a
+// LastTransitionTime grows steadily rather than resetting each reconcile.
+func (c *durationCollector) Cache(key conditionKey, status, reason string, ts time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if ts.IsZero() {
+		if prev, exists := c.entries[key]; exists && prev.status == status && prev.reason == reason {
+			return
+		}
+		ts = c.now()
+	}
+	c.entries[key] = conditionValue{status: status, reason: reason, ts: ts}
+}
+
+// EvictKey removes the entry for a single condition type on an instance.
+func (c *durationCollector) EvictKey(key conditionKey) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	delete(c.entries, key)
+}
+
+func (c *durationCollector) EvictInstance(gvr, namespace, name string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for k := range c.entries {
+		if k.gvr == gvr && k.namespace == namespace && k.name == name {
+			delete(c.entries, k)
+		}
+	}
+}
+
+func (c *durationCollector) EvictGVR(gvr string) {
+	c.mu.Lock()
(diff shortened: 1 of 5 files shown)


pkg/metrics/duration_collector.go
```diff
@@ -0,0 +1,159 @@
...
+			c.desc,
+			prometheus.GaugeValue,
+			seconds,
+			e.key.labelValues()...,
+		)
+	}
+}
+
+// Cache stores ts under key. A zero ts is treated as now, so the series
+// starts at 0 and grows until a real LastTransitionTime overwrites it.
```
cheeseandcereal: This comment on `Cache` says a zero ts "starts at 0 and grows until a real LastTransitionTime overwrites it." But `EmitConditionMetrics` calls `Cache` on *every* reconcile, so a condition that never gets a `LastTransitionTime` is reset to `now` each reconcile so the series grows, snaps back to 0, grows again (sawtooth). The old code pinned it at 0, which was at least consistent.

I would recommend to skip caching zero-ts entries when the key already exists (cache-if-absent), or don't emit at all for conditions without a transition time.

pkg/metrics/duration_collector.go
```diff
@@ -0,0 +1,159 @@
...
+	return []string{k.gvr, k.namespace, k.name, k.conditionType, k.conditionStatus, k.reason}
+}
+
+// durationCollector computes instance_condition_current_status_seconds at
+// scrape time. It caches each condition's LastTransitionTime and returns
+// time-since on Collect, so the value stays accurate while the controller
+// is idle — unlike a GaugeVec set at reconcile time, which would freeze.
+type durationCollector struct {
+	desc       *prometheus.Desc
+	labelNames []string
```
cheeseandcereal: This is never read `Collect` uses `key.labelValues()` (`pkg/metrics/duration_collector.go:97`). Doesn't seem like this field is needed.

pkg/metrics/duration_collector.go
```diff
@@ -0,0 +1,159 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for k := range c.entries {
+		if k.gvr == gvr {
+			delete(c.entries, k)
+		}
+	}
+}
+
+func (c *durationCollector) reset() {
```
cheeseandcereal: this `reset()` and `size()` are only used in tests. Recommend moving these functions to the _test.go file since it's in the same package anyways.

pkg/metrics/instance.go
```diff
@@ -169,35 +177,21 @@ func EmitConditionMetrics(
...
-			})
+			InstanceConditionCurrentStatusSeconds.EvictConditionType(
+				gvrKey, ns, name, string(oldCond.Type),
+			)
 		}
 	}
 }
 
-// deleteInstanceMetrics removes all gauge metrics for a specific instance.
-// Called when an instance is deleted (not found).
```
cheeseandcereal, line 186: Is there a reason this comment is being removed instead of updated?

pkg/metrics/instance.go
```diff
@@ -169,35 +177,21 @@ func EmitConditionMetrics(
...
-	InstanceConditionCurrentStatusSeconds.DeletePartialMatch(prometheus.Labels{
-		labelGVR:       gvr.String(),
-		labelNamespace: namespace,
-		labelName:      name,
-	})
+	InstanceConditionCurrentStatusSeconds.EvictInstance(gvr.String(), namespace, name)
 }
 
-// DeleteGVRMetrics removes all gauge metrics for all instances of a GVR.
-// Called when an RGD is deregistered (deleted).
```
cheeseandcereal, line 196: Is there a reason this comment is being removed instead of updated?

pkg/metrics/instance.go
```diff
@@ -129,28 +127,38 @@ func EmitConditionMetrics(
...
-					string(oldCond.Type), string(oldCond.Status), oldReason,
-				)
+				InstanceConditionCurrentStatusSeconds.EvictKey(durationKey{
+					gvr:             gvrKey,
+					namespace:       ns,
+					name:            name,
+					conditionType:   string(oldCond.Type),
+					conditionStatus: string(oldCond.Status),
+					reason:          oldReason,
+				})
```
cheeseandcereal: Eviction relies entirely on the reconcile-time diff between `initialConditions` (read from etcd) and `finalConditions` (in-memory). If they ever disagree with what's cached, stale entries are never removed:

- In `pkg/controller/instance/controller.go:223`, `finalConditions` come from `rcx.Instance` even when the status update to the API server **failed**. The cache then holds an entry for a state that never persisted. If the next reconcile transitions to a different reason/status, `EvictKey` (`pkg/metrics/instance.go:140`) only removes the etcd-known old key and the phantom entry survives **forever**, emitting an ever-growing duration for the same instance + condition_type alongside the real series.
- Same failure mode if any other writer touches the conditions between reconciles.

Related to my comments around the eviction functions in the collector, I'd recommend to key the map by some composite key `(gvr, namespace, name, conditionType)` and store `{status, reason, ts}` as the value. That enforces the invariant "at most one series per condition type per instance," makes transitions O(1), and collapses `EvictKey`/`EvictConditionType` into one operation. `Collect` still has all six label values available from key + value.

pkg/metrics/duration_collector.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Cache stores ts under key. A zero ts is treated as now, so the series
+// starts at 0 and grows until a real LastTransitionTime overwrites it.
+func (c *durationCollector) Cache(key durationKey, ts time.Time) {
+	if ts.IsZero() {
+		ts = c.now()
+	}
+	c.mu.Lock()
+	c.entries[key] = ts
+	c.mu.Unlock()
```
cheeseandcereal: Style nit:

In a bunch of places you are using `defer c.mu.Unlock()`, and in other places (like here), you directly run `c.mu.Unlock()` as the last line in the function.

It would be good to pick one of these approaches (I would probably lean towards defer) and do that consistently across this whole file.

pkg/metrics/duration_collector.go
lalitg327: I agree. updated to skip caching zero-ts entries when the key already exists

pkg/metrics/duration_collector.go
lalitg327: yes, it is not being used. removed in the updated version.

pkg/metrics/duration_collector.go
lalitg327: moved to duration_collector_test.go

pkg/metrics/instance.go
lalitg327: It got removed when I was trying to trim some comments. added back with updated comment

pkg/metrics/instance.go
lalitg327: It got removed when I was trying to trim some comments. added back with updated comment

pkg/metrics/instance.go
lalitg327: Nice catch! Interestingly this issue was also present in the existing GaugeVec logic. But I agree that one series per condition type per instance avoids phantom series when the cache diverges from etcd. Reworked the cache to key by (gvr, namespace, name, conditionType) with {status, reason, ts} as the value, as you suggested. Now a transition overwrites rather than relying on the etcd diff

pkg/metrics/duration_collector.go
lalitg327: yeah make sense. updated to use defer c.mu.Unlock() style throughout in duration_collector.go

pkg/metrics/instance.go
```diff
@@ -94,22 +94,20 @@ var (
...
 	)
 )
 
-// emitConditionMetrics computes the diff between initial and final conditions,
-// emits gauge metrics for all current conditions, emits structured logs for
-// changed conditions, and cleans up gauges for disappeared conditions.
+// EmitConditionMetrics syncs the duration cache with an instance's
+// conditions: it caches each final condition's LastTransitionTime, evicts
+// the previous series when a condition's status/reason changed, and evicts
+// conditions that no longer appear. It also logs status transitions.
```
cheeseandcereal: I think the comment is stale

cheeseandcereal:
Can you update the PR description? It has some stale parts with the latest changes now

pkg/metrics/instance.go
lalitg327: updated comment

lalitg327:
> Can you update the PR description? It has some stale parts with the latest changes now

updated stale section of PR description

cheeseandcereal (approved):
lgtm. thanks!

jlbutler:
LGTM, thanks @lalitg327 for the careful work, and @cheeseandcereal for the review.

For release note purposes... worth calling out that (when the gate is enabled) the value now advances continuously instead of freezing (so threshold alerts on it will fire on real elapsed time).

Thanks for closing out the leaking RGDs in test as well. 

/lgtm

jlbutler (approved):

PR #1331, opened 2026-07-21
Fix custom conditions status flow and tighten build-time validation

cheeseandcereal:
Follow-up to #1301 (author-defined custom status conditions), fixing correctness issues in the instance status flow and gaps in build-time validation.

Note: Most of the large line count in the diff is general refactoring (adjusting comments, adding tests, and moving around some helper functions). The core functionality changes are more scoped and described below.

## Main Fix

Overriding a built-in condition type such as `ResourcesReady` caused an infinite reconcile loop. `lastTransitionTime` preservation compared against the in-memory value the condition markers had just written instead of what was persisted, so whenever the two disagreed every reconcile stamped a fresh `lastTransitionTime`, bumped `resourceVersion`, and re-enqueued itself.

## Status flow fixes (`pkg/controller/instance`, `pkg/runtime`)

- Snapshot `.status` as fetched from the API server (`ReconcileContext.WireStatus`), before the condition markers mutate the object in memory. `lastTransitionTime` preservation and the skip-write comparison now run against what is actually persisted.
- Finalizer/label helpers return nil when no write was issued, so the already-managed fast path (the steady state) doesn't re-capture the snapshot from the marker-mutated object. That pollution defeated the skip-write guard and let incomplete evaluations merge built-ins onto the wire as `Unknown`.
- Skip-write guard now works: it previously compared against marker-written built-ins (never equal) and a typed `InstanceState` vs the wire's plain string (never equal, could panic `DeepCopyJSON`); `state` is now written as a plain string.
- Inject only kro's four built-ins into author CEL; the full marker list leaked the previous reconcile's author conditions into computed `runtime.condition(schema, _)` lookups.
- When an expression produces no output (data pending, eval error, duplicate drop), keep its previously persisted condition instead of dropping it from the wire. A fully successful evaluation still replaces the wire, so removed types clean up.
- Write only built-ins when no author conditions are declared, cleaning up leftovers from a removed `conditions:` block.
- On graph-resolve failure with author conditions, leave persisted conditions untouched and only set `state: Error` (no graph to evaluate expressions; the old filtering dropped built-in-typed overrides).
- Record the state-transition metric from the state actually written; reject empty condition types at evaluation time.

## Build-time validation (`pkg/graph`, `pkg/cel/library`)

- Condition expressions must return a Condition or `list(Condition)`; anything else previously built an Active RGD whose instances could only reach `state: Error`.
- Reject the `runtime` variable outside `status.conditions` (templates, `forEach`, etc.), where it previously built fine and failed every reconcile.
- Scope the self-reference rule to `runtime.condition(schema, ...)`: child-resource lookups may share author type names; unknown literal `schema` lookups are rejected (typo guard).
- Reject empty literal types and literal types declared by more than one `conditions:` entry (guaranteed wire duplicates); repeats within one entry (ternary branches) stay legal.
- Move the validator from `pkg/cel/library` into `pkg/graph` (drops the CEL library's `api/v1alpha1` dep), parse each expression once, and use cel-go's `PreOrderVisit` instead of a hand-rolled walker.

## Cleanups

- Fold the `hasAuthorConditions` constructor flag into `ReconcileConfig`.
- Deduplicate condition decoding helpers; key built-in checks off `v1alpha1.KROBuiltinConditionTypes`.
- Trim oversized comments, remove unreachable macro code and unused synthesized condition fields, standard `metav1` alias.
- Delete leaked cluster-scoped RGDs in the integration suite (they made parallel runs time out).

## Behavior changes

Note: All of the behavior changes are only related to custom status conditions, and don't impact the rest of kro/existing RGD usage/etc.

- RGDs that built but could never work now fail the build (`Inactive`): non-Condition output, `runtime` outside `status.conditions`, duplicate literal types.
- Computed `runtime.condition(schema, _)` lookups of non-built-in types return an empty condition instead of one-reconcile-stale author data.
- `runtime.condition(resource, 'X')` with `X` colliding with an author type: previously rejected, now allowed.
- Conditions persist through data-pending/failing evaluations instead of vanishing; removing a `conditions:` block returns the wire to kro's built-ins.

## Testing

- Unit (`go test -race ./pkg/...`): regressions for the `lastTransitionTime` churn, incomplete-evaluation merging, leftover cleanup, the fast-path snapshot pollution (via full `Reconcile`), and the build-time rejections.
- Integration (envtest): specs for override `lastTransitionTime` stability, data-pending preservation (built-ins stay off the wire), `conditions:` block removal, and rejection messages on the RGD `Ready` condition. Full core suite 151/151 in ~90s (previously timed out on the leaked RGDs).
- Manual (kind): `ResourcesReady`-override RGD holds a stable `resourceVersion`/`lastTransitionTime` over 90s (no hot loop); `runtime.condition(schema, 'ResourcesReady')` mirrors kro's internal value; generation bumps advance `observedGeneration` without touching `lastTransitionTime`; deletion cleans up children; build rejections surface readable RGD errors.


pkg/graph/conditions.go
```diff
@@ -0,0 +1,201 @@
...
+	// second pass can detect self-references, rejecting types declared by
+	// more than one entry along the way.
+	declared := map[string]struct{}{}
+	for _, root := range asts {
+		if root == nil {
+			continue
+		}
+		for typ := range literalConditionTypes(root) {
+			if _, dup := declared[typ]; dup {
+				return fmt.Errorf("condition type %q is declared by more than one conditions entry", typ)
```
jlbutler, line 63: I think this check ends up being more conservative than the comment at line 36 implies (I don't think this is branch-aware, so it'll over-reject right?). It's ok to be conservative, but maybe soften that comment for posterity?

pkg/controller/instance/status.go
```diff
@@ -232,17 +222,15 @@ func (c *Controller) updateStatus(rcx *ReconcileContext) error {
...
+	// Mirror the computed status onto the instance so the deferred emitters
+	// read what's on the wire, not the marker's built-ins.
 	rcx.Instance.Object["status"] = status
 
-	if equality.Semantic.DeepEqual(oldStatus, status) {
+	// Skip the API server write if the persisted status already matches.
+	// This prevents an infinite reconcile loop where every unconditional
+	// UpdateStatus bumps resourceVersion, which triggers a watch event,
+	// which re-enqueues the instance, which reconciles again.
+	if equality.Semantic.DeepEqual(rcx.WireStatus, status) {
```
jlbutler: Is there a gap here for number types? I'm not super deep on this, but tracing both sides of the comparison... the wire side comes through apimachinery and is typed by value (e.g. a whole value like 1.0 gets int64), while the built side from CEL is structural (a double with a whole value lands as float64). If I have that right, then there's a window that fails the DeepEqual when we don't want it to. Not a loop issue (api-server no-ops the identical bytes and status writes don't bump generation, so it won't re-trigger), but is a redundant UpdateStatus every reconcile for this corner case. Maybe worth a test that seeds the wire status as int64 (the fake dynamic client doesn't round-trip numbers, so a naive test would pass and miss it).

jlbutler (approved):
A couple of nits, nothing blocking.

pkg/controller/instance/status.go
cheeseandcereal: Yup, good catch. Fixed in update: https://github.com/kubernetes-sigs/kro/compare/76ca1568375322eca5e3a3a7cb0314245becf0a8..c883a95c2f89dc5cd05e65e0e3cf798fb73d0f28

pkg/graph/conditions.go
cheeseandcereal: Good call, updated

pkg/graph/conditions.go
michaelhtm, line 1: nit: this should land in validation.go

pkg/graph/conditions_test.go
```diff
@@ -0,0 +1,157 @@
...
+			name: "self-reference rejected",
+			exprs: []string{
+				`runtime.newCondition({type: 'PrimaryReady', status: 'True', reason: '', message: ''})`,
+				`runtime.newCondition({type: 'Ready',
+					status: runtime.condition(schema, 'PrimaryReady').status,
+					reason: '', message: ''})`,
+			},
+			wantErrs: []string{"custom conditions cannot reference each other", `"PrimaryReady"`},
+		},
+		{
```
michaelhtm, line 104: not sure i understand how these conditions are referencing each other

michaelhtm (commented):
👍

pkg/graph/conditions_test.go
cheeseandcereal: "Ready" becomes a "custom condition" once it's defined here. It's referencing "PrimaryReady". Custom status conditions can't reference each other because they are not ordered like the regular spec section of the graph. This is an intentional design trade-off due to the limitations of implementing conditions, and treating them as termninal things evaluated at the end of the graph/etc.

Conditions all evaluate at the same time, so reading another author condition could only ever return a stale value from the previous reconcile and evaluation order can't be fixed by sorting them, because condition types can be computed at runtime. So this rejects them, and the condition should recompute the value from the same inputs instead (i.e. copy/pasted cel expression or variable, etc. Conditions themselves shouldn't be used as variables).

pkg/graph/conditions.go
cheeseandcereal: The functions here are really specific to validating conditions (except for the few generic AST helpers isRuntimeCall, literalString, mapLiteralEntryStringValue, which aren't really validations but generic helpers), I think it might be more confusing to mix them together, since this file separation makes it much more clear they are specific to the condition part of the processing as opposed to the general graph.

Not super opinionated, so if you think it should move I can do that, but I don't think it necessarily needs to be in validation.go, especially given it's already a ~400 line file.

PR #1332, opened 2026-07-26
fix: use background propagation when deleting managed resources

### LINKED ISSUE #1330: Deleting an instance orphans the Pods of a managed Job
### Description

First, I would like to determine whether this behavior constitutes a "bug" or is intentional. If it is indeed a bug, I will submit a pull request.

**Observed Behavior**:

When you define an RGD that creates a Job and create its instance, not only is the Job created but also a Pod is created with this Job as its parent.
If an RGD instance is deleted under such circumstances, the Pod will remain intact and continue running—it must be explicitly removed using `kubectl delete pod`.

**Expected Behavior**:

Deleting the instance should delete the Job and its Pods, consistent with kubectl delete job semantics.

**Reproduction Steps**

### Prerequisites

I am running a Kubernetes cluster on my local machine using Kind@0.32.0.

### Install kro CRDs
```sh
$ kubectl apply -f ./helm/crds
```

### Start the Controller
```sh
$ go run ./cmd/controller
```

### Define a test RGD called SleepJob
Create a `sleep-rgd.yaml` file and apply it.

```sh
$ kubectl apply -f sleep-rgd.yaml
```

```yaml
# sleep-rgd.yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: sleepjob.kro.run
spec:
  schema:
    apiVersion: v1alpha1
    kind: SleepJob
    spec:
      durationSeconds: integer | default=3600
  resources:
  - id: job
    template:
      apiVersion: batch/v1
      kind: Job
      metadata:
        name: ${schema.metadata.name}-job
      spec:
        template:
          spec:
            restartPolicy: Never
            containers:
            - name: sleeper
              image: busybox
              command: ["sleep", "${string(schema.spec.durationSeconds)}"]
```

### Creating an instance of SleepJob
```sh
$ echo '{"apiVersion":"kro.run/v1alpha1","kind":"SleepJob","metadata":{"name":"orphaned-pod-repro","namespace":"default"},"spec":{}}' | kubectl apply -f -
```

In addition to creating the SleepJob instance, a Job and Pod are also created.

```sh
$ kubectl get sleepjobs
NAME                 STATE    READY   AGE
orphaned-pod-repro   ACTIVE   True    18s

$ kubectl get job,pod
NAME                               STATUS    COMPLETIONS   DURATION   AGE
job.batch/orphaned-pod-repro-job   Running   0/1           37s        37s

NAME                               READY   STATUS    RESTARTS   AGE
pod/orphaned-pod-repro-job-2d9w9   1/1     Running   0          37s
```

### Deleting a SleepJob instance

```sh
$ kubectl delete sleepjob orphaned-pod-repro
```

This will remove the SleepJob instance and the Job, but the Pod will remain intact.

```sh
$ kubectl get sleepjobs
No resources found in default namespace.

$ kubectl get job,pod
NAME                               READY   STATUS    RESTARTS   AGE
pod/orphaned-pod-repro-job-2d9w9   1/1     Running   0          5m
```

When deleting a SleepJob instance, the Controller outputs the following message. As indicated here, the `propagationPolicy` setting is responsible for this behavior.
```
Warning: child pods are preserved by default when jobs are deleted; set propagationPolicy=Background to remove them or set propagationPolicy=Orphan to suppress this warning
```

**Versions**:
- kro version: [commit 1762f16](https://github.com/kubernetes-sigs/kro/tree/1762f16d30670376a5b1a1892152f8fed8c7df18) - This was just the `HEAD` commit I was working on at the time—there's no particular significance to this specific commit.
- Kubernetes Version (`kubectl version`): v1.36.1

Others (please specify), Instance (Create, Update, Deletion)

numb86:
Fixes #1330 .

I confirmed that the Pods are deleted when running the "Reproduction Steps" in #1330 with this fix applied.

The cause is that `propagationPolicy` has not been specified when deleting resources.
When `propagationPolicy` is not specified, the `defaultGCPolicy` of the resource is used, and for Job (and ReplicationController) it is `OrphanDependents`. In this case, the child resources (Pods, in the case of Job) become orphans and remain.
Therefore, setting `propagationPolicy` to `Background` fixes this problem.

The `defaultGCPolicy` of resources other than Job and ReplicationController is equivalent to `Background`, so setting `propagationPolicy` to `Background` has no impact on them.

- Implementation showing that `defaultGCPolicy` is used when `propagationPolicy` is not specified, and that orphaning happens only when `defaultGCPolicy` is `OrphanDependents` (`shouldOrphanDependents`)
    - https://github.com/kubernetes/kubernetes/blob/v1.36.0/staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go#L891-L932
- Job's `defaultGCPolicy` is `OrphanDependents` for batch/v1 (`DefaultGarbageCollectionPolicy`)
    - https://github.com/kubernetes/kubernetes/blob/v1.36.0/pkg/registry/batch/job/strategy.go#L60-L74
- ReplicationController's `defaultGCPolicy` is `OrphanDependents` for core/v1 (`DefaultGarbageCollectionPolicy`)
    - https://github.com/kubernetes/kubernetes/blob/v1.36.0/pkg/registry/core/replicationcontroller/strategy.go#L59-L73

### DIFF
--- a/pkg/controller/instance/deletion.go
+++ b/pkg/controller/instance/deletion.go
@@ -181,12 +181,16 @@ func (c *Controller) deleteTarget(
 		return nil
 	}
 
+	// Background must be explicit: an unspecified policy falls back to the
+	// type's default GC policy, which for batch/v1 Jobs orphans their Pods.
+	propagationPolicy := metav1.DeletePropagationBackground
+
 	// Track whether any delete request was accepted. a successful Delete does NOT
 	// mean the object is gone yet, just that deletion is in progress.
 	anyDeleted := false
 	for _, target := range targets {
 		rc := resourceClientFor(rcx, node.Spec.Meta, target.GetNamespace())
-		err := rc.Delete(rcx.Ctx, target.GetName(), metav1.DeleteOptions{})
+		err := rc.Delete(rcx.Ctx, target.GetName(), metav1.DeleteOptions{PropagationPolicy: &propagationPolicy})
 		if apierrors.IsNotFound(err) {
 			// Already gone: leave anyDeleted as is and keep checking others.
 			continue
--- a/pkg/controller/instance/applyset/applyset.go
+++ b/pkg/controller/instance/applyset/applyset.go
@@ -467,8 +467,12 @@ func (a *ApplySet) ListOrphans(ctx context.Context, opts PruneOptions) ([]Orphan
 // to avoid deleting a resource that was recreated since listing.
 func (a *ApplySet) DeleteOrphan(ctx context.Context, candidate OrphanCandidate) (DeleteOrphanResult, error) {
 	uid := candidate.Object.GetUID()
+	// An unspecified propagation policy would fall back to the type's default
+	// GC policy, which orphans dependents for some types (e.g. batch/v1 Job).
+	propagationPolicy := metav1.DeletePropagationBackground
 	deleteOpts := metav1.DeleteOptions{
-		Preconditions: &metav1.Preconditions{UID: &uid},
+		Preconditions:     &metav1.Preconditions{UID: &uid},
+		PropagationPolicy: &propagationPolicy,
 	}
 
 	var err error
--- a/pkg/controller/instance/applyset/applyset_test.go
+++ b/pkg/controller/instance/applyset/applyset_test.go
@@ -779,6 +779,57 @@ func TestPrune_UIDPrecondition(t *testing.T) {
 	})
 }
 
+func TestDeleteOrphan_BackgroundPropagation(t *testing.T) {
+	ctx := t.Context()
+	mapper := newTestRESTMapper()
+	parent := newTestParent(schema.GroupVersionKind{
+		Group: "kro.run", Version: "v1alpha1", Kind: "TestKind",
+	})
+	applySetID := ID(parent)
+
+	orphan := newConfigMap("orphan-cm", "default")
+	orphan.SetLabels(map[string]string{
+		ApplysetPartOfLabel: applySetID,
+	})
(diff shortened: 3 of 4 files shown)


pkg/controller/instance/deletion.go
```diff
@@ -181,12 +181,16 @@ func (c *Controller) deleteTarget(
...
+	// type's default GC policy, which for batch/v1 Jobs orphans their Pods.
+	propagationPolicy := metav1.DeletePropagationBackground
+
 	// Track whether any delete request was accepted. a successful Delete does NOT
 	// mean the object is gone yet, just that deletion is in progress.
 	anyDeleted := false
 	for _, target := range targets {
 		rc := resourceClientFor(rcx, node.Spec.Meta, target.GetNamespace())
-		err := rc.Delete(rcx.Ctx, target.GetName(), metav1.DeleteOptions{})
+		err := rc.Delete(rcx.Ctx, target.GetName(), metav1.DeleteOptions{PropagationPolicy: &propagationPolicy})
```
Skarlso, line 193: This isn't enough here. And this is overriding everything that is potentially created in an RGD for a CRD. 🤔 

Setting a finalizers overwrites the policy:
```go
	for _, f := range finalizers {
		switch f {
		case metav1.FinalizerOrphanDependents:
			return true
		case metav1.FinalizerDeleteDependents:
			return false
		}
	}
```
in the apiserver.

If we DO want to do this, shouldn't this actually be Forground policy? Wouldn't that be something that Kro community would desire in this case since anything that gets orphaned is actually something that Kro will lose visibility on? 🤔

PR #1337, opened 2026-08-03
Allow custom status conditions to reference each other

lalitg327:
## Summary

Lets one author-defined condition read another. `runtime.condition(schema, 'X')`
where `X` is declared by a `runtime.newCondition({type: 'X', ...})` in the same
RGD is now legal, and kro evaluates entries in dependency order so the reader
observes the value published this reconcile.

## Why

The custom-conditions feature registered `runtime.condition(obj, type)` for
reading conditions, but a lookup of an author's own type was rejected at build
time:

```
runtime.condition(schema, "PrimaryReady"): custom conditions cannot reference
each other
```

Authors could read kro's four built-ins but not compose their own conditions.
That forces duplication: an aggregate like `Ready` has to restate every
sub-expression it depends on, and the two copies drift.

```yaml
# before: the primary check is written twice
status:
  conditions:
    - ${runtime.newCondition({
        type: 'PrimaryReady',
        status: primary.status.phase == 'Running' ? 'True' : 'False',
        reason: 'PrimaryPhase', message: ''})}
    - ${runtime.newCondition({
        type: 'Ready',
        status: primary.status.phase == 'Running'
          && replicas.status.readyReplicas > 0 ? 'True' : 'False',
        reason: 'Aggregate', message: ''})}
```

```yaml
# after: Ready composes the condition it depends on
status:
  conditions:
    - ${runtime.newCondition({
        type: 'PrimaryReady',
        status: primary.status.phase == 'Running' ? 'True' : 'False',
        reason: 'PrimaryPhase', message: ''})}
    - ${runtime.newCondition({
        type: 'Ready',
        status: runtime.condition(schema, 'PrimaryReady').status == 'True'
          && replicas.status.readyReplicas > 0 ? 'True' : 'False',
        reason: 'Aggregate', message: ''})}
```

## How it works

Declaration order does not have to match dependency order. An entry that reads
`PrimaryReady` may be declared before the entry that declares it; the build
computes an evaluation rank per entry from a topological sort over the reference
edges, and the runtime evaluates in rank order.

**Emission stays in declaration order.** Evaluation order is derived at runtime
construction and never reaches `status.conditions[]`, so a future change to how
ranks are computed cannot rewrite status on every instance in the cluster.

The RGD compiles once at admission, before any instance exists, so a `type:`
expression is only statically knowable when fully literal. A computed name like
`'Shard-' + s.metadata.name` is known only by its shape and is matched as a
pattern, which is why a literal read of `Shard-web` resolves against a fan-out
entry. A pattern matches more names than the expression will produce; that is
the safe direction, since a spurious edge costs a false cycle at build time
while a missing edge would cause silent staleness at runtime.

### Rejected at build time

A `type:` expression with more than 32 possible values, rather than widening to
match every name (which would disable typo detection for the whole RGD):

```
conditions[0]: condition type expression has more than 32 possible values;
split it across separate conditions entries
```

Self-references, with the type named rather than surfacing as an opaque cycle:

```
runtime.condition(schema, "Loop"): a conditions entry cannot depend on a
condition type it may itself declare (matches Loop)
```

Reference cycles, attributed to both entries. Fan-out entries are rendered by
pattern, since two of them can produce the same glob and only the index tells
them apart:

```
conditions entries form a reference cycle: entry 0 (A) -> entry 1 (B) -> entry 0 (A)
conditions entries form a reference cycle: entry 0 (Alpha-*) -> entry 1 (Beta-*) -> entry 0 (Alpha-*)
```

### Unavailable dependencies

An entry whose dependency did not resolve is skipped rather than evaluated
against an empty read, which would publish a confident wrong value. Two cases
propagate to dependents, transitively:

- a producer entry that failed to evaluate, so the type was never published
- a type withdrawn because two entries published it with different values

The result is marked incomplete, so the caller preserves the last persisted
value for the affected types instead of blanking them.

## Test plan

- `go test ./pkg/graph/... ./pkg/runtime/... ./pkg/controller/...` passes.
- Build-time cases: a cross-reference is accepted whichever order the two
  entries are declared in; a self-reference is rejected; cycles spanning two,
  three, and four entries are rejected; a cycle through a fan-out or ternary
  entry names the pattern rather than an opaque index; a literal read resolves
  against a computed type; the alternation cap is enforced.
- Runtime cases: an entry declared before the one it reads still observes this
  reconcile's value; emission order is unchanged across three different
  evaluation orders; a failed producer propagates to its dependents
  transitively; a withdrawn type drops the entries depending on it; a pending
  dependency preserves the last persisted value rather than blanking it; an
  override of a built-in type does not shadow it for lookups.

### DIFF
--- a/pkg/controller/instance/status_test.go
+++ b/pkg/controller/instance/status_test.go
@@ -172,12 +172,12 @@ func TestUpdateStatusMirrorsPersistedConditionsOntoInstance(t *testing.T) {
 		Template: &unstructured.Unstructured{
 			Object: map[string]interface{}{"status": map[string]interface{}{}},
 		},
-		Conditions: []*krocel.Expression{
+		Conditions: conditionEntries(
 			mustCompileControllerExpr(t,
 				`runtime.newCondition({type: 'PrimaryReady', status: 'True', reason: 'Healthy', message: 'all good'})`,
 				library.RuntimeVarName,
 			),
-		},
+		),
 	}
 
 	controller, rcx, raw := newControllerAndContext(t, instance, newTestGraphWithInstance(instanceNode))
@@ -314,6 +314,14 @@ func TestStampAuthorConditionsEmptyReasonAndMessage(t *testing.T) {
 	assert.Nil(t, stamped[0].Message, "empty message should serialize as nil pointer")
 }
 
+func conditionEntries(exprs ...*krocel.Expression) []graph.ConditionEntry {
+	entries := make([]graph.ConditionEntry, len(exprs))
+	for i, e := range exprs {
+		entries[i] = graph.ConditionEntry{Expr: e}
+	}
+	return entries
+}
+
 // authorConditionsInstanceNode returns an instance node declaring the given
 // author condition expressions.
 func authorConditionsInstanceNode(t *testing.T, exprs ...*krocel.Expression) *graph.Node {
@@ -328,7 +336,7 @@ func authorConditionsInstanceNode(t *testing.T, exprs ...*krocel.Expression) *gr
 		Template: &unstructured.Unstructured{
 			Object: map[string]interface{}{"status": map[string]interface{}{}},
 		},
-		Conditions: exprs,
+		Conditions: conditionEntries(exprs...),
 	}
 }
 
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -724,7 +724,7 @@ func buildInstanceNode(
 	namespaced bool,
 	statusVariables []variable.FieldDescriptor,
 	statusTemplate map[string]interface{},
-	conditions []*krocel.Expression,
+	conditionEntries []ConditionEntry,
 	inspector *ast.Inspector,
 ) (*Node, error) {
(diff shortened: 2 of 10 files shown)


jakobmoellerdev:
I think its a great idea but ill have to check the code for the conditions in more detail

fabianburth:
Hey, great proposal. I'm still trying to get a thorough understanding. As initial feedback, I think the biggest conceptual issue is the regex based dependency computation. 
I'm wondering whether we could model conditions analogously to resources. Specifically, I'm thinking about introducing a stable `id` per condition and modeling the fan-out use case with collections.  So, we'd use `forEach` and make similar assumption for `type` uniqueness as resource collections make for resource `name` uniqueness.

What do you think?

cheeseandcereal:
I'm wondering how necessary this change would be once we have something like more proper support for variables? This is a lot of extra machinery to maintain just to enable "no need to copy/paste an expression" [for conditions] which is the more general problem that [variables](https://github.com/kubernetes-sigs/kro/pull/1035) already poses to solve.

Do we think it's worth it to implement this to enable this use case just for conditions as opposed to the more general variables solution which would also solve this use-case, but for everything? (Not just conditions).

PR #1340, opened 2026-08-04
fix(graph): key extended CEL env cache by schema, not just var name

### LINKED ISSUE #1339: `readyWhen` on multiple `forEach` resources: all collections share one cached `each` type, causing spurious `undefined field` errors
### Description

**Observed Behavior**:

When an RGD has two or more `forEach` (collection) resources that each use `each` in `readyWhen`, all of them are type-checked against the *same* schema — whichever collection happens to be processed first. Any collection whose `readyWhen` references a field absent from that first schema fails with `undefined field '<name>'`, even though the field exists in its own template kind.

The RGD goes `Inactive` with `GraphAccepted=False`, which stops reconciliation for every instance of the kind.

Because resources are iterated in Go map order, *which* collection wins is nondeterministic. On our RGD this made the same RGD spec, CRD schemas, and kro image compile on some clusters and not others, and flip between valid and invalid on a single cluster with no input change and no controller restart (errors recur on the reconcile backoff, each retry re-rolling the order).

Adding optional-field markers (`each.status.?ready.orValue(...)`, `has(each.status.ready)`) does not suppress it — the type genuinely lacks the field, so the compiler rejects it before optionality applies.

**Expected Behavior**:

Each collection's `each` variable is typed from its own template kind's schema. The example below should compile.

**Reproduction Steps**:

Two `forEach` resources, built-in kinds only. `Pod.status` has `phase` but no `ready`; `Job.status` has `ready` but no `phase` — so neither is a superset and the build fails every time, with the error naming whichever resource lost the race.

`resourcegraphdefinition.yaml`:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: eachcollision
spec:
  schema:
    apiVersion: v1alpha1
    kind: EachCollision
    spec:
      names: "[]string | required=true minItems=1"
  resources:
    - id: pods
      forEach:
        - entry: ${schema.spec.names}
      readyWhen:
        - ${each.status.phase == "Running"}
      template:
        apiVersion: v1
        kind: Pod
        metadata:
          name: ${entry + "-pod"}
          namespace: ${schema.metadata.namespace}
        spec:
          containers:
            - name: main
              image: busybox

    - id: jobs
      forEach:
        - entry: ${schema.spec.names}
      readyWhen:
        - ${each.status.ready == 1}
      template:
        apiVersion: batch/v1
        kind: Job
        metadata:
          name: ${entry + "-job"}
          namespace: ${schema.metadata.namespace}
        spec:
          template:
            spec:
              restartPolicy: Never
              containers:
                - name: main
                  image: busybox
```

`instance.yaml`:

```yaml
apiVersion: kro.run/v1alpha1
kind: EachCollision
metadata:
  name: eachcollision-sample
  namespace: default
spec:
  names:
    - alpha
    - beta
```

```
kubectl apply -f resourcegraphdefinition.yaml
kubectl get rgd eachcollision
```

The RGD reports `Inactive` / `GraphAccepted=False`; the instance is not required to reproduce, since this fails at RGD compilation.

Narrowing it down: either resource **alone** compiles fine. A single collection plus a non-collection resource referencing a kind by id also compiles fine. Only two-or-more collections with `each` in `readyWhen` breaks.

**Versions**:
- kro version: v0.9.1 (also present in v0.9.2 / v0.9.3 — that code path is unchanged)
- Kubernetes Version: v1.32.13-eks

**Error Logs**:
```
failed to validate resource "jobs": failed to type-check readyWhen expression "each.status.ready == 1" in resource "jobs": ERROR: <input>:1:12: undefined field 'ready'
 | each.status.ready == 1
 | ...........^
```

Consecutive reconciles of the *same unchanged RGD*, one second apart, naming different resources:

```
11:46:59Z  ... resource "jobs": ERROR: <input>:1:12: undefined field 'ready'
11:47:00Z  ... resource "pods": ERROR: <input>:1:12: undefined field 'phase'
11:47:02Z  ... resource "jobs": ERROR: <input>:1:12: undefined field 'ready'
```

**Proposed fix**:

The per-build cache of extended CEL environments is keyed on the parent environment and the variable name, but not on the schema the typed variable is derived from. Since every collection uses the variable name `each` and the same parent env, they all collide on one entry: the first collection populates it and the rest get a cache hit, never consulting their own schema. The in-tree comment notes the cache is for collections that "share the same schema," but nothing enforces that precondition.

Adding the schema to that cache key fixes it. I verified this locally: the repro above goes from failing every build to passing every build, and the existing `pkg/graph` and `pkg/cel` suites pass unchanged. Happy to open a PR, along with a regression test for two collections with differing template kinds — I couldn't find existing coverage for that combination.

ResourceGraphDefinition (Create, Update, Deletion)

pujitha24:
Motivation:
When a ResourceGraphDefinition has two or more forEach (collection)
resources that each reference `each` in readyWhen, the RGD randomly
failed to compile with a spurious "undefined field" error, flipping
between valid and invalid across reconciles with no input change.

Approach:
buildContext.extendWithTypedVar caches the CEL environment it builds
for a collection's `each` variable, keyed on (parent env, var name).
Every collection uses the same variable name "each" and the same
parent env, so all collections in one build collided on a single
cache entry: whichever collection populated it first "won", and every
other collection's readyWhen was then type-checked against the wrong
kind's status schema. Because Go map iteration order is randomized,
which collection won was nondeterministic across reconciles.

The fix adds the schema pointer to the cache key (extendedEnvKey),
matching the existing declTypes map in the same file which is already
keyed by schema pointer on the same rationale: buildContext is
documented as scoped to a single NewResourceGraphDefinition() call,
so schema pointers stay live and stable for the whole build.

Validation:
Added a regression test in pkg/graph/builder_test.go
(TestGraphBuilder_CELTypeChecking) with two forEach collections
templating VPC and Subnet, whose status schemas each have a field the
other lacks (vpcID vs subnetID) - mirroring the issue's Pod/Job repro.
Confirmed the test fails with the same "undefined field" error when
run against the pre-fix code (via git stash) and passes with the fix.

Ran the repo's documented unit test command and it passed:
  go test -race ./pkg/... -coverprofile unit-cover.out

Report: https://github.com/kubernetes-sigs/kro/issues/1339

---

Fixes #1339

```release-note
fix(graph): key extended CEL env cache by schema, not just var name
```

### DIFF
--- a/pkg/graph/build_context.go
+++ b/pkg/graph/build_context.go
@@ -49,8 +49,8 @@ type buildContext struct {
 	// inference, then compile reuses the cached AST.
 	checkedASTs map[checkedASTKey]*cel.Ast
 
-	// (parent env, varName) → extended env. Deduplicates readyWhen on
-	// collection nodes that share the same schema.
+	// (parent env, varName, schema pointer) → extended env. Deduplicates
+	// readyWhen on collection nodes that share the same schema.
 	extendedEnvs map[extendedEnvKey]*cel.Env
 }
 
@@ -62,6 +62,7 @@ type checkedASTKey struct {
 type extendedEnvKey struct {
 	parent  *cel.Env
 	varName string
+	schema  *spec.Schema
 }
 
 // schemaDeclType returns a DeclType for the given schema, memoized by pointer.
@@ -124,7 +125,7 @@ func (bc *buildContext) compile(env *cel.Env, expr *krocel.Expression) (*cel.Ast
 // single typed variable declaration derived from the given schema. Uses
 // schemaDeclType for memoized schema→DeclType conversion.
 func (bc *buildContext) extendWithTypedVar(parent *cel.Env, varName string, s *spec.Schema) (*cel.Env, error) {
-	key := extendedEnvKey{parent: parent, varName: varName}
+	key := extendedEnvKey{parent: parent, varName: varName, schema: s}
 	if env, ok := bc.extendedEnvs[key]; ok {
 		return env, nil
 	}
--- a/pkg/graph/builder_test.go
+++ b/pkg/graph/builder_test.go
@@ -2206,6 +2206,52 @@ func TestGraphBuilder_CELTypeChecking(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			// Regression test: readyWhen on multiple forEach collections must type-check
+			// "each" against each collection's own template kind, not whichever
+			// collection's schema happened to populate the cache first. VPC.status has
+			// vpcID but no subnetID, Subnet.status has subnetID but no vpcID, so neither
+			// is a superset of the other.
+			name: "readyWhen on multiple forEach collections with different template kinds",
+			resourceGraphDefinitionOpts: []generator.ResourceGraphDefinitionOption{
+				generator.WithSchema(
+					"Test", "v1alpha1",
+					map[string]interface{}{
+						"names": "[]string",
+					},
+					nil,
+				),
+				generator.WithResourceCollection("vpcs", map[string]interface{}{
+					"apiVersion": "ec2.services.k8s.aws/v1alpha1",
+					"kind":       "VPC",
+					"metadata": map[string]interface{}{
+						"name": "${name}-vpc",
+					},
+					"spec": map[string]interface{}{
+						"cidrBlocks": []interface{}{"10.0.0.0/16"},
+					},
+				},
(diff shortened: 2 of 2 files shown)


jakobmoellerdev (commented):
/ok-to-test
/lgtm

jakobmoellerdev (approved):

PR #1345, opened 2026-08-07
Use stdlib maps and slices instead of x/exp

v47:
## Summary

`golang.org/x/exp` was a direct dependency for exactly two calls, both
of which the standard library has covered since Go 1.21/1.23:

- `pkg/apis/condition_types.go` — `x/exp/slices` for one `slices.Contains`
- `pkg/graph/builder.go` — `x/exp/maps` for one `maps.Keys`

After the swap, `go mod tidy` moves `golang.org/x/exp` out of the direct require block. It remains an indirect dependency via cel-go, antlr, and the k8s libraries, so no `go.sum`.

Also collapses the hand-rolled loop in `ConditionTypes.DependsOn` to `slices.Contains`, since the import was already being touched.

The `maps.Keys` swap is not purely mechanical. `x/exp`'s `maps.Keys` returned a slice with `cap == len`, so `append` onto it always reallocated. Stdlib `slices.Collect(maps.Keys(m))` grows in powers of two and returns spare capacity, which makes the existing
```go
allIdentifiers := append(nodeNames, SchemaVarName, EachVarName, library.RuntimeVarName)
```
an aliasing hazard — it would write into nodeNames' backing array. This is a similar bug fixed in this file, so this follows that precedent and uses `slices.Concat`.

`slices.Sorted` over `slices.Collect` is free at the same line count and makes the identifier list deterministic.

### DIFF
--- a/pkg/graph/builder.go
+++ b/pkg/graph/builder.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/google/cel-go/cel"
-	"golang.org/x/exp/maps"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


pkg/graph/builder.go
```diff
@@ -250,8 +250,8 @@ func (b *Builder) NewResourceGraphDefinition(originalCR *v1alpha1.ResourceGraphD
 	// Create a single expression inspector for all AST inspection operations.
 	// This uses a lightweight env that only declares identifier names (no full schemas) -
 	// sufficient for parsing and finding references, but NOT for type-checking or compilation.
-	nodeNames := maps.Keys(nodes)
-	allIdentifiers := append(nodeNames, SchemaVarName, EachVarName, library.RuntimeVarName)
+	nodeNames := slices.Sorted(maps.Keys(nodes))
+	allIdentifiers := slices.Concat(nodeNames, []string{SchemaVarName, EachVarName, library.RuntimeVarName})
```
Skarlso: Technically, this is a logical change, since now nodeNames is always sorted which is another performed operation on the CPU, for no apparent reason. This part of the codebase did not rely on the nodeNames being stable. Further, you sort the nodeNames here, but then don't sort whatever is SchemaVarName, EachVarName and RuntimeVarName. So why sort it in the first place?

I would rather want to use `slices.Collect` instead.

That said... Alternatively, I would argo, that here, using Concat and maps.Keys and iters is overkill. :) 

I think it's even nicer to just collect the keys to oldefashioned way doing:
```go
nodeNames := make([]string, 0, len(nodes)
for name := range nodes {
    nodeNames = append(nodeNames, name)
}
allIdentifiers := slices.Concat(nodeNames, []string{SchemaVarName, EachVarName, library.RuntimeVarName})
...
```
and done. 

maps.Keys is nice if you hand over the Seq to something expecting the Seq. Otherwise, it's an overhead.

Skarlso: Nice, thanks. 🙇‍♀️

Skarlso:
/lgtm

for what it's worth. :)

jakobmoellerdev (approved):

PR #1346, opened 2026-08-10
fix(cel): prevent panic when uint/bytes CEL values reach deep-copy

### LINKED ISSUE #1308: Instance controller panics ("cannot deep copy …") on an RGD whose CEL expression resolves to a non-JSON-representable type

pujitha24:
Motivation:
A ResourceGraphDefinition whose status or resource-template CEL
expression resolves to CEL uint or bytes (or a list/map/optional
containing them) crashed the kro instance controller on first
reconcile. GoNativeType (pkg/cel/conversion/conversions.go) converted
CEL uint to Go uint64 and CEL bytes to Go []byte, but apimachinery's
runtime.DeepCopyJSONValue - used by unstructured.Unstructured.DeepCopy()
for resource templates and unstructured.NestedMap/DeepCopyJSON for
status - only accepts string, int64, bool, float64, nil, json.Number,
map[string]interface{} and []interface{}. Anything else hits its
default case and panics ("cannot deep copy uint64" / "cannot deep copy
[]uint8"). The RGD is accepted and its CRD installed either way, so
the panic only surfaces once an instance is reconciled.

Approach:
This is a narrower, surgical fix, not the full build-time rejection of
non-representable RGD schemas that the report ultimately asks for
(GraphAccepted=False before a CRD is ever installed). That requires
reasoning about static CEL types versus actual runtime values and is a
larger, more invasive change. The report itself calls out a runtime
coercion/guard in the conversion path as an acceptable complementary
approach.

An earlier version of this PR put the guard directly inside
GoNativeType. A reviewer pointed out that GoNativeType is a shared
conversion function with more than one caller, and putting the guard
there was too broad: json.marshal() (pkg/cel/library/json.go) also
calls GoNativeType, but goes through encoding/json afterwards, which
*can* represent []byte (base64-encoded) and uint64 beyond the int64
range just fine. Guarding inside GoNativeType broke that legitimate
use case (verified with a failing test before this revision: bytes
and >int64 uint marshaled fine on main, and were incorrectly rejected
by the earlier version of this PR).

The fix now lives at the point where the JSON-safety constraint
actually applies: Expression.Eval (pkg/cel/expression.go), which is
the evaluation entrypoint whose result flows into apimachinery's
unstructured deep-copy for resource templates and status fields.
GoNativeType itself is unchanged (CEL uint -> Go uint64, CEL bytes ->
Go []byte, as before). A new conversion.EnsureJSONSafe function
recursively walks GoNativeType's output and:
  - converts uint64 -> int64 (JSON-safe), erroring instead of
    overflowing when the value exceeds math.MaxInt64.
  - errors on []byte, asking the author to encode explicitly (e.g.
    base64.encode(...)) rather than returning a value that panics
    downstream.
Expression.Eval calls EnsureJSONSafe after GoNativeType, so this
becomes an actionable error on the RGD/instance instead of a
controller crash for every caller that goes through Expression.Eval
(resource templates, status fields, readiness/include-when
conditions). json.marshal calls GoNativeType directly and is
unaffected by this guard, since encoding/json doesn't have the same
restriction.

This does not implement the build-time GraphAccepted=False rejection
the report describes as the ultimate expected behavior - it only
prevents the controller from panicking on values that would trigger
it, surfacing a clear error instead.

Validation:
- go build ./...
- go test ./pkg/... (all packages pass)
- go test -race ./pkg/cel/... (passes)
- go vet ./...
- gofmt -l on all changed files (no output)
- pkg/cel/conversion/conversions_test.go: TestGoNativeType_Bytes and
  TestGoNativeType_Uint now assert GoNativeType returns the raw
  []byte/uint64 unchanged (no guard); new
  TestEnsureJSONSafe_Bytes/Uint/UintOverflow/NestedList/NestedMap
  cover the guard itself, including nested lists/maps.
- pkg/cel/library/json_test.go: added "marshal bytes" and "marshal
  uint beyond int64 range" cases to TestJSONMarshal. Confirmed these
  are a real regression check for the reviewer's concern: with the
  guard still inside GoNativeType (prior revision), both cases failed
  with "unsupported type" errors; with the guard moved to
  Expression.Eval, both pass, matching pre-existing main-branch
  behavior for json.marshal.
- Did not run an envtest/integration reproduction of the actual
  controller panic (no envtest bringing up an RGD with the issue's
  `${uint(...)}`/`${hash.sha256(...)}` expressions and reconciling an
  instance). This repo's CONTRIBUTING.md only requires local tests to
  pass for a change at this scope; the panic's root cause is fully
  contained in the Expression.Eval -> GoNativeType/EnsureJSONSafe
  path, and the targeted unit tests reproduce and fix it directly.

Report: https://github.com/kubernetes-sigs/kro/issues/1308

---

Fixes #1308

```release-note
fix(cel): prevent panic when uint/bytes CEL values reach deep-copy
```

### DIFF
--- a/pkg/cel/conversion/conversions.go
+++ b/pkg/cel/conversion/conversions.go
@@ -17,6 +17,7 @@ package conversion
 import (
 	"errors"
 	"fmt"
+	"math"
 	"reflect"
 	"time"
 
@@ -44,13 +45,25 @@ func GoNativeType(v ref.Val) (interface{}, error) {
 	case types.IntType:
 		return v.Value().(int64), nil
 	case types.UintType:
-		return v.Value().(uint64), nil
+		// apimachinery's unstructured deep-copy only accepts int64 for
+		// integers (not uint64), so convert here rather than let a raw
+		// uint64 reach DeepCopyJSONValue and panic. Values that don't fit
+		// in an int64 have no JSON-safe representation, so surface an
+		// actionable error instead.
+		u := v.Value().(uint64)
+		if u > math.MaxInt64 {
+			return nil, fmt.Errorf("%w: uint value %d overflows int64; convert it explicitly, e.g. string(...)", ErrUnsupportedType, u)
+		}
+		return int64(u), nil
 	case types.DoubleType:
 		return v.Value().(float64), nil
 	case types.StringType:
 		return v.Value().(string), nil
 	case types.BytesType:
-		return v.Value().([]byte), nil
+		// Raw bytes have no JSON-safe representation and panic apimachinery's
+		// unstructured deep-copy (DeepCopyJSONValue). Require the author to
+		// encode explicitly rather than silently pick an encoding for them.
+		return nil, fmt.Errorf("%w: bytes value cannot be used directly in a resource template or status field; encode it explicitly, e.g. base64.encode(...)", ErrUnsupportedType)
 	case types.DurationType:
 		return v1.Duration{Duration: v.Value().(time.Duration)}.ToUnstructured(), nil
 	case types.TimestampType:
--- a/pkg/cel/conversion/conversions.go
+++ b/pkg/cel/conversion/conversions.go
@@ -45,25 +45,13 @@ func GoNativeType(v ref.Val) (interface{}, error) {
 	case types.IntType:
 		return v.Value().(int64), nil
 	case types.UintType:
-		// apimachinery's unstructured deep-copy only accepts int64 for
-		// integers (not uint64), so convert here rather than let a raw
-		// uint64 reach DeepCopyJSONValue and panic. Values that don't fit
-		// in an int64 have no JSON-safe representation, so surface an
-		// actionable error instead.
-		u := v.Value().(uint64)
-		if u > math.MaxInt64 {
-			return nil, fmt.Errorf("%w: uint value %d overflows int64; convert it explicitly, e.g. string(...)", ErrUnsupportedType, u)
-		}
-		return int64(u), nil
+		return v.Value().(uint64), nil
 	case types.DoubleType:
 		return v.Value().(float64), nil
 	case types.StringType:
 		return v.Value().(string), nil
 	case types.BytesType:
-		// Raw bytes have no JSON-safe representation and panic apimachinery's
-		// unstructured deep-copy (DeepCopyJSONValue). Require the author to
-		// encode explicitly rather than silently pick an encoding for them.
-		return nil, fmt.Errorf("%w: bytes value cannot be used directly in a resource template or status field; encode it explicitly, e.g. base64.encode(...)", ErrUnsupportedType)
+		return v.Value().([]byte), nil
 	case types.DurationType:
 		return v1.Duration{Duration: v.Value().(time.Duration)}.ToUnstructured(), nil
 	case types.TimestampType:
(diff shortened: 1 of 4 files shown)


jakobmoellerdev (commented):
/ok-to-test

pkg/cel/conversion/conversions.go
```diff
@@ -44,13 +45,25 @@ func GoNativeType(v ref.Val) (interface{}, error) {
...
+			return nil, fmt.Errorf("%w: uint value %d overflows int64; convert it explicitly, e.g. string(...)", ErrUnsupportedType, u)
+		}
+		return int64(u), nil
 	case types.DoubleType:
 		return v.Value().(float64), nil
 	case types.StringType:
 		return v.Value().(string), nil
 	case types.BytesType:
-		return v.Value().([]byte), nil
+		// Raw bytes have no JSON-safe representation and panic apimachinery's
```
jakobmoellerdev: Im not sure this is the right place to guard here. Shouldnt this be done at top level when evaluating the result, and not here in conversions?

jakobmoellerdev (approved):

jakobmoellerdev (commented):
/lgtm

pujitha24:
@jakobmoellerdev — unrelated to this PR, a quick ask: I'm applying for Kubernetes org membership, which needs two sponsors who are reviewers/approvers (and from different companies). You reviewed and approved #1340 and #1346 here. Would you be open to sponsoring? My merged contributions span kro, kueue, external-dns and the cluster-api providers. No problem at all if you'd prefer not to.

PR #1349, opened 2026-08-12
helm: add extraEnv support to deployment template

n3rdc4ptn:
## Why

The chart declares `deployment.extraVolumes` and `deployment.extraVolumeMounts` in values.yaml and templates them (added recently), but there is no equivalent `deployment.extraEnv` for injecting additional environment variables into the kro container.

This is needed for use cases like pointing kro at a remote API server via `KUBECONFIG` env var when deploying kro on a separate cluster from the one it manages.

## What changed

- `helm/templates/deployment.yaml`: render `deployment.extraEnv` after the built-in env vars
- `helm/values.yaml`: add `deployment.extraEnv: []` default


jakobmoellerdev (commented):
/lgtm
/ok-to-test

jakobmoellerdev (approved):

PR #1350, opened 2026-08-12
chore(test): increase deletion timeout and add debug message

Skarlso:
Hardening the test run and fixing it to wait for enough time for the deletion to occur.

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kro/1348/presubmits-integration-tests/2087638655068803072

For more information, please read the comment I left.

### DIFF
--- a/test/integration/suites/core/crd_test.go
+++ b/test/integration/suites/core/crd_test.go
@@ -236,12 +236,20 @@ var _ = Describe("CRD", func() {
 			// Delete ResourceGraphDefinition
 			Expect(env.Client.Delete(ctx, rgd)).To(Succeed())
 
-			// Verify CRD is deleted
+			// On numerous occasions when the apiserver is overloaded the deletion
+			// just doesn't appear to be happening quickly enough.
+			// Therefore, here we increase the time a bit and log the finalizer,
+			// the deletion timestamp and the condition to make it easier to debug this.
+			// The deletion was issued by Kro correctly, it just didn't happen.
+			// https://github.com/kubernetes-sigs/kro/pull/1350
+			// The above link contains a related kro test run failure.
 			Eventually(func(g Gomega, ctx SpecContext) {
-				err := env.Client.Get(ctx, types.NamespacedName{Name: crdName},
-					&apiextensionsv1.CustomResourceDefinition{})
-				g.Expect(err).To(MatchError(errors.IsNotFound, "crd should be deleted"))
-			}, 20*time.Second, time.Second).WithContext(ctx).Should(Succeed())
+				crd := &apiextensionsv1.CustomResourceDefinition{}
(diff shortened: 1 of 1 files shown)


test/integration/suites/core/crd_test.go
```diff
@@ -236,12 +236,20 @@ var _ = Describe("CRD", func() {
...
-				err := env.Client.Get(ctx, types.NamespacedName{Name: crdName},
-					&apiextensionsv1.CustomResourceDefinition{})
-				g.Expect(err).To(MatchError(errors.IsNotFound, "crd should be deleted"))
-			}, 20*time.Second, time.Second).WithContext(ctx).Should(Succeed())
+			// On numerous occasions when the apiserver is overloaded the deletion
+			// just doesn't appear to be happening quickly enough.
+			// Therefore, here we increase the time a bit and log the finalizer,
+			// the deletion timestamp and the condition to make it easier to debug this.
+			// The deletion was issued by Kro correctly, it just didn't happen.
+			// https://github.com/openmcp-project/service-provider-kro/pull/143
```
jakobmoellerdev: please remove this link

test/integration/suites/core/crd_test.go
```diff
@@ -236,12 +236,20 @@ var _ = Describe("CRD", func() {
...
+			// The above link contains a related kro test run failure.
+			Eventually(func(g Gomega, ctx SpecContext) {
+				crd := &apiextensionsv1.CustomResourceDefinition{}
+				err := env.Client.Get(ctx, types.NamespacedName{Name: crdName}, crd)
+				g.Expect(err).To(MatchError(errors.IsNotFound, "crd should be deleted"),
+					"crd still present: deletionTimestamp=%v finalizers=%v conditions=%v",
+					crd.DeletionTimestamp, crd.Finalizers, crd.Status.Conditions)
+			}, time.Minute, time.Second).WithContext(ctx).Should(Succeed())
```
jakobmoellerdev, line 252: we increased the timeout before already and this will not fix the flake. we have prs incoming that should solve this issue. https://github.com/kubernetes-sigs/kro/pull/763 maybe is worth a look here as well as

test/integration/suites/core/crd_test.go
Skarlso: Ah wrong link 😳

test/integration/suites/core/crd_test.go
Skarlso: Hmmm. Okay. Maybe the debug line at least helps in understanding what makes this lock? I could add in some more information? Do you think that would be beneficial?

Also I'm hoping the label deletion pr might be able to make this a bit better.

PR #1352, opened 2026-08-13
fix(logging): route klog to zap and make zap-devel configurable in Helm

jakobmoellerdev:
- Configure klog to use the root zap logger so that client-go/kube logs (such as leader election) are printed in consistent JSON format.
- Expose `config.development` in Helm values and pass it as `--zap-devel` to the controller to cleanly toggle development vs production logger modes.


cheeseandcereal (approved):
Nice! lgtm

PR #1353, opened 2026-08-13
fix(runtime): identify the waited-on resource in readyWhen readiness errors

### LINKED ISSUE #1253: Improve `ResourcesReady` condition messages.
### Feature Description

**Problem Statement**:
When a node isn't fulfilling one of its `readyWhen` conditions, the `ResourcesReady` condition is set to `False` with a reason highlighting the node and failing condition in the following format: 
```
resource reconciliation failed: apply results contain errors: waiting for external reference 
"$NODE": readyWhen condition evaluated to false: "$CONDITION" (waiting for readiness)
```

This doesn't seem ideal for at least 2 reasons:
- It exposes some implementation details of the RGD to the user
- It requires some knowledge about the RGD implementation in order to find the resource that might be hanging in an error/pending state, waiting to be ready. The node name doesn't really help the user find the actual resource that is being waited on.

**Proposed Solution**:
Quick win solution:
- The `ResourcesReady` condition should at least point to the actual resource that is being waited on, providing the following information: group, kind, namespace and name. The node name can probably stay as it carries some valuable information for RGD maintainers.
- The "waiting for readiness" trailing statement should be remove, it doesn't provide value
 
Potential follow-up issue, probably open for a KREP:
Potential rework of the `readyWhen` field, making it an array of objects with the following structure:
```yaml
readyWhen:
- condition: string | required=true description="current cel expression to validate readiness" 
  message: string | required=false description="new field, potentially a cel expression to allow pointing 
    to nodes fields (allowing for a bubble up of some of the nodes conditions?)
``` 

pujitha24:
When a node's readyWhen condition evaluates to false, the ResourcesReady
condition only named the internal RGD node ID (e.g. "vpc"), not the actual
resource being waited on, and appended a redundant "(waiting for readiness)"
suffix from the sentinel error's own text.

This adds a `resourceIdentity()` helper that formats group/version/kind and
namespace/name from the observed object, and includes it in the readiness
message. To drop the redundant suffix while keeping `errors.Is(err,
ErrWaitingForReadiness)` working, `waitingForReadinessError` now implements
the standard `Unwrap()` interface instead of a custom `Is()`.

No change to reconciliation behavior or timing — only the message text.

AI assistance: drafted with Claude Code.

Fixes #1253

```release-note
fix(runtime): identify the waited-on resource in readyWhen readiness errors
```


pkg/runtime/errors.go
```diff
@@ -26,6 +27,24 @@ var ErrDataPending = errors.New("data pending")
 
 var ErrWaitingForReadiness = errors.New("waiting for readiness")
 
+// waitingForReadinessError formats a readiness-wait message that satisfies
+// errors.Is(err, ErrWaitingForReadiness) without appending the sentinel's own
+// "waiting for readiness" text, which duplicates the NodeState/condition
+// reason and adds no value for users reading the message.
+type waitingForReadinessError struct {
```
jakobmoellerdev, line 34: I think this should align more to the standard unwrap interface from go stdlib no?

jakobmoellerdev (commented):
the idea from your ai is not bad, but you should really make shorter descriptions and be more considerate of human reviewers by doing your own review as well.

pujitha24:
Sorry for going quiet here. I did push a fix for this the same day, 45213ef — swapped the custom `Is(target error) bool` for the standard `Unwrap() error` returning `ErrWaitingForReadiness`, so `errors.Is` works through the normal stdlib path. I just never came back to confirm it in the thread, which is on me. That's the current state of the PR, tests still pass.

pujitha24:
/assign @cheftako

This has lgtm and is green — assigning for approval whenever you have a moment.

PR #1365, opened 2026-08-18
refactor: scaffolding for graph development work; incremental from #1355

jlbutler:
Scaffolding to kick off Graph development work.

This PR cherrypicks the first eleven commits of #1355 onto main so they can be reviewed and merged independently. Authorship is preserved; all eleven are @jakobmoellerdev split out to simplify review.

Every commit intends to refactor while preserving behavior of existing code. None of them touch `pkg/graphengine` or the Graph API types from the active dev branch, so nothing about the Graph work is being decided by merging this PR. 

The entire point is to take 11 commits and 28 files off the top of a 189-file review so the remaining diff is closer to the part that needs real scrutiny. A few similar PRs against main are likely to follow.

## Commits

- `engine-prep: characterization tests for Runtime.ApplyOrder + minimal consumer`
- `engine-prep: rename graph.RGDConfig to graph.Config (compat alias kept)`
- `engine-prep: introduce runtime.KroBuiltinCondition, drop v1alpha1 from private helpers`
- `engine-prep: de-parameterise validation helpers from v1alpha1 types`
- `engine-prep: promote revisions.Resolver interface + OwnerKey alias`
- `engine-prep: extract stripStatusFromSchema from getSchemaWithoutStatus`
- `engine-prep: extract buildResourceNode + inferStatusSchema from RGD wrappers`
- `engine-prep: buildInstanceNode takes GroupVersionResource directly`
- `engine-prep: introduce graph.Source + Builder.CompileSource`
- `engine-prep: FromGraph functional-options tail (WithInstance, WithMaxCollectionSize)`
- `engine-prep: rename revisions.Entry.RGDName to OwnerKey`

28 files changed, 719 insertions, 462 deletions.

The shape of the series is loosening the coupling between the graph builder, the runtime and the `v1alpha1` types: extracting helpers that were inlined in RGD-specific wrappers, de-parameterising validation from the API types, and adding functional options where a call site previously took a fixed argument list. The first commit adds characterization tests for `Runtime.ApplyOrder` before anything moves, which is the right order to do this in.

Two renames carry compatibility aliases (`graph.RGDConfig` to `graph.Config`, and `revisions.Entry.RGDName` to `OwnerKey`), so nothing downstream breaks on this alone.

## Verification

On this branch, against current main:

- `go build ./...` clean
- `go vet ./...` clean
- `go test ./pkg/... ./cmd/...` green, no failures
- `golangci-lint run ./...` with the repo's pinned v2.11.4: 0 issues

The cherry-pick applied with no conflicts, which is itself some evidence of independence: these commits do not rely on anything later in #1355.

## Why separately

#1355 is 189 files and its interesting content is the RGD-on-Graph adapter, the shared engine, and the removal of the classic runtime. This series is none of that. Reviewing it here means it can be checked quickly on its own terms, and #1355 gets correspondingly smaller. After this merges, a rebase there should drop these commits automatically.


jakobmoellerdev (approved):
/lgtm

PR #1369, opened 2026-08-19
test: add coverage for unasserted RGD reconcile, ownership and upgrade contracts

cheeseandcereal:
## What this does

Adds test coverage for a set of behaviors the RGD reconcile path relies on today but does not assert anywhere. Tests only — no production code changes, no CI configuration changes.

The unifying theme is that every contract here **fails silently**. None of these behaviors produce an error, an event, or a condition when they break: a status field simply never appears, an orphaned object simply stays, a metric series simply stops being scraped, a field removed from a template simply doesn't go away. That combination — load-bearing, unasserted, and quiet on failure — is what makes them worth pinning.

These came out of a coverage-gap pass over the instance reconcile path, comparing what the suites assert against what the code actually guarantees.

## What's covered

| Area | Contract | File |
| --- | --- | --- |
| Instance data | Text containing `${...}` in a spec field or annotation is carried through verbatim | `test/integration/suites/core/instance_expression_isolation_test.go` |
| Instance data | An instance value that names one of the definition's resources stays a plain string | same |
| Status | A status field declared as an array is projected as an array, top level and nested | `.../status_array_projection_test.go` |
| Ownership | Managed resources are applied under a stable server-side-apply field manager | `.../resource_ownership_test.go` |
| Ownership | Two instances rendering the same object do not silently overwrite each other | same |
| Pruning | A resource removed from a definition is pruned even while another resource is not ready | `.../resource_pruning_test.go` |
| Collections | The configured collection size limit is the one enforced | `.../collection_behavior_test.go` |
| Collections | Drift is corrected in every namespace a collection spans | same |
| Resource shapes | Templates and external refs work with API versions outside the `vN[alpha\|beta]N` convention | `.../resource_shape_compatibility_test.go` |
| Resource shapes | A definition may create a `CustomResourceDefinition` | same |
| Resource shapes | An unresolvable child namespace names the field the author has to fix | same |
| Metrics | `instance_reconcile_errors_total` counts reconcile-level failures, not retried per-resource applies | `.../reconcile_error_metrics_test.go` |
| Metrics | The registered metric family names stay registered | `pkg/metrics/register_families_test.go` |
| Upgrade | Removing a field from a template removes it from resources created before the upgrade | `test/upgrade/template_shrink_test.go` + `fixtures/template-shrink/` |

## Why these specific ones

A few deserve a word on motivation, since the value isn't obvious from the test name alone:

**Instance data is data, not a template.** An instance is written by the consumer of a generated API, who knows the schema but not the definition behind it. Users legitimately store shell snippets, container commands and downstream templating config in spec fields, so `${...}` has to survive a round trip. Separately, an instance value that happens to name a resource id must stay a string — resource ids are an internal detail of the definition, and resources can hold cluster state the instance's author cannot read.

**API versions are opaque strings.** Kubernetes only requires a CRD version to be a valid DNS label, and widely used operators pick names outside the usual convention: Azure Service Operator versions every type as `vNapiYYYYMMDD`, Google Config Connector ships versions like `v1p1beta1`. Both templates and external references are covered.

**CRDs are the most demanding template payload.** Shipping CRDs or an operator as a kro API is an established pattern, and a CRD's `openAPIV3Schema` is a deeply nested, self-referential document.

**The field manager is a compatibility surface.** Which manager owns a field decides whether dropping that field from a template removes it from the live object, because server-side apply only clears fields the *same* manager set previously. If the manager name changes between releases, objects created by an earlier release keep their fields owned by a manager that never applies again, and template removals silently stop taking effect on exactly those objects. New objects behave correctly, which makes the divergence easy to miss — hence the upgrade-time spec as well as the in-cluster one.

**Metric names are operator-facing.** Dashboards, alerts and recording rules reference them by name. Removing or renaming one is invisible at compile time and at scrape time: the series just stops appearing, so panels go blank and alerts stop evaluating rather than failing loudly. The counter-scale spec is the same concern from the other side — if a retried per-resource apply failure incremented `instance_reconcile_errors_total`, any absolute threshold on that series would start firing for instances that are merely waiting on something external.

## Two existing files touched

`test/upgrade/{mutation,upgrade}_test.go` — 7 lines total.

Both keep an explicit list of RGDs that post-upgrade suites deliberately mutate, so their spurious-GraphRevision invariants can distinguish intentional changes from unexpected ones. The new template-shrink suite mutates a second RGD, so it is registered alongside the existing `mutationRGDName` and `retentionRGDName` entries, following the pattern already in those files. Comments updated to match.

## Verification

All green on this branch:

```
make build
make verify-codegen                                            # up to date, no churn
make lint                                                      # 0 issues
make test WHAT=unit
make test WHAT=integration                                     # core: 173/173 (+16 specs)
make test-e2e-upgrade-kind KRO_UPGRADE_FROM_VERSION=v0.9.3      # 46 pre-upgrade / 79 post-upgrade (+4 specs)
```

The new tests land in suites the presubmits already run (`integration-tests`, `e2e-upgrade-tests`), and the upgrade fixture is picked up by the existing `fixtures/*/rgd.yaml` glob, so no CI configuration changes are needed.

Each spec was also checked to fail for the intended reason, not incidentally — by breaking the behavior it covers and confirming the assertion is the one that trips, with a failure message that names the problem rather than just reporting a timeout.

## Notes for reviewers

- **The metric list and the field-manager name are spelled out literally, on purpose.** Referencing `applyset.FieldManager` or deriving the metric names from the registry would make each assertion follow a rename instead of catching it. Both are meant to force a deliberate, reviewable decision — if the value has to change, update it in the test and treat pre-existing objects or existing dashboards as the migration concern they are.
- **`configuredMaxCollectionSize` in `collection_behavior_test.go` mirrors the value in `test/integration/environment/setup.go`.** The environment builds its `RGDConfig` inline rather than exposing it, so the expected limit is a documented constant with a pointer to its source.
- **The shared-resource and multi-namespace-drift specs sample over a window** (`Consistently`, and repeated tampering) rather than reading once, because a single read can land between competing writes and look stable.
- **Spec-level isolation:** every spec creates its own namespace and uses randomized names, groups and schema kinds, so nothing collides in the shared envtest cluster. The two `DescribeTable` entries use distinct schema kinds for the same reason — sharing one would make their generated CRDs contend.
- **Helpers reuse what the package already has** (`waitForRGDActive`, `instanceConditions`); the few new ones (`newInstance`, `waitForInstanceState`, `installTestCRD`) are additive.

### DIFF
--- a/test/integration/suites/core/collection_behavior_test.go
+++ b/test/integration/suites/core/collection_behavior_test.go
@@ -1,0 +1,213 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package core_test
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	krov1alpha1 "github.com/kubernetes-sigs/kro/api/v1alpha1"
+	"github.com/kubernetes-sigs/kro/pkg/testutil/generator"
+)
+
+// configuredMaxCollectionSize mirrors the RGDConfig the integration environment
+// installs (see test/integration/environment/setup.go). The collection cap is
+// an operator-facing setting, so the value the controller enforces has to be
+// the value it was given rather than a built-in default.
+const configuredMaxCollectionSize = 1000
+
+// Collection behavior that the existing forEach coverage does not reach.
+//
+//   - The configured collection cap is what gets enforced. Operators raise or
+//     lower it deliberately; silently enforcing a different number makes a
+//     tuned deployment behave unlike its configuration, and the effective limit
+//     can end up lower than what was asked for.
+//   - Drift correction has to cover every namespace a collection spans. A
+//     collection can template items into different namespaces, and an item is
+//     no less managed for living in the second one. If only part of a
+//     collection is watched, the rest drifts unnoticed while the instance keeps
+//     reporting that everything is ready.
+var _ = Describe("CollectionBehavior", func() {
+	var namespace string
+
+	BeforeEach(func(ctx SpecContext) {
+		namespace = fmt.Sprintf("test-%s", rand.String(5))
+		Expect(env.Client.Create(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Name: namespace},
+		})).To(Succeed())
+	})
+
+	AfterEach(func(ctx SpecContext) {
+		Expect(env.Client.Delete(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Name: namespace},
+		})).To(Succeed())
+	})
+
+	It("enforces the configured collection size limit", func(ctx SpecContext) {
+		// 34 x 34 = 1156 items, above the configured cap. Nothing should be
+		// created, and the instance should say which limit it hit.
+		rgd := generator.NewResourceGraphDefinition("test-collection-cap",
+			generator.WithSchema(
+				"TestCollectionCap", "v1alpha1",
+				map[string]interface{}{
+					"name": "string",
+				},
+				nil,
+			),
+			generator.WithResourceCollection("configmaps", map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "ConfigMap",
+				"metadata": map[string]interface{}{
+					"name": "${schema.spec.name}-${string(i)}-${string(j)}",
+				},
+			},
+				[]krov1alpha1.ForEachDimension{
+					{"i": "${lists.range(34)}"},
+					{"j": "${lists.range(34)}"},
+				},
+				nil, nil),
+		)
+		Expect(env.Client.Create(ctx, rgd)).To(Succeed())
+		DeferCleanup(func(ctx SpecContext) {
+			Expect(env.Client.Delete(ctx, rgd)).To(Succeed())
+		})
+		waitForRGDActive(ctx, rgd.Name)
+
+		name := "collection-cap"
+		instance := newInstance("TestCollectionCap", name, namespace, map[string]interface{}{
+			"name": name,
+		})
+		Expect(env.Client.Create(ctx, instance)).To(Succeed())
+
+		Eventually(func(g Gomega, ctx SpecContext) {
+			g.Expect(env.Client.Get(ctx, types.NamespacedName{
+				Name:      name,
+				Namespace: namespace,
+			}, instance)).To(Succeed())
+
+			conds := instanceConditions(instance)
+			g.Expect(conds).To(ContainSubstring(fmt.Sprintf("%d", configuredMaxCollectionSize)),
+				"the instance should report the configured collection limit of %d; conditions: %s",
+				configuredMaxCollectionSize, conds)
+		}, 60*time.Second, 2*time.Second).WithContext(ctx).Should(Succeed())
+
+		state, _, err := unstructured.NestedString(instance.Object, "status", "state")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(state).ToNot(Equal("ACTIVE"),
+			"an instance whose collection exceeds the cap must not report ACTIVE")
+	})
+
+	It("corrects drift on collection items in every namespace the collection spans", func(ctx SpecContext) {
+		altNamespace := fmt.Sprintf("alt-%s", namespace)
+		Expect(env.Client.Create(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Name: altNamespace},
+		})).To(Succeed())
+		DeferCleanup(func(ctx SpecContext) {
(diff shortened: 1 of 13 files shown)


test/integration/suites/core/collection_behavior_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+// configuredMaxCollectionSize mirrors the RGDConfig the integration environment
+// installs (see test/integration/environment/setup.go). The collection cap is
+// an operator-facing setting, so the value the controller enforces has to be
+// the value it was given rather than a built-in default.
+const configuredMaxCollectionSize = 1000
+
+// Collection behavior that the existing forEach coverage does not reach.
+//
+//   - The configured collection cap is what gets enforced. Operators raise or
```
jakobmoellerdev, line 41: lowering the collection limit is currently unsupported behavior so it is completely fine for me that this produces weird results. maybe this should be documented.

test/integration/suites/core/resource_ownership_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+	"github.com/kubernetes-sigs/kro/pkg/controller/instance/applyset"
+	"github.com/kubernetes-sigs/kro/pkg/testutil/generator"
+)
+
+// How kro takes ownership of the resources it manages.
+//
+// Two properties, both about not losing track of what we own:
+//
+//   - The server-side-apply field manager is a compatibility surface. Which
```
jakobmoellerdev, line 37: this should likely be an upgrade test

jakobmoellerdev (commented):
This is a great test set. One thing I would ask is if some of these behaviors really must be within integration or if we cannot move them to unit tests / upgrade tests. Apart from that this looks good to me!

test/upgrade/template_shrink_test.go
```diff
@@ -0,0 +1,158 @@
...
+// version-sensitive: if a release changes the field manager it applies under,
+// objects created by an earlier release keep their fields owned by a manager
+// that never applies again, and dropping a field from a template silently stops
+// removing it from exactly those objects — the pre-existing ones. New objects
+// behave correctly, which makes the divergence easy to miss.
+//
+// Adding a field is already covered by the RGD mutation checks. This is the
+// other direction, which is the one that depends on field ownership carrying
+// across the upgrade.
+var _ = ginkgo.Describe("Post-Upgrade Template Shrink", ginkgo.Ordered, func() {
```
Skarlso, line 58: Can you please add a fixture testing that the configmap was actually there BEFORE the upgrade?

There is no way knowing if it actually failed previously, and then the eventually just asserts that it was there and done.

And for an upgrade test I would say it's kinda important to verify the previous state.

test/integration/suites/core/collection_behavior_test.go
```diff
@@ -0,0 +1,213 @@
...
+		Expect(env.Client.Create(ctx, instance)).To(Succeed())
+
+		Eventually(func(g Gomega, ctx SpecContext) {
+			g.Expect(env.Client.Get(ctx, types.NamespacedName{
+				Name:      name,
+				Namespace: namespace,
+			}, instance)).To(Succeed())
+
+			conds := instanceConditions(instance)
+			g.Expect(conds).To(ContainSubstring(fmt.Sprintf("%d", configuredMaxCollectionSize)),
```
Skarlso, line 109: I'd expand this contains check. Just checking that 1000 is in that text is a bit brittle. :) I would like to check a bit more here to make sure that we are actually seeing what we expect please.

test/integration/suites/core/resource_shape_compatibility_test.go
```diff
@@ -0,0 +1,422 @@
...
+			}
+		}
+		g.Expect(established).To(BeTrue(), "CRD %s is not Established", crd.Name)
+	}, 30*time.Second, time.Second).WithContext(ctx).Should(Succeed())
+}
+
+func ptrTrue() *bool {
+	v := true
+	return &v
+}
```
Skarlso, line 417: This is superfluous. Also, please use `new` instead now. :) Thanks 🙇‍♀️.

test/integration/suites/core/resource_shape_compatibility_test.go
```diff
@@ -0,0 +1,422 @@
...
+
+// schemaGVK builds a GroupVersionKind for the test CRDs above.
+func schemaGVK(group, version, kind string) schema.GroupVersionKind {
```
Skarlso, line 420: This is used in one place only. I don't really see the point of it. Can you collapse this please?

test/integration/suites/core/instance_expression_isolation_test.go
```diff
@@ -0,0 +1,284 @@
...
+			continue
+		}
+		typ, _, _ := unstructured.NestedString(cond, "type")
+		status, _, _ := unstructured.NestedString(cond, "status")
+		reason, _, _ := unstructured.NestedString(cond, "reason")
+		msg, _, _ := unstructured.NestedString(cond, "message")
+		out += fmt.Sprintf("[%s=%s(%s): %s] ", typ, status, reason, msg)
+	}
+	return out
+}
```
Skarlso, line 284: These are used in other tests as well. Could you please put them into a suite_test file? Thanks! :)

test/integration/suites/core/instance_expression_isolation_test.go
```diff
@@ -0,0 +1,284 @@
...
+	for _, c := range conds {
+		cond, ok := c.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		typ, _, _ := unstructured.NestedString(cond, "type")
+		status, _, _ := unstructured.NestedString(cond, "status")
+		reason, _, _ := unstructured.NestedString(cond, "reason")
+		msg, _, _ := unstructured.NestedString(cond, "message")
+		out += fmt.Sprintf("[%s=%s(%s): %s] ", typ, status, reason, msg)
```
Skarlso, line 281: This is a very good case for a string builder. :)

test/integration/suites/core/collection_behavior_test.go
```diff
@@ -0,0 +1,213 @@
...
+		Expect(env.Client.Create(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Name: namespace},
+		})).To(Succeed())
+	})
+
+	AfterEach(func(ctx SpecContext) {
+		Expect(env.Client.Delete(ctx, &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Name: namespace},
+		})).To(Succeed())
+	})
```
Skarlso, line 64: This is duplicated amongst all the test cases. Please put them into a suite test.

Skarlso (commented):
I would like a couple of things to be looked at more carefully, but I'm okay with the general approach. :) Thanks! 🙇‍♀️

PR #1372, opened 2026-08-19
fix(graph/schema): add invalidation epoch tracking to CachedSchemaResolver and add schema tests

jakobmoellerdev:
## Description
- Adds per-GroupKind invalidation epoch tracking to `CachedSchemaResolver` to prevent in-flight `singleflight.Do` fetches from caching stale schemas if an `Invalidate()` event occurs concurrently.
- Adds comprehensive unit test coverage for schema caching, CEL type conversion, type fallbacks, and schema field descriptors.

## Type of change
- [x] Bug fix
- [x] Test coverage

## Testing
- `make build`
- `make lint`
- `make test WHAT=unit`
- Integration test suite passed.

### DIFF
--- a/pkg/graph/schema/resolver/cached.go
+++ b/pkg/graph/schema/resolver/cached.go
@@ -1,0 +1,114 @@
+// Copyright 2026 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resolver
+
+import (
+	"sync"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+	"golang.org/x/sync/singleflight"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/cel/openapi/resolver"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+// CachedSchemaResolver wraps an underlying schema resolver with a
+// bounded LRU cache and push-driven invalidation. Entries live until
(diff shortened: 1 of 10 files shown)


pkg/graph/schema/resolver/cached.go
```diff
@@ -0,0 +1,114 @@
...
+// a CRD content change is observed).
+//
+// There is no time-based TTL: because the schema watcher fires precise
+// invalidations on CRD changes, periodic refetches are redundant. The cache
+// stays correct as long as the watcher pushes every relevant change.
+//
+// Concurrent fetches for the same GVK are deduplicated through
+// singleflight so the underlying delegate sees one call per unique
+// key, even under high concurrency.
+type CachedSchemaResolver struct {
```
Skarlso, line 40: This doesn't emit any metrics while the old one does. 🤔

pkg/graph/schema/resolver/cached.go
```diff
@@ -0,0 +1,114 @@
...
+// the supplied value. The schema watcher calls this on CRD content
+// changes so subsequent ResolveSchema calls re-fetch the new shape.
+// Idempotent: calling for an unknown GK is a no-op.
+func (c *CachedSchemaResolver) InvalidateGroupKind(gk schema.GroupKind) {
+	c.mu.Lock()
+	c.epochs[gk]++
+	for _, k := range c.cache.Keys() {
+		if k.GroupKind() == gk {
+			c.cache.Remove(k)
+			c.sf.Forget(k.String())
```
Skarlso, line 75: This is incorrect here. 🤔 It means that sf will never forget keys that are in-flight. You need to move Forget out of the loop I think, but that would mean tracking the in flight keys. Ugh.

Also, also, this is of course, not used yet, right? So what's the purpose here? How are you going to utilize this? :)

Skarlso: Actually, this self-heals on the next invalidate run. So don't care. :)

pkg/graph/schema/resolver/cached.go
```diff
@@ -0,0 +1,114 @@
...
+		sch, err := c.delegate.ResolveSchema(gvk)
+		if err != nil {
+			return nil, err
+		}
+		c.mu.RLock()
+		curEpoch := c.epochs[gk]
+		c.mu.RUnlock()
+		if curEpoch == epoch {
+			c.cache.Add(gvk, sch)
+		}
```
Skarlso, line 107: I think you need a write lock inside the epoch since you released the read lock.

PR #1373, opened 2026-08-19
feat(cel): add AST inspector, program cost limits, and CEL test coverage

jakobmoellerdev:
## Description
- Introduces `pkg/cel/ast/inspector.go` providing AST traversal helpers for extracting identifiers, member accesses, and optional chaining expressions.
- Enforces `DefaultCostLimit` (10,000,000) and `DefaultInterruptCheckFrequency` (100) on CEL programs to protect against unbounded execution.
- Adds comprehensive CEL test coverage across environment, type conversions, and expressions.
- Add dedicated optional tracking so it can be used for soft dependency declaration

## Type of change
- [x] Feature / Improvement
- [x] Test coverage

## Testing
- `make build`
- `make lint`
- `make test WHAT=unit`
- Integration test suite passed.

### DIFF
--- a/pkg/cel/ast/inspector.go
+++ b/pkg/cel/ast/inspector.go
@@ -73,6 +73,13 @@ type ExpressionInspection struct {
 	// ResourceDependencies lists all known resources and their access paths
 	// used in the expression
 	ResourceDependencies []ResourceDependency
+	// OptionalResourceDependencies lists known resources that appear only as
+	// the direct receiver of CEL optional chaining (id.?field / id[?"k"]) —
+	// the id is arg[0] of a `_?._` or `_[?_]` call. This is a strict subset of
+	// the occurrences also recorded in ResourceDependencies; callers compare
+	// occurrence counts to decide whether every reference to an id is optional
+	// (a soft dependency) or at least one is a hard reference.
+	OptionalResourceDependencies []ResourceDependency
 	// FunctionCalls lists all known function calls and their arguments found
 	// in the expression
 	FunctionCalls []FunctionCall
@@ -95,6 +102,7 @@ func (e *ExpressionInspection) UsesOmit() bool {
 
 func (e *ExpressionInspection) merge(other ExpressionInspection) {
 	e.ResourceDependencies = append(e.ResourceDependencies, other.ResourceDependencies...)
+	e.OptionalResourceDependencies = append(e.OptionalResourceDependencies, other.OptionalResourceDependencies...)
 	e.FunctionCalls = append(e.FunctionCalls, other.FunctionCalls...)
 	e.UnknownResources = append(e.UnknownResources, other.UnknownResources...)
 	e.UnknownFunctions = append(e.UnknownFunctions, other.UnknownFunctions...)
@@ -262,6 +270,24 @@ func (a *Inspector) inspectCall(ast *celast.AST, call celast.CallExpr, path stri
 
 	for _, arg := range call.Args() {
 		out.merge(a.inspectExpr(ast, arg, ""))
+	}
+
+	// Optional chaining: `id.?field` (_?._) and `id[?"k"]` (_[?_]) parse with
+	// the made-optional receiver as arg[0]. When that receiver is a bare,
+	// known-resource identifier, the whole reference is non-gating: if the
+	// resource has not published yet the expression yields optional.none() and
+	// the field is dropped. Record it so the caller can tell a
+	// receiver-of-optional reference apart from a hard access on the same id.
+	if fn == "_?._" || fn == "_[?_]" {
+		args := call.Args()
+		if len(args) >= 1 && args[0].Kind() == celast.IdentKind {
+			name := args[0].AsIdent()
+			if _, isLoop := a.loopVars[name]; !isLoop {
+				if _, ok := a.resources[name]; ok {
+					out.OptionalResourceDependencies = append(out.OptionalResourceDependencies, ResourceDependency{ID: name, Path: name})
+				}
+			}
+		}
 	}
 
 	// Namespaced (member) function: target.method
--- a/pkg/cel/coverage_test.go
+++ b/pkg/cel/coverage_test.go
@@ -1,0 +1,733 @@
+// Copyright 2025 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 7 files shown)


pkg/cel/expression.go
```diff
@@ -24,6 +24,23 @@ import (
...
+	// DefaultCostLimit bounds the maximum evaluation cost for CEL programs to
+	// prevent unbounded resource consumption from pathological expressions.
+	DefaultCostLimit = 10_000_000
+	// DefaultInterruptCheckFrequency defines the instruction step frequency for checking
+	// cancellation and interrupt signals during evaluation.
+	DefaultInterruptCheckFrequency = 100
+)
+
+// DefaultProgramOptions returns the standard execution bound options for CEL programs.
+func DefaultProgramOptions() []cel.ProgramOption {
```
Skarlso, line 37: @jakobmoellerdev this isn't used yet anywhere. Did you want to use it later in a follow-up?

pkg/cel/environment.go
```diff
@@ -92,66 +110,97 @@ func WithListVariables(names []string) EnvOption {
...
+	baseDeclarationsOnce[idx].Do(func() {
+		decls := coreDeclarations()
+		if includeRuntime {
 			// Runtime() registers the `runtime` CEL variable used to author
 			// custom status conditions. The graph builder rejects it outside
 			// the schema's status.conditions block.
-			library.Runtime(),
+			decls = append(decls, library.Runtime())
 		}
+		cachedBaseDeclarations[idx] = decls
```
Skarlso, line 178: You need to make a copy of this slice here or do something like `decls[:len(decls):len(decls)]` otherwise, since you appended runtime to this it expanded the capacity and length, meaning now, you're are sharing the underlying array in the cachedBaseDeclarations.

pkg/cel/environment.go
```diff
@@ -233,14 +283,27 @@ func defaultEnvironment(options ...EnvOption) (*cel.Env, *DeclTypeProvider, erro
...
+	opts := append([]EnvOption{WithTypedResources(schemas)}, options...)
+	return defaultEnvironment(opts...)
+}
+
+// TypedEnvironmentWithIDsAndProvider builds the typed CEL environment with
+// the supplied typed resources, additionally declaring the given identifiers
+// as untyped (dyn) variables. Useful when some identifiers carry no schema
+// but must still resolve at type-check time.
+func TypedEnvironmentWithIDsAndProvider(schemas map[string]*spec.Schema, dynIDs []string, options ...EnvOption) (*cel.Env, *DeclTypeProvider, error) {
+	opts := append([]EnvOption{WithTypedResources(schemas), WithResourceIDs(dynIDs)}, options...)
```
Skarlso, line 305: I'm not a 100% sure here, but I think this needs a dedup but I need to understand the code more to be more precise here. :) Maybe you know it better. Can it happen that the schema declares the same id as dynIDs?

Skarlso (commented):
Found three things that I'm unsure about the rest look okay. :)

pkg/cel/ast/inspector.go
```diff
@@ -264,6 +272,24 @@ func (a *Inspector) inspectCall(ast *celast.AST, call celast.CallExpr, path stri
...
+	// the made-optional receiver as arg[0]. When that receiver is a bare,
+	// known-resource identifier, the whole reference is non-gating: if the
+	// resource has not published yet the expression yields optional.none() and
+	// the field is dropped. Record it so the caller can tell a
+	// receiver-of-optional reference apart from a hard access on the same id.
+	if fn == "_?._" || fn == "_[?_]" {
+		args := call.Args()
+		if len(args) >= 1 && args[0].Kind() == celast.IdentKind {
+			name := args[0].AsIdent()
+			if _, isLoop := a.loopVars[name]; !isLoop {
```
cheeseandcereal, line 285: This new `loopVars` check inherits the inspector's over-broad comprehension scope: iterator and accumulator names are installed before `IterRange` and `AccuInit`, where CEL still resolves an outer identifier. With resource `x`, `x.?items.orValue([]).map(x, x)` evaluates against outer `x` but inspection records no dependency. You may want to inspect the full loop and initializer before pushing scoped names.

pkg/cel/coverage_test.go
```diff
@@ -0,0 +1,733 @@
...
+	}
+}
+
+// --- schemas.go: SchemaDeclTypeWithMetadata across schema shapes ---
+
+func declType(t *testing.T, s *spec.Schema, root bool) *apiservercel.DeclType {
+	t.Helper()
+	if s == nil {
+		// Pass a typed-nil common.Schema to exercise the early nil check.
+		return SchemaDeclTypeWithMetadata(nil, root)
```
cheeseandcereal, line 207: `nil` here is a nil `common.Schema` interface, whereas `defaultEnvironment` wraps map entries as non-nil `*openapi.Schema` values. A nil schema entry still panics in `IsXIntOrString`.

pkg/cel/environment.go
```diff
@@ -233,14 +283,27 @@ func defaultEnvironment(options ...EnvOption) (*cel.Env, *DeclTypeProvider, erro
...
 	return env, provider, err
 }
 
 // TypedEnvironmentWithProvider creates a typed CEL environment.
 // It returns both the environment and the DeclTypeProvider.
-func TypedEnvironmentWithProvider(schemas map[string]*spec.Schema) (*cel.Env, *DeclTypeProvider, error) {
-	return defaultEnvironment(WithTypedResources(schemas))
+func TypedEnvironmentWithProvider(schemas map[string]*spec.Schema, options ...EnvOption) (*cel.Env, *DeclTypeProvider, error) {
+	opts := append([]EnvOption{WithTypedResources(schemas)}, options...)
+	return defaultEnvironment(opts...)
```
cheeseandcereal, line 297: `WithCustomDeclarations` in these newly accepted options can carry `cel.Types(...)`, but the OpenAPI provider is built over `types.NewEmptyRegistry()` and replaces those types. A declaration that works in `DefaultEnvironment` then fails here with `unexpected failed resolution`. You may want to layer the OpenAPI provider over the already-resolved provider.

pkg/cel/environment.go
```diff
@@ -233,14 +283,27 @@ func defaultEnvironment(options ...EnvOption) (*cel.Env, *DeclTypeProvider, erro
...
+}
+
+// TypedEnvironmentWithIDsAndProvider builds the typed CEL environment with
+// the supplied typed resources, additionally declaring the given identifiers
+// as untyped (dyn) variables. Useful when some identifiers carry no schema
```
cheeseandcereal, line 302: `dynIDs` are forwarded to `WithResourceIDs`, which declares `cel.AnyType` (protobuf `Any`), not CEL `dyn`. `items.map(x, x)` is rejected, but type-checks with `cel.DynType`; use `DynType` or change this API contract.

PR #1375, opened 2026-08-21
docs: note that lists built from maps need sort() for stable ordering

### LINKED ISSUE #1374: docs: note that lists built from maps need sort() for stable ordering
**Location**:

`website/docs/docs/concepts/rgd/04-cel-libraries.md`, near the `sort()` / `sortBy()` entries (currently around lines 258-259). Possibly also worth a line in `03-cel-expressions.md`.

**Current State**:

The CEL libraries page documents `<list>.sort()` and `<list>.sortBy(var, keyExpr)` as sorting helpers, and separately documents `<map>.merge(map)`. Nothing anywhere says that **a list built from a map has non-deterministic element order**, or that this matters for reconciliation.

**Suggested Changes**:

Add a short note wherever `sort()` is introduced: when a template field is a list of objects derived from a map, sort it, or the rendered desired state changes between reconciles even when nothing has changed.

The pattern that hits this is common with AWS providers. ACK `ec2.services.k8s.aws` resources (`VPC`, `Subnet`, `SecurityGroup`) take `tags` as `[{key, value}]` rather than a map, so authors converting a `map[string]string` schema field into that shape write something like:

```yaml
# unstable: element order comes from map iteration
tags: '${schema.spec.awsTags.map(k, {"key": k, "value": schema.spec.awsTags[k]})}'
```

```yaml
# stable
tags: '${schema.spec.awsTags.map(k, k).sort().map(k, {"key": k, "value": schema.spec.awsTags[k]})}'
```

**Additional Context**:

This is measured, not theoretical. With a 4-key map, deleting the child object and letting kro rebuild it three times produced:

```
auto-delete cost-center project team
cost-center project team auto-delete
auto-delete cost-center project team
```

With `.sort()` inserted, all three rebuilds produced identical output.

The consequence is worse than cosmetic: the rendered desired state differs between reconciles, so kro re-applies the field every cycle and a provider controller can see a perpetual diff on a resource nobody touched. It is also quietly self-inflicted and hard to attribute, since the RGD reports healthy throughout.

Came out of #1314, where an author was building exactly this map-to-tag-list conversion for ACK EC2 resources. Two other items from that thread are already being handled there (a non-compiling `merge()` example in the single-resource-RGD guide, and a misleading SimpleSchema parse error); this one is separate and purely a docs gap, so filing it on its own.

/area documentation
/area cel
/good-first-issue

---

## Scope addition: document both conversion directions and the `merge()` constraint

Expanding this per discussion on #1314. The underlying gap is that nothing documents how to move between the two `tags` shapes, or the constraint that decides which one to declare. All of the following is verified but undocumented.

**Map-shaped schema field to list-shaped resource** (needs `.sort()`, per the original description above):

```yaml
tags: '${schema.spec.awsTags.map(k, k).sort().map(k, {"key": k, "value": schema.spec.awsTags[k]})}'
```

**List-shaped schema field to map-shaped resource**, via `transformMapEntry`, available because `ext.TwoVarComprehensions()` is enabled in `pkg/cel/environment.go`:

```yaml
tags: '${schema.spec.awsTags.transformMapEntry(i, t, {t.key: t.value})}'
```

**Declaring the list shape** requires a named custom type; there is no literal inline-struct type, and `[]struct{key string, value string}` fails with `unknown type: struct{key string, value string}`:

```yaml
schema:
  types:
    Tag:
      key: string
      value: string
  spec:
    awsTags: '[]Tag | default=[{"key":"auto-delete","value":"no"}]'
```

**The constraint worth documenting alongside:** `merge()` is registered map-to-map only (`cel.MemberOverload` in `pkg/cel/library/maps.go`), so it does not apply to a list-shaped field. An author who declares the list shape to avoid sorting then cannot use `merge()` to inject mandatory platform tags. That trade-off is what should drive the choice of shape, and it is currently invisible.

Reasonable home for all of this is a short subsection in `concepts/rgd/04-cel-libraries.md`, or a page under `building-abstractions` if it wants worked examples.

allamand:
Lists derived from maps have non-deterministic element order because map iteration order is not guaranteed. When such a list feeds a template field, the rendered desired state can change between reconciles even when nothing changed, causing kro to re-apply the field and downstream controllers to see a perpetual diff.

Changes:
- Add a warning near the sort()/sortBy() entries in the CEL libraries page with the common ACK map-to-tag-list example.
- Add a cross-referencing note in the transformList section of the CEL expressions page.
- Fix the broken merge() example in the single-resource-RGD guide: merge is a member function, so it must be called as schema.spec.tags.merge({...}), not as a global merge(schema.spec.tags, {...}), which fails to compile.

Fixes #1374

### DIFF
--- a/website/docs/docs/concepts/rgd/04-cel-libraries.md
+++ b/website/docs/docs/concepts/rgd/04-cel-libraries.md
@@ -280,6 +280,27 @@ indices: ${lists.range(schema.spec.count)}
 flat: ${schema.spec.nestedPorts.flatten()}
 ```
 
+:::warning Sort lists built from maps
+A list built from a map has **non-deterministic element order** — the order
+comes from map iteration and can change from one reconcile to the next. When a
+template field is a list of objects derived from a map, call `sort()` on the
+keys first, or the rendered desired state changes between reconciles even when
+nothing has actually changed. kro then re-applies the field every cycle and a
+downstream controller can see a perpetual diff on a resource nobody touched.
+
+This commonly happens when a provider models a key/value collection as a list
+of objects (for example `[{key, value}]`) rather than a map, so converting a
+`map[string]string` schema field into that shape needs a sort to stay stable:
+
+```kro
+# Unstable: element order comes from map iteration
+tags: ${schema.spec.tags.map(k, {"key": k, "value": schema.spec.tags[k]})}
+
+# Stable: sort the keys before building the list
+tags: ${schema.spec.tags.map(k, k).sort().map(k, {"key": k, "value": schema.spec.tags[k]})}
+```
+:::
+
 ### Encoders
 
 Base64 encoding and decoding from [cel-go/ext](https://pkg.go.dev/github.com/google/cel-go/ext#Encoders).
--- a/website/docs/docs/building-abstractions/10-single-resource-rgd.md
+++ b/website/docs/docs/building-abstractions/10-single-resource-rgd.md
@@ -234,15 +234,18 @@ resources:
         name: ${schema.spec.name}
       spec:
         # other required fields omitted for brevity
-        tags: ${merge(schema.spec.tags, {
+        tags: ${schema.spec.tags.merge({
           "managed-by": "kro",
           "tenant": schema.metadata.?labels["tenant"].orValue("shared"),
           "namespace": schema.metadata.namespace
         })}
 ```
 
-Putting the platform tags in the second argument to `merge()` makes them win on
-key conflicts, so the wrapper always injects the tags you require.
+Passing the platform tags as the argument to `merge()` makes them the second
+map, so they win on key conflicts and the wrapper always injects the tags you
+require. Note that `merge()` is a member function, so it must be called on the
+receiving map (`schema.spec.tags.merge({...})`) rather than as a global
+function (`merge(schema.spec.tags, {...})`), which does not compile.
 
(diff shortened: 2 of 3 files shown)


website/docs/docs/concepts/rgd/04-cel-libraries.md
```diff
@@ -280,6 +280,28 @@ indices: ${lists.range(schema.spec.count)}
...
 
+:::warning Sort lists built from maps
+A list built from a map has **non-deterministic element order** — the order
+comes from map iteration and can change from one reconcile to the next. When a
+template field is a list of objects derived from a map, call `sort()` on the
+keys first, or the rendered desired state changes between reconciles even when
+nothing has actually changed. kro then re-applies the field every cycle and a
+downstream controller can see a perpetual diff on a resource nobody touched.
+
+This is common with AWS providers. ACK `ec2.services.k8s.aws` resources
```
jakobmoellerdev: please drop the AWS specific note for vendor neutrality or add other vendors

jakobmoellerdev (commented):
please fixup your commit message
/ok-to-test

website/docs/docs/concepts/rgd/04-cel-libraries.md
allamand: fixed

jakobmoellerdev (approved):
/lgtm

PR #1376, opened 2026-08-21
KREP-025 Cel time library for Kro

NicholasBlaskey:
Describes a potential cel time library for Kro

### DIFF
--- a/docs/design/proposals/time-functions.md
+++ b/docs/design/proposals/time-functions.md
@@ -1,0 +1,295 @@
+# KREP-025: Time Functions for CEL
+
+## Problem statement
+
+An operator written in Go can easily check the current time and decide not to
+make an update because it is outside of a deployment window. It can monitor the
+duration between events and decide to take an action on that.
+
+kro has no way to access time. The workaround today would be to have an external
+system pass in the time through a schema value. This adds complexity and
+potentially causes additional reconciles beyond what is necessary.
+
+kro's CEL environment has limited time helpers. Date and time math is notoriously
+difficult, and without the proper library to handle it, it is basically
+impossible to get correct in CEL.
+
+This primitive is orthogonal to the graph engine design (KREP-024). Both the RGD
+model and the proposed Graph Kind want it, with no major difference in
+implementation or consequences for either.
+
+## Non goals
+
+- Fully designing every possible helper someone might need in the time library.
+  We can iterate on what is needed easily.
+- Stopping the user from making bad decisions. We should make it easy to do the
+  right thing and hard to do the wrong thing, but we aren't aiming to stop
+  people from misusing or abusing the time package.
+
+## Solution
+
+We are proposing a CEL standard library. The most interesting and novel aspect
+of this is the impure section, a single `time.now()` function. It always takes
+one argument.
+
+`time.now(null)` returns the current timestamp. This form will not cause the
+instances of the RGD to have any additional reconciles. There is no guarantee of
+when it will be updated or how stale it can be.
+
+`time.now(evaluateAfter)` also returns the current timestamp, and additionally
+causes the instance to be reconciled after `evaluateAfter`. If the CEL
+evaluation time has already passed `evaluateAfter`, the argument is ignored and
+no special scheduling consideration is given.
+
+Both forms return the standard CEL
+[`timestamp`](https://github.com/google/cel-spec/blob/master/doc/langdef.md)
+type, the same type as `schema.metadata.creationTimestamp`, so the result
+composes with `duration()` arithmetic and the comparison operators used in the
+examples.
+
+## Examples
+
+### Startup grace period
+
+To illustrate how this works, here is a condition that has a 5 minute grace
+period. Suppose we have a component that has a long startup period. As a result
+of this it may be acceptable to ignore issues (for example in a custom status
+condition) for a brief period after the object is created.
+
+```cel
+${time.now(schema.metadata.creationTimestamp + duration("5m"))
+  >= schema.metadata.creationTimestamp + duration("5m")}
+```
+
+### Periodic rollout
+
+A ticker rolls a workload on a fixed interval, for example to restart the pods
+every 10 minutes to work around a memory leak.
+
+The ticker is built by passing `time.now()` into itself with an offset added,
+which schedules the next reconcile one interval out:
+```cel
+${time.now(time.now(null) + duration("10m"))}
+```
+
+To actually roll the pods we put a value that changes once per interval into the
+pod template. A rolling restart happens whenever the pod template changes, so we
+set an annotation to a monotonic bucket number: the current time divided by the
+interval.
+```cel
+# pod template annotation; bumps once per 10-minute bucket and rolls the pods
+spec:
+  template:
+    metadata:
+      annotations:
+        kro.run/restart-bucket: ${string(int(time.now(time.now(null) + duration("10m"))) / 600)}
+```
+
+`int(time.now(null))` is the current time in seconds since the epoch. Dividing by
+the interval in seconds (600 for 10 minutes) gives a counter that increments by
+one every interval and never repeats.
+
+### Certificate renewal
+
+We can include a renewal job that will renew a certificate only if the
+certificate is going to expire within 5 days.
+```cel
+includeWhen:  # renew within 5d of a 90d expiry
+  - ${cel.bind(renewAt,
+        timestamp(credential.spec.renewedTime) + duration("2160h") - duration("120h"),
+        time.now(renewAt) >= renewAt)}
+```
+
+Note that kro can't commit to a strong guarantee of when things get rescheduled.
+It would be a bad idea to try to renew the certificate 5 seconds before it
+expires, because kro may be busy with other work and so on. This strict
+guarantee is nearly impossible in a K8s operator world.
+
+### Business hours
+
+Time window blockers are a good illustration of the power of this primitive.
+
+Suppose you have a job that should run only during 9am-5pm, so the consequences
+of the job failing can be dealt with during business hours. First we define a
+ConfigMap (note this can and should be just kro variables after the variables
+KREP is merged).
+
+```yaml
+# ConfigMap (id: clock) computing the window boundaries
+data:
+  nextWindowOpen: ${cel.bind(now, time.now(null),
+    cel.bind(o, now.withTime({hours: 9}, schema.spec.timezone),
+      cel.bind(nextOpen, now < o ? o : o.addDays(1, schema.spec.timezone),
+        cel.bind(_, time.now(nextOpen), string(nextOpen)))))}
+  nextWindowClose: ${cel.bind(now, time.now(null),
+    cel.bind(c, now.withTime({hours: 17}, schema.spec.timezone),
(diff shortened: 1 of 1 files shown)


docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+
+We are proposing a CEL standard library. The most interesting and novel aspect
+of this is the impure section, a single `time.now()` function. It always takes
+one argument.
+
+`time.now(null)` returns the current timestamp. This form will not cause the
+instances of the RGD to have any additional reconciles. There is no guarantee of
+when it will be updated or how stale it can be.
+
+`time.now(evaluateAfter)` also returns the current timestamp, and additionally
```
jakobmoellerdev, line 39: this function is steering kro behavior in a way that no other function can right now. I really think if you want to introduce reconcile steering behavior it should be its own function. Additionally because its within time.now(x) a function description like (time.now(duration("5m")) would not indicate that there would be a requeue attempt in earliest 5m (its not exactly 5 as we still have to respect queue priority and fairness)

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+requeue period from 10 seconds to 10 hours should not change how user RGDs
+mostly behave.
+
+### decouple requeue and time.now()
+
+An alternative design could be having `time.now` and a separate
+`time.requeueAfter` to give finer-grained control.
+
+The issue with this solution is it allows RGD authors to easily ignore the
+concept of requeuing. Time becomes a potential for hard-to-debug unexpected
```
jakobmoellerdev, line 243: You write before

> Stopping the user from making bad decisions. We should make it easy to do the
  right thing and hard to do the wrong thing, but we aren't aiming to stop
  people from misusing or abusing the time package.

I think your call here to bring time and requeue together like this is gonna lead people to make bad decisions on how to write their expressions.

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+center.
+
+### time.after(evaluateTime)
+
+This was my initial thinking for `time.now()`. `time.after` would return false
+before the time specified then true afterwards. It would requeue after the
+evaluation time.
+
+It is a hard-to-misuse primitive but lacks power. The ability to pass
+`time.now()` to itself gives the ability to build tickers and on-off cycles like
```
jakobmoellerdev, line 255: these tickers are non-obvious behavior and the description gives me a headache when I think how to explain it to others. I really think we need to reconsider more expressive syntax for such functionalities.

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+those specified in the examples.
+
+### Offset instead of a moment
+
+`time.now` could take an offset such as `duration("5m")` as its argument to know
+when to requeue, rather than an absolute instant. This makes some things easier
+to represent; the ticker becomes more obvious. It makes others more difficult to
+represent, like the time window blockers.
```
jakobmoellerdev, line 263: whats the downside? if there is none, the proposal should probably switch to offsets

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+## Alternatives
+
+### Not having time in CEL
+
+Time adds potential for tons of foot guns and complexity. We remove an
+assumption that we can calculate everything based on just inputs to the graph.
+
+Most of these examples can be solved with another operator or a cronjob or some
+other resource. While this is true it ends up being workarounds for kro's lack
+of feature support. kro becomes more powerful with the option of using this.
```
jakobmoellerdev, line 218: I think we can argue whether KROs job is to be come a cron master.
More powerful is not always good.

All controllers in kubernetes are written for eventual consistency. thus, for things like rollout or maintenance window control it has become the norm to define windows and work with queues. (as you state yourself in https://github.com/kubernetes-sigs/kro/pull/1376/changes#diff-5db7bf89655a5f1a7ed0cd7ec6dca5a4835ca6ccfbd41dedcb3b6a15ed93e55bR228)

However, I think it is perfectly possible to reconsider this KREP here and think about how to actually control gating and reconcile behavior with other mechanisms than the ones listed here. For example, we have not considered adding a separate control gate (such as what we did with readyWhen/includeWhen)

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+`EarliestRequeue()` and the controller translates into a `RequeueAfter` on the
+reconcile result.
+
+If the controller restarts we lose all the information about when to requeue.
+This is ok because we will reconcile every instance again and compute when they
+next need to requeue.
+
+## Limitations
+
+- Can't stop infinite reconciles. A user can easily write an RGD that requeues
```
jakobmoellerdev, line 188: If you say you can easily "break the law", This is breaking the proposals own non-goal IMHO:

> [...] We should make it easy to do the
  right thing and hard to do the wrong thing, but we aren't aiming to stop
  people from misusing or abusing the time package.

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Certificate renewal
+
+We can include a renewal job that will renew a certificate only if the
+certificate is going to expire within 5 days.
+```cel
+includeWhen:  # renew within 5d of a 90d expiry
+  - ${cel.bind(renewAt,
+        timestamp(credential.spec.renewedTime) + duration("2160h") - duration("120h"),
+        time.now(renewAt) >= renewAt)}
```
jakobmoellerdev, line 100: the time now function would cause a requeue at renewAt as per https://github.com/kubernetes-sigs/kro/pull/1376/changes#diff-5db7bf89655a5f1a7ed0cd7ec6dca5a4835ca6ccfbd41dedcb3b6a15ed93e55bR39

jakobmoellerdev (commented):
While I in general support supporting time-gated behavior in KRO, I really believe this KREPs scope is much larger than it seems as it is effectively also a KREP for gating reconcile requeue behavior with CEL expressions, similar in intent to the `omit` sentinel but with even more loose characteristics. I'm very cautious to accept the KREP as is for 3 reasons:

- It needs to be gated by default and define graduation criteria (easy fix)
- It needs to define very clearly how to ensure fairness and showcase its mapping on the queue priority of instances
- It needs to honestly probably have a better language expression definition, because the current `time.now(requeueAt)` semantics are not easy enough to understand for me. I believe kro's functions should be easier to understand and reason about

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
+# KREP-025: Time Functions for CEL
```
ellistarn-ant, line 1: Wanted to make you aware of the work I did on time for Graph: https://github.com/ellistarn/kro/blob/krocodile/experimental/docs/design/001-graph.md#time. You might find some useful use cases or mental models in there.

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+  right thing and hard to do the wrong thing, but we aren't aiming to stop
+  people from misusing or abusing the time package.
+
+## Solution
+
+We are proposing a CEL standard library. The most interesting and novel aspect
+of this is the impure section, a single `time.now()` function. It always takes
+one argument.
+
+`time.now(null)` returns the current timestamp. This form will not cause the
```
ellistarn-ant, line 35: why null vs overloading w/ time.now() ?

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Non goals
+
+- Fully designing every possible helper someone might need in the time library.
+  We can iterate on what is needed easily.
+- Stopping the user from making bad decisions. We should make it easy to do the
+  right thing and hard to do the wrong thing, but we aren't aiming to stop
+  people from misusing or abusing the time package.
+
+## Solution
```
ellistarn-ant, line 29: When I worked through this problem, I found that all I needed was a single time.now() primitive, plus the ability to do arithmetic (e.g. time.now() + duration(...)). The semantics I chose was that the evaluation of any expression that results in a `time` type has an implicit requeue behavior.

e.g. time.Now() > timestamp(string) would do the subtraction and requeue at the future time it might flip.

---

Semantically, you are essentially signaling to KRO that the expression has meaning at a specific point in time, and that the graph should be evaluated then. This works across complex expressions like addition/substraction/comparison, etc.

docs/design/proposals/time-functions.md
ellistarn-ant: I think what you have here is close, but I am not convinced by the `evaluateAfter` choice vs just doing `time.now() + duration('5s')` and triggering an evaluation at that time.

NicholasBlaskey: I wrote a section on the solver as an alternative
https://github.com/kubernetes-sigs/kro/pull/1376/changes/BASE..efc16b3d420e6c8c0eeab1bfd6562fd07fb21c2b#diff-5db7bf89655a5f1a7ed0cd7ec6dca5a4835ca6ccfbd41dedcb3b6a15ed93e55bR264-R295

Curious if you have any thoughts on that section

docs/design/proposals/time-functions.md
```diff
@@ -0,0 +1,295 @@
...
+requeue at. It's clever and would be a magical, great user experience if it
+could work.
+
+There is nothing to solve for in the example of setting an environment variable
+to `time.now()`. When should this be requeued? A solver would say never. The
+proposed design is asking the user to explicitly decide. Making the user make an
+explicit choice avoids a false sense of security that we are able to figure out
+what users intended from their RGD.
+
+Solving is complicated. Every function a user passes into time needs a backwards
```
ellistarn-ant, line 281: I had a fairly simple rule for this.

time.now() returns a special type whose operators enqueue the graph. This is essentially "time overloads the operators" from C if you are familiar. So `<` `>` `==` all know how to do the relevant math and enqueue the future reconcile. If you have a function like `someFunction(now())`, it's inert until it causes the comparison ` > x`.

docs/design/proposals/time-functions.md
ellistarn-ant: I think it's workable. Left a comment.

docs/design/proposals/time-functions.md
ellistarn-ant: If you want to do manipulation to the time primitive in another cel function, that function would necessarily need to handle the enqueue itself.

ellistarn-ant:
> @jakobmoellerdev @ellistarn-ant Rewrote the KREP to use a new solver based approach.

Awesome!

jakobmoellerdev (commented):
looks much better than previous iterations. do you have a working branch or something to share here to get a feeling on how this could work?

NicholasBlaskey:
@jakobmoellerdev wrote a draft implementation up 
https://github.com/kubernetes-sigs/kro/pull/1434

jakobmoellerdev (approved):
this is a good interface now. probably needs a bit of tuning here and there for QOL settings and other implications but we can cover that with testing and docs. generally approved from my side.

looking for another lgtm

cheeseandcereal:
/assign
I will take a look at this next week!

PR #1379, opened 2026-08-22
fix(simpleschema): give a clear error for OpenAPI-shaped fields

allamand:
When a SimpleSchema field is written as a nested OpenAPI/structural schema instead of the inline string form, kro treated the mapping as an inline struct and eventually surfaced a confusing error such as "unknown type: no" from two levels deeper (the default value read as a field type).

Detect a nested mapping that looks like an OpenAPI schema (a "type" key where every key is an OpenAPI keyword) and return an actionable error that names the field and suggests the inline form, e.g. 'map[string]string | default={"key":"value"}'. A genuine inline struct that merely contains a field named "type" or "default" is not affected.

Adds unit tests for the new error, the detection heuristic, and the false-positive guard.

### DIFF
--- a/pkg/simpleschema/transform.go
+++ b/pkg/simpleschema/transform.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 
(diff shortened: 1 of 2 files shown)


pkg/simpleschema/transform.go
```diff
@@ -171,12 +172,63 @@ func (t *transformer) buildFieldSchema(name string, spec interface{}, parent *ex
 	case string:
 		return t.buildFieldFromString(name, val, parent)
 	case map[string]interface{}:
+		// A nested mapping is treated as an inline struct. If it instead looks
+		// like a hand-written OpenAPI/structural schema, surface an actionable
+		// error here rather than letting an OpenAPI keyword or a default value
+		// be misinterpreted as a field type two levels down (e.g. the
+		// confusing "unknown type: no").
+		if looksLikeOpenAPISchema(val) {
```
jakobmoellerdev, line 180: while the idea here is nice this heuristic is really not the way to solve this I believe.

PR #1384, opened 2026-08-26
chore: update e2e test for termination and fix ALLOW_CRD_DELETION correctly

Skarlso:
When I run these tests locally, these always failed with timeouts.

```
    --- FAIL: chainsaw/check-instance-creation (43.83s)
    --- FAIL: chainsaw/check-instance-update (46.57s)
    --- FAIL: chainsaw/check-multi-resource-rgd (57.83s)
    --- FAIL: chainsaw/check-rgd-deletion (74.13s)
FAIL
Tests Summary...
- Passed  tests 7
- Failed  tests 4
```

Eventually, I tracked it down to two things. One: chainsaw setting that increases the deletion timeout.
Two: reducing the deletion grace period for the deployments.

These together lead to a consistent test passing on each run:

```
PASS
Tests Summary...
- Passed  tests 11
- Failed  tests 0
- Skipped tests 0
Done.
```

Also, incidentally, I found that `make deploy-kind-helm ALLOW_CRD_DELETION=false` actually didn't work. :)

### DIFF
--- a/.chainsaw.yaml
+++ b/.chainsaw.yaml
@@ -1,0 +1,15 @@
+# Instance deletion in Kro is behind a finalizer flag. Deletion remove the
+# finalizer only once every child resource is gone. A Delete, therefore,
+# together with the terminationGracePeriodSeconds can take longer then that
+# on a machine with cold kind nodes. Here, we set the cleanup to longer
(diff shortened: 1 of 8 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1386, opened 2026-08-30
fix(cli): generate valid CRD manifests

immanuwell:
`kro generate crd` emitted YAML with `typemeta` and `objectmeta`. 
Both YAML and JSON also missed `apiVersion` and `kind`, so `kubectl` rejected the output

This sets the CRD GVK and uses Kubernetes-aware YAML marshaling. 
Small fix, but generated manifests work now

Repro before this fix:

```sh
kro generate crd -f rgd.yaml > generated.yaml
kubectl apply -f generated.yaml
```

Result: `apiVersion not set, kind not set`.

Test: `cd cmd/kro && go test -race ./...`

### DIFF
--- a/cmd/kro/commands/generate/crd.go
+++ b/cmd/kro/commands/generate/crd.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/kubernetes-sigs/kro/api/v1alpha1"
 	"github.com/spf13/cobra"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/kro/commands/generate/crd.go
```diff
@@ -71,3 +72,8 @@ func generateCRD(rgd *v1alpha1.ResourceGraphDefinition) error {
 
 	return nil
 }
+
+func prepareCRDForOutput(crd *extv1.CustomResourceDefinition) {
+	crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
+	crd.SetAnnotations(map[string]string{"kro.run/cli-version": "dev"})
```
Skarlso, line 78: Hmmm, not sure about this dev version here. 🤔 Shouldn't this be overridable?

PR #1392, opened 2026-09-04
chore(fix): encode long instance-names and refactor node-id hashing

Skarlso:
# Description

Graph PR changed how node-ids are actually encoded. This PR now is now threading in the event recorder in to graph and updating the docs about the node-id hashing mechanism.

It also fixes an issue with instance-names which can overflow in the same way as node-ids.

# Relevant Issue

Part of https://github.com/kubernetes-sigs/kro/issues/1055.

### DIFF
--- a/website/docs/docs/concepts/15-instances.md
+++ b/website/docs/docs/concepts/15-instances.md
@@ -102,18 +102,18 @@ Resources created by kro (Deployments, Services, ConfigMaps, etc.) receive label
 
 **Labels:**
 
-| Label | Description |
-|-------|-------------|
-| `kro.run/owned` | Set to `"true"` to indicate kro manages this resource |
-| `kro.run/kro-version` | Version of kro managing the resource |
-| `kro.run/instance-id` | UID of the instance that created this resource |
-| `kro.run/instance-name` | Name of the instance |
-| `kro.run/instance-namespace` | Namespace of the instance (only for namespaced instances) |
-| `kro.run/instance-group` | API group of the instance |
-| `kro.run/instance-version` | API version of the instance |
-| `kro.run/instance-kind` | Kind of the instance |
-| `app.kubernetes.io/managed-by` | Set to `"kro"` |
-| `kro.run/node-id` | Resource ID from the RGD |
+| Label                            | Description                                                                                    |
+|----------------------------------|------------------------------------------------------------------------------------------------|
+| `kro.run/owned`                  | Set to `"true"` to indicate kro manages this resource                                          |
+| `kro.run/kro-version`            | Version of kro managing the resource                                                           |
+| `kro.run/instance-id`            | UID of the instance that created this resource                                                 |
+| `kro.run/instance-name`          | Name of the instance                                                                           |
+| `kro.run/instance-namespace`     | Namespace of the instance (only for namespaced instances)                                      |
+| `kro.run/instance-group`         | API group of the instance                                                                      |
+| `kro.run/instance-version`       | API version of the instance                                                                    |
+| `kro.run/instance-kind`          | Kind of the instance                                                                           |
+| `app.kubernetes.io/managed-by`   | Set to `"kro"`                                                                                 |
+| `kro.run/node-id`                | Label-safe token for the resource ID from the RGD (see [long resource IDs](#long-resource-ids))|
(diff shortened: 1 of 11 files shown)


pkg/metadata/labelvalue.go
```diff
@@ -0,0 +1,61 @@
...
+// truncated prefix joined to a fnv32a hash of the whole input when it does not.
+//
+// This is used conditionally on purpose. It will leave labels that already work
+// alone, preventing a possible backwards compatibility issue, but will encode
+// values that are too long, keeping that from breaking the flow.
+func EncodeLabelValue(s string) string {
+	// no-op if overflow is satisfied
+	if !LabelValueOverflows(s) {
+		return s
+	}
```
tomasaschan: This handles presumptive label values that are invalid because of length - but there are a bunch of other requirements on label values too. Should we use this mechanism to handle all of them equally?

If so, we also need to stop including _any part_ of the original value below.

Skarlso: hmmmmmmm maybe. I'm open to a general validation here. 🤔

Skarlso: Thinking about this a bit more, I would be reluctant to introduce that in this PR though to not further complicate things. WDYT?

tomasaschan: I think as long as whatever approach we use to guarantee the replacement value is valid is backwards compatible (i.e. if we later add full validation + cleanup, values that were already valid except for their length get treated the same as today) there's very little risk in doing it in steps as needed.

The current approach works for that - values that need more than shortening won't be valid after this anyway, so there's no risk of having something valid change behavior if we later amend to full validation. But we do have to be a little mindful about it then, so it _might_ be easier to just rip the bandaid now and not have to worry about it later.

Skarlso: That's a solid argument.

Skarlso: @tomasaschan Actually. Sorry. :D By the time we reach this part of the code a validity check has already been done in some form for valid characters. So there is nothing else that can be done here. :) Especially since validity for characters is encoded in the api and in validation.go as well 
~~~go
lowerCamelCaseRegex = regexp.MustCompile(`^[a-z][a-zA-Z0-9]*$`)
~~~
 (validation.go:43)

So not sure there is anything else here that we can do with regards to cleanup and things like that. :)

website/docs/docs/concepts/15-instances.md
```diff
@@ -123,11 +123,78 @@ Resources created by kro (Deployments, Services, ConfigMaps, etc.) receive label
...
+kro emits an `InstanceLabelEncoded` warning event on the instance naming the
+value any selector needs.
+
+:::tip
+`kro.run/instance-id` holds the instance UID, which always fits in a label
+value and is never hashed. Prefer it when you need a selector that identifies
+one specific instance:
+
+```bash
+kubectl get all -l kro.run/instance-id=<uid>
```
tomasaschan: nit: `get all` here is a little misleading, as despite its name it's not really everything - just everything that Kubernetes decides to add to the `all` category... Many users reading this when figuring out how to find custom resources instantiated as part of a resource graph are going to think the label filter isn't working, when in fact it's the resource type that isn't matching. Maybe we can use a more specific example, that make it more obvious to replace `all` (or whatever is there instead) with the resource type(s) one is actually looking for?

tomasaschan (commented):
I haven't dug through the test code in detail, but I think the prod code paths look reasonable.

I'm not 100% convinced that emitting an event is the right mechanism to signal this, though; it'll be pretty noisy with information that is effectively constant, since it emits an event every time the resource is reconciled, not just when it's created.

website/docs/docs/concepts/15-instances.md
Skarlso: That is a fair point! I will add something specific instead.

tomasaschan (commented):
/lgtm

jakobmoellerdev:
> I mean, the fact that hashed values are very obviously hashed, could well be enough for users to look elsewhere. Paired with good docs (which you've already included! 👍) and a release note about it I think it'll be fine.
> 
> Does kro have a system for installation wide feature flags? I'm NOT suggesting we build one for this, but it might be a good candidate to jack into one if it already exists.

we support the kubernetes feature flag system on the controllers, see https://github.com/kubernetes-sigs/kro/tree/main/pkg/features

jakobmoellerdev (approved):
/lgtm

PR #1396, opened 2026-09-05
Mention allow-breaking-changes annotation in CRD update error

### LINKED ISSUE #1395: Breaking-change error should mention kro.run/allow-breaking-changes instead of implying a recreate
### Description

When KRO blocks a breaking CRD schema update, the error does not mention `kro.run/allow-breaking-changes`, the annotation that exists specifically to allow it. Operators reasonably conclude the only way forward is to delete and recreate the RGD, which is far more disruptive and can be genuinely unsafe on a live cluster.

**Observed**

`pkg/client/crd.go:172`:

```go
return fmt.Errorf("cannot update CRD %s: breaking changes detected: %s", desired.Name, report)
```

Surfaced on the RGD as `KindReady: cannot update CRD <name>: breaking changes detected: MinLength constraint 1 was added; ...`.

**Expected**

The message names the supported escape hatch, e.g.:

```
cannot update CRD <name>: breaking changes detected: <report>.
Set the kro.run/allow-breaking-changes annotation to "true" on the
ResourceGraphDefinition to apply it; this can invalidate existing instances that
violate the new schema.
```

**Why it matters**

We tightened some string fields with `minLength=1` and the RGD went terminal on a long-lived cluster. Reading only the error, the apparent options were "fix the schema" or "recreate the RGD" — and the RGD in question drives per-AZ `ENIConfig`, so recreating it risks pod IP allocation. The annotation was already supported in the version we run (v0.9.3) and made it a one-line, deliberate change instead. We only found it by reading the source and #186.

For comparison, the annotation is well documented in the code itself (`api/v1alpha1/groupversion_info.go`), including its caveat — that text just never reaches the operator who hits the error.

Related: #186 (introduced the detection and the annotation), #428 (a user hitting this class and ending up with wedged finalizers).

pujitha24:
Motivation:
When kro blocks a breaking CRD schema update, the error surfaced on the
ResourceGraphDefinition only said "breaking changes detected: <report>"
and never mentioned kro.run/allow-breaking-changes, the annotation that
exists specifically to let an operator apply the change deliberately.
Reading only the error, an operator could reasonably conclude the only
way forward is to delete and recreate the RGD, which is far more
disruptive and can be unsafe on a live cluster (e.g. RGDs that manage
per-AZ resources).

Approach:
Extend the error returned from CRDWrapper.Ensure in pkg/client/crd.go
to name the v1alpha1.AllowBreakingChangesAnnotation constant and note
its caveat, so the escape hatch is visible directly on the RGD's
condition instead of only in source comments.

Validation:
- go build ./... passes.
- go test ./pkg/client/... -v: all tests pass, including a new
  TestEnsureBreakingChangeErrorMentionsAnnotation that triggers a real
  breaking change (version.served true->false) and asserts the error
  contains the annotation name. Verified this test fails on the
  pre-change code and passes after the change.
- go vet ./pkg/client/... is clean.
- golangci-lint run ./pkg/client/... (v2.12.2, the version pinned in
  this repo's Makefile) reports 0 issues.
- Confirmed this branch's HEAD already contains upstream/main's tip,
  and upstream/main's 5 most recent CI runs are all green.

This is a copy change only; it does not alter when breaking changes
are blocked or allowed, only the text of the error message.

Report: https://github.com/kubernetes-sigs/kro/issues/1395
Assisted-by: claude-sonnet-5 (via Claude Code)

---

Fixes #1395

```release-note
Mention allow-breaking-changes annotation in CRD update error
```

### DIFF
--- a/pkg/client/crd.go
+++ b/pkg/client/crd.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	logr "sigs.k8s.io/controller-runtime/pkg/log"
 
+	"github.com/kubernetes-sigs/kro/api/v1alpha1"
 	crdcompat "github.com/kubernetes-sigs/kro/pkg/graph/crd/compat"
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
 )
@@ -171,7 +172,12 @@ func (w *CRDWrapper) Ensure(ctx context.Context, desired v1.CustomResourceDefini
 		if !report.IsCompatible() {
 			log.Info("Breaking changes detected in CRD update", "name", desired.Name, "breakingChanges", len(report.BreakingChanges), "summary", report)
 			if !allowBreakingChanges {
-				return fmt.Errorf("cannot update CRD %s: breaking changes detected: %s", desired.Name, report)
+				return fmt.Errorf(
+					"cannot update CRD %s: breaking changes detected: %s. "+
+						"Set the %s annotation to \"true\" on the ResourceGraphDefinition "+
+						"to apply it; this can invalidate existing instances that violate the new schema",
+					desired.Name, report, v1alpha1.AllowBreakingChangesAnnotation,
+				)
 			}
 			log.Info("Allowing breaking changes", "name", desired.Name)
 		}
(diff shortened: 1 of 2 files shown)


pujitha24:
Fair, and noted for next time — I filed this straight off the issue without waiting for a maintainer to weigh in on whether it's worth fixing. Happy to leave it here for review, or close it if you'd rather it wait until there's confirmation on the issue first.

pujitha24:
Fair — closing this. It was a small speculative fix filed without waiting for maintainer confirmation on the underlying issue, same thing you flagged the first time around. Sorry for the noise, won't repeat the pattern here.

PR #1400, opened 2026-09-08
fix: detect enum= and validation= markers added to existing schema fields

### LINKED ISSUE #1391: Enum and CEL validation additions to existing fields aren't detected as CRD schema changes
### Description

**Observed Behavior**:

  When a `ResourceGraphDefinition`'s instance schema has a field with no `enum=` marker (or no `validation=` CEL marker), and a later update adds one of these constraints to that *same, already-existing* field, `kro` accepts the updated RGD (`GraphAccepted: True`, `Ready: True`, no errors or warnings) but never patches the live CRD's `openAPIV3Schema` to add the constraint. The CRD keeps silently accepting values that should now be rejected.

  Root cause (found by reading `pkg/graph/crd/compat/schema.go`): `compareEnumValues()` returns immediately if *either* schema's `Enum` is empty:

  ```go
  func compareEnumValues(path string, oldSchema, newSchema *v1.JSONSchemaProps, result *Report) {
        // Use length checks instead of nil checks
        if len(oldSchema.Enum) == 0 || len(newSchema.Enum) == 0 {
                return
        }
        ...
  }
  ```

  So the "no enum → has enum" transition is never reported as a change, `Report.HasChanges()` comes back false, and the CRD update is skipped entirely. There's also no `compareXValidations`-equivalent function in this file at all, so adding a `validation=` CEL marker to an existing field is never detected as a change either, regardless of nil-transition.

  This looks like the same class of bug as #1088 ("CRD not updated when schema default values change"), which was fixed by explicitly handling the nil/empty-to-non-empty transition for `Default`. `Enum` (and `XValidations`, which isn't compared at all) appear to need the same treatment.

  **Expected Behavior**:

  Adding an `enum=` or `validation=` marker to an existing schema field should be detected as a (non-breaking, since it only restricts newly-written values, similar to how `PatternAdded` is already classified) change, and should cause `kro` to patch the live CRD's `openAPIV3Schema` accordingly.

  No `Instance` is needed to reproduce this — it's observable purely at the CRD level.

  1. Apply an initial RGD with an unconstrained string field:
  ```yaml
  apiVersion: kro.run/v1alpha1
  kind: ResourceGraphDefinition
  metadata:
    name: enum-repro
  spec:
    schema:
      apiVersion: v1alpha1
      kind: EnumRepro
      spec:
        color: string
    resources: []
  ```
  ```
  kubectl apply -f enum-repro.yaml
  kubectl get crd enumrepros.kro.run -o jsonpath='{.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.color}'
  # {"type":"string"}
  ```

  2. Update the *same* field to add an enum:
  ```yaml
  apiVersion: kro.run/v1alpha1
  kind: ResourceGraphDefinition
  metadata:
    name: enum-repro
  spec:
    schema:
      apiVersion: v1alpha1
      kind: EnumRepro
      spec:
        color: 'string | enum="red,green,blue"'
    resources: []
  ```
  ```
  kubectl apply -f enum-repro.yaml
  kubectl get resourcegraphdefinition enum-repro -o jsonpath='{.status.conditions}'
  # GraphAccepted: True, "resource graph and schema are valid"; Ready: True
  kubectl get crd enumrepros.kro.run -o jsonpath='{.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.color}'
  ```

  **Expected**: `{"enum":["red","green","blue"],"type":"string"}`
  **Actual**: `{"type":"string"}` — unchanged. Re-applying the identical manifest again reports `unchanged`, and the CRD's `resourceVersion` never
  moves.

  The same reproduces with a `validation=` CEL marker instead of `enum=` on a field that previously had no `validation=` marker (e.g. `arns:
  '[]string'` → `arns: '[]string | validation="self.all(item, item.matches(...))"'`) — the rule never appears under `x-kubernetes-validations` in
  the live CRD.

  **Versions**:
  - kro version: v0.9.2 (`registry.k8s.io/kro/kro:v0.9.2`)
  - Kubernetes Version (`kubectl version`): Client v1.31.4, Server v1.34.9-eks-bca9cf6

  **Involved Controllers**:
  - N/A — reproduces with kro's built-in RGD/CRD reconciliation alone, no other controllers involved.

  ```
  (none — no error or warning is logged; the controller silently treats the schema as up-to-date)
  ```

  * Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to 
  the original issue to help the community and maintainers prioritize this request

_No response_

dmarquez-splunk:
## Summary

Fixes #1391.

`compareEnumValues` returned immediately whenever either schema's `Enum` was empty, so the "no enum → has enum" transition on an already-existing field was never reported as a change: `Report.HasChanges()` came back `false` and the controller never patched the live CRD's `openAPIV3Schema`, even though the RGD update itself was accepted (`GraphAccepted: True`, `Ready: True`).

There was also no `compareXValidations`-equivalent function at all, so adding a `validation=` CEL marker to an existing field had the same silent-no-op behavior, regardless of the nil/empty transition.

Both transitions are now detected and classified **breaking**, consistent with how `PatternAdded`/`MinLengthAdded`/etc. already treat a new restriction added to an existing field: it doesn't invalidate already-stored objects, but it does restrict what new writes can set the field to, so it requires the `kro.run/allow-breaking-changes` annotation like every other case in this file. The reverse transitions (enum/CEL rule removed entirely) are non-breaking, mirroring `PatternRemoved`.

## Changes

- `pkg/graph/crd/compat/schema.go`: fixed `compareEnumValues`'s nil/empty-transition handling (mirrors the `compareFloatConstraint`/`compareIntConstraint` pattern), and added `compareXValidations`, wired into `compare()`.
- `pkg/graph/crd/compat/changes.go`: added `EnumAdded`, `EnumRemoved`, `XValidationAdded`, `XValidationRemoved` change types and their descriptions.
- `pkg/graph/crd/compat/schema_test.go`, `changes_coverage_test.go`: unit tests for the new transitions, including through the top-level `compare()` entrypoint on nested properties (the issue's exact repro shape).
- `test/integration/suites/core/crd_test.go`: two envtest-backed integration tests reproducing the issue end-to-end — apply an RGD, add an `enum=`/`validation=` marker to an existing field on update, verify the change is blocked without `allow-breaking-changes`, then verify the CRD is actually patched (and the constraint is enforced by the apiserver against a real instance) once allowed.

## Test plan

- [x] `go test ./pkg/graph/crd/compat/...` — all existing and new unit tests pass
- [x] `go test ./pkg/...` — full unit suite passes
- [x] `go vet ./...`, `gofmt -l .` — clean
- [x] `go tool ginkgo --focus "CRD Creation|CRD Breaking Changes|CRD Watch Reconciliation|CRD Ownership Conflicts" ./test/integration/suites/core/...` (envtest) — 13/13 pass, including the two new cases reproducing the issue


test/integration/suites/core/crd_test.go
```diff
@@ -687,6 +687,223 @@ var _ = Describe("CRD", func() {
...
+				}
+				g.Expect(raws).To(ConsistOf(`"red"`, `"green"`, `"blue"`))
+			}, 30*time.Second, 250*time.Millisecond).WithContext(ctx).Should(Succeed())
+
+			// The constraint must actually be enforced by the apiserver now, not merely
+			// present as unread schema text.
+			invalid := newInstance("EnumAdded", "enum-added-invalid", namespace, map[string]any{
+				"color": "purple",
+			})
+			Expect(env.Client.Create(ctx, invalid)).To(HaveOccurred())
```
Skarlso, line 801: This passes on a `NotFound` I believe? That would be bad. Add something like apierrors.IsInvalid(err) or something like that.

Skarlso: Also, I think should be an eventually instead of Expect.

pkg/graph/crd/compat/schema.go
```diff
@@ -247,6 +266,52 @@ func compareEnumValues(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
...
+			result.AddBreakingChange(path+".x-kubernetes-validations", XValidationAdded, "", rule)
+		}
+	}
+}
+
+// toValidationRuleSet converts a list of CEL validation rules to a set of rule expressions.
+func toValidationRuleSet(rules v1.ValidationRules) map[string]bool {
+	set := make(map[string]bool, len(rules))
+	for _, r := range rules {
+		set[r.Rule] = true
```
Skarlso, line 300: What about if the `Message` field changes? You are ignoring that here. Does it matter?

test/integration/suites/core/crd_test.go
```diff
@@ -687,6 +687,223 @@ var _ = Describe("CRD", func() {
...
+			Eventually(func(g Gomega, ctx SpecContext) {
+				err := env.Client.Get(ctx, types.NamespacedName{Name: rgd.Name}, rgd)
+				g.Expect(err).ToNot(HaveOccurred())
+
+				if rgd.Annotations == nil {
+					rgd.Annotations = make(map[string]string)
+				}
+				rgd.Annotations[krov1alpha1.AllowBreakingChangesAnnotation] = "true"
+				rgd.Spec.Schema.Spec = toRawExtension(map[string]any{
+					"color": `string | enum="red,green,blue"`,
```
Skarlso, line 770: Errr. There is something wrong here. you are reassigning your spec to the same value here. There is no update. Your change actually comes from `annotationChangedPredicate` change in the controller because you updated the breaking changes annotation.

Please redesign this test / reasses its validity. And there are several other weird things in here I'm unsure about. :)

Skarlso (changes requested):
I found a few discrepencies in the test cases and a missing question about the messages. :) Otherwise, structurally this looks okay.

PR #1401, opened 2026-09-09
feat(cel): add strings.plural() for English pluralization

cheeseandcereal:
## Summary

Adds a kro CEL library exposing **`strings.plural(string) -> string`**, backed by the same `flect.Pluralize` inflection kro already uses to derive CRD resource names from kinds (`pkg/graph/crd/crd.go`, `pkg/metadata/groupversion.go`, `pkg/controller/resourcegraphdefinition/controller_cleanup.go`).

Motivation: expressing the RGD `schema` -> CRD translation as a graph node needs a way to turn a kind into its plural resource name, e.g.

```yaml
metadata:
  name: ${strings.plural(schema.spec.kind.lowerAscii()) + "." + schema.spec.apiVersion.split("/")[0]}
spec:
  names:
    kind: ${schema.spec.kind}
    plural: ${strings.plural(schema.spec.kind.lowerAscii())}
```

## Behaviour

| Expression | Result |
|---|---|
| `strings.plural('Pod')` | `Pods` |
| `strings.plural('ClusterPolicy')` | `ClusterPolicies` |
| `strings.plural('Ingress')` | `Ingresses` |
| `strings.plural('deployment')` | `deployments` |
| `strings.plural('Endpoints')` | `Endpoints` (already plural; heuristic) |
| `strings.plural('')` | `` (empty) |
| `strings.plural(1)` | compile error (`found no matching overload`) |

- **General purpose**, lowercase and CamelCase input keeps its case. All-caps input is *not* handled well (`POLICY` -> `POLICYs`) and irregular nouns come back in dictionary case (`PERSON` -> `People`), so lowercase first when the case of the input is not under your control.
- **Not trimmed or normalised**: only the last word is inflected, separators are kept (`cluster-policy` -> `cluster-policies`), leading/trailing whitespace stays in place and whitespace-only input returns `""`. Call `.trim()` first on user-supplied values.
- **Lowercase first** to reproduce kro's CRD naming exactly: `strings.plural(kind.lowerAscii())`. The order matters for kinds ending in an upper-case acronym or written in all caps (`ServiceDNS` -> `servicedns` vs `servicednses`; `POLICY` -> `policies` vs `policys`). A contract test pins the CEL recipe to the CRD name produced by `crd.SynthesizeCRD` and the resource of `metadata.GetResourceGraphDefinitionInstanceGVR` across ~25 kinds, and the docs call the ordering out.
- For valid RGD kinds (`^[A-Z][a-zA-Z0-9]*$`) `lowerAscii()` and Go's `strings.ToLower` are equivalent, so the CEL recipe and kro's Go code agree by construction.

## Design notes

- Registered as library **`kro.strings`** so it coexists with cel-go's `ext.Strings()` (`cel.lib.ext.strings`), which also owns the `strings.` namespace via `strings.quote`. cel-go de-duplicates singleton libraries by `LibraryName()`, so a clashing name would be silently dropped; the `kro.` prefix follows `kro.lists` / `kro.omit` / `kro.runtime`. Tests build an env with both libraries and chain them (`strings.plural(k).lowerAscii()`, `strings.quote(strings.plural(k))`), and a `DefaultEnvironment()` test covers the production registration order in both cached base environments.
- The AST inspector derives known functions from `env.Functions()`, so `strings.plural` is recognised as a function (not a resource reference) with no allowlist changes. Tests cover a resource / def node literally named `strings` alongside `strings.plural` and `strings.quote` at the inspector, RGD builder and Graph compiler levels.
- Non-string arguments are rejected at RGD/Graph build time by the type checker (`found no matching overload for 'strings.plural' applied to '(int)'`, also against typed `schema` fields). At runtime, `dyn` values of the wrong type hit cel-go's overload guard (`no such overload: strings.plural(int)`) before the binding runs.
- Only `plural()` is added. `singular()`, whitespace trimming, and refactoring the three existing Go `flect.Pluralize` call sites onto a shared helper are intentionally out of scope.
- No new dependencies (`github.com/gobuffalo/flect` is already direct).

### DIFF
--- a/pkg/cel/library/strings.go
+++ b/pkg/cel/library/strings.go
@@ -1,0 +1,113 @@
+// Copyright 2026 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package library
+
+import (
+	"reflect"
+
+	"github.com/gobuffalo/flect"
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+)
+
+// Strings returns a CEL library that provides kro-specific string functions.
+//
+// The functions live in the `strings.` namespace alongside cel-go's
+// ext.Strings() (which provides strings.quote and the <string>.method
+// helpers). CEL has no real namespaces: both libraries simply register
+// fully-qualified function names, and cel-go de-duplicates libraries by
+// LibraryName(), so the two can coexist in one environment.
+//
+// Library functions:
+//
+// strings.plural(word: string) -> string
+//
+// Returns the English plural form of word, using the same inflection rules
+// kro uses to derive CRD resource names from kinds. The input is not trimmed
+// or otherwise normalised: only the final word is inflected, so separators
+// are kept ('cluster-policy' -> 'cluster-policies') and whitespace-only or
+// symbol-only input yields an empty string. Input that already looks plural
+// is usually returned unchanged, but that detection is heuristic.
+//
+//	strings.plural('Pod')            // 'Pods'
+//	strings.plural('ClusterPolicy')  // 'ClusterPolicies'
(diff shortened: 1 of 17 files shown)


pkg/cel/library/strings.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+func (l *stringsLibrary) ProgramOptions() []cel.ProgramOption {
+	return nil
+}
+
+// stringsPlural implements strings.plural(string) -> string.
+func stringsPlural(value ref.Val) ref.Val {
+	native, err := value.ConvertToNative(reflect.TypeFor[string]())
```
jakobmoellerdev, line 102: I dont think this is the right way to pluralize. pluralization for CRD names should be backed by the OpenAPI Schema

cheeseandcereal:

PR #1402, opened 2026-09-09
docs: document the Graph API and restructure concepts for v0.10.0

cheeseandcereal:
## Summary

End-user documentation for the Graph API (KREP-024, shipped in d2171fc), a restructure of the concepts section so material shared by ResourceGraphDefinitions and Graphs is documented once, an `Upgrading` section for v0.10.0, and a validation pass over every YAML/CEL block in the docs against the code on `main`.

Only the current (`main`) docs are changed. `versioned_docs/` content is untouched apart from a one-line prop change in the five `overview.mdx` files that render `EngineCapabilities` (see Infra).

## What's new

| Area | Pages |
| --- | --- |
| `concepts/00-overview` | Composition in kro: RGD vs Graph, one shared engine (with diagram) |
| `concepts/graph/` | Overview (alpha banner, evaluation model, enablement steps), Nodes, Modifiers, Scopes and Nesting, Status and Lifecycle |
| `upgrading/` | Overview and `v0.10.0`: manual CRD apply, stricter `${` parsing, aggregation-mode `/status` RBAC. Assumes a separate changelog; covers only what an operator must know or do. |
| `api/crds/graph` | Graph CRD reference via the existing `RGDReference` component; `make manifests` now copies `kro.run_graphs.yaml` |
| `examples/graph/` | namespace-decorator, ingress-fanin, coredns, singleton, generated from the CI-compiled `examples/graph/*.yaml` (`rgd.yaml` excluded: it is self-labelled aspirational) |

## Restructure

```
concepts/
  00-overview.md
  expressions/      CEL Expressions, CEL Libraries, Dependencies & Ordering
  reconciliation/   Conditionals, Readiness, Collections, External References
  rgd/              Overview, Schema, Resource Basics, Status Conditions, Static Analysis, Instances
  graph/            Overview, Nodes, Modifiers, Scopes and Nesting, Status and Lifecycle
```

Hoisted pages gained Graph-specific content only where behavior genuinely differs: readiness gating (dependents wait in an RGD, not in a Graph), no `schema` variable, `ref` vs `externalRef`, `forEach` on `patch` nodes, `runtime.*` being RGD-only. Every filename stem is preserved so only parent paths change.

**Redirects are deliberately not in this PR.** With `lastVersion: "0.9.4"`, the old paths are still live routes served from `versioned_docs/version-0.9.4/`, and `@docusaurus/plugin-client-redirects` rejects a `from` that collides with an existing route. The 9-entry redirect map is recorded in `website/README.md` under Versioning as a step for the 0.10.0 docs-cut PR.

## Existing pages

- Feature Gates: `GraphKind` (v0.10.0) and the previously undocumented `InstanceConditionMetrics`
- Controller Tuning: `--graph-concurrent-reconciles`, `--apply-concurrency`, `--cel-cost-limit`, `--controller-namespace`, `--controller-service-account`; soft-not-ready backoff; rate-limiter settings are Helm values, not flags-only
- Metrics: the four `graphengine_*` series and `instance_condition_current_status_seconds`
- Access Control: `/status` subresource in the aggregation examples (was missing, and kro writes status through it), a "granting users access to Graph" section
- Graph Revisions, Instance Deletion, API reference, overview, FAQ

## Validation pass

Every complete RGD, Graph, instance, and manifest in `website/docs` was extracted and checked against the code on `main` with a throwaway harness (not committed): RGDs through `graph.Builder` and the graph-engine `rgdadapter`, Graphs through `compiler.Compile`, and all objects dry-run-created against an envtest kube-apiserver with kro's and the integration suite's ACK CRDs installed. 63/63 pass; the only remaining CEL parse failures are the intentionally invalid "✗" examples.

Fixed along the way:

- CEL that was invalid YAML: unquoted ternaries and map literals (`mapping values are not allowed`) across CEL Expressions, CEL Libraries, Feature Gates, Schema, Status Conditions, Single-Resource RGD
- Optional-chaining comparisons that fail type checking: `.?x != ""`, `.?readyReplicas > 0`, `.?ingress.size()`; now unwrapped with `.orValue()`, and the "returns `null`" explanation corrected on three pages
- Duplicate YAML keys in the ACK controller (`ports`) and CoreDNS (`selector`) examples; the CoreDNS one is also fixed in `examples/kubernetes/core-dns/core-dns.yaml`
- `kind: test` (not UpperCamelCase), an unquoted SimpleSchema `default=` with spaces, a boolean in a string template, `Complete Example` templates missing `metadata`, an instance `apiVersion` without a group
- Stale facts: GraphRevision label is `kro.run/…` not `internal.kro.run/…`; `FAILED` instance state no longer exists; `kubectl get` headers are `STATE`/`True`; underscores in IDs are rejected; reserved resource IDs listed; `validation="<CEL>"` marker documented; markers are space-separated

## Diagrams

Four new components matching the existing `RGDProcessFlow`/`RevisionFlow` conventions: `CompositionFlow` (concepts overview), `GraphProcessFlow` (Graph overview; `RGDProcessFlow` now exports a shared `ProcessFlow`), `ReadinessGating` (Readiness page), `ScopeFrames` (Scopes and Nesting). Reference pages intentionally have none.

## Infra

- `.github/workflows/website-presubmit.yaml`: `npm ci && typecheck && test && build` on PRs touching `website/`. The site has `onBrokenLinks: "throw"`, but until now a broken link was only caught by `deploy-docs.yaml` after merge to `main`.
- `EngineCapabilities` takes a `links` prop and resolves against `useActiveVersion()`. This is why the five `versioned_docs/version-0.9.x/docs/overview.mdx` files each have a one-line change (they pass their own, unchanged paths). It also fixes the pre-existing bug where cards on `/next/docs/overview` linked to the stable version.
- `kro` code-fence highlighter learns `nodes`, `patch`, `ref`, `def`, `graph`, `serviceAccountName`, and the `listType`/`listMapKey`/`validation` markers.
- Latent fixes: four `examples/*/category.json` renamed to `_category_.json` (Docusaurus never read them), empty `examples/azure/` removed, `job-example.md` got an H1.

## Reviewing locally

```bash
cd website && npm ci && npm run build && npm run serve
```

New content is under `/next/`: `/next/docs/concepts/overview`, `/next/docs/concepts/graph/overview`, `/next/docs/concepts/reconciliation/readiness`, `/next/docs/upgrading/v0.10.0`, `/next/api/crds/graph`, `/next/examples/graph`.

## Follow-up (not in this PR)

- 0.10.0 docs cut: add `@docusaurus/plugin-client-redirects` with the map in `website/README.md`


jakobmoellerdev (commented):
## Review

Excellent, high-effort docs PR — the validation pass shows. I verified the technical claims against the code on `main` (metrics, controller flags, feature gates, the Graph CRD, and the compiler/runtime semantics) and built the site locally with the exact presubmit sequence (`npm ci && typecheck && test && build`): **the build is green and `onBrokenLinks: "throw"` does not fire**, so the concepts restructure introduced no broken internal links.

Two things to fix (one blocking), then this is good to go:

**🔴 Blocking** — `concepts/graph/03-modifiers.md`: the support table lists `readyWhen` as **Yes** on `patch` nodes, but the compiler hard-rejects it (`pkg/graphengine/compiler/validation.go:117-118`). Inline comment below.

**🟡 Minor** — `05-status-and-lifecycle.md` reason table omits the reachable `FieldManagerConflict` reason; `upgrading/01-v0.10.0.md` understates a missing-CRD failure. Inline comments below.

Everything else checks out against code:
- **Metrics**: the four `graphengine_*` series and `instance_condition_current_status_seconds` match `pkg/metrics/{watchrouter,instance,duration_collector}.go` exactly (names, types, labels).
- **Controller tuning**: `--graph-concurrent-reconciles`=1, `--apply-concurrency`=20, `--cel-cost-limit`=0, `--controller-namespace/-service-account`, plus rate-limiter/collection/QPS defaults all match `cmd/controller/main.go` and `helm/values.yaml`.
- **Feature gates**: `GraphKind` and `InstanceConditionMetrics` are both Alpha, `Default: false` (`pkg/features/features.go`).
- **Graph CRD reference**: `kro.run/v1alpha1`, `Kind: Graph`, `Namespaced`; the copied `kro.run_graphs.yaml` is byte-identical to `helm/crds/`, so `make manifests` is a no-op.
- **Concept semantics**: node ID grammar `^[A-Za-z][A-Za-z0-9]*$` (underscores rejected), reserved IDs, exactly-one-of node kinds, `ref` vs `externalRef`, `forEach` allowed on `patch`/rejected on `ref`, **no `schema` variable**, **no `runtime.*`** (`WithRuntimeLibrary(false)`), `readyWhen` not gating dependents, and scope capture/shadowing/no-mix — all confirmed against `pkg/graphengine/compiler/` and `.../runtime/`.
- **Restructure**: 16 renames, zero surviving old-path references in live docs/config/sidebars; all relative links and cross-file anchors resolve.

Non-blocking note: the `needs-rebase` label appears stale — `git merge-tree` against current `main` produces no conflicts. A trivial rebase should clear the bot.

website/docs/docs/concepts/graph/03-modifiers.md
```diff
@@ -0,0 +1,80 @@
...
+meaning, in a Graph node and an RGD resource. Their full semantics are
+documented once, under [Reconciliation](../reconciliation/01-conditional-creation.md).
+This page covers only what is specific to Graphs.
+
+## Support by Node Kind
+
+| | `template` | `ref` | `def` | `patch` | `graph` |
+| --- | --- | --- | --- | --- | --- |
+| `includeWhen` | Yes | Yes | Yes | Yes | No |
+| `readyWhen` | Yes | Yes | Yes | Yes | No |
```
jakobmoellerdev: **Blocking:** this cell is wrong — `readyWhen` is **not** supported on `patch` nodes. The compiler rejects it at compile time:

```go
// pkg/graphengine/compiler/validation.go:117-118
if len(n.ReadyWhen) > 0 {
    return fmt.Errorf("readyWhen is not supported on patch nodes: a patch contributes fields to its target and publishes no value into scope, so there is nothing for readyWhen to evaluate; to wait for the target's state, add a ref node for the target and put readyWhen on it")
}
```

The comment at `validation.go:99-101` also states patch may carry `forEach` and `includeWhen` "but not readyWhen", and the prose below (lines 27-29) never lists `readyWhen` for patch — so the table contradicts both the code and its own page.

Please change the `readyWhen` × `patch` cell to **No**, and consider adding the workaround the error message suggests (add a `ref` node for the target and put `readyWhen` on it).

website/docs/docs/concepts/graph/05-status-and-lifecycle.md
```diff
@@ -0,0 +1,186 @@
...
+| `Accepted` | `True` | `Compiled` | The spec is valid: node IDs are unique, expressions parse and type-check, and the dependency graph is acyclic. |
+| `Accepted` | `False` | `InvalidGraph` | The spec failed validation or compilation. The message says why. Nothing is applied. |
+| `ResourcesConverged` | `True` | `Applied` | Every node has been applied and satisfies its `readyWhen` conditions. |
+| `ResourcesConverged` | `False` | `WaitingForReadiness` | Resources were applied, but at least one `readyWhen` condition is not yet true. |
+| `ResourcesConverged` | `False` | `DataPending` | A node references data that does not exist in the cluster yet, so it and its dependents cannot be resolved. |
+| `ResourcesConverged` | `False` | `ApplyFailed` | A resource could not be applied. The message carries the API server error. |
+| `ResourcesConverged` | `False` | `PruneFailed` | A retired resource could not be deleted. |
+| `ResourcesConverged` | `False` | `ReleaseFailed` | A retired `patch` contribution could not be released from its target. |
+| `ResourcesConverged` | `False` | `DeleteFailed` | During Graph deletion, a managed resource or contribution could not be removed. |
```
jakobmoellerdev, line 33: **Minor (completeness):** this reason table reads as exhaustive but omits `FieldManagerConflict`, which is reachable: `ResourcesConverged=False` / reason `FieldManagerConflict` when a field a node wants to write is owned by another field manager kro refuses to steal (`pkg/controller/graph/controller.go:367-368,858-864`; tested in `controller_test.go:696` and `coverage_test.go:120-126`). Worth adding a row (status `False`).

website/docs/docs/upgrading/01-v0.10.0.md
```diff
@@ -0,0 +1,93 @@
...
+Three CRDs change in this release. Apply them before upgrading the controller:
+
+```bash
+kubectl apply --server-side -f https://raw.githubusercontent.com/kubernetes-sigs/kro/v0.10.0/helm/crds/kro.run_resourcegraphdefinitions.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/kubernetes-sigs/kro/v0.10.0/helm/crds/internal.kro.run_graphrevisions.yaml
+kubectl apply --server-side -f https://raw.githubusercontent.com/kubernetes-sigs/kro/v0.10.0/helm/crds/kro.run_graphs.yaml
+```
+
+- `graphs.kro.run` is new. It must exist before the `GraphKind` feature gate is
+  enabled, or the controller will fail to start its Graph watch. Installing it
```
jakobmoellerdev: **Minor (precision):** a missing `graphs.kro.run` CRD with `GraphKind` enabled doesn't just fail "its Graph watch" — the whole controller exits (`cmd/controller/main.go:365-379`, `os.Exit(1)` on `setupGraphController` failure, which includes cache/watch setup). Suggest wording like "…or the controller will fail to start" to reflect the actual failure mode.

website/docs/docs/concepts/graph/05-status-and-lifecycle.md
```diff
@@ -0,0 +1,186 @@
...
+Resources created by a `template` node carry:
+
+| Key | Type | Value |
+| --- | --- | --- |
+| `kro.run/node-id` | label | The node's `id`. For a node inside a `graph` node, the dot-joined path (`backend.deployment`); hashed if longer than 63 characters. |
+| `kro.run/collection-index` | label | Position within a `forEach` expansion. Collection items only. |
+| `kro.run/collection-size` | label | Total item count of the expansion. Collection items only. |
+| `kro.run/instance-id` | label | The Graph's UID. Collection items only. |
+| `internal.kro.run/node-path` | annotation | The full slash-separated node path (`backend/deployment`). |
+| `internal.kro.run/apply-order` | annotation | The node's position in the dependency order. Deletion runs highest first. |
```
Skarlso: I believe this is incorrect. Graphs delete in order of managedResources in reverse.

website/docs/docs/concepts/graph/05-status-and-lifecycle.md
```diff
@@ -0,0 +1,186 @@
...
+deletes it. This happens when a node is removed from the spec, when a
+`forEach` collection shrinks, when an `includeWhen` becomes false, or when a
+resource is renamed.
+
+A resource is never pruned because its node is merely waiting. If a node cannot
+be resolved this reconcile (`DataPending`), the resources it applied previously
+are kept.
+
+Retired `patch` contributions are **released** rather than pruned: kro applies
+an empty payload under the contribution's field manager, which removes the
```
Skarlso, line 128: We apply an identity-only object and not an empty one I believe. 🤔

website/docs/docs/concepts/graph/01-overview.md
```diff
@@ -0,0 +1,215 @@
...
+Graphs are off by default. Enabling them is a cluster-level decision with
+several steps.
+
+1. **Install the Graph CRD.** Helm does not install or update CRDs on
+   `helm upgrade`. Apply it directly:
+
+   ```bash
+   kubectl apply --server-side -f https://raw.githubusercontent.com/kubernetes-sigs/kro/main/helm/crds/kro.run_graphs.yaml
+   ```
+
```
Skarlso, line 184: It does _install_ them from `crds` folder, but it doesn't apply them that's correct.

website/docs/docs/concepts/reconciliation/03-collections.md
```diff
@@ -574,7 +579,8 @@ be required.
 
 Each resource is limited to **10 forEach dimensions** by default. If a resource
 defines more forEach iterators than this limit, validation fails at RGD
-processing time before any resources are created.
+processing time before any resources are created. For a Graph the same limit is
```
Skarlso: Actually don't think we do this. `config.rgd.maxCollectionDimensionSize` is not shared at all.

website/docs/docs/concepts/reconciliation/04-external-references.md
```diff
@@ -339,8 +339,53 @@ For collection external refs, watches also detect:
...
+    def:
+      running: ${allPods.filter(p, p.status.phase == 'Running').size()}
+```
+
+Two capabilities are specific to Graph `ref` nodes:
+
+- **Dynamic types.** `apiVersion` and `kind` may themselves be CEL expressions,
+  so a Graph can read a kind whose type is only known from another node. See
+  [Dynamic Types](../graph/02-nodes.md#dynamic-types).
+- **Missing targets are not errors.** A `ref` whose target does not exist yet
```
Skarlso, line 379: This isn't specific to graphs.

website/docs/docs/concepts/graph/03-modifiers.md
cheeseandcereal: Ah good catch. Updated the table

website/docs/docs/concepts/graph/05-status-and-lifecycle.md
cheeseandcereal: Ah yeah good catch. That was a follow up PR committed after the original docs PR landed but before now. Added that.

website/docs/docs/upgrading/01-v0.10.0.md
cheeseandcereal: Sure makes sense. Updated the wording here. Didn't explicitly say "controller will fail to start" because it actually does run for ~2 minutes before exiting, but regardless updating this wording to be more clear.

website/docs/docs/concepts/graph/01-overview.md
```diff
@@ -0,0 +1,215 @@
...
+   If you deploy kro without the Helm chart, you must also pass
+   `--controller-namespace` and `--controller-service-account` to the
+   controller. With `GraphKind` enabled and either flag missing, the controller
+   refuses to start.
+
+3. **Create the applier ServiceAccount and its RBAC.** The ServiceAccount named
+   in `spec.serviceAccountName` (or the namespace's `default` ServiceAccount)
+   needs permission for every resource type the Graph creates, reads, patches,
+   or deletes, plus `list` and `watch` on anything the Graph reads through a
```
Skarlso: Watches aren't erroring atm they are just logged. And the watch is happening with Kro's identity, right?

website/docs/docs/concepts/graph/01-overview.md
```diff
@@ -0,0 +1,215 @@
...
+## Enabling Graphs
+
+Graphs are off by default. Enabling them is a cluster-level decision with
+several steps.
+
+1. **Install the Graph CRD.** Helm does not install or update CRDs on
+   `helm upgrade`. Apply it directly:
+
+   ```bash
+   kubectl apply --server-side -f https://raw.githubusercontent.com/kubernetes-sigs/kro/main/helm/crds/kro.run_graphs.yaml
```
Skarlso, line 182: This pulls from main. Which is probably fine Just flagging it.

Skarlso (commented):
A few comments so far I found but I'm still reading to verify integrity. :)

website/docs/docs/concepts/graph/01-overview.md
cheeseandcereal: This is specifically the behavior of "helm upgrade". It doesn't "install" new CRDs or "upgrade" existing CRDs when you run helm upgrade, regardless of what's in the `crds` folder, so I think this is accurate. Are you trying to say that 'helm upgrade' will install newly added things into the `crds` folder? Because I don't believe that's correct.

website/docs/docs/concepts/graph/05-status-and-lifecycle.md
cheeseandcereal: I think the code comments for the implementation are a bit weird and are calling an empty object an "identity-only object".

If you take a look: https://github.com/kubernetes-sigs/kro/blob/main/pkg/graphengine/executor/simple.go#L1803

You can see that the actual server-side apply that's happening (the `Patch` call) is applying `obj`, which is set with the code:

https://github.com/kubernetes-sigs/kro/blob/9a84c408bf6a39e83adadbe4da31d5bc8c0c8293/pkg/graphengine/executor/simple.go#L1771-L1775

This _is_ an empty object (only the metadata set to identify the unique object, but no other metadata, spec, etc).

The way that server-side apply works, is that for you to relinquish your field ownership (over any field), you have to make an apply to that object, containing _only_ the fields you currently care about. So, for example, removing 1 field would require you applying every other field except that one.

To remove your ownership from all fields, you apply an empty object. Whether or not this actually translates to fields being removed depends on the ownership of those fields. For example if there was a join owner, it might not change anything on the underlying object, but, you will no longer have field ownership. If there are no other owners, then kubernetes will try to remove that field.

But yeah, I believe the wording here is accurate (and clearer than the code comments in the implementation area for this). Let me know if you think I'm still missing something

website/docs/docs/concepts/reconciliation/04-external-references.md
cheeseandcereal: Are you sure?

This functionality is built into the graph engine, but the way that RGD is adapted/built/converted into a graph node, cel expressions in these fields (if in an RGD definition) are interpreted as literal strings (which maintains exact parity with how it works on 0.9.4 as well):


```
kubectl apply -f - <<'EOF'
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: dynamic-gvk-repro
spec:
  schema:
    apiVersion: v1alpha1
    kind: DynamicGVKRepro
  resources:
    - id: cm
      template:
        apiVersion: ${'v1'}
        kind: ${'ConfigMap'}
        metadata:
          name: dynamic-gvk-repro
EOF

kubectl wait rgd/dynamic-gvk-repro --for=condition=GraphAccepted=False --timeout=60s
kubectl get rgd dynamic-gvk-repro -o jsonpath='{.status.conditions[?(@.type=="GraphAccepted")].message}{"\n"}'

failed to build resource "cm": failed to extract GVK from resource cm: invalid kind "${'ConfigMap'}": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?') (must be a valid DNS-1035 label when lowercased)
```

website/docs/docs/concepts/graph/01-overview.md
cheeseandcereal: I'm open to pinning it to a tagged ref, but I think main might be better anyways, personally.

website/docs/docs/concepts/graph/01-overview.md
cheeseandcereal: > Watches aren't erroring atm they are just logged.

Was there something in here that contradicts that?

> And the watch is happening with Kro's identity, right?

Oh shoot, I think you're right. And I think this is an actual problem we need to fix in the code. We can't/shouldn't be using the controller identity to read the objects created by a graph. It leads to possible trust boundary issues (use graph to read something cross namespace your SA wouldn't have permissions for, somehow get an error with some content or something into your graph, etc), and is also a weird UX for configuring permissions too. I'll add it to the "todo" list...

website/docs/docs/concepts/reconciliation/03-collections.md
cheeseandcereal: The _default_ limit of 10 exists in both RGD and graph. You can see it in graph here: https://github.com/kubernetes-sigs/kro/blob/856829311f00b74a8480f84cdbf0bde8abf3f240/pkg/graphengine/runtime/collection.go#L136-L138

I think the confusion may be coming from the wording below. This limit cannot be adjusted on graph, but can be with `--rgd-max-collection-dimension-size` for RGDs. I can clarify the wording below to indicate that it's just for RGDs.

But the text here about the limit of 10 for both by default is accurate.

I will update the text here to be more clear that these are different.

website/docs/docs/concepts/graph/05-status-and-lifecycle.md
cheeseandcereal: Ah yeah. This is correctly documented on this page in line 141, but this is a vestige from copying over. This annotation is still applied for graphs but is effectively unused by the controller for any real work. I will update this to say it's informational only.

.github/workflows/website-presubmit.yaml
```diff
@@ -0,0 +1,45 @@
+name: Website presubmit
```
jakobmoellerdev, line 1: please drop this in favor of a real presubmit in prow! thanks!

jakobmoellerdev (approved):
/lgtm

PR #1404, opened 2026-09-09
fix(graphengine): retry transient collection update failures instead of tolerating them

cheeseandcereal:
## Problem

When the server-side-apply UPDATE of an already-existing `forEach` collection member is rejected, `applyCollectionItem` tolerates it: the live object is recorded as applied and the node converges. That is right for a permanent rejection (Invalid/BadRequest, e.g. an immutable field), where retrying can never help and one bad member must not wedge the node. But the same path swallowed transient failures — 5xx, timeouts, throttling, 409, a `failurePolicy: Fail` admission webhook whose backend is unreachable — because `classifyRejection`'s permanent/transient flag only fed the log line and the `OnToleratedRejection` hook.

A user sees the instance `ACTIVE` with `ResourcesReady=True "all resources are created and ready"` while one member still holds the old value; a `Warning UpdateRejected` event on the child is the only trace. Nothing is requeued, so the member stays stale until an unrelated event or the multi-hour resync, even after the outage clears. The pre-rewrite engine reported such an update error as `ResourcesReady=False` and converged as soon as the webhook went away.

## Fix

- `pkg/graphengine/executor/simple.go` `applyCollectionItem`: branch on `classifyRejection`'s flag. A permanent rejection takes the existing tolerate path (`recordUpdateRejected`, hook fires, node converges). Anything else keeps the live identity in `Applied` (`recordApplied`, so the object stays in inventory and is never pruned) but is also recorded as an item failure (`recordFailure`), so the node is held not-ready with the cause, dependents gate, the instance stays `IN_PROGRESS`, and the reconcile requeues with the existing capped backoff until the update lands; the hook does not fire. A new envtest spec blocks UPDATEs of one member with a `ValidatingWebhookConfiguration` pointing at an unreachable loopback URL, asserts `IN_PROGRESS` with the webhook error on `ResourcesReady`, then deletes the webhook and asserts convergence without touching the instance.
- `pkg/graphengine/executor/executor.go` `ToleratedRejection`: the `Permanent` field is removed — only permanent rejections reach the hook now, so it would be a constant, and its only consumer (the instance controller's `Warning UpdateRejected` event) never read it.
- `classifyRejection` is unchanged in behaviour; its comment now states why "permanent" is kept narrow (Invalid/BadRequest): misreading a transient cause as permanent tolerates a stale member as converged, while the reverse only costs a visible retry.

## Behaviour changes

- An UPDATE of an existing collection member that fails with anything other than Invalid/BadRequest now holds the instance `IN_PROGRESS` (`ResourcesReady=False`, reason `NotReady`, message naming the item and cause), gates dependents, and is retried with backoff (3 s doubling to a 5 min cap) instead of reporting `ACTIVE` with a stale member. No `Warning UpdateRejected` event is emitted for these anymore; the cause is on the condition.
- A no-change re-apply is an admission-checked UPDATE too: while such a webhook keeps failing, an instance whose members are already at the desired state also flips from `ACTIVE` to `IN_PROGRESS` on its next reconcile, and returns to `ACTIVE` on the first successful re-apply. A scalar template behind the same webhook is already reported this way (hard error, `ERROR`).
- 403 `Forbidden` is on the retried side: a 403-coded admission denial (controller-runtime `admission.Denied`, Gatekeeper, a `ValidatingAdmissionPolicy` validation with `reason: Forbidden`) or missing RBAC now surfaces as not-ready + retry rather than silent convergence, matching the create path and the pre-rewrite engine. A denial that can never pass keeps the instance visibly `IN_PROGRESS` with the denial message rather than falsely `ACTIVE`.
- Unchanged, now documented: a denial the apiserver codes 400 (its default when a webhook returns `allowed: false` without a status code, e.g. Kyverno's default deny) or 422 (a `ValidatingAdmissionPolicy` with the default reason `Invalid`) is classified permanent and still tolerated with the instance `ACTIVE`. The same policy engine's denial therefore lands on either side depending on the status code it sets.
- Permanent rejections of an UPDATE to an existing member are tolerated exactly as before; create failures, scalar templates, terminating objects and field-manager conflicts are unchanged.
- Existing tests that pinned the old behaviour were updated: `TestSimple_ApplyTemplate_CollectionApplyTolerance`'s Conflict subtest now expects `ErrNotReady` instead of success; `TestSimple_ApplyTemplate_LargeCollectionTolerance` expects `20 item(s) failed to apply` (the 10 conflicting updates now count) with `Applied` still 50; `TestSimple_ApplyTemplate_CollectionHardErrors` uses `BadRequest` instead of `Forbidden` for its second permanently-rejected member.

### DIFF
--- a/pkg/controller/instance/controller.go
+++ b/pkg/controller/instance/controller.go
@@ -177,15 +177,14 @@ func NewController(
 		})
 	}
 	// Surface a tolerated collection-update rejection (an already-existing item
-	// whose SSA update was rejected — e.g. an immutable field — that we keep-live
-	// and converge on) as a Warning Event on the affected CHILD object, so the
-	// dropped desired change is visible in `kubectl describe` and not just the
-	// controller log. Observational only: this hook never influences readiness or
-	// requeue (the node still converges), so it cannot reintroduce the wedge an
-	// unfixable update would otherwise cause. Set at construction (not per
-	// reconcile) because the executor is shared across concurrent reconcile
-	// workers; the event subject is derived from the rejection's own target
-	// identity rather than any per-reconcile instance state.
+	// whose SSA update was permanently rejected — e.g. an immutable field — that
+	// we keep-live and converge on) as a Warning Event on the affected CHILD
+	// object, so the dropped desired change is visible in `kubectl describe`.
+	// Observational only: the hook never influences readiness or requeue.
+	// Transient failures never reach it (they hold the node not-ready and their
+	// cause is on the ResourcesReady condition). Set at construction because the
+	// executor is shared across concurrent reconcile workers; the event subject
+	// is derived from the rejection's own target identity.
 	if eventRecorder != nil {
 		exec.OnToleratedRejection = func(r executor.ToleratedRejection) {
 			ref := &corev1.ObjectReference{
--- a/pkg/graphengine/executor/apply_template_test.go
+++ b/pkg/graphengine/executor/apply_template_test.go
@@ -207,12 +207,11 @@ func TestSimple_ApplyTemplate_WatchRegistrationFailureIsSoftFail(t *testing.T) {
 func TestSimple_ApplyTemplate_CollectionApplyTolerance(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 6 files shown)


cheeseandcereal:

PR #1405, opened 2026-09-09
fix(graphengine): fail closed on malformed external-collection selectors

cheeseandcereal:
## Problem

An external collection's `metadata.selector` is schemaless and, since #1230, may be a whole CEL expression such as `selector: ${schema.spec.sel}`, so its shape is first seen at reconcile time. `refCollectionSelector` decoded the rendered value leniently: unknown keys were dropped and a non-object value was treated as absent, so a misspelled `{matchLabel: {tier: db}}`, a bare `{tier: db}` label map, or a plain string became an empty `LabelSelector`, which `LabelSelectorAsSelector` turns into `labels.Everything()`.

The instance then listed every object of the kind in the namespace — or cluster-wide when `metadata.namespace` is omitted — registered a matching all-objects drift watch, reported `ACTIVE`, and published foreign objects into its status. The pre-rewrite engine (`7458023`) decoded strictly and failed the node; the test pinning that was dropped with the rewrite. Graph `ref:` nodes had no compile-time shape check at all, so a literal typo compiled too.

## Fix

- `pkg/graphengine/compiler/selector.go` (new): `DecodeLabelSelector` strictly decodes a selector map (`FromUnstructuredWithValidation` in strict mode) and runs `metav1validation.ValidateLabelSelector` on it; `validateRefSelector` checks a ref node's literal selector at compile time against the `LabelSelector` OpenAPI schema (a standalone `${...}` expression is accepted and left to the runtime decode), mirroring `pkg/graph`'s RGD-level check.
- `pkg/graphengine/compiler/context.go`: `projectPayload` calls `validateRefSelector` on the shared ref projection path, so a malformed literal selector is a compile error (`Accepted=False`) for both static- and dynamic-GVK Graph `ref:` nodes.
- `pkg/graphengine/executor/simple.go` `refCollectionSelector`: an absent selector still selects everything; a present one must be an object and is decoded through `DecodeLabelSelector`. Any malformed shape is a hard error naming the node, raised before the drift watch is registered and the list runs, so the instance reports `ResourcesReady=False` / `state: ERROR` with an actionable message instead of matching everything. Pinned by unit tests at helper and `Apply` level and by an envtest spec in which an instance-supplied `{matchLabel: ...}` selector fails hard and converges once corrected.
- `website/docs/.../05-external-references.md`: a warning that an instance-supplied selector must be a valid `LabelSelector` object and never falls back to matching everything.

## Behaviour changes

- An external collection whose rendered `metadata.selector` is present but malformed (unknown keys, wrong value types, invalid operator or label syntax) now fails the node hard instead of listing everything. Valid selectors, `{}`, `{matchLabels: {}}` and an absent selector behave as before; `ValidateLabelSelector` enforces the same rules `LabelSelectorAsSelector` already did, so no previously accepted selector is rejected. The hard error is still requeued at the flat default interval, but the instance is no longer reported `InProgress` under a never-ending not-ready backoff.
- Unlike the pre-rewrite decoder, a selector that is present but not an object (including `null`) is an error rather than `labels.Everything()`. A collection node always compiles with a non-null selector, so this only occurs when a CEL expression resolves to a non-object value.
- Unlike the pre-rewrite decoder, key matching is case-sensitive: a wrongly-cased key such as `MatchLabels` is an unknown field (the previous `encoding/json` decoder matched field names case-insensitively).
- A Graph `ref:` node with a malformed literal selector is now rejected at compile time. RGDs were already rejected by the `pkg/graph` builder with the same parser and validation, so nothing an RGD author could publish before is newly rejected.

### DIFF
--- a/pkg/graphengine/compiler/context.go
+++ b/pkg/graphengine/compiler/context.go
@@ -220,7 +220,7 @@ func (ctx *CompilationContext) buildDefNode(n *expv1alpha1.Node, order int, payl
 }
 
 func (ctx *CompilationContext) buildNode(p *parser.Parser, n *expv1alpha1.Node, order int) (*Node, *spec.Schema, error) {
-	kind, payload, err := projectPayload(n)
+	kind, payload, err := projectPayload(p, n)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -589,8 +589,9 @@ func validateRefMetadata(payload map[string]any) error {
 }
 
 // projectPayload converts the discriminated-union API node into a single
-// unstructured map suitable for CEL extraction.
-func projectPayload(n *expv1alpha1.Node) (NodeKind, map[string]any, error) {
+// unstructured map suitable for CEL extraction. Ref metadata contracts are
+// checked here so they cover both the static and dynamic-GVK ref paths.
+func projectPayload(p *parser.Parser, n *expv1alpha1.Node) (NodeKind, map[string]any, error) {
 	switch {
 	case n.Template != nil:
 		obj, err := unmarshalRaw(n.Template.Raw)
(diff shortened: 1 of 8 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1406, opened 2026-09-09
fix(graphengine): restore RGD validator / compiler / runtime parity

cheeseandcereal:
## Problem

On the RGD path the RGD reconciler still validates with `pkg/graph`, while instances are compiled and executed by the graphengine compiler/runtime. Wherever the two disagree, the RGD becomes `Active` and every instance goes `ERROR`, with nothing on the RGD hinting at the cause. Four such divergences exist, all regressions against the pre-rewrite engine (`7458023`):

- `graphengine` is reserved as a node id by the compiler and, to keep the two lists in sync, by the RGD validator; an existing RGD with `id: graphengine` flips to `Inactive` on upgrade and its instances freeze.
- The RGD validator accepts `optional<bool>` conditions such as `readyWhen: ["${cm.?immutable}"]` or `includeWhen: ["${schema.spec.?enabled}"]`, but the compiler rejects them and the runtime hard-errors on the `nil` an empty optional evaluates to (the old runtime read it as `false`).
- The runtime caps forEach axes at the package constant (10) regardless of `--rgd-max-collection-dimension-size`: with the flag at 12 an 11-axis RGD is `Active` while every instance fails with `exceeds the maximum of 10`.
- The adapter seeds the `schema` node as a schema-typed CEL value, but the executor republishes it through `Runtime.Set`, which only wrapped Template/Ref nodes, so `schema` becomes a raw map and `${schema.metadata.creationTimestamp.getFullYear()}` fails with `no such overload getFullYear(string)`.

## Fix

- `pkg/graph/validation.go`, `pkg/graphengine/compiler/validation.go`: `graphengine` is removed from both reserved-id lists; nothing references such an identifier.
- `pkg/graphengine/compiler/typecheck.go` `validateAndCompileConditions`: accepts `bool`, `optional<bool>` (the same `conversion.IsBoolOrOptionalBool` check the RGD validator uses) or `dyn`; other types are still rejected. `pkg/graphengine/runtime/node.go` `conditionResult`: an empty optional (`nil`, or `sentinels.Omit` under `CELOmitFunction`) reads as `false` for includeWhen and readyWhen, singleton and collection.
- `pkg/graphengine/runtime`: per-Runtime `maxCollectionDimensions` (`WithMaxCollectionDimensions`, `<= 0` keeps the default) enforced by `cartesianProduct`, wired from `--rgd-max-collection-dimension-size` through the instance controller, the RGD reconciler, the Graph controller and subgraph child runtimes. An envtest spec with 11 axes (the integration environment's RGD-level cap is 11) pins the RGD reconciler → runtime plumbing.
- `pkg/graphengine/runtime/runtime.go` `Set`: also wraps a Def node carrying an `objectOverride` — the adapter's `schema` node, whose schema is declared rather than inferred — so the published value keeps the seed's typing. `rgdadapter.BuildRuntimeForInstance` registers the same override as the cached path, and the instance controller's pre-apply `candidateMetadata` drops its now-redundant re-wrap.
- Comment corrections: references to the removed `ProjectInstanceStatus` point at the synthesized author-status patch node; `patchFieldManager` notes that `graphSeg` is a constant on the RGD/instance path; `reconcileApplySetInventory` no longer claims to mirror the deletion-only `applyset.Project`; `.golangci.yml` no longer points at a nonexistent file.

## Behaviour changes

- `graphengine` is usable as a resource/node id again; an RGD that was `Inactive` because of it returns to `Active` on its next reconcile.
- `optional<bool>` includeWhen/readyWhen expressions compile on the instance path; an empty optional means "not included" / "not ready". A CEL `null` result (reachable only from `dyn`-typed conditions) reads as `false` the same way — the old engine did too; on `main` it was a hard `want bool` error.
- The runtime's forEach axis cap follows `--rgd-max-collection-dimension-size` (default 10, so unchanged unless the flag is set); standalone Graphs get the same cap from the same flag.
- `${schema.*}` stays typed by the RGD's declared schema plus ObjectMeta for the whole reconcile, as on the old engine; fields without a declared schema still resolve to native values as before.
- Flipped assertions: `TestIsKROReservedWord` now expects `graphengine` → unreserved; the two compiler cases that pinned the `optional<bool>` rejection now assert acceptance. The integration environment's RGD-level dimension cap is 11 instead of 10; no existing spec asserted a cap rejection.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,15 +39,15 @@ linters:
           - unparam
           - gocyclo
         path: (_test\.go|^test/integration/.*\.go|^pkg/testutil/.*\.go)$
-      # Pre-existing staticcheck findings deferred to a dedicated cleanup PR
-      # (tracked in docs/pr-1355-open-todos.md). SA1019 deprecated-API
-      # migrations (scheme.Builder, client.Apply→SSA ApplyConfiguration,
-      # GetEventRecorderFor, sets.String, FieldsV1.Raw) change behavior/ripple
-      # through generated scaffolding, and two exported-symbol renames (ST1003:
-      # V1ApplySetIdFormat, ApiExtensionsClient) ripple across consumers — both
-      # are out of scope here. staticcheck is otherwise fully enabled (it was
-      # previously disabled entirely via checks: [-QF1008]); the follow-up PR
-      # does the SA1019 migration and drops these exclusions.
+      # Pre-existing staticcheck findings deferred to a dedicated cleanup PR.
+      # SA1019 deprecated-API migrations (scheme.Builder, client.Apply→SSA
+      # ApplyConfiguration, GetEventRecorderFor, sets.String, FieldsV1.Raw)
+      # change behavior/ripple through generated scaffolding, and two
+      # exported-symbol renames (ST1003: V1ApplySetIdFormat,
+      # ApiExtensionsClient) ripple across consumers — both are out of scope
+      # here. staticcheck is otherwise fully enabled (it was previously
+      # disabled entirely via checks: [-QF1008]); the follow-up PR does the
+      # SA1019 migration and drops these exclusions.
       - linters:
           - staticcheck
         text: "SA1019"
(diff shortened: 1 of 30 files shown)


cheeseandcereal:

PR #1407, opened 2026-09-09
fix(graphengine): run CRD schema invalidation regardless of the GraphKind gate

cheeseandcereal:
## Problem

On a default install (`GraphKind` off), every `ResourceGraphDefinition` instance whose graph references a Kind that did not exist at the controller's first compile — a second RGD's own generated kind (the target of its synthesized status patch node), or a CRD installed after kro started — fails every reconcile with `GraphResolved=False ResolutionFailed: graph-engine build failed: rgdadapter: compile: build node "…": rest mapping for …: no matches for kind "…"` until the controller is restarted, while the RGD itself reports `Active` (it is validated by the old `pkg/graph.Builder`, whose dynamic mapper reloads on `NoMatch`). An RGD with only built-in template kinds and no `status:` block is unaffected.

The graph-engine compiler maps kinds through a `DeferredDiscoveryRESTMapper` over a memory-cached discovery client. That mapper self-heals a `NoMatch` only while its cache is unpopulated; after the first successful refresh it is a frozen snapshot until `Reset()`. The only `Reset()` path is `Compiler.InvalidateSchema`, driven solely by the CRD `SchemaWatcher`, which was created inside the gated Graph controller — yet the same compiler is injected into the RGD reconciler and every instance controller regardless of the gate. The pre-#1355 engine used controller-runtime's dynamic mapper on the instance path and never hit this.

On the Graph path (gate on) a related staleness exists: the impersonated executor clients used the manager's mapper (which never reloads on a same-GVK plural/scope change) and a controller-runtime client memoizes each GVK's REST mapping for its lifetime, so deleting and recreating a CRD under the same GVK with a different plural kept applying through the old plural until restart.

## Fix

- `pkg/graphengine/compiler/context.go`: `buildNode` resolves mappings through the new `restMapping`, which on a `NoMatch` calls `meta.MaybeResetRESTMapper` and retries exactly once (other errors are returned as-is). This alone fixes the default-install failure: a Kind created after the first compile is mappable on the next compile. `TestCompile_KindAddedAfterDiscoveryWarmsIsMappable` pins it against the real deferred mapper.
- `cmd/controller/graphengine.go`, `main.go`: `setupGraphEngine` builds the compiler, the Graph compile cache and the `SchemaWatcher` unconditionally and `setupGraphController` reuses them, so CRD events reach `Compiler.InvalidateSchema` on the RGD path too — evicting a changed CRD's schema ahead of the resolver cache's TTL and resetting the mapper on a same-GVK plural/scope change, which is not a `NoMatch`. `TestCRDEventWithNoSubscribersStillInvalidatesSchema` pins that the watcher invalidates with zero Graph subscribers.
- `pkg/graphengine/compiler/compiler.go`: new `Compiler.RESTMapper()`; `InvalidateSchema` doc now carries the single explanation of the frozen-snapshot behaviour.
- `cmd/controller/graphengine.go`, `pkg/controller/graph/impersonation.go`, `pkg/graphengine/schemawatcher/watcher.go`: the impersonated executor clients are built over `Compiler.RESTMapper()` instead of `mgr.GetRESTMapper()`, and `impersonationCache` implements `SchemaInvalidator` (purge all cached executors), registered through the new `SchemaWatcher.AddSchemaInvalidator` since the cache exists only when the gate is on. Both are needed for a recreated CRD to be applied through its new plural/scope.
- `test/integration/environment/setup.go`, `test/integration/suites/impersonation/confinement_test.go`: mirror the production wiring (ungated watcher, compiler mapper for impersonated clients). `docs/design/proposals/deferred-schema-resolution.md`: the RGD path is now described as not *subscribing* to the watcher rather than not using it.

## Behaviour changes

- Default installs now run the CRD `SchemaWatcher`. It needs full CRD objects, so it adds a structured CRD informer next to the RGD reconciler's metadata-only CRD watch — one extra list/watch and the CRD objects resident in memory; no new RBAC. Its log name is `graph-engine.schema-watcher` (was `graph-schema-watcher.schema-watcher` with the gate on).
- On every leader start the CRD informer's initial list replays an Add per CRD, each resetting the compiler's discovery cache (previously only with the gate on); afterwards every CRD add, schema-changing update and delete does so once, and a compile that hits a `NoMatch` performs one rediscovery per attempt.
- Graph path (gate on): every schema-changing CRD event purges all cached impersonated executors, which are rebuilt on the next reconcile; impersonated clients now see a newly installed CRD after the watcher's reset rather than via controller-runtime's on-demand group reload (dynamic Graphs are re-enqueued on that event).
- Remaining gap: on the RGD/instance path the executor still applies through the manager client and the per-instance program cache keeps the compiled GVR, so a CRD recreated under the same GVK with a different plural/scope still needs a controller restart there (follow-up).

### DIFF
--- a/cmd/controller/graphengine.go
+++ b/cmd/controller/graphengine.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"fmt"
+	"net/http"
 
 	"github.com/go-logr/logr"
 	"k8s.io/client-go/kubernetes"
@@ -37,15 +38,48 @@ import (
 // api/v1alpha1's SchemeBuilder, which main.go already adds — no separate
 // registration is needed here.
 
-// setupGraphController wires the Graph controller into the
-// manager alongside the ResourceGraphDefinition stack. It builds the compile
-// cache, the resource-drift watch router, and the CRD schema watcher, then
-// registers the reconciler. The router and schema watcher are added as manager
-// Runnables so their event channels feed the same work queue as Graph spec
-// changes.
+// setupGraphEngine builds the graph-engine compiler, the Graph compile cache
+// and the CRD schema watcher (added to the manager) and returns them. It runs
+// regardless of the GraphKind gate: the compiler serves every RGD instance
+// controller, and the watcher is what drives Compiler.InvalidateSchema.
+func setupGraphEngine(
+	mgr ctrl.Manager,
+	restConfig *rest.Config,
+	httpClient *http.Client,
+	logger logr.Logger,
+	celCostLimit uint64,
+) (*compiler.Compiler, *registry.Registry, *schemawatcher.SchemaWatcher, error) {
+	cmp, err := compiler.NewCompiler(restConfig, httpClient)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("build graph-engine compiler: %w", err)
+	}
+	cmp.WithCostLimit(celCostLimit)
+
+	// The registry lives here so the watcher can invalidate it (empty while the
+	// gate is off). The watcher needs full CRD objects, so this adds a structured
+	// CRD informer next to the RGD reconciler's metadata-only one.
+	reg := registry.New()
+	sw := schemawatcher.New(logger.WithName("graph-engine"), schemawatcher.Config{
+		Cache:   mgr.GetCache(),
+		Graphs:  reg,
+		Schemas: cmp,
+	})
+	if err := mgr.Add(sw); err != nil {
+		return nil, nil, nil, fmt.Errorf("add graph-engine schema watcher: %w", err)
+	}
+	return cmp, reg, sw, nil
+}
+
+// setupGraphController wires the Graph controller into the manager alongside
+// the ResourceGraphDefinition stack, reusing the compiler, compile cache and
+// CRD schema watcher from setupGraphEngine. It builds the resource-drift watch
+// router (a manager Runnable feeding the Graph work queue) and the impersonated
+// executor clients, then registers the reconciler.
 func setupGraphController(
 	mgr ctrl.Manager,
 	cmp *compiler.Compiler,
+	reg *registry.Registry,
+	sw *schemawatcher.SchemaWatcher,
 	metaClient metadata.Interface,
 	logger logr.Logger,
 	concurrentReconciles int,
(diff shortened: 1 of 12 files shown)


cmd/controller/graphengine.go
```diff
@@ -37,15 +38,48 @@ import (
...
+	// CRD informer next to the RGD reconciler's metadata-only one.
+	reg := registry.New()
+	sw := schemawatcher.New(logger.WithName("graph-engine"), schemawatcher.Config{
+		Cache:   mgr.GetCache(),
+		Graphs:  reg,
+		Schemas: cmp,
+	})
+	if err := mgr.Add(sw); err != nil {
+		return nil, nil, nil, fmt.Errorf("add graph-engine schema watcher: %w", err)
+	}
```
Skarlso, line 69: Wait, why are we adding another informer? 🤔

pkg/controller/graph/impersonation.go
```diff
@@ -76,6 +77,18 @@ type impersonationCache struct {
 	newExec func(user string) (executor.Interface, error)
 }
 
+// InvalidateSchema purges every cached executor: a controller-runtime client
+// memoizes each GVK's REST mapping for life, so a recreated CRD needs fresh clients.
+func (c *impersonationCache) InvalidateSchema(_ schema.GroupKind) {
```
Skarlso, line 82: Errr... is it safe to ignore the GK here...?

pkg/graphengine/compiler/context.go
```diff
@@ -266,7 +266,7 @@ func (ctx *CompilationContext) buildNode(p *parser.Parser, n *expv1alpha1.Node,
 	if err != nil {
 		return nil, nil, fmt.Errorf("resolve schema for %s: %w", gvk, err)
 	}
-	mapping, err := ctx.restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+	mapping, err := ctx.restMapping(gvk)
```
Skarlso, line 269: There is a subtle difference here when things fail now they aren't cached and your initMu will lock everything on a failed RGD.

So a failing RGD now breaks the entire process.

cmd/controller/graphengine.go
Skarlso: Even more interesting, this is a full cache, right? And we already have a Partial cache as far as I see the code here: `pkg/controller/resourcegraphdefinition/controller.go:156`. So why is this now needed? 🤔

cheeseandcereal:

PR #1408, opened 2026-09-09
fix(instance): persist status on early returns and project author conditions while suspended

cheeseandcereal:
## Problem

Three early exits of the RGD instance controller (`pkg/controller/instance/`) return without writing any status. When `stampInstanceMetadata` fails persistently — e.g. an instance restored from a backup whose ApplySet inventory fails `ValidateParentInventory` — the instance has no `.status` at all while the controller error-loops, and the guard's remediation text is only in the controller log. When `preApplyApplySetInventory` fails (a malformed `applyset.kubernetes.io/contains-group-kinds` annotation, or a rejected inventory patch), an instance that was `Ready=True`/`ACTIVE` at generation N keeps reporting exactly that after a spec change at N+1 whose children never converge. The pre-rewrite engine persisted status on both.

While suspended via `kro.run/reconcile: suspended`, an RGD that declares author conditions got kro's built-in `ResourcesReady=False/ReconciliationSuspended` injected into the author-owned condition list, and the author conditions were carried verbatim from the wire, so a spec change delivered together with the annotation was not reflected (old `observedGeneration`, stale status). The pre-rewrite engine re-evaluated author conditions while suspended.

## Fix

- `controller.go`, `controller_graph_engine.go`: a `stampInstanceMetadata` failure on either the graph-engine or the suspend path now marks `InstanceManaged=False/ManagementFailed` (`finalizer/labeling failed: <cause>`) and persists it through the existing `updateConditionsStatus` early-exit writer (`persistInstanceNotManaged`) before returning the stamp error; a persist error is only logged.
- `controller_graph_engine.go`: a `preApplyApplySetInventory` failure now marks `ResourcesReady=False/NotReady` (`pre-apply applyset inventory failed: …`) and persists through `persistGraphEngineStatus` with `degraded=true` before returning the same requeue error; with author conditions they are projected as on the normal path.
- `controller.go` `reconcileSuspended`: with author conditions, the list is re-projected against a runtime built from the latest issued revision, which must be Active (`activeRuntime`, sharing `runtimeOptions` with the apply path), through `projectAuthorConditions`, factored out of `persistGraphEngineStatus`. Conditions that read only `schema` resolve at the current generation, conditions referencing managed resources are data-pending and keep their previous value, and no built-in is appended; when no runtime can be built the previous author conditions are carried forward. Suspension stays visible to authors through `runtime.condition(schema, 'ResourcesReady')`.
- `status.go` `persistNodeFreeStatus`: takes the author-owned condition list as a parameter; deletion keeps passing `deletionConditions`.

## Behaviour changes

- Stamp failure (both paths): the instance now carries `InstanceManaged=False/ManagementFailed` with the cause, `Ready=False`, `GraphResolved` and `ResourcesReady` as `Unknown/AwaitingReconciliation`, and `state: ERROR`. The pre-rewrite engine wrote `state: ACTIVE` here (its state manager had seen no nodes); `ERROR` matches the other `updateConditionsStatus` early exits. With author conditions only `state: ERROR` is written. The finalizer is still not installed and the returned error is unchanged.
- Pre-apply inventory failure: `ResourcesReady=False/NotReady` and `Ready=False` at the current generation, `InstanceManaged`/`GraphResolved` `True`, `state: ERROR`. The pre-rewrite engine wrote `state: IN_PROGRESS` here; `ERROR` is the classification a hard apply error gets. The requeue error and metrics are unchanged.
- Suspension with author conditions: only author conditions are on the wire, schema-only ones follow the current spec at the current generation, `state` stays `ACTIVE` (`ERROR` only for a degraded projection, as on the normal path). Upgrade caveat: an instance that was already suspended under the previous release keeps the `ResourcesReady/ReconciliationSuspended` entry that release injected until it is resumed — condition types the projection cannot produce while suspended are carried forward from the wire. Suspension without author conditions and deletion are unchanged.
- The pre-existing stamp-failure and pre-apply unit tests, which asserted only the returned error, now also assert the persisted status; no assertion was removed.

### DIFF
--- a/pkg/controller/instance/controller.go
+++ b/pkg/controller/instance/controller.go
@@ -43,6 +43,7 @@ import (
 	"github.com/kubernetes-sigs/kro/pkg/graphengine/executor"
 	"github.com/kubernetes-sigs/kro/pkg/graphengine/registry"
 	"github.com/kubernetes-sigs/kro/pkg/graphengine/rgdadapter"
+	geruntime "github.com/kubernetes-sigs/kro/pkg/graphengine/runtime"
 	"github.com/kubernetes-sigs/kro/pkg/metadata"
 	"github.com/kubernetes-sigs/kro/pkg/metrics"
 	"github.com/kubernetes-sigs/kro/pkg/requeue"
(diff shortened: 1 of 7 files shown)


cheeseandcereal:

PR #1409, opened 2026-09-09
fix(graphengine): replace author status via Update under legacy kro manager

cheeseandcereal:
## Problem

After an in-place upgrade from any released kro to `main`, an instance's author status field (`status.<field>` from `spec.schema.status`) whose CEL expression later stops resolving is never removed: it keeps its last value forever (for example `status.availableReplicas` stays `1` after the Deployment is scaled to zero), while the instance still reports `Ready=True`. Instances created on `main` are not affected, and downgrading corrects it.

Released controllers wrote the whole instance status with `UpdateStatus` and no explicit field manager, so every upgraded instance carries a `{manager: kro, operation: Update, subresource: status}` managedFields entry owning its author fields. On `main` the synthesized status `patch:` node writes only the resolved author fields by forced server-side apply under a per-node manager; the first post-upgrade apply sends identical values, so it merely co-owns the fields, and SSA removes a field only once its last owner drops it — which the legacy entry never does.

## Fix

- `pkg/graphengine/compiler` (`compiler.go`, `program.go`, `context.go`): new `WithStatusReplace(nodeID)` compile option → `Node.StatusReplace`, plumbed like `WithSelfWatchExempt` (including child-frame propagation); the per-node flag assignments in `compileFrame` move to `applyNodeAffordances` to stay under the `gocyclo` limit. `runtime.Node` gains the `StatusReplace()` accessor.
- `pkg/graphengine/rgdadapter/runtime.go` `schemaCompileOpts`: sets `WithStatusReplace(StatusPatchNodeID)` on the synthesized author-status node.
- `pkg/graphengine/executor/simple.go`: `applyPatchOne` routes a `StatusReplace` status node to the new `replaceStatus`, which replaces the instance status through `Status().Update` under field manager `kro` (`legacyStatusFieldManager`), carrying the live `conditions`/`state` over unchanged so their ownership stays with `kro-instance-status`. Writing under the same manager and operation continues the legacy entry instead of adding a sibling owner, and an Update prunes an omitted field from every manager. A JSON-identical status (numbers compared by canonical encoding) issues no write; a resourceVersion 409 re-reads the target and retries, surfacing as soft not-ready ("status replace conflict …") if it persists. The recorded `Contribution` keeps the per-node patch manager so the ledger on the instance is unchanged and nothing is released on upgrade.
- `docs/design/proposals/graph.md`: notes the RGD/instance-path exception to forced status SSA.
- Tests: envtest cases for a field owned by a legacy `kro/Update` entry and by a head-created patch manager (both removed, ownership converges on `kro/Update`, conditions/state untouched) plus the unchanged Graph path; fake-client cases for the no-write guard and the conflict retry; a core integration spec asserting author fields are owned by `kro/Update` on the wire and that a field a foreign co-owner still holds is removed once its expression stops resolving.

## Behaviour changes

- RGD instances: author status fields are now written by `Update` under manager `kro` (`{kro, Update, status}`) instead of SSA under `kro-graphengine.patch.<graph>.<node>`; a field whose expression stops resolving is removed regardless of managedFields ownership, as before the graph engine. `conditions`/`state` are still written by the controller via SSA under `kro-instance-status` on the reconcile path (the early-exit and deletion/suspend paths already `UpdateStatus` under `kro`).
- Upgrade from a released version: the first post-fix reconcile that changes or removes an author field continues the legacy `kro/Update` entry; nothing is written while the rendered fields are unchanged.
- Upgrade from unfixed `main`: author fields stay owned by the per-node patch manager until a value changes (moves to `kro/Update`) or a field disappears (pruned from every owner); the contribution ledger annotation is unchanged, so no author fields are released on upgrade.
- Standalone Graph `patch:` nodes with `status:` keep forced SSA under the per-node manager.
- Not covered: if an RGD update removes all author status fields, no status node is synthesized and stale author values are only removed by the regenerated CRD's schema pruning on the next status write.


cheeseandcereal:

PR #1410, opened 2026-09-09
fix(graphengine): enforce resource-identity uniqueness before any write

cheeseandcereal:
## Problem

When two nodes of one graph render the same object (same GVK, namespace and name — e.g. two ConfigMap resources both named `${schema.spec.name}`), the executor only noticed when the second node was reached in the topological walk. By then the first node had already been server-side applied. On the RGD path that apply is forced under the shared field manager, so if a user-owned object with that name already existed it was adopted (applyset `part-of` label, `kro.run/owned`, data overwritten); the instance then went `ERROR`, and deleting it pruned the user's object. The pre-rewrite engine projected the whole desired set (`applyset.Project`) before applying anything, so nothing was written.

Two smaller gaps in the same area: a forEach `patch` node whose axis contains duplicate values (`${["a","a"]}`) reached `Ready=True` while patching one target twice and recording two identical contribution rows — the byte-identical `template` variant was rejected — and the compile-time iterator-coverage rule only counted `metadata.name`/`metadata.namespace`, so `forEach: [{k: ${["ConfigMap","Secret"]}}]` with `kind: ${k}` and a fixed name was rejected although every rendered identity is distinct.

## Fix

- `executor/simple.go`: `Simple.Apply` runs a pre-walk pass (`reserveIdentities`) at the top of every frame, root and subgraph child alike. It renders each template node that is resolvable before the walk, namespace-defaults the objects as `prepareItem` does, and reserves their identities in the per-Apply claim set; any collision returns the hard `ErrDuplicateIdentity` before anything is written, with `ApplyResult{}` empty. The claim set is two-phase: the walk's per-item claim consumes the node's own reservation once and rejects any other prior entry.
- `executor/simple.go`: the pass only considers templates whose hard dependencies are, transitively, Def nodes. The runtime seeds soft-dependency targets (templates the synthesized status node reads) with an empty object, so an identity field like `${a.?data.k.orValue("fallback")}` would otherwise resolve against the placeholder and reserve an identity the walk never renders, false-positiving against a node that really renders it. Pinned by `TestApply_SoftDepPlaceholderIsNotReserved`.
- `runtime/node.go`: `Resolve` runs `validateUniqueIdentities` for patch collections as well as template collections, so duplicate axis values fail with `duplicate identity in collection` before any write.
- `compiler/compiler.go`: `isIdentityFieldPath` counts `apiVersion`, `kind`, `metadata.name`, and `metadata.namespace` when the target is namespaced or the node is dynamic-GVK (scope unknown at compile time); the coverage error lists these fields. `omit()` is therefore also rejected in `apiVersion`/`kind`.

## Behaviour changes

- A graph with two resolvable nodes rendering the same object now fails the whole frame before any write: `ResourcesReady=False` / `resource reconciliation failed: apply "cmb": executor: duplicate resource identity across nodes: nodes "cma" and "cmb" both render v1/ConfigMap/<ns>/<name>`, nothing created, adopted or pruned. Because no node in the frame runs, the RGD's synthesized author-status patch node does not run that cycle either (controller-owned conditions and state are still written).
- A collision the pass cannot see — a node whose identity depends on another node's published output — is still caught only per item during the walk, so the first colliding node is applied (and a pre-existing object it names is adopted) before the second is refused, as before.
- Upgrade: an object adopted by an earlier duplicate-identity apply stays adopted; deleting that instance still prunes it. The fix prevents new adoptions, it cannot un-adopt.
- forEach patch collections with duplicate axis values now fail hard instead of converging while double-patching one target.
- Graphs/RGDs with a forEach iterator only in `kind`/`apiVersion`, or only in `metadata.namespace` on a dynamic-GVK node, now compile. The compile error for an uncovered iterator changed from `every forEach iterator must appear in metadata.name (or metadata.namespace for namespaced resources) …` to `every forEach iterator must appear in an identity field (apiVersion, kind, metadata.name, or metadata.namespace for namespaced (or dynamic-GVK) resources) …`; the two existing tests that pinned the old wording (`compiler_test.go`, `patch_test.go`) assert the new one.

### DIFF
--- a/pkg/controller/instance/controller_graph_engine_test.go
+++ b/pkg/controller/instance/controller_graph_engine_test.go
@@ -1835,12 +1835,18 @@ func TestReconcileViaGraphEngine_PatchContributions(t *testing.T) {
 		watcher := &fakeInstanceWatcher{}
 		err := c.reconcileViaGraphEngine(context.Background(), inst, watcher)
 		require.Error(t, err)
-		// The collision is now caught PRE-WRITE by the executor's identity-claim
-		// guard (before cm2's SSA write clobbers cm1), so the error is
-		// ErrDuplicateIdentity rather than the post-apply validateAppliedIdentities
-		// message. (Like any hard apply error it is still delayed-requeued so the
-		// instance retries; the state below reflects the degraded outcome.)
+		// The collision is caught by the executor's pre-walk identity pass (both
+		// names are static), so the error is ErrDuplicateIdentity rather than the
+		// post-apply validateAppliedIdentities message and neither node is applied.
+		// (Like any hard apply error it is still delayed-requeued so the instance
+		// retries; the state below reflects the degraded outcome.)
 		assert.Contains(t, err.Error(), "duplicate resource identity across nodes")
+
+		live := &unstructured.Unstructured{}
+		live.SetAPIVersion("v1")
+		live.SetKind("ConfigMap")
+		getErr := fakeRuntimeCl.Get(context.Background(), client.ObjectKey{Namespace: "default", Name: "shared-cm"}, live)
+		assert.True(t, apierrors.IsNotFound(getErr), "shared-cm must not exist after a pre-walk duplicate rejection, got err=%v", getErr)
 
 		stored := getStoredParentObject(t, raw)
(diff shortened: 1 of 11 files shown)


cheeseandcereal:

PR #1411, opened 2026-09-09
test(integration): add default-gates lane and restore deleted unit guards

cheeseandcereal:
## Problem

The envtest integration suite unconditionally enabled the alpha feature gates `CELOmitFunction` and `GraphKind` before booting, so `make test WHAT=integration` — the only integration lane CI runs — never exercised the configuration a stock kro install actually runs (both gates default off; the helm chart ships `featureGates: {}`). Anything that only happens with a gate off, such as the component set wired outside the `GraphKind` block or the `optional.none()` → `null` rendering path with `CELOmitFunction` off, had no coverage.

Two unit guards were also deleted in the engine rewrite without replacement: the status-schema tests for array-index paths (`status.parents[0].child` must yield `type: array` with an `items` schema, or the generated CRD rejects the list kro then writes), and the test that one failing author-condition expression is dropped and reported as degraded while its well-formed siblings still surface. The test helper `isConflictErr` also matched a 409 by message substring instead of by status reason.

## Fix

- `test/integration/environment/featuregates.go`: `KRO_INTEGRATION_FEATURE_GATES` selects the gates the core suite enables before it starts the manager (same `Name=bool,…` syntax as `--feature-gates`): unset keeps the suite's own set, empty or `none` enables nothing, anything else is passed through verbatim. `TestFeatureGatesFromEnv` pins the parsing.
- `test/integration/suites/core/setup_test.go`: `TestCore` reads the variable (unset behaves exactly as before: `CELOmitFunction=true,GraphKind=true`). `requiresFeatureGate(gate)` labels a container `feature-gate:<Name>`, and a top-level `BeforeEach` skips labelled specs whose gate is off and fails on an unknown gate name; because it is top-level, a skipped spec runs neither its container's `BeforeEach` nor its `AfterEach`.
- Gate labels on the 15 Graph-kind containers (`graph_*_test.go` except `graph_rgd_*`, which drive the engine directly and need no controller) and on the `Omit` container — 42 specs, every one of which creates a `Graph` object or templates `omit()`.
- `Makefile`: new `make test WHAT=integration-default-gates` runs the same suites with the variable exported empty and its own coverprofile; `WHAT=integration` is unchanged; `WHAT=unit` now also runs `./test/integration/environment/...`. `scripts/ci/presubmits/integration-tests-default-gates` mirrors `integration-tests`, and `test/README.md` documents the two lanes.
- `pkg/graph/schema/field_descriptor_test.go` (restored): `GenerateSchemaFromCELTypes` with indexed paths — nested array, deep mixed path, index as the leaf, and sibling fields merging into one items schema.
- `pkg/graphengine/rgdadapter/status_projection_test.go`: `TestProjectInstanceConditions_FatalErrorInOneExpressionSkippedNotAborted` drives an evaluation error, a scalar non-condition and a list of non-conditions through `ProjectInstanceConditions` ahead of a good condition, asserting `ErrConditionProjectionDegraded`, `incomplete=true` and the surviving condition.
- `test/integration/environment/graph_helpers.go`: `isConflictErr` is now `apierrors.IsConflict(err)`; `TestIsConflictErr` covers the Update and server-side-apply 409 shapes, a non-409 API error, and a plain error that merely carries the message text.

## Behaviour changes

- `make test WHAT=integration` runs the same specs with the same gates and skips none; spec names now carry `[feature-gate:…]` labels. `make test WHAT=unit` additionally runs the `test/integration/environment` package (no envtest, about a second).
- `make test WHAT=integration-default-gates` is red on today's `main`: 31 specs fail (a few fewer in some runs, since the trigger depends on discovery-cache timing), all with `rest mapping for …: no matches for kind …`, because with `GraphKind` off nothing resets the graph-engine compiler's REST mapper — the bug fixed in #8. With that fix applied on top, the lane is fully green. CI should run the new presubmit as non-blocking until #8 merges, then make it blocking.
- `isConflictErr`: an error that is not an API `Status` but whose text contains "the object has been modified" is no longer treated as a conflict; the controller-runtime client never produces such a value.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,28 +130,36 @@ go-generate: ## Run go generate against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
+# Ginkgo flags shared by both integration lanes; they differ only in the
+# feature gates enabled (KRO_INTEGRATION_FEATURE_GATES, see test/README.md).
+INTEGRATION_GINKGO_FLAGS = -p --timeout=8m --cover -coverpkg=github.com/kubernetes-sigs/kro/pkg/...
+
 .PHONY: test
-test: manifests generate fmt vet envtest ## Run tests. Use WHAT=unit or WHAT=integration, pass extra args after --
+test: manifests generate fmt vet envtest ## Run tests. WHAT=unit|integration (alpha gates on)|integration-default-gates (production gate defaults)|upgrade; extra args after --
 ifeq ($(WHAT),integration)
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_VERSION) --bin-dir $(LOCALBIN) -p path)" \
-		go tool ginkgo -p \
-		--timeout=8m \
-		--cover \
+		go tool ginkgo $(INTEGRATION_GINKGO_FLAGS) \
 		--coverprofile=integration-cover.out \
-		-coverpkg=github.com/kubernetes-sigs/kro/pkg/... \
+		$(filter-out $@,$(MAKECMDGOALS)) \
+		./test/integration/suites/...
+else ifeq ($(WHAT),integration-default-gates)
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_VERSION) --bin-dir $(LOCALBIN) -p path)" \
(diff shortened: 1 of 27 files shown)


cheeseandcereal:

PR #1412, opened 2026-09-09
fix(graph): surface status-write failures, cap inventories, project forEach patches

cheeseandcereal:
## Problem

The standalone Graph controller (`GraphKind` gate) keeps two status inventories, `status.managedResources` and `status.contributions`, and tears a Graph down from them. Three status-write failures around those inventories were invisible or misreported. When the post-apply write of the contribution ledger was refused (a dozen `forEach` patch nodes over a thousand targets exceed etcd's request size), the error was returned but `ResourcesConverged`/`Ready` stayed `True` with an empty ledger, so deleting the Graph left every contributed field and field manager on its targets. When the pre-apply write-ahead was refused (six 1000-item collection nodes exceed the `managedResources` cap of 5000), no condition was set and the in-memory Graph still held the refused list, so the conditions write was refused for the same reason: a fresh Graph had no `status` at all and a blank `READY` column. `status.contributions` also had no `maxItems`, so nothing bounded the ledger.

Separately, the write-ahead contribution projection only handled singleton patches (`len(desired) != 1` was skipped), so a `forEach` patch node had zero write-ahead rows while the executor recorded one contribution per target. And `spec.serviceAccountName` documented "empty means the namespace default" but its pattern rejected the explicit `""` that Helm and Kustomize render for an unset value.

## Fix

- `api/v1alpha1/graph_types.go`: new `GraphInventoryMaxItems = 5000`; `status.contributions` gains `MaxItems=5000` like `managedResources`; the `serviceAccountName` pattern admits the empty string (`^$|…`). `helm/crds/kro.run_graphs.yaml` regenerated. `TestGraphCRD_MarkersMatchConstants` asserts the generated CRD agrees with the constant and the regex under test.
- `tracking.go`: `projectContributions` emits one write-ahead `Contribution` per rendered patch target under the node's field manager; new `checkInventoryCaps` (typed `inventoryTooLargeError`) and `adoptUIDs` (copies observed UIDs onto write-ahead entries without adding identities).
- `controller.go` `writeAheadIntent`: checks both inventories against the cap before writing or applying (fail closed), and on a refused managed-resource write restores `g.Status.ManagedResources` to what the server holds so the conditions write is not refused too.
- `controller.go` `reconcileGraph`: a refused write-ahead sets `ResourcesConverged=False` with reason `InventoryTooLarge` or `WriteAheadFailed`; a post-apply cap check (the executor can apply identities the projection cannot see) sets `InventoryTooLarge`, keeps the write-ahead union with observed UIDs adopted, skips prune/release and returns the cap error joined with a hard apply error; a refused clean-path ledger write sets `StatusWriteFailed`. The error branches persist `ledgerUnion` (the ledger the write-ahead left on the server plus the observed rows) instead of `prior ∪ observed`, so a patch whose target does not exist yet no longer has its write-ahead rows written and erased on every reconcile, each status write re-enqueuing the Graph. `TestReconcile_SoftPathKeepsWriteAheadLedger` pins one ledger write per Graph and no shrink for forEach and singleton patches over absent targets.
- `docs/design/proposals/graph.md`: the three new condition reasons; `""` documented as unset.

## Behaviour changes

- New `ResourcesConverged=False` reasons: `WriteAheadFailed` (pre-apply inventory write refused, nothing applied), `StatusWriteFailed` (clean apply but the contribution ledger did not persist), `InventoryTooLarge` (an inventory would exceed 5000 entries; the controller fails closed and names the field, count and cap). `Ready` follows.
- `status.contributions` is now capped at 5000 entries like `managedResources`. Because the pre-apply inventory is `previous ∪ intended`, a `forEach` patch set whose union exceeds the cap — for example renaming or retargeting more than 2500 contribution rows in one step — is now refused with `InventoryTooLarge` where it previously succeeded; rename such nodes incrementally. For `managedResources` the same shape already failed with a raw 422.
- A patch node whose target does not exist yet now holds a stable write-ahead ledger and `WaitingForReadiness` until the target appears; before, a singleton patch in that state rewrote the ledger twice per reconcile and hot-looped.
- `spec.serviceAccountName: ""` is admitted and behaves exactly like an omitted field (impersonates `system:serviceaccount:<ns>:default`); malformed values are still rejected.
- `api/v1alpha1/graph_types_test.go`: the `"empty rejected"` row is flipped to allowed, since it pinned the rejected-empty-string behaviour this PR removes.

### DIFF
--- a/api/v1alpha1/graph_types.go
+++ b/api/v1alpha1/graph_types.go
@@ -51,14 +51,23 @@ type GraphSpec struct {
 	// of the Graph's namespace, confining resource access to that namespace by
 	// default.
 	//
+	// An explicit empty string is equivalent to omitting the field (templating
+	// tools such as Helm and Kustomize render an unset value as "").
+	//
 	// The kro controller ServiceAccount must be granted the "impersonate" verb
 	// on serviceaccounts for this to take effect.
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	// +kubebuilder:validation:Pattern=`^$|^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
 	// +kubebuilder:validation:MaxLength=253
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 }
+
+// GraphInventoryMaxItems is the MaxItems cap on GraphStatus.ManagedResources and
+// GraphStatus.Contributions. The controller checks inventories against it before
+// a status write so an oversized inventory becomes a condition, not a rejection.
+// Markers cannot reference constants: keep the two MaxItems=5000 markers equal.
+const GraphInventoryMaxItems = 5000
 
 // GraphStatus defines the observed state of a Graph.
 type GraphStatus struct {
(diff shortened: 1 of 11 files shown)


cheeseandcereal:

PR #1413, opened 2026-09-09
fix(graph): make Graph teardown safe against gone CRDs, UID-free entries and partial applies

cheeseandcereal:
## Problem

Deleting a standalone Graph (`GraphKind`) runs `Simple.Delete` over `Status.ManagedResources` from the finalizer, and three cases went wrong there. A managed resource whose CRD had been removed wedged the finalizer forever once kro had restarted: the cold REST mapper surfaces the missing mapping as a raw `*meta.NoKindMatchError`, which `Delete` tolerated neither as NotFound nor as Conflict, so the Graph sat in `Terminating` with `DeleteFailed … no matches for kind` until the CRD was recreated.

The reconciler write-aheads UID-free inventory entries before Apply so a status write lost after Apply still leaves teardown something to delete — but `Delete` skipped every UID-free entry as an anti-forgery guard, so in exactly that crash window the finalizer came off and every child was orphaned (Graph children carry no `ownerReference`).

`Status.AppliedServiceAccount` was recorded only after a clean or soft not-ready apply. A hard failure that still created resources (a ConfigMap lands, a ClusterRole gets 403) persisted those entries with no identity, so after the user switched `spec.serviceAccountName` to a rights-less ServiceAccount and deleted, teardown ran as that ServiceAccount and wedged on `DeleteFailed … cannot delete resource "configmaps"`. The RGD/instance path never calls `executor.Delete` and is untouched.

## Fix

- `executor.Interface.Delete` gains an `ownerUID types.UID` parameter (the Graph UID); both controller call sites (teardown and prune) pass `g.GetUID()`.
- `Simple.Delete` / `deletePreconditioned`: a NoMatch that survives one mapper refresh (`retryOnNoMatch` — a no-op for controller-runtime's lazy mapper, which already reloads on the first miss; it matters for a resettable mapper) is treated as "already gone" alongside NotFound and Conflict, matching `Release`, `mappingFor` and the applyset prune path.
- `Simple.Delete` / `verifyUIDFreeEntry`: a UID-free entry is resolved against the live object and deleted, preconditioned on the live UID, only when `ownedByGraph` holds — the object's `managedFields` carry this Graph's template field manager (per-Graph segment, so a legacy per-node manager counts), or for collection members the `kro.run/instance-id` label equals the Graph UID and `kro.run/node-id` equals the entry's node token. An absent object, a type no longer served, an unmarked object, or a Forbidden GET (an identity that cannot read a type cannot have applied it) are skipped with an Info log; any other GET failure is returned so the reconcile retries. `TemplateFieldManager` is exported so tests can stage objects with the real marker.
- `reconcileGraph`: `Status.AppliedServiceAccount` is recorded whenever anything reached the cluster — clean, soft not-ready, or a hard failure with a non-empty `Applied`/`Contributions`; a hard failure that applied nothing keeps the last-good identity.
- Tests: executor unit and envtest tests for the NoMatch retry and the UID-free verification (real SSA `managedFields`, a cold lazy mapper against a deleted CRD); controller tests for the identity recording and the UID plumbing; two core integration specs (CRD removed then Graph deleted; staged UID-free inventory where only this Graph's child is deleted); the RBAC-enforcing impersonation suite now also deletes its refused Graph, whose inventory names a type the ServiceAccount may not read.

## Behaviour changes

- Teardown or prune of a managed resource whose CRD no longer exists releases the finalizer instead of reporting `DeleteFailed … no matches for kind` forever.
- A UID-free (write-ahead) inventory entry is deleted on teardown or prune when the live object verifiably belongs to the Graph; an entry naming an object without those markers, or one the Graph's ServiceAccount may not `get`, is still never deleted (now with an Info log). One GET per UID-free entry; entries with a recorded UID are unaffected.
- `Status.AppliedServiceAccount` is also written after a hard apply failure that landed at least one resource or patch contribution, so such a Graph is torn down under the identity that created its resources even if `spec.serviceAccountName` changed afterwards. The field remains a single scalar: after a mid-life ServiceAccount change the most recent identity that reached the cluster wins.
- Go API change, no CRD change: `executor.Interface.Delete(ctx, ownerUID, resources)`. The unit test "empty UID is skipped and does not delete pre-existing object" keeps its assertion under the name "empty UID naming an unmarked object is skipped and does not delete it"; `TestDelete_UIDPrecondition` now pins that a verified UID-free entry is deleted with the live UID as precondition instead of producing no Delete call.

### DIFF
--- a/pkg/controller/graph/controller.go
+++ b/pkg/controller/graph/controller.go
@@ -248,7 +248,7 @@ func (r *Reconciler) reconcileDelete(ctx context.Context, logger logr.Logger, re
 	}
 
 	if len(g.Status.ManagedResources) > 0 {
-		if err := ex.Delete(ctx, g.Status.ManagedResources); err != nil {
+		if err := ex.Delete(ctx, g.GetUID(), g.Status.ManagedResources); err != nil {
 			return teardownFailed(err, "executor delete", "failed to persist teardown delete-failure condition")
 		}
 	}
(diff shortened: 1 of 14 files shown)


cheeseandcereal:

PR #1414, opened 2026-09-09
fix(graphengine): type dynamic-GVK collections as list(dyn), not any

cheeseandcereal:
## Problem

A Graph `ref:` whose `apiVersion`/`kind` is a CEL expression and that carries `metadata.selector` (the graph.md "Dynamic GVKs" collection pattern) has no compile-time schema, so the compiler declared it through `krocel.WithResourceIDs`, which emitted an opaque `any` (`google.protobuf.Any`) variable despite its doc promising `dyn`. cel-go treats `any` like `dyn` for field selection, indexing and assignability but rejects it as a comprehension range, and kro's forEach check accepts only a list or `dyn`. So `forEach: [{w: ${insts}}]` failed with `Accepted=False: forEach "w" must return a list, got "google.protobuf.Any"` and `${insts.map(...)}` / `.filter` / `.all` / `.exists` with `expression of type 'any' cannot be range of a comprehension`, while `size(insts)`, `insts[0].metadata.name` and the `${dyn(insts)}` workaround compiled. Captured ancestor IDs got the same declaration, so a subgraph `forEach` over a parent collection was rejected too. `examples/graph/rgd.yaml`'s L1 Graph uses exactly this pattern, but it is a stamped child Graph that `TestExamplesGraphCompile` never compiles.

The pre-rewrite engine (`7458023`) is not a comparison point: `metadata.ExtractGVKFromUnstructured` and schema resolution rejected any `${` in `apiVersion`/`kind`, so dynamic GVKs did not exist, and `WithResourceIDs` only fed the parse-only inspector environment.

## Fix

- `pkg/cel/environment.go`: `WithResourceIDs` declares `cel.DynType` instead of `cel.AnyType`, as its doc already said. The compiler and legacy RGD-builder inspector environments only parse, so nothing changes there, and the legacy builder type-checks against `TypedEnvironmentWithProvider`, so the set of RGD expressions accepted at RGD validation is unchanged. The RGD adapter's per-reconcile `status.conditions` environment (`rgdadapter/status.go`) now lets a comprehension range over a bare scalar node id, but that is unreachable for any RGD that reaches `Active`: the legacy builder type-checks the same expressions against the typed environment first and rejects them.
- `pkg/graphengine/compiler/compiler.go`: `compileFrame` partitions schemaless identifiers via `schemalessIdentifiers`: dynamic-GVK collections (selector refs, forEach templates) are declared `list(dyn)` through `krocel.WithListVariables` — the `[]any` the executor publishes — while dynamic single templates/refs, subgraph nodes and captured ancestors stay `dyn`; Patch nodes are skipped as before. The helper is extracted because inlining it pushes `compileFrame` over the `gocyclo` limit. `TestCompile_DynamicRefCollectionTyping` pins forEach, comprehensions and subgraph captures over a dynamic collection, that a bare collection is still rejected where a bool or string is expected, and that a bare dynamic single ref compiles in `includeWhen` and in a typed field.
- Doc corrections in `compiler/context.go` (`buildDynamicNode` also compiles Refs; a collection is `list(dyn)`) and `rgdadapter/status.go` (scalar nodes are `dyn`).
- New executor test (dynamic selector ref feeding a forEach template and comprehensions against a fake client) and envtest Graph spec (static CRD ref → dynamic selector ref → forEach ConfigMaps + summary, then a third instance appears).

## Behaviour changes

- Graphs (including RGD-adapter Graphs) that reference a dynamic-GVK collection in `forEach` or in a comprehension now compile and converge; previously `Accepted=False`. `${dyn(insts)}` keeps working.
- A bare dynamic collection reference is precisely typed: `includeWhen: ["${insts}"]` is rejected with `must return bool, got list(dyn)` and `${insts}` in a string field with `returns "list(dyn)" but expected "string"` (same outcome as before, clearer message).
- A bare dynamic non-collection identifier (single dynamic ref/template, subgraph node, captured ancestor) is `dyn`: `includeWhen: ["${target}"]` now passes compilation and fails at runtime with `includeWhen "target" returned map[string]interface {}, want bool` where it used to fail compilation, and `${target}` into a typed field passes the structural check. This is the documented intent of the cross-frame `dyn` seam.
- No RGD-validation change; no existing test was modified or removed.

### DIFF
--- a/pkg/cel/environment.go
+++ b/pkg/cel/environment.go
@@ -36,7 +36,7 @@ type EnvOption func(*envOptions)
 // envOptions holds all the configuration for the CEL environment.
 type envOptions struct {
 	// resourceIDs will be converted to CEL variable declarations
-	// of type 'any'.
+	// of type 'dyn'.
 	resourceIDs []string
 	// typedResources maps resource names to their OpenAPI schemas.
 	// These will be converted to typed CEL variables with field-level
@@ -72,7 +72,10 @@ func WithRuntimeLibrary(enabled bool) EnvOption {
 	}
 }
 
-// WithResourceIDs adds resource ids that will be declared as CEL variables.
+// WithResourceIDs adds resource ids that will be declared as untyped (`dyn`)
+// CEL variables. They must be `dyn`, not `any`: the checker treats both as
+// wildcards for assignability and field access, but rejects `any` as a
+// comprehension range, so `x.map(...)` over an `any` identifier fails to compile.
 func WithResourceIDs(ids []string) EnvOption {
 	return func(opts *envOptions) {
 		opts.resourceIDs = append(opts.resourceIDs, ids...)
@@ -98,9 +101,9 @@ func WithTypedResources(schemas map[string]*spec.Schema) EnvOption {
 	}
 }
 
-// WithListVariables adds list-typed variable declarations to the CEL environment.
-// Used for collection resources so they support list operations/macros like all()
-// exists(), filter(), and map() etc...
+// WithListVariables adds list(dyn) variable declarations to the CEL environment.
+// Used for collection resources without an element schema so they support list
+// operations/macros like all(), exists(), filter(), and map().
 func WithListVariables(names []string) EnvOption {
 	return func(opts *envOptions) {
 		for _, name := range names {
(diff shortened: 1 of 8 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1415, opened 2026-09-09
fix(graphengine): report patch field conflicts as FieldManagerConflict; reject readyWhen on patch nodes

cheeseandcereal:
## Problem

A main-resource `patch:` node applies its contribution with unforced server-side apply, so a field already owned by another manager (`kubectl`, another controller, a peer Graph) is refused with a field-level 409. The executor wrapped that 409 only in `ErrNotReady`, and the Graph controller reported it as `ResourcesConverged=False/WaitingForReadiness` — documented as "applied, but readyWhen not yet met" — although nothing was applied and, for a `kubectl`-owned field, never would be. The template path's peer-Graph refusal already carried `ErrFieldManagerConflict` but was reported as `WaitingForReadiness` too, and any other soft not-ready node earlier in the walk hid the conflict entirely.

`readyWhen` was documented as supported on `patch:` nodes, but a patch publishes no value into scope: referencing the patch itself or any other node is rejected, and the only expressions that compiled — identifier-free ones, or `each`-based ones on a forEach patch — were evaluated against the rendered contribution, never the target. A realistic `readyWhen: ["${w.status.phase == 'Ready'}"]` on a patch of `w` failed the whole Graph with a misleading error.

## Fix

- `executor.applyPatchOne`: a 409 carrying `FieldManagerConflict` causes is wrapped with both `ErrFieldManagerConflict` and `ErrNotReady` (still a soft requeue, now distinguishable); a 409 without such causes (e.g. an optimistic-concurrency precondition) stays plain soft not-ready. New helper `hasFieldManagerConflictCause`.
- `executor.Simple.Apply`: among several soft not-ready nodes, a field-manager conflict takes precedence as the returned error, so it is reported even while an upstream node is still converging (`TestPatch_FieldManagerConflictSoftRequeue` places a never-ready template ahead of the conflicting patch).
- Graph controller: new marker `ResourcesFieldManagerConflict` ⇒ `ResourcesConverged=False`, reason `FieldManagerConflict`, selected before `DataPending`/`WaitingForReadiness` inside the `ErrNotReady` branch, so requeue/backoff and inventory handling are unchanged. The template peer-Graph conflict gets the same reason.
- Instance controller: the same conflict (cross-engine `ownedByGraphTemplate` guard) keeps reason `NotReady` — the instance markers have no custom reasons — but its message is prefixed `field manager conflict:` instead of `waiting for unresolved resource:`.
- `compiler.validateKindCompatibility`: `readyWhen` on a `patch` node is rejected at compile time with an actionable message (put the `readyWhen` on a `ref:` node for the target). `forEach` and `includeWhen` stay supported. The executor's `SetObserved` for patch nodes is removed (nothing evaluates it any more).
- `docs/design/proposals/graph.md`: `FieldManagerConflict` added to the `ResourcesConverged` reasons; the patch section, banner and KREP-002 row now say `forEach`/`includeWhen` are supported on patch nodes and `readyWhen` is not.

## Behaviour changes

- New `ResourcesConverged` reason `FieldManagerConflict` on Graph objects, for a refused main-resource patch contribution and for a template object owned by a peer Graph; anything keying on `WaitingForReadiness` for those cases will now see `FieldManagerConflict`. It wins over `WaitingForReadiness`/`DataPending` from other nodes. `Ready=False` roll-up, requeue and backoff are unchanged.
- Instance path: reason stays `NotReady`; only the message prefix changes to `field manager conflict:`.
- Upgrade caveat: a Graph whose `patch:` node carries a `readyWhen` — including a forEach patch with an `each`-based expression, which compiled and reached `Ready=True` before — now fails compilation (`Accepted=False/InvalidGraph`) until the `readyWhen` is removed; its existing contributions stay in place meanwhile. No shipped example or test fixture had such a node, and the RGD adapter's synthesized status patch never carries one.
- Executor error text: a causeless patch 409 now reads `patch conflict on …` instead of `patch field conflict on …` (still `ErrNotReady`).
- `TestPatch_FieldManagerConflictSoftRequeue` now expects two applied resources (an upstream template was added ahead of the patch node); no other existing assertion changed.

### DIFF
--- a/docs/design/proposals/graph.md
+++ b/docs/design/proposals/graph.md
@@ -1,6 +1,6 @@
 # KREP-024: Graph
 
-> **Implementation status (as of v1alpha1):** This proposal document has been partially reconciled with the shipped implementation in `pkg/graphengine/` and `pkg/controller/graph/`. Key architectural differences from earlier drafts — including unified `ref:` for collections (superseding `watch:`), explicit `graph:` subgraph nodes, the raw-manifest `patch:` API (a `Patch *runtime.RawExtension` field authored exactly like `template:` — no `PatchSpec` wrapper, no `subresource`/`body` fields, no `forEach`; the target endpoint is derived from field presence, see `api/v1alpha1/graph_types.go:208`), collection expansion caps (`MaxCollectionSize`), and the `Accepted`/`ResourcesConverged`/`Ready` condition model — are reflected below. Note also that RGD is no longer a sibling engine: the RGD controller now unconditionally routes through the graph engine (it issues `GraphRevision` objects and serves compiled graphs through the shared runtime; see `pkg/controller/resourcegraphdefinition/controller_reconcile.go`). Features marked as "Planned" or "Not yet implemented" (such as KREP-006 `propagateWhen` and lifecycle signals) remain future work. Remaining design claims may still drift.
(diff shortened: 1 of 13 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1347, opened 2026-08-10
feat: add shortNames and categories in the RGD

fabianburth:
Builds on https://github.com/kubernetes-sigs/kro/pull/132, addressing the PR comments.

PR #1417, opened 2026-09-09
feat: add plural form next to shortnames and categories as override option

### LINKED ISSUE #1416: `Podinfo` is plurazied as `podinfoes` because of https://github.com/gobuffalo/flect - Expand `Categories` and `ShortNames` section with `Plural`
### Description

**Observed Behavior**:

The plural for of podinfo should not be podinfoes.

With an RGD like this:

```yaml
apiVersion: kro.run/v1alpha1
kind: ResourceGraphDefinition
metadata:
  name: podinfo
spec:
  schema:
    apiVersion: v1alpha1
    kind: Podinfo
    spec:
      image: string
      message: string | default="Hello from OCM and kro"
      replicas: integer | default=1 minimum=0
```

Kro right now generates a weird name for pluralized form of this name:

```
$ kubectl api-resources --api-group=kro.run
NAME                       SHORTNAMES   APIVERSION         NAMESPACED   KIND
podinfoes                               kro.run/v1alpha1   true         Podinfo
resourcegraphdefinitions   rgd          kro.run/v1alpha1   false        ResourceGraphDefinition
systems                                 kro.run/v1alpha1   true         System
```

because of this library:

https://github.com/gobuffalo/flect

This is fine for most of the cases, but in case the English doesn't really work out, we should have the ability to overwrite the plural form of the created CRD by adding a `plural` field to it and using that instead of if it is defined.

_Edit_:

Turns out https://github.com/kubernetes-sigs/kro/pull/1347/ already added ShortNames and Categories and now we should just expand that with `Plural`. :)

Where should 

**Expected Behavior**:

**Versions**:
- kro version: 0.9.4
- Kubernetes Version (`kubectl version`):

**Involved Controllers**:

```
Paste any relevant logs here
```

_No response_

Skarlso:
Issue number https://github.com/kubernetes-sigs/kro/issues/1416 found that some form of Kind's pluralization can be very weird.

This PR adds `plural` form next to `shortNames` and `categories` like https://github.com/kubernetes-sigs/kro/pull/1347 did.

The only caveat being that plural always needs to be correctly given back AND it can _never_ be changed again.

To that effect, I extracted some of the common functions to be able to put the determination of the name into one place. :)

Fixes https://github.com/kubernetes-sigs/kro/issues/1416.


tomasaschan (commented):
/lgtm

Adding this next to `categories` and `shortNames` makes sense, but I wonder if we should consider a breaking change sometime before 1.0 to move them all under a `names` node, to avoid cluttering the "root" of the definition of a new type.

jakobmoellerdev (approved):

PR #1424, opened 2026-09-16
Let new resource kinds work without restarting kro

cheeseandcereal:
After an RGD instance successfully creates its resources, a user can add another RGD with a different kind and an ordinary status projection. On a default installation, the new RGD becomes Active, but its instance reports `GraphResolved=False/ResolutionFailed` with `no matches for kind`, creates no children, and never fills the projected status.

The compiler's discovery snapshot can still be marked fresh after new CRDs appear. On a missing-kind mapping error, it now resets the existing mapper and retries once. Other errors retain their wrapping, and explicit schema invalidation is preserved.

An isolated regression test with default feature gates reconciles one RGD before introducing another. It checks that the second instance creates its child and projects status without restarting the controllers. Focused tests cover persistent misses and other mapping errors.

Rediscovery can block other compiles sharing the mapper. One retry limits added attempts, not elapsed time.


jakobmoellerdev (approved):
/lgtm

PR #1425, opened 2026-09-16
Retry collection member updates after transient API rejections

cheeseandcereal:
An instance can report `ACTIVE` while an unavailable admission webhook prevents an existing `forEach` member from receiving an update. Healthy siblings advance, but the rejected member stays stale because the failed update is treated as success and no failure-driven retry is scheduled.

Use the existing rejection classification to retain the live member's identity and UID while returning not-ready for errors outside `Invalid`/`BadRequest`. Siblings continue applying, dependents wait, and the controller's existing backoff retries the update. This includes Forbidden and unclassified failures. Validation rejections keep their existing live-value tolerance for immutable updates; the tolerated-event payload loses its now-constant `Permanent` field. Existing field-manager-conflict guards retain their precedence.

The regression fails on the baseline's false success and passes with this change: removing only the failed webhook repairs the stale member under its original UID and returns the unchanged instance to `ACTIVE`. Targeted race tests also cover identity retention, dependent gating, and rejection boundaries. Real-API checks preserve ownership-conflict handling and immutable-Pod scale-up/down behavior.

### DIFF
--- a/pkg/graphengine/executor/apply_template_test.go
+++ b/pkg/graphengine/executor/apply_template_test.go
@@ -201,33 +201,25 @@ func TestSimple_ApplyTemplate_WatchRegistrationFailureIsSoftFail(t *testing.T) {
 	})
 }
 
-// Collection per-item apply tolerance. The two failure shapes are treated
-// differently on purpose, and the distinction is what keeps a collection from
-// wedging on one bad member.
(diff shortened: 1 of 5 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1426, opened 2026-09-16
Keep conditional RGD children while their inputs are unready

cheeseandcereal:
An RGD can create a child whose `includeWhen` reads a dependency's phase, then delete that child when the dependency briefly changes from `Running` to `Starting`. The instance reports not-ready, but the child has already been classified as excluded and pruned.

Move the existing opt-in readiness gate before inclusion evaluation. A dependent with unready hard inputs is recorded as `Unresolved`, preserving its existing child until inclusion can be evaluated against ready inputs. Standalone Graph keeps its ungated behavior.

The all-hard-dependency tradeoff is intentional: even a schema-disabled child waits, potentially indefinitely, when a template-only dependency is unready. With this fix on the pinned base, the instance-wide prune gate also retains unrelated retired resources. With completed-template pruning, withheld/skipped owners remain protected while unrelated completed templates can prune retired members.

Focused race tests and two envtest cases pass. The same tests on the baseline reproduce missing `Unresolved` entries and actual child deletion. The primary example uses no author status or optional access, preserves child UID/data through the readiness dip, and converges afterward. The second verifies the schema-disabled tradeoff and pruning when readiness returns.

### DIFF
--- a/pkg/graphengine/executor/gating_test.go
+++ b/pkg/graphengine/executor/gating_test.go
@@ -162,6 +162,147 @@ func TestSimple_GateReadinessDoesNotReportWithheldAsApplied(t *testing.T) {
 	}
 }
 
+func TestSimple_GateReadinessBeforeIncludeWhen(t *testing.T) {
+	t.Parallel()
+	const resourcePredicate = `${db.data.phase == "Running"}`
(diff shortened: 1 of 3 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1427, opened 2026-09-16
Preserve CEL timestamp typing when publishing RGD schema values

cheeseandcereal:
An RGD using `${string(schema.metadata.creationTimestamp.getFullYear())}` in a resource template can compile successfully yet fail during instance reconciliation with `no such overload: getFullYear(string)`. The expression should return the instance's creation year.

The cached runtime seeds a typed schema value, but publishing its overridden Def replaces that value with a raw map. Extend `Runtime.Set` to use the existing schema-aware wrapper for overridden Def nodes with a schema.

The regression runs the real executor with an in-memory Kubernetes client and asserts year `"2024"` before and after schema publication. It fails on the base and passes with this fix. A computed bool/map/list case protects ordinary Defs; temporarily wrapping every Def makes that guard fail. Runtime and adapter package tests, plus focused executor tests, pass with the race detector.

### DIFF
--- a/pkg/graphengine/rgdadapter/runtime_cache_test.go
+++ b/pkg/graphengine/rgdadapter/runtime_cache_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1428, opened 2026-09-16
Reject schema-only RGD duplicate identities before adopting objects

cheeseandcereal:
When two RGD resources render the same object identity, the first can overwrite and label a pre-existing user object before the second reports a duplicate. Deleting the failed instance can then delete the adopted object.

Reuse the existing ApplySet projection to check included, fully resolved, static-GVK templates whose whole-node dependencies are empty or only `schema`. Include collection rows and default empty namespaced namespaces before checking. A collision stops child application and pruning, preserves prior inventories, and persists failure conditions and `ERROR` state at the current generation. Author conditions are still projected; author-status fields retain their previous values or remain absent for that rejected cycle.

The accompanying API regression fails on the baseline because the user's ConfigMap is overwritten. With the fix, its data and complete metadata, including UID and resourceVersion, remain unchanged through completed instance deletion. The same result holds with the alpha feature gates disabled. Targeted controls cover namespace aliases, collection/scalar overlap, ignored writers, placeholder names, and status/inventory retention.

Resource-dependent templates and standalone Graphs retain the existing per-walk duplicate guard and its first-write limitation.

### DIFF
--- a/pkg/controller/instance/controller_graph_engine.go
+++ b/pkg/controller/instance/controller_graph_engine.go
@@ -176,6 +176,10 @@ func (c *Controller) reconcileViaGraphEngine(
 	// but the parent has no inventory tracking them.
 	supersetMeta, applier, preErr := c.preApplyApplySetInventory(ctx, log, inst, rt)
 	if preErr != nil {
+		if errors.Is(preErr, applyset.ErrDuplicateResource) {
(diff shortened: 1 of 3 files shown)


jakobmoellerdev (approved):
/lgtm

PR #1430, opened 2026-09-17
fix(crd): Fail closed on schema changes

tomasaschan:
Currently, there are effectively breaking CRD changes that the comparator lets through. Most of those cover features that can't be generated by kro's SimpleSchema, but there are also a couple of cases that are effectively breaking changes but which the comparator reports as no changes.

Downstream of the comparator, this means that for some schema changes, kro skips applying new CRD config to the cluster. An RGD change reports like it has gone through, but the schema is not updated on the CRD.

This PR extends the comparator to report a lot more types of schema changes as breaking or non-breaking, including fixing the enum classification bug that led to the report of #1391. 

Additionally, the comparator is now made to fail closed, so that new features need to be explicitly implemented here (if not already covered). The behavior if not, is changed from "successful RGD reconciliation with CRD changes ignored" to "blocked RGD reconciliation", so that problematic schema changes are surfaced to RGD authors instead of silently ignored. The RGD author can always use the documented `allow-breaking-changes` annotation as an escape hatch to force kro to roll out the CRD changes even in case the comparator is too strict.

The stricter behavior is guarded by the Alpha feature gate `strict-crd-compatibility-checks`, which is disabled by default. With the gate disabled, the comparator retains its previous behavior.


jakobmoellerdev (commented):
TBH I love this. I would like to go ahead on this feature but it would need some docs on the feature flag, would you mind adding this? You have my support for it in 0.10. WDYT @cheeseandcereal

pkg/graph/crd/compat/schema.go
```diff
@@ -268,6 +331,190 @@ func compareArrayItems(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
...
+		return
+	}
+
+	fieldPath := path + ".format"
+	if oldSchema.Format != "" && newSchema.Format == "" {
+		result.AddNonBreakingChange(fieldPath, FormatRemoved, oldSchema.Format, "")
+		return
+	}
+
+	result.AddBreakingChange(fieldPath, FormatChanged, oldSchema.Format, newSchema.Format)
```
cheeseandcereal, line 378: This may be too strict. A format can change and still be backwards compatible. For example, int32 changing to int64 is not a breaking change, but would get flagged as such here.

pkg/graph/crd/compat/schema.go
```diff
@@ -268,6 +331,190 @@ func compareArrayItems(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
...
+
+	fieldPath := path + ".additionalProperties"
+	oldHasSchema := oldAdditional != nil && oldAdditional.Schema != nil
+	newHasSchema := newAdditional != nil && newAdditional.Schema != nil
+	if oldHasSchema && newHasSchema {
+		appendReport(result, compare(fieldPath, oldAdditional.Schema, newAdditional.Schema, strictChecks))
+		return
+	}
+
+	result.AddBreakingChange(
```
cheeseandcereal, line 359: Is this case accurate? If there were no additional properties allowed before, and now there are, that's not a breaking change, but it would get to this branch and return as a breaking change, so you would never be able to add "additionalProperties" to anything, right? Or am I missing something here.

pkg/features/features.go
```diff
@@ -40,16 +40,22 @@ const (
 	// Graph). When disabled (the default), the Graph controller is not started
 	// and existing ResourceGraphDefinition behavior is unaffected.
 	GraphKind featuregate.Feature = "GraphKind"
+
+	// StrictCRDCompatibilityChecks enables conservative compatibility checks
+	// for schema facets that the original CRD comparator did not cover. When
+	// disabled, the comparator retains its previous behavior.
+	StrictCRDCompatibilityChecks featuregate.Feature = "strict-crd-compatibility-checks"
```
cheeseandcereal, line 47: kebab-casing for this isn't following the style of the rest of the feature gates. Value should just be `StrictCRDCompatibilityChecks`

pkg/graph/crd/compat/schema.go
```diff
@@ -268,6 +331,190 @@ func compareArrayItems(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
 	}
 }
 
+// compareAdditionalProperties checks map value schemas recursively. We cannot
+// safely classify changes between absent, boolean, and schema forms because
+// their compatibility depends on both validation and structural-schema pruning.
+// Fail closed for those changes instead of claiming they are always breaking.
```
cheeseandcereal, line 337: "instead of claiming they are always breaking" seems contradictory to "fail closed" early in the sentence unless I'm reading this weird.

pkg/graph/crd/compat/schema.go
```diff
@@ -268,6 +331,190 @@ func compareArrayItems(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
...
+	return value != nil && *value
+}
+
+func compareTopology(path string, oldSchema, newSchema *v1.JSONSchemaProps, result *Report) {
+	fields := []struct {
+		name string
+		old  any
+		new  any
+	}{
+		{"x-kubernetes-list-map-keys", oldSchema.XListMapKeys, newSchema.XListMapKeys},
```
cheeseandcereal, line 421: This incorrectly results in a few non-breaking things getting flagged as breaking:

 nil -> "atomic" is a no-op (apiserver default for arrays), and x-kubernetes-list-map-keys order is irrelevant. Adding listType=atomic or reordering listMapKey= markers gets flagged as breaking here even though they are not breaking changes.

pkg/graph/crd/compat/schema.go
```diff
@@ -268,6 +331,190 @@ func compareArrayItems(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
...
+			path+"."+field.name,
+			TopologyChanged,
+			formatSchemaFieldValue(field.old),
+			formatSchemaFieldValue(field.new),
+		)
+	}
+}
+
+func compareValidationRules(path string, oldSchema, newSchema *v1.JSONSchemaProps, result *Report) {
+	if len(oldSchema.XValidations) == 0 && len(newSchema.XValidations) == 0 {
```
cheeseandcereal, line 440: Is there a reason to do this length check separately when `equality.Semantic.DeepEqual` will do the same check below this?

pkg/graph/crd/compat/schema.go
```diff
@@ -268,6 +331,190 @@ func compareArrayItems(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
...
+		return
+	}
+	if equality.Semantic.DeepEqual(oldSchema.XValidations, newSchema.XValidations) {
```
cheeseandcereal, line 443: This check isn't sufficient. Removing validation rules would be classified as a breaking change under this code, even though it's not.

pkg/graph/crd/compat/schema.go
```diff
@@ -268,6 +331,190 @@ func compareArrayItems(path string, oldSchema, newSchema *v1.JSONSchemaProps, re
...
+	"XListMapKeys":           true,
+	"XListType":              true,
+	"XMapType":               true,
+	"XValidations":           true,
+}
+
+func compareUnclassifiedFields(path string, oldSchema, newSchema *v1.JSONSchemaProps, result *Report) {
+	schemaType := reflect.TypeOf(*oldSchema)
+	oldValue := reflect.ValueOf(*oldSchema)
+	newValue := reflect.ValueOf(*newSchema)
```
cheeseandcereal, line 483: Nit: you could use `reflect.Type/ValueOf(schema).Elem()` to avoid de-referencing the pointers and copying the whole schemas in memory here.

cheeseandcereal (changes requested):

cheeseandcereal:
Thanks for the PR! Yes agreed with Jakob, I like the changes, and as mentioned, it will also need docs to be mergeable. No opposition to getting this in for 0.10 personally.

Also, I think some of these checks (enum update, nullable removed, preserve fields removed, etc) should just be added as general checks, not even gated behind the feature flag. I think if we can definitively say "this is for sure a breaking change", then I would consider it a bug that we're not checking for it now, and that would be fine to add in general.

There are a few other checks added here that aren't "definitive" and default to fail closed (i.e. format change can be forward compatible but is always considered breaking here), so I think those should stay behind the feature flag (in addition to the actual "comparator fails closed" bit).

jakobmoellerdev:
the only reason to gate it is due to the natural behavior being breaking to users. Agreed that it should be the new default

tomasaschan:
Thanks for the reviews, and the encouragement!

The comparator is intentionally conservative; it is arguably safer to flag any change that isn't known safe as breaking, even if it's not necessarily so, than to err in the other direction. The format example is illustrative: a format change is not _generally_ safe, and is therefore flagged even if there are examples of format changes which are safe.

One nice thing about this is that adding support for recognizing more types of changes as non-breaking, isn't itself a non-breaking change to kro, so improving the logic to recognize more types of changes as safe can be done iteratively. The format change is, again, a good example: we can merge these changes as-is, and add support for changing smaller integer types to larger ones later.

And since there is already a documented escape hatch, false positives are annoying but not blocking. Users who want to make changes which are incorrectly flagged as breaking can still do so, they just have to annotate their RGD. 

Failing closed - i.e. considering any change that can't be classified as unsafe - is also intended to make this safe when extending SimpleSchema: when a new feature is added there, the generated CRD changes won't silently fall through here and result in stale CRDs (like the result of the enum bug) but instead be flagged as unsafe until support is added here.

I agree that some of these classifications are maybe overly strict, though, so I'll see if I can relax them a little. I'll also address the constant casing inconsistency, try to improve comment wording where needed, and add some user docs about the feature flag.

pkg/graph/crd/compat/schema.go
tomasaschan: Fixed.

pkg/graph/crd/compat/schema.go
tomasaschan: I fixed this to be a little more accurate.

pkg/graph/crd/compat/schema.go
tomasaschan: Nope, no reason. Gone.

pkg/graph/crd/compat/schema.go
tomasaschan: Yeah, the total set of possible changes here was actually small enough to implement proper validation of them. I think it should be accurate enough now!

pkg/graph/crd/compat/schema.go
tomasaschan: I just removed this docstring entirely; the other unexported helpers don't have one, and this one was more confusing than helpful.

pkg/features/features.go
tomasaschan: Fixed.

pkg/graph/crd/compat/schema.go
tomasaschan: Fixed, including support for relaxing schema requirements on additional properties.

pkg/graph/crd/compat/schema.go
tomasaschan: I'm deliberately punting this one, since the number of variations to implement here can be vast. This way it is at least never too lenient; adding support for detecting some types of format changes as non-breaking can be added as follow-ups (by whoever wants/needs them).

tomasaschan:
I pushed a couple of changes; nothing that materially alters the approach, but that I think make it easier to roll this out in 0.10 without too much of an issue:

- Split the feature gate into two: one that enables the extended checks, and one that enables failing on anything unclassified. This lets users opt in to e.g. only the extended checks, without forcing them to also accept the conservative interpretation of "we don't know how this works".

- Went a little bit deeper with the new checks, to make more cases that were classified as breaking in a previous version of this PR more accurate now.

I also added some user-facing docs for the feature gates, and expanded the docs on the `allow-breaking-changes` annotation a bit.

pkg/features/features.go
```diff
@@ -40,6 +40,15 @@ const (
...
 	GraphKind featuregate.Feature = "GraphKind"
+
+	// ExtendedCRDComparison enables compatibility checks for schema facets that
+	// the original CRD comparator did not cover. When disabled, the comparator
+	// retains its previous behavior.
+	ExtendedCRDComparison featuregate.Feature = "ExtendedCRDComparison"
+
+	// ConservativeCRDComparison treats changes to OpenAPI schema fields without
+	// an explicit compatibility classification as breaking.
+	ConservativeCRDComparison featuregate.Feature = "ConservativeCRDComparison"
```
Skarlso, line 46: I'm a bit confused by these two now. :D This says they are independent and this calssifies them as breaking, but `ConservativeCRDComparison` alone fails open, right? I'm confused by the docs and the code not matching up I think?

pkg/graph/crd/compat/schema_facets_test.go
```diff
@@ -0,0 +1,393 @@
...
+			name:              "preserve unknown fields added",
+			old:               v1.JSONSchemaProps{Type: "object"},
+			new:               v1.JSONSchemaProps{Type: "object", XPreserveUnknownFields: &trueValue},
+			nonBreakingType:   PreserveUnknownFieldsAdded,
+			expectedFieldPath: ".x-kubernetes-preserve-unknown-fields",
+		},
+		{
+			name:              "list type changed",
+			old:               v1.JSONSchemaProps{Type: "array", XListType: &atomic},
+			new:               v1.JSONSchemaProps{Type: "array", XListType: &set},
```
Skarlso: Isn't this a breaking change with regards to how types are interpreted?

website/docs/docs/concepts/rgd/01-overview.md
```diff
@@ -177,19 +177,23 @@ kro supports the following annotations on ResourceGraphDefinitions:
...
-  annotations:
-    kro.run/allow-breaking-changes: "true"
-spec:
-  # ...
+```bash
+kubectl annotate resourcegraphdefinition my-rgd \
+  kro.run/allow-breaking-changes="true" --overwrite
+kubectl apply -f my-rgd.yaml
+kubectl annotate resourcegraphdefinition my-rgd \
+  kro.run/allow-breaking-changes-
```
cheeseandcereal: Did this get cut off? This doesn't look right.

Regardless, I personally think the previous example with the actual yaml was easier to grok. The full example yaml doesn't imply using kubectl, or having to think about what `kubectl annotate` is doing, etc.

I don't think any of this behavior is really changing anyways, so I think everything from here up and can just remain unchanged, personally.

website/docs/docs/concepts/rgd/01-overview.md
```diff
@@ -177,19 +177,23 @@ kro supports the following annotations on ResourceGraphDefinitions:
...
 ```
 
+The annotation bypasses every breaking-change check, including checks enabled
+by the CRD comparison feature gates. Remove it after the intended update so
```
cheeseandcereal: > Remove it after the intended update so later accidental breaking changes are blocked again.

I don't think we have to spell this out, nor be specific to the use case of performing a "single" breaking change. It's possible you may just want to always disable this check forever. No need to be over-specific. I think documenting the general use-case (as it was) was fine. Adding the sentence that this also bypasses the strict checks too is fine. But no need to re-write this part IMO.

pkg/features/features.go
```diff
@@ -40,6 +40,15 @@ const (
 	// Graph). When disabled (the default), the Graph controller is not started
 	// and existing ResourceGraphDefinition behavior is unaffected.
 	GraphKind featuregate.Feature = "GraphKind"
+
+	// ExtendedCRDComparison enables compatibility checks for schema facets that
+	// the original CRD comparator did not cover. When disabled, the comparator
+	// retains its previous behavior.
+	ExtendedCRDComparison featuregate.Feature = "ExtendedCRDComparison"
```
cheeseandcereal: I don't think we need 2 feature flags for this.

My take:
For the new checks that are _not_ definitive (i.e. checking if a format is forward compatible is out of scope, so we just conservatively consider all format changes breaking), keep those checks behind the conservative comparison feature flag.

For the checks that _are_ definitive (i.e. removing an enum value), just add those as normal checks alongside the other checks with no conditions. While possibly technically breaking, it is a "bug" that kro allows an RGD to make a breaking change, so definitive checks that add coverage are in scope to add without a feature flag, IMO.

pkg/graph/crd/compat/schema.go
cheeseandcereal: Makes sense, similar to some of the other added checks this is just a general conservative, non-definitive check. In this case I think it's fine to keep this check behind the conservative check feature flag.

pkg/graph/crd/compat/schema.go
```diff
@@ -100,20 +119,41 @@ func compare(path string, oldSchema, newSchema *v1.JSONSchemaProps) *Report {
...
-	compareArrayItems(path, oldSchema, newSchema, result)
+	compareArrayItems(path, oldSchema, newSchema, result, options)
+
+	if options.extendedComparison {
+		// For maps, check the value schema.
+		compareAdditionalProperties(path, oldSchema, newSchema, result, options)
+
+		compareFormat(path, oldSchema, newSchema, result)
+		compareNullable(path, oldSchema, newSchema, result)
+		comparePreserveUnknownFields(path, oldSchema, newSchema, result)
```
cheeseandcereal: IMO I think these 2 checks are "definitive", so we can probably just add them in general (don't need this extended comparison feature flag gate).

pkg/graph/crd/compat/schema.go
```diff
@@ -223,9 +266,37 @@ func compareRequiredFields(path string, oldSchema, newSchema *v1.JSONSchemaProps
...
+		result.AddBreakingChange(
+			path+".enum",
+			EnumConstraintAdded,
+			"",
+			formatSchemaFieldValue(newSchema.Enum),
+		)
+		return
+	}
+	if len(newSchema.Enum) == 0 {
+		if !extendedComparison {
```
cheeseandcereal: IMO there's no reason to gate these 2 new checks behind a feature flag. These are just correct to always flag

cheeseandcereal (approved):
Thanks for the contribution!

jakobmoellerdev (approved):

PR #1431, opened 2026-09-17
Keep inventory projection from suppressing wanted children

cheeseandcereal:
An RGD that reads a dependency in author status and uses an absence-tolerant `includeWhen` can report `ACTIVE` without creating its wanted child. Inventory projection evaluates the condition against an empty placeholder and caches that verdict for execution.

Run projection on a disposable runtime reusing the compiled Program, effective instance schema, seed scope, and collection limit. Execution reevaluates inclusion against applied dependencies. This adds one runtime construction per projection without recompilation.

Controller regressions cover both verdict directions and reproduce the wrong cached verdict without isolation. A default-gates envtest example verifies app/leaf creation, child data, author status, and timestamp-based inclusion with this patch. No `readyWhen` is required. Conditional write-ahead inventory can still be incomplete until the post-apply update.


jakobmoellerdev (approved):
/lgtm

PR #1435, opened 2026-09-18
fix(graphengine): RGD external ref pending causes errors instead of waits

NicholasBlaskey:
Regression on RGD in graph engine 

Previously if a instance status referencing a pending external ref the graph execution would be in progress with a trimmed status.

After graph update we get an error saying we can't apply an empty object.

Fix this regression by omitting the empty field that would cause this error.


pkg/graphengine/runtime/node.go
```diff
@@ -363,6 +365,17 @@ func (n *Node) renderOne(bindings map[string]any) (*unstructured.Unstructured, e
 	// down), so we drop the whole enclosing array field after Resolve — see
 	// below.
 	var omitArrayFields [][]string
+	// omitObjectParents collects the map-key paths of the enclosing object
```
cheeseandcereal, line 368: Can you please cut down/simplify/distill these code comments?

You have 22 lines of comments for 4 lines of code in these first 4 changed blocks.

pkg/graphengine/runtime/node.go
```diff
@@ -407,9 +424,45 @@ func (n *Node) renderOne(bindings map[string]any) (*unstructured.Unstructured, e
...
+		return
+	}
+	// Deduplicate and sort deepest-first (longest path first).
+	seen := make(map[string]struct{}, len(parents))
+	uniq := make([][]string, 0, len(parents))
+	for _, p := range parents {
+		if len(p) == 0 {
+			continue
+		}
+		key := strings.Join(p, "\x00")
```
cheeseandcereal, line 451: What is this line for? Why do you need to add a null byte to a string?

pkg/graphengine/runtime/node.go
```diff
@@ -407,9 +424,45 @@ func (n *Node) renderOne(bindings map[string]any) (*unstructured.Unstructured, e
...
+		if _, ok := seen[key]; ok {
+			continue
+		}
+		seen[key] = struct{}{}
+		uniq = append(uniq, p)
+	}
+	sort.SliceStable(uniq, func(i, j int) bool { return len(uniq[i]) > len(uniq[j]) })
```
cheeseandcereal, line 458: Maybe I'm misunderstanding, but this is sorting based on the length of the path string? But what you're trying to do is order based on the _depth_ of the path, right? I don't think those things are equivalent, so I don't think this is doing what's intended.

pkg/graphengine/runtime/node.go
NicholasBlaskey: simplified comments

pkg/graphengine/runtime/node.go
NicholasBlaskey: this null byte was added to be the path separator to track duplicates

Removed the duplication check and handled it when we do the removal

pkg/graphengine/runtime/node.go
NicholasBlaskey: Updated this code to be more clear

cheeseandcereal (approved):

cheeseandcereal:
Ok, after further looking, I think actually the current behavior for arrays is wrong too. I'm going to suggest we go with a more holistic approach to refactor/match the current behavior in 0.9.4, which I've implemented here: #1441

PR #1439, opened 2026-09-21
docs: improve example for stable ordering on lists built from maps

omus:
Follow up to: https://github.com/kubernetes-sigs/kro/pull/1375. Provides an improved example for the stable ordering for lists built from map by using `transformList` and `sortBy`. The advantage of this approach is it avoids having to reference the original map to perform the `[{key, value}]` transformation. I think this pattern is better as it allows for additional entries to be merged into the map before the transformation.


omus:
cc: @allamand

PR #1441, opened 2026-09-22
fix(graphengine): restore progressive object and array status

cheeseandcereal:
## Summary

Restore progressive object and array status projection when expressions are data-pending.

Replace field-path-based omission with a private pending marker and one recursive cleanup in the runtime:

- Omit object properties and parent containers whose children are all pending.
- Keep resolved array entries at their original indices, using null placeholders for earlier pending entries and trimming trailing pending entries.
- Keep partially resolved objects inside arrays, explicitly empty values, and resolved nulls.

Production changes are confined to `pkg/graphengine/runtime/node.go`.

## Behavior comparison

`main` below refers to behavior before this fix. `P` denotes a data-pending expression, and `R(x)` denotes an expression that successfully resolves to `x`. Each example is the value of an authored status field; **absent** means that entire field is omitted.

### Objects

| Authored field value | v0.9.4 rendered value | `main` rendered value |
| --- | --- | --- |
| `{a: P, b: P}` — all children pending | **absent** | `{}` |
| `{a: R("a"), b: P}` — one child resolves | `{"a":"a"}` | `{"a":"a"}` |
| `{nested: {value: P}}` — nested object entirely pending | **absent** | `{"nested":{}}` |
| `{values: [P]}` — nested array entirely pending | **absent** | `{}` |
| `R({})` — expression successfully returns an empty object | `{}` | `{}` |

### Arrays

| Authored field value | v0.9.4 rendered value | `main` rendered value |
| --- | --- | --- |
| `[R("a"), R("b")]` — all elements resolve | `["a","b"]` | `["a","b"]` |
| `[R("a"), P]` — trailing element pending | `["a"]` | **absent** |
| `[P, R("b")]` — leading element pending | `[null,"b"]` | **absent** |
| `[R("a"), P, R("c")]` — middle element pending | `["a",null,"c"]` | **absent** |
| `[P, P]` — all elements pending | **absent** | **absent** |
| `[{name: R("a"), value: P}]` — partially resolved object in an array | `[{"name":"a"}]` | **absent** |
| `[P, R(null), P]` — resolved null between pending elements | `[null,null]` | **absent** |
| `R([])` — expression successfully returns an empty array | `[]` | `[]` |

A pending expression supplying an entire object or array is omitted in both versions. For the cases above, this PR produces the v0.9.4 shapes. These are rendered payloads: admission still depends on the target schema, including whether array items allow null placeholders.

## Testing

- Runtime and external-reference e2e tests cover all cases in the tables above, including partial objects, array gaps, and resolved empty/null values.
- Kind e2e coverage includes deletion, spec updates while pending, and recovery, using schemas that allow null array entries and reject empty status objects.
- Verified against v0.9.4 and pre-fix `main`: the fix matches v0.9.4 for all 13 cases. Focused runtime tests with race detection and all 11 e2e tests passed.


jakobmoellerdev (approved):

Skarlso:
Programatically this looks okay. But I think I have a hunch where this might break and I think the e2e test didn't catch it because of the way the ordering is happening.

So, the e2e test patches the CRD. Where as the RGD's crd generator actually rejects nil values that are pushed here:

```go
			if _, pending := resolved.(pendingValue); pending {
				v[i] = nil
			} else {
				v[i] = resolved
				length = i + 1
			}

```

RGD:

```yaml
  apiVersion: kro.run/v1alpha1
  kind: ResourceGraphDefinition
  metadata:
    name: pending-null-repro
  spec:
    schema:
      apiVersion: v1alpha1
      kind: PendingNullRepro
      spec:
        tag: string | default=v1
      status:
        # this should be resolving
        canary: ${schema.spec.tag}
        # simple type items:{type:string}.
        sparseValues:
          - ${input.data.ECHO_VALUE}
          - ${schema.spec.tag}
    resources:
      - id: input
        externalRef:
          apiVersion: v1
          kind: ConfigMap
          metadata:
            name: pending-null-repro-src
            namespace: default
```

And then the instance:

```yaml
apiVersion: kro.run/v1alpha1
kind: PendingNullRepro
metadata:
  name: repro
  namespace: default
```

And the error:

```
kubectl get pendingnullrepro repro -o jsonpath='{.status}'
{"conditions":[{"lastTransitionTime":"2026-09-22T06:12:27Z","message":"instance is properly managed with finalizers and labels","observedGeneration":1,"reason":"Managed","status":"True","type":"InstanceManaged"},{"lastTransitionTime":"2026-09-22T06:12:27Z","message":"runtime graph created and all resources resolved","observedGeneration":1,"reason":"Resolved","status":"True","type":"GraphResolved"},{"lastTransitionTime":"2026-09-22T06:12:27Z","message":"resource reconciliation failed: apply \"instance\" (patch): PendingNullRepro.kro.run \"repro\" is invalid: status.sparseValues[0]: Invalid value: \"null\": sparseValues[0] in body must be of type string: \"null\"","observedGeneration":1,"reason":"NotReady","status":"False","type":"ResourcesReady"},{"lastTransitionTime":"2026-09-22T06:12:27Z","message":"resource reconciliation failed: apply \"instance\" (patch): PendingNullRepro.kro.run \"repro\" is invalid: status.sparseValues[0]: Invalid value: \"null\": sparseValues[0] in body must be of type string: \"null\"","observedGeneration":1,"reason":"NotReady","status":"False","type":"Ready"}],"state":"ERROR"}%
```

cheeseandcereal:
Yes, this code still applies null values within an array in the case that there is a resolved value after a pending value (the value is set to null instead of omitted to maintain ordering). Note that this is intentional, because it's the _same_ behavior that's in 0.9.4 right now which is what this is trying to reproduce.

Agreed, it would probably make sense to take another look at this behavior at some point, but I'm not trying to assert what the "correct" behavior here should be besides replicating the exact behavior before the refactor (in 0.9.4) for now. This does mean that there are cases where kro will try to apply a null value in an array where "null" is not an actual valid value, and can reject as you've seen in your example, but this is consistent with the behavior in 0.9.4 as well.

Skarlso:
/lgtm

@cheeseandcereal Ah I might have misunderstood the intention of the this PR then. 🤦‍♀️ Sorry about that.

